PR #3, opened 2014-11-14
usrloc_dmq: add module for usrloc sync via dmq

AndreyRybkin:
New module for synchronizing usrloc contacts via dmq.


miconda:
I will merge this. There are few things to be sorted out in short term:

-  variables to be declared at the beginning of functions or blocks, to be sure it still compiles with old strict C compilers
- add license at the top of the c/h files - it must be GPLv2 or BSD, at your choice

We will grant you direct commit access to main kamailio repository, so you can push those updates directly.

PR #13, opened 2014-12-23
dispatcher: fixing some misspelled words and adding support for custom number of successful probing requests

alezzandro:
First of all excuse me if in the previous pull request I didn't follow guidelines by devel-git-commit guide :)
As anticipated in the previous request I made some changes to dispatcher module for counting successful ping requests before moving a destination from inactive to active state.
While doing this changes I noticed some misspelled variables (in code and docs too) and I fixed them. So I've used similar variables (with the corrected words) also for the new feature I developed. 
In the case we leave the old variables introducing the new one we will have in the code:
"ds_probing_threshhold" vs "ds_inactive_threshold"
This behaviour could be confusing while editing the configuration file and also while reading the code.

As you requested I've separated the misspelled fixes in two commits one for the code and one for the docs.

@linuxmaniac suggested to create two different pull request: I tried but I had no success. I've submitted the first pull request only with two commits about the misspelled words, by the way when I've pushed the other two commits regarding the new developed feature they has been added to the pending pull request automatically.

Please note: You will notice a rename of the var "failure_count" in "message_count" (dispatch.h) because I use the same counter var for failure and successful requests.

### DIFF
--- a/modules/dispatcher/README
+++ b/modules/dispatcher/README
@@ -26,15 +26,17 @@ Edited by
 
 Alessandro Arrichiello
 
-   Hewlett-Packard
+   Hewlett Packard
 
-   Copyright � 2004 FhG FOKUS
+   Copyright © 2004 FhG FOKUS
 
-   Copyright � 2005 Voice Sistem
+   Copyright © 2005 Voice Sistem
 
-   Copyright � 2010 Daniel-Constantin Mierla (asipto.com)
+   Copyright © 2010 Daniel-Constantin Mierla (asipto.com)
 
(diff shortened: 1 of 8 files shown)


miconda:
I made some comments inline to get the patches merged clean. There is a mismatch in the code and documentation about the default value of the new parameter. Then, trying to keep the lines within 80chars size is rather important. Few other bits and pieces I commented about.

alezzandro:
@miconda I've just followed all your suggestions and pushed all the commits, please let me know if I need to change something more.
Thanks.

PR #14, opened 2014-12-24
Improve for correct working with all RADIUS AV pairs

borikinternet:
In this pull request I`d correct work with retrieved from RADIUS AV pairs for modules auth_radius, misc_radius and peering. Now this modules correctly put to $avp() all retrieved params:
- integer name and integer value
- integer name and string value
- string name and integer value
- string name and string value

All above said and for SIP-AVP params too.

### DIFF
--- a/modules/auth_radius/sterman.c
+++ b/modules/auth_radius/sterman.c
@@ -87,20 +87,26 @@ static inline int extract_avp(VALUE_PAIR* vp, unsigned short *flags,
 		p = strchr(vp->strvalue,'#');
 		q = strchr(vp->strvalue,':');
 		r = strrchr(vp->strvalue,'#');
-		if ( p != NULL && p == r && p == vp->strvalue && q != NULL ) {
+		if ( p == vp->strvalue && q != NULL ) {
 			// int name and str value
 			*flags |= AVP_VAL_STR;
 			p++;
 			name->n = strtol(p,&q,10);
 			q++;
 			value->s.len = strlen(q);
-			value->s.s = malloc(value->s.len + 1);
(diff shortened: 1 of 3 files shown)


borikinternet:
Thanks, your recomendations are implemented in new commits. Realy, I think, that free() is called in function, that destroys avp, but when I inspect usr_avp, I understood that I done a mistake. Similary, I added correct processing of malloc() errors, malloc() and free() was replaced to pkg_malloc() and pkg_free() respectively. I look at SIP-AVP processing and found one mistake (for '#' in string value case, I`m not considered this option), and this corrected too in new commits. In my oppinion, not using memory allocation in this section will be very traumatic for kamailio, because we don`t know, in which place in memory we place our string values.

Discussion question: some vendors (as Cisco) put name in value too. Cisco placed '=' between name part of value and value part of value. Some other vendors may put some other symbols. In my code I checking '=',':' and '#' by analogy with SIP-AVP pair syntax. Whether it is necessary or not?

miconda:
pkg_malloc() for name and value is not needed. It is more traumatic to allocate in pkg, then copy there and then free after adding as Kamailio AVP (there is no benefit, only a lot of operations here for nothing good). When adding as a K AVP, the name and value are cloned in shared memory. Copying the values directly from the RADIUS AVP is ok (like it was so far) and faster.

The new patches are on top of the old one, this is not what we want to commit, because there must be a patch per module, starting from the existing code. Now it is hard to review the contribution, because we have to analyse patches to patches, not being easy to spot eventual bugs. These patches in the pull request must be discarded and create new ones against Kamailio master.

PR #17, opened 2015-01-05
dialplan: skip check subexpressions on expressions with pvs

linuxmaniac:
subexpressions should not be checked when the expression has private variables

output of the error
```
 0(10175) DEBUG: dialplan [dp_db.c:496]: build_rule(): building rule for [2:^(00|\+)([1-9][0-9]+)$/^(00|\+)([1-9][0-9]+)$/\2]
 0(10175) DEBUG: dialplan [dp_db.c:533]: build_rule(): attrs are: ''
 0(10175) DEBUG: dialplan [dp_db.c:580]: add_rule2hash(): new dpl_id 1
 0(10175) DEBUG: dialplan [dp_db.c:593]: add_rule2hash(): new index , len 0
 0(10175) DEBUG: dialplan [dp_db.c:629]: add_rule2hash(): added the rule id 1 index 0 pr 1 next (nil) to the index with 0 len
 0(10175) DEBUG: dialplan [dp_repl.c:146]: repl_exp_parse(): replacement expression is [\1]
 0(10175) DEBUG: dialplan [dp_db.c:102]: dpl_check_pv(): parsing [^$avp(s:caller_cc)$avp(s:caller_ac)([1-9][0-9]+)$]
 0(10175) DEBUG: <core> [pvapi.c:419]: pv_spec_lookup(): PV <$avp(s:caller_cc)> is not in cache
 0(10175) DEBUG: <core> [usr_avp.c:882]: parse_avp_ident(): Parsing 's:caller_cc'
 0(10175) DEBUG: <core> [pvapi.c:293]: pv_cache_add(): pvar [$avp(s:caller_cc)] added in cache
 0(10175) DEBUG: dialplan [dp_db.c:124]: dpl_check_pv(): string [^$avp(s:caller_cc)$avp(s:caller_ac)([1-9][0-9]+)$] has variables
 0(10175) DEBUG: dialplan [dp_db.c:496]: build_rule(): building rule for [2:^$avp(s:caller_cc)$avp(s:caller_ac)([1-9][0-9]+)$/^$avp(s:caller_cc)$avp(s:caller_ac)([1-9][0-9]+)$/\1]
 0(10175) ERROR: dialplan [dp_db.c:502]: build_rule(): repl_exp \1 refers to 1 sub-expressions, but subst_exp ^$avp(s:caller_cc)$avp(s:caller_ac)([1-9][0-9]+)$ has only 0
[...]

PR #18, opened 2015-01-05
dialplan: using $(avp("key")[*]) on match/subst rules

### LINKED ISSUE #5: dialplan: using $(avp("key")[*]) on match/subst rules
Enhancement to dialplan module.
A rule having an avp with index all`([*])` will be checked with the avp values one by one.
Example:
Having `$(avp(s:list)[*]) -> "111", "222"`
`'^(00|\+)?$(avp(s:list)[*])$'` will be checked as:
*  `'^(00|\+)?111$'`
*  `'^(00|\+)?222$'`

linuxmaniac:
This closes #5 . Be aware that this needs the #17 applied in order to work

### DIFF
--- a/modules/dialplan/dialplan.h
+++ b/modules/dialplan/dialplan.h
@@ -103,7 +103,7 @@ dpl_id_p select_dpid(int id);
 struct subst_expr* repl_exp_parse(str subst);
 void repl_expr_free(struct subst_expr *se);
 int translate(struct sip_msg *msg, str user_name, str* repl_user, dpl_id_p idp, str *);
-int rule_translate(struct sip_msg *msg, str , dpl_node_t * rule,  str *);
+int rule_translate(struct sip_msg *msg, str , dpl_node_t * rule, pcre *subst_comp,  str *);
 
 pcre *reg_ex_comp(const char *pattern, int *cap_cnt, int mtype);
 #endif
--- a/modules/dialplan/dp_repl.c
+++ b/modules/dialplan/dp_repl.c
@@ -202,30 +202,6 @@ pcre *dpl_dyn_pcre_comp(sip_msg_t *msg, str *expr, str *vexpr, int *cap_cnt)
 	return re;
 }
 
-pcre *dpl_dynamic_pcre(sip_msg_t *msg, str *expr, int *cap_cnt)
-{
-	pv_elem_t *pelem = NULL;
-	str vexpr;
-
-	if(expr==NULL || expr->s==NULL || expr->len<=0)
-		return NULL;
-
-	if(pv_parse_format(expr, &pelem)<0){
-		LM_ERR("parsing pcre expression: %.*s\n",
-				expr->len, expr->s);
-		return NULL;
-	}
-	if(pv_printf_s(msg, pelem, &vexpr)<0){
-		LM_ERR("cannot get pcre dynamic expression value: %.*s\n",
-				expr->len, expr->s);
-		pv_elem_free_all(pelem);
-		return NULL;
-	}
-	pv_elem_free_all(pelem);
-
-	return dpl_dyn_pcre_comp(msg, expr, &vexpr, cap_cnt);
-}
-
 dpl_dyn_pcre_p dpl_dynamic_pcre_list(sip_msg_t *msg, str *expr)
 {
 	pv_elem_p elem = NULL;
@@ -402,11 +378,10 @@ error:
 #define MAX_PHONE_NB_DIGITS		127
 static char dp_output_buf[MAX_PHONE_NB_DIGITS+1];
 int rule_translate(sip_msg_t *msg, str string, dpl_node_t * rule,
-		str * result)
+		pcre *subst_comp, str * result)
 {
 	int repl_nb, offset, match_nb, rc, cap_cnt;
 	struct replace_with token;
-	pcre *subst_comp;
 	struct subst_expr * repl_comp;
 	str match;
 	pv_value_t sv;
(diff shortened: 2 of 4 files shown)


miconda:
The patch is quite big and I am trying to figure out the purpose. Was $(avp(x)[*]) not printed properly in the match or subst rules? It should, because its value is taken via common pv framework which evaluates it like for an xlog() parameter.

Can you give an example of a rule that doesn't work before this patch and works after?

linuxmaniac:
This is the route that I'm using for testing this feature:
```
route[TEST] {
	pv_unset("$(avp(s:pbx_cli)[*])");
	$avp(s:pbx_cli) = "1111";
	$avp(s:pbx_cli) = "2222";
	$avp(s:pbx_cli) = "3333";
	$var(num) = "11118765";
	if(dp_translate("2", "$var(num)/$avp(s:dest)"))
	{
		xlog("$var(num) translated to var $avp(s:dest) \n");
	}
	$var(num) = "0022228765";
	if(dp_translate("2", "$var(num)/$avp(s:dest)"))
	{
		xlog("$var(num) translated to var $avp(s:dest) \n");
	}
	$var(num) = "00333387";
	$(avp(s:cloud_pbx_cli)[*]) = "666";
	if(dp_translate("2", "$var(num)/$avp(s:dest)"))
	{
		xlog("$var(num) translated to var $avp(s:dest) \n");
	}
	if(dp_translate("1", "$var(num)/$avp(s:dest)"))
	{
		xlog("$var(num) translated to var $avp(s:dest) \n");
	}
}
```
This is the rules at dialplan:
```
sqlite> select * from dialplan;
1|1|1|2|^(00|\+)([1-9][0-9]+)$|0|^(00|\+)([1-9][0-9]+)$|\2|
2|2|1|2|^$avp(s:caller_cc)$avp(s:caller_ac)([1-9][0-9]+)$|0|^$avp(s:caller_cc)$avp(s:caller_ac)([1-9][0-9]+)$|\1|
3|1|2|0|^(00|\+)([1-9][0-9]+)$|0|^(00|\+)([1-9][0-9]+)$|\2|
4|2|2|0|^$avp(s:caller_cc)$avp(s:caller_ac)([1-9][0-9]+)$|0|^$avp(s:caller_cc)$avp(s:caller_ac)([1-9][0-9]+)$|\1|
5|1|2|1|^(00|\+)([1-9][0-9]+)$|0|^(00|\+)([1-9][0-9]+)$|\2|
6|2|2|1|^$avp(s:caller_cc)$avp(s:caller_ac)([1-9][0-9]+)$|0|^$avp(s:caller_cc)$avp(s:caller_ac)([1-9][0-9]+)$|\1|
7|2|2|1|^(00|\+)?$(avp(s:pbx_cli)[*])([0-9]{3})$|0|^(00|\+)?$(avp(s:pbx_cli)[*])([0-9]{3})$|\2|
8|2|2|1|^(00|\+)?$(avp(s:pbx_cli)[*])([1-9]{2})$|0|^(00|\+)?$(avp(s:pbx_cli)[*])([1-9]{2})$|$avp(s:cloud_pbx_cli)|
```
As you can see if you use $(avp(s:pbx_cli)[*]) you will get "333
3, 2222, 1111". So the normal behavior for print but not what we want. 
```
 4(1425) DEBUG: dialplan [dp_repl.c:77]: dpl_dynamic_pcre(): compiled dynamic pcre expression: ^(00|\+)?$(avp(s:pbx_cli)[*])([0-9]{3})$$ (^(00|\+)?333
3, 2222, 1111([0-9]{3})$) 2

```

With the patch we are checking every value of the avp, one by one:
```
 4(5674) DEBUG: dialplan [dp_repl.c:249]: dpl_dynamic_pcre_list(): parsed pcre expression: ^(00|\+)?$(avp(s:pbx_cli)[*])([0-9]{3})$$
 4(5674) DEBUG: dialplan [dp_repl.c:222]: dpl_dyn_pcre_comp(): compiled dynamic pcre expression: ^(00|\+)?3333([0-9]{3})$ 2
 4(5674) DEBUG: dialplan [dp_repl.c:222]: dpl_dyn_pcre_comp(): compiled dynamic pcre expression: ^(00|\+)?2222([0-9]{3})$ 2
 4(5674) DEBUG: dialplan [dp_repl.c:222]: dpl_dyn_pcre_comp(): compiled dynamic pcre expression: ^(00|\+)?1111([0-9]{3})$ 2
 4(5674) DEBUG: dialplan [dp_repl.c:628]: translate(): match check: [^(00|\+)?1111([0-9]{3})$] -1
 4(5674) DEBUG: dialplan [dp_repl.c:628]: translate(): match check: [^(00|\+)?2222([0-9]{3})$] -1
 4(5674) DEBUG: dialplan [dp_repl.c:628]: translate(): match check: [^(00|\+)?3333([0-9]{3})$] -1
 4(5674) DEBUG: dialplan [dp_repl.c:613]: translate(): regex operator testing over [00333387]
```

linuxmaniac:
Rebased and changed dpl_get_avp_val() implementation using get_avp_val() as @miconda suggested

miconda:
OK, I see now what you were looking for. I will think a bit about, because you remove the option of being able to use $(avp(x)[*]) as it returns for print.

linuxmaniac:
What about having the $(avp(x)[*]) as it returns for print as the first value to check and then one value at the time?

miconda:
It still changes the expected behaviour and somehow adds unnecessary complexity as I expect one wants a variant, not the both at the same time.

Perhaps the best is to have another index notation for this case, like:
```
$(avp(x)[+])
```
or
```
$(avp(x)[**])
```

The parser function of the index needs to be extended, but that should be easy. Then there are not many pvs using '*' index, thus would not be hard to propagate.

To decide: what to return in case of using the new index in print function (e.g., xlog) -- I would say the first value. CSV is available with '*'.

linuxmaniac:
OK, ```$(avp(x)[+])``` seems nice and I will go for returning the first value on xlog. Agree?

miconda:
ok - it could be defined as PV_IDX_ITR (from iterator).

linuxmaniac:
Initial work introducing PV_INDX_ITR. Is there anything else to take care of?

miconda:
It can be merge, testing will be easier afterwards ...

linuxmaniac:
Merged manually

PR #23, opened 2015-01-09
registrar: check for local path on lookup

charlesrchance:
When performing a lookup, the Path (if present) is evaluated and if the first hop is local we skip it to avoid unnecessary looping.

It is useful if multiple servers are sharing a common database, each saving contacts with their local address as the Path.


miconda:
You can merge it!

PR #28, opened 2015-01-12
siputils: fix for e164_check()

seudin:
- fix improper length check

### DIFF
--- a/modules/siputils/checks.c
+++ b/modules/siputils/checks.c
@@ -417,7 +417,7 @@ static inline int e164_check(str* _user)
 	char c;
 
 	if ((_user->len > 2) && (_user->len < 17) && ((_user->s)[0] == '+')) {
-		for (i = 1; i <= _user->len; i++) {
+		for (i = 1; i < _user->len; i++) {
 			c = (_user->s)[i];
 			if (c < '0' || c > '9') return -1;
 		}


miconda:
Note for the future -- be sure the comments on the commits are providing proper details.

While the description here was correct, respectively:

"fix improper length check"

The commit has something else:

"- the condition for non-digit matching was always false"

Which can be a side effect in some cases, but not the real bug that was fixed. The commit didn't change anything that tested if a char is between 0 and 9.

In the future, be sure the commit message is giving the proper details, otherwise it can mislead. Commit logs are very important, more than the comments on issue tracker.

linuxmaniac:
My bad. You're totally right.

0x6e6562:
This regression appears to be in the 4.2.2 release, but not in 4.2.1 - are you likely to be releasing a 4.2.3 soon, or should we just go back to 4.2.1 for now?

miconda:
It is not a regression, the fix was for an issue that is there for long time (including 4.2.1). The description is wrong and the fix doesn't relate to the previous patch fixing the 0-9 range check. There will be a 4.2.3, you can use git branch or nightly deb builds of branch 4.2 from apt repository.

0x6e6562:
I'm not 100% sure that the issue was in 4.2.1, but I might be mistaken. I upgraded from 4.2.1 to 4.2.2 today and the first and only thing that broke was the `is_e164()` function always returning false. That was with inputs that had previously returned true in 4.2.1. But potentially I got something else wrong in my particular setup.

miconda:
The issue solved by a previous patch was that is_e164() was returning always true -- it was checking that (c<'0' && c>'9') to say it is not e164, which never happens.

The issue solved by the patch on this item was about going over the length of the parameter with 1, which was even before the first patch.

After the first patch, the effect of second issue going over the length, results in checking an extra character which very likely is not a digit.

In other words, the first fix revealed another bug by enabling a side effect of it in most of the cases.

miconda:
Btw, the check can be done with regex, like matching $rU is starting with + followed by [1-9] followed by many [0-9].

PR #31, opened 2015-01-13
nathelper: contact_only option

linuxmaniac:
without this patch nathelper build request-uri from received
if it's set and falls back to contact.

Now we have an option to send the Contact (that has the username part) no matter what.

We used this in order to fix problems with FritzBox subscribers but seeing https://www.mail-archive.com/sr-users@lists.sip-router.org/msg13081.html maybe this is useful for everybody

with @sipwise hat on

### DIFF
--- a/modules/nathelper/nathelper.c
+++ b/modules/nathelper/nathelper.c
@@ -2051,6 +2051,8 @@ nh_timer(unsigned int ticks, void *timer_idx)
 	int rval;
 	void *buf, *cp;
 	str c;
+	str recv;
+	str *dst_uri;
 	str opt;
 	str path;
 	str ruid;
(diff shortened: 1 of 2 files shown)


linuxmaniac:
The rebase just removed the comments here. @miconda I think this works for us because we send the messages through our lb and the contact has the received parameter included so lb sends the message to the proper IP with the proper contact info.

But I'm kind of stuck here, what should be do here in order to get contact and received from usrloc and build the proper message in nathelper?

miconda:
You need to extend both usrloc and nathelper. usrloc has to pass both contact and received fields.

The data passed from usrloc to nathelper is done by printing in a buffer the size of the field (as integer, in 4 bytes) followed by the field value. Both modules must use the same order of the fields (usrloc to write, nathelper to read). usrloc must be extended in both functions: the one fetching from database and the one fetching from memory.

In usrloc, you can look at how path is added, same should be done for received. In nathelper, if path is not set, then the destination is taken from received if set, if not from contact.

The new parameter might not be needed afterwards, as the r-uri for options should be always the contact address.

linuxmaniac:
reworked as @miconda suggested
usrloc.get_all_*_ucontacts now sends received and contact fields
nathelper.nh_timer reads both and uses received for building the dst but contact for building r-uri

miconda:
Can be merged.

linuxmaniac:
Merged
wrong number on commit fixes :-(

PR #45, opened 2015-01-18
modules/cdp: added src_addr parameter in peer definition

grumvalski:
This patch add the possibility to specify the src ip addr we want to use when connecting to a peer.
Useful when we have several ip addresses in the same machine and we want diameter traffic goimg out from a specific one.


miconda:
Apparently it doesn't work to assign more than one developer -- i wanted to ping the ims guys on this item.

kamailio-sync:
Hey Daniel,

Received. will look into it!

miconda:
Any update on reviewing this? The patch seems rather ok.

miconda:
Going to merge this pull request, waiting here for a while. Patch looks ok. If any adjustments, then can be done directly on the merged code.

grumvalski:
Thanks Daniel.

Cheers,

Federico

PR #46, opened 2015-01-18
Improoved working with RADIUS AV pairs

borikinternet:
- added modparam import_all_avp parameter (int, true if not zero)
- reading all RADIUS AV-pairs to kamailio avp (not only SIP-AVP, if import_all_avp set to true)
- skiping name part of value (example: from RADIUS we recieve name - h323-x-header, value -
  "h323-x-header=blah-blah-blah", in kamailio we have name
  h323-x-header, value - "blah-blah-blah"), only for non SIP-AVP
- all old comments are taken into account


miconda:
I merged the patch and then renamed the new parameter import_all_avp to radius_avps_mode to be more suggestive about what kind of AVPs are controlled with this parameter.

PR #49, opened 2015-01-21
auth: use correct parameter for pv_auth_check checks

mamikk:
The pv_auth_check has a flags and a checks parameter but the flags
parameter is used for both. This means that if flags has the 1-bit set
it will enable both HA1-passwords and from/to URI checks and setting the
1-bit in the checks parameter does not do anything.

This fixes the parameters so that the checks parameter is used for the
to/from URI checks.


miconda:
Thanks for reporting the bug and providing the patch!

PR #52, opened 2015-01-27
Geoip2 module

sokhapkin:
Implementation of geoip2 module based on the new Maxmind geoip2 database format.

### DIFF
--- a/modules/geoip2/Makefile
+++ b/modules/geoip2/Makefile
@@ -1,0 +1,15 @@
+# $Id$
+#
+# WARNING: do not run this directly, it should be run by the master Makefile
+
+include ../../Makefile.defs
+auto_gen=
+NAME=geoip2.so
+LIBS= -lmaxminddb
+
+DEFS+=-DKAMAILIO_MOD_INTERFACE
+
+SERLIBPATH=../../lib
+SER_LIBS+=$(SERLIBPATH)/kmi/kmi
+
+include ../../Makefile.modules
--- a/modules/geoip2/README
+++ b/modules/geoip2/README
@@ -1,0 +1,146 @@
+geoip2 Module
+
+Sergey Okhapkin
+
+   callwithus.com
+
+Edited by
+
+Daniel-Constantin Mierla
+
+   <miconda@gmail.com>
+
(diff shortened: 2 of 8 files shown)


miconda:
Couple of clarifications to be made, if you are aware of the answers:

* does the new geoip database/api work with both IPv4 and IPv6?
* would be there cases to use old geoip module and the new one? I see they depend on different libraries, I assume the new one is not available in old distros, but then, can be there situations when both libraries are available?

Right now, the new module exports same functions and variables as old geoip module. In the case both can be used, then the functions/vars should be renamed to reflect the new module name. The config needs to be changed anyhow to load the new module.

sokhapkin:
Geoip2 databases provide both v4 and v6 addresses lookup.
I see no use cases when both modules might be used, that's why I use the same 
exported function name to minimize changes in kamailio cfg script. The only 
change required is to load/set parameter of another module.

Maxmind pushes new developments to use geoip2 databases and API, the old API 
is still supported by maxmind but is obsoleted.

miconda:
It appears that some older versions (but still in use), have the old libs, so I think this new module should come have the variables and name of the functions changed, because the module name is also different, therefore the configuration file must be edited anyhow.

sokhapkin:
I renamed exported fumction to geoip2_match()

miconda:
Rename also the variable exported as well as the C functions and variables that are not static, otherwise it can result in conflicts if both geoip modules are used. Moreover, makes development harder because searching for a function/symbol (and related things such as auto completion, tags jumping) gives more results.

sokhapkin:
I made global geoip->geoip2 names change.

miconda:
I will commit your patches individually with you as author, but changing the commit messages to match project requirements.

Let me know if you want to maintain the module directly, we can grant you write access to the repository.

If yes, be sure for future commits you follow the guidelines from:

* http://www.kamailio.org/wiki/devel/git-commit-guidelines

sokhapkin:
I will issue pull request again if changes will be needed, no need for write access.

miconda:
I pushed the new module to kamailio repository.

PR #54, opened 2015-01-28
presence : fix send_notify_request

lazedo:
aux_body_processing after get get_p_notify_body
can we backport this commit to 4.2 ?

also, please delete branch after merge.


miconda:
I applied the patch manually and pushed it with you as author.

For the future: the message of the commit must be about the commit only, not with questions about back porting and remarks to delete branches -- those do not belong to a commit history. Make the commit message as you would push directly to master. All the other remarks belong to the discussion here in the pull request, so you can add them as comments in this thread/page.

If the issue is in 4.2 and you tested, then the commit can be backported.

PR #55, opened 2015-01-28
presence_dialoginfo : optionally send dummy dialog

lazedo:


lazedo:
this seems to also affect astra phones besides Bria

miconda:
You can merge it to master branch.

lazedo:
Hi Daniel,

can i backport to 4.2 ?
would like to fix astra / bria - blf issues

miconda:
First send a dedicated email on sr-dev with details about the issue and the fix, proposing to backport. It is good to get the eyes of other people on the changes and the opinions on the fix. I would like to avoid breaking in any way the expectations from a stable branch. If the opinions on sr-dev are for backporting, then I am fine as well.

kamailio-sync:
Hi List,

some sip clients like Bria and some sip phones like Astra
expect a <dialog-info><dialog></dialog></dialog-info>
on the NOTIFY message sent.

presence sends a NOTIFY after the SUBSCRIBE with NULL body if it doesn't find the presentity.

the patch already in master allows to configure the creation of a dummy dialog to be sent when presence doesn't find the presentity.

this behavior was only observed in dialog subscriptions so the patch was made against presence_dialoginfo.

the new param force_dummy_dialog defaults to 0 to preserve existing behavior.

does anyone have anything against backporting this to 4.2 ?

Best

miconda:
You should send a new email, with an explicit subject -- that is what I meant with a 'dedicated' email -- keeping discussion on this thread is not highlighting the purpose. This thread here is marked as closed.

Issue #48, opened 2015-01-20
multiple /tmp file vulnerabilities


linuxmaniac:
Reported by: Helmut Grohne <helmut@subdivi.de>

The kamailio package now installs /etc/kamailio/kamailio-basic.cfg which
can be selected via the CFGFILE= setting in /etc/default/kamailio. The
configuration contains:
```
modparam("mi_fifo", "fifo_name", "/tmp/kamailio_fifo")
```
This setting is insecure and may allow local users to elevate privileges
to the kamailio user.

The issue extends to kamailio-advanced.cfg. It seems that this is due to
an incomplete fix of #712083. Looking further, the state of /tmp file
vulnerabilities in kamailio looks worrisome. Most of the results of the
following command (to be executed in the kamailio source) are likely
vulnerable if executed:
```
grep '/tmp/[a-z0-9_.-]\+\(\$\$\)\?\([" ]\|$\)' -r .
```
Granted, some of the results are examples, documentation or obsolete.
But quite a few reach the default settings:

 * kamcmd defaults to connecting to unixs:/tmp/kamailio_ctl.
 * The kamailio build definitely is vulnerable as can be seen in
   utils/kamctl/Makefile.

More research clearly is required here.  Given these findings, the
security team may want to veto the inclusion of kamailio in a stable
release, which would be very unfortunate as kamailio is quite a unique
piece of software with little competitors in its field.

Helmut

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775681

miconda:
I think this can be fixed by setting user/group for fifo/ctl-socket files. Each of these modules have afferent parameter. Would that work?

linuxmaniac:
Changes I'm planning to do in the Debian packaging:

> kamcmd defaults to connecting to unixs:/tmp/kamailio_ctl.

I'm going to change the default of ctl module to /var/run/kamailio/kamailio_ctl so kamcmd will use it by default and I'm going to set explicitly the binrpc parameter on the etc/kamailio/*.cfg files  

> The kamailio build definitely is vulnerable as can be seen in utils/kamctl/Makefile.

I'm going to use basedir Makefile config instead of /tmp

oej:
On 24 Jan 2015, at 13:20, Victor Seva <notifications@github.com> wrote:

> Changes I'm planning to do in the Debian packaging:
> 
> kamcmd defaults to connecting to unixs:/tmp/kamailio_ctl.
> 
[...]

PR #56, opened 2015-01-28
utils/kamctl: use $(basedir) as tmp dir on building instead of insecure /tmp

linuxmaniac:
Reported by: Helmut Grohne <helmut@subdivi.de> See #48
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775681


miconda:
You can merge it.

PR #64, opened 2015-02-02
Coudot/rtpengine avp sdp

camilleoudot:
As proposed on the mailing list (http://lists.sip-router.org/pipermail/sr-dev/2014-December/026514.html), here is a patch to be able to retrieve the new SDP in an avp

### DIFF
--- a/modules/rtpengine/README
+++ b/modules/rtpengine/README
@@ -67,7 +67,7 @@ Richard Fuchs
               4.5. extra_id_pv (string)
               4.6. setid_avp (string)
               4.7. force_send_interface (string)
-              4.8. write_sdp_avp (string)
+              4.8. write_sdp_pv (string)
               4.9. rtp_inst_pvar (string)
 
         5. Functions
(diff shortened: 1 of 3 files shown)


kamailio-sync:
On Monday 02 February 2015 17:20:29 Camille Oudot wrote:
> As proposed on the mailing list (http://lists.sip-router.org/pipermail/sr-dev/2014-December/026514.html), here is a patch to be able to retrieve the new 
SDP in an avp
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/kamailio/kamailio/pull/64
> 
> -- Commit Summary --
> 
>   * modules/rtpengine: return the new SDP in an AVP
>   * rtpengine: README update

The patch is mixing shm_malloc and pkg_free.


Why is it limited to AVP's?

Maybe i misread, but it seems there is only one value written in the variable 
and it doesn't need to be stored in the transaction. So there seems to be no 
use for storing in an AVP and i think it could be extended to allow any 
writable PV.
-- 
Alex Hermann

camilleoudot:
Thanks for the review, there is a shm/pkg mixup indeed. Seems like avp_add() aready takes care of shm_malloc() anyway.

There is no also absolute need to restrict to AVP, i will add support for script $var() also.
-- 
Camille

camilleoudot:
Closing this pull request. I will post a new clean patch.

PR #66, opened 2015-02-04
usrloc fix

sokhapkin:
When multiple instances of kamailio share location table, kamailio on starup loads all records from the table and lately removes records which belongs to another kamailio instance(s), bringing the table out of sync with running instances.

### DIFF
--- a/Makefile.groups
+++ b/Makefile.groups
@@ -124,7 +124,7 @@ mod_list_perldeps=app_perl db_perlvdb
 mod_list_python=app_python
 
 # - modules depending on geoip library
-mod_list_geoip=geoip
+mod_list_geoip=geoip geoip2
 
 # - modules depending on sqlite library
 mod_list_sqlite=db_sqlite
--- a/Makefile.groups
+++ b/Makefile.groups
@@ -124,7 +124,10 @@ mod_list_perldeps=app_perl db_perlvdb
 mod_list_python=app_python
 
 # - modules depending on geoip library
-mod_list_geoip=geoip geoip2
+mod_list_geoip=geoip
+
+# - modules depending on new geoip library
+mod_list_geoip2=geoip2
 
 # - modules depending on sqlite library
 mod_list_sqlite=db_sqlite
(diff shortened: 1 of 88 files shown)


miconda:
The ignoring was made on purpose, allowing to handle the registrations even the socket is different. Think about the case when kamailio is running on a server changing the IP (eg, on a DSL link), being restarted when the provider changes the address.

In current form, the patch will not be accepted - solutions:

  * save server_id in usrloc and rely on that for loading only own records (it requires adding a new column to location database)
  * make a parameter to enable/disable this behavior

The parameter should be done even with server_id, because I think that two servers are configured to use the same database table when they want to share the location. If you don't want the two (or more) servers to share the location, then you can create many location tables with different name but same structure (and add a record to version table) and each server is using a different one.

As for the pull request, you have to re-sync your repository with the main one, because in pull requests you push now also the patches for geoip2 module, which is already part of the kamailio.

I am going to close this item here.

sokhapkin:
I strictly disagree with the purpose of the ignoring.

miconda:
Can you give more details? I think I presented the reasons and I know it is used that way. Also, I gave you suggestions on how to get what you want. The proposed patch is breaking existing used behavior, so you have to provide a patch that preserves it.

kamailio-sync:
On Wednesday 04 February 2015, Daniel-Constantin Mierla wrote:
>   * save server_id in usrloc and rely on that for loading only own records
> (it requires adding a new column to location database) 

What would be the use case of this? If you would want to restrict records to 
only one proxy, just use separate location tables. No need to clutter the 
module code for this.

-- 
Greetings,

Alex Hermann

linuxmaniac:
I just wanted to point that I have a WIP 4.2.X https://github.com/linuxmaniac/kamailio/tree/vseva/nathelper_socket_filter that we use (4.1.X) in order to share usrloc.location between proxies, using dbmode=3, splitting the work of natpinger between them filtering the users by socket.

So nathelper has a new module parameter ```filter_socket``` to control this behavior

Maybe we can get to common point to deal with different scenarios

sokhapkin:
What the parameter will do? I use db_mode 2.

linuxmaniac:
Please move the discussion to sr-users@lists.sip-router.org

Issue #37, opened 2015-01-17
call_control :: Depends on pv


oej:
Call_control depends on pv, but it's not documented in the README

 0(70806) ERROR: <core> [pvapi.c:790]: pv_parse_spec2(): error searching pvar "avp"
 0(70806) ERROR: <core> [pvapi.c:994]: pv_parse_spec2(): wrong char [s/115] in [$avp(s:can_uri)] at [5 (5)]
 0(70806) : call_control [call_control.c:1076]: mod_init(): invalid AVP specification for canonical_uri_avp: `$avp(s:can_uri)'
 0(70806) ERROR: <core> [sr_module.c:945]: init_mod(): Error while initializing module call_control

PR #67, opened 2015-02-04
add pv module as call_control dependence in documentation

linuxmaniac:
As @oej noticed in #37 pv is missing in the list of module dependences


oej:
Fine with me. Thank you!

PR #75, opened 2015-02-10
For dbmongo update, use an update document using the set update operator

mikomarrache:
...instead of replacing the entire document with partial data.


miconda:
Thanks. Patch applied manually with you as author. For future reference, when submitting a commit to kamailio, follow the guidelines from next link for the commit message:

Commit message is used for email notifications as well as changelogs, therefore is quite important in the ecosystem.

PR #78, opened 2015-02-11
presence : add min_expires param

lazedo:
some UAs send very low values for subscriptions causing unwanted flooding of the server with subscribe/notify messages
this new parameter allows to set a minimum value for the subscription.
default value is 0 to preserve existing behaviour

### DIFF
--- a/modules/presence/doc/presence_admin.xml
+++ b/modules/presence/doc/presence_admin.xml
@@ -322,7 +322,7 @@ modparam("presence", "expires_offset", 10)
        <section>
                <title><varname>max_expires</varname> (int)</title>
                <para>
-               The the maximum admissible expires value for PUBLISH/SUBSCRIBE
+               The maximum admissible expires value for PUBLISH/SUBSCRIBE
                message (in seconds).
                </para>
                <para>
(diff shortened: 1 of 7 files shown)


juha-h:
lazedo writes:

> some UAs send very low values for subscriptions causing unwanted
> flooding of the server with subscribe/notify messages
> this new parameter allows to set a minimum value for the subscription.

is that supported by the standards, i.e., can presence server tell the
ua the minimum allowed subscription expires time so that the ua can
adapt to that?

-- juha

lazedo:
Hi juha,

i don't think that its part of any standard.
what we observed is that UAs do respect the expires we send with
subscription=active
this acts like the max_expires params, if its out of bounds it corrects the
value and subscribes with that value in expires.
when it send the subscription=active, it will send the expires accepted and
saved in the server.

does that help ?

miconda:
I think this is covered by SIP RFC - a 423 reply with Min-Expires header has to be sent back.

* https://tools.ietf.org/html/rfc3261#section-20.23

I am not sure sending back 200ok with higher expires is RFC compliant, but if it helps with some user agents, it can be made optional.

lazedo:
thanks for pointing that Daniel.

i think we could add another parameter to set the action 
min_expires_action = 1 => force min_expires
min_expires_action = 2 => send 423 with Min-Expires header

is this acceptable ?

it is already optional, it acts only when min_expires > 0 and the default is zero.

oej:
Section 3.1.1. in RFC 6665 is the correct reference here. But it does point to RFC 3261.
"200-class responses to SUBSCRIBE requests also MUST contain an
   "Expires" header field.  The period of time in the response MAY be
   shorter but MUST NOT be longer than specified in the request."

lazedo:
well, the experience we had so far is that UAs do accept the Expires sent independently of the requested value.

 but to keep kamailio defaults rfc compliant, we can set the default action to 2 send 423.

miconda:
@lazedo that should cover all needs, keeping the option of RFC compliancy available. You can implement that.

miconda:
Thanks! Can be merged to master branch.

PR #79, opened 2015-02-11
Possible hash_entry,hash_id duplicate due to bad loading of dialogs at startup

mikomarrache:
At startup, assume d_table->entries[5].next_id is initialized at 25 (see function init_dlg_table in dlg_hash.c) and assume you have only one dialog in the database with hash_entry=5 and hash_id=25. Since the condition next_id < dlg->h_id is false (because next_id == dlg->h_id == 25), the next available hash_id for a future dialog that matches this entry will be 25. Therefore, there will be duplicated (hash_entry,hash_id) which will cause errors when the dialog is inserted in the database.


miconda:
Thanks. Patch applied manually with you as author. For future reference, when submitting a commit to kamailio, follow the guidelines from next link for the commit message:

Commit message is used for email notifications as well as changelogs, therefore is quite important in the ecosystem.

PR #84, opened 2015-02-13
registrar : add lookup_to_dset optional uri param as in lookup

lazedo:
lookup function has an optional uri param for looking up based on that param.
lookup_to_dset is used if we don't want to overwritte the Request-URI but there is no option to lookup based on a uri.

during tests i noticed that the append_branch was being called when i did a lookup_to_dset("location", "$fu") and kamailio exited because of that.

i made the append_branch to depend on the _mode ( 0 = lookup ) which seemed to fixed.

### DIFF
--- a/modules/registrar/lookup.c
+++ b/modules/registrar/lookup.c
@@ -399,22 +399,24 @@ int lookup_helper(struct sip_msg* _m, udomain_t* _d, str* _uri, int _mode)
 				path_dst.len = 0;
 			}
 
-			/* The same as for the first contact applies for branches 
-			 * regarding path vs. received. */
-			LM_DBG("instance is %.*s\n",
-			       ptr->instance.len, ptr->instance.s);
-			if (append_branch(_m, &ptr->c,
-					  path_dst.len?&path_dst:&ptr->received,
-					  path_dst.len?&path_str:0, ptr->q, ptr->cflags,
-					  ptr->sock,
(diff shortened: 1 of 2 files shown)


grumvalski:
I didn't have the time to test the changes, but it seems to me that making append_branch depend on the _mode=0, breaks what the lookup_to_dset function had been thought for, adding new branches to the dset so that new branches can be added to a transaction via t_append_branches (tsilo relies on those two functions to work). Maybe I don't understand what are you trying to achieve calling lookup_to_dset with a ruri, but what's the purpose of calling a lookup without rewriting the r-uri nor appending new branches?

miconda:
I second the fact that disabling the append_branch() execution is not clear. More details about what use cases is fixing might sort out the right solution.

Few other remarks:
* apparently the indentation was changed (whitespaces instead of tabs?)
* if there is a fix + new feature, split the commit in two, because the fix might need to be backported

lazedo:
i do agree about the append branch part, i didn't find a way (lack of knowledge) to make it work without it.
when using the uri param, $fu in this case,, kamailio exited. 
i think that there must be a better option instead of _mode.
help appreciated.

miconda:
If you get a crash, get the backtrace from the core file to see what is the problem. Probably the new parameter is not handled properly.

lazedo:
closing this, will make implementation of requirement in lookup registered(uri) function

PR #88, opened 2015-02-16
Cleanup CURL handle when the POST value can't be fetched from the PVAR

mikomarrache:


miconda:
Is the effective change the addition at line 121?

```
+		curl_easy_cleanup(curl);
```

The rest seems to be indentation.

For the pull request to be accepted automatically, write the commit message according to the guidelines at:

mikomarrache:
Yes, the effective change is the CURL cleanup.

miconda:
Patch applied with the fix, to be backported.

Issue #80, opened 2015-02-11
introduce RUN_DIR


linuxmaniac:
In order to get rid of any occurrence of ```/tmp``` in modules, introduce a ```RUN_DIR``` make option and every module should honor it

default could be ```/var/run/$(NAME)```

Related to #48

miconda:
I added few patch for Makefile to define RUN_DIR at compile time. The folder is also created during install. If no PREFIX is provided, then the path is /var/run/kamailio/, otherwise is $(PREFIX)/var/run/kamailio/

Now need to be propagated in the modules using /tmp hardcoded.

linuxmaniac:
This seems to be done for 4.3

PR #89, opened 2015-02-17
set defaults of fifo and ctl to /var/run/kamailio

linuxmaniac:
Following #48 and #80


miconda:
The patches are ok, but before merging, the 'make install' commands must be updated to change the paths in configs and kamctl based on PREFIX.

linuxmaniac:
As discussed on IRC, no changes needed on install-cfg rule

PR #90, opened 2015-02-17
nathelper - add received_ui pvar

lazedo:
allows comparison with persisted value in usrloc.

if($ulc(exp=>received) == $received_uri) {
  ....
}

### DIFF
--- a/modules/nathelper/nathelper.c
+++ b/modules/nathelper/nathelper.c
@@ -302,6 +302,7 @@ static int fixup_fix_sdp(void** param, int param_no);
 static int fixup_add_contact_alias(void** param, int param_no);
 static int add_rcv_param_f(struct sip_msg *, char *, char *);
 static int nh_sip_reply_received(sip_msg_t *msg);
+static int pv_get_received_uri(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
 
 static void nh_timer(unsigned int, void *);
 static int mod_init(void);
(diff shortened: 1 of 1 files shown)


miconda:
This looks like $su from pv module.

lazedo:
yes it does, but there is a difference. receive_uri returns the value as is persisted in usrloc,the proto is uppercase.
$su, $received_uri - sip:192.168.27.2:8305;transport=tcp - sip:192.168.27.2:8305;transport=TCP

miconda:
String comparison should be lower case.

Besides that, there are string transformation 'tolower' and 'toupper'.

It is no need for another variable. Eventually the two functions should be made to use a common function stored in core.

lazedo:
i agree, but not when is part of a select statement.
"select * from usrloc where received = $su" would not work.

miconda:
I am sure sql has functions for tolower or toupper or case insensitive match.

lazedo:
that depends on the engine, right ? the point is, the value is stored in mixed case. i'm trying to get a match of the value that is stored. 
nat_traversal also has a source_uri pv that matches what it internally calculates, nathelper should do the same.

lazedo:
...or we can patch nathelper to lower case what it puts in received_avp ? and also make sure $su returns lowercase ?

miconda:
I dont know about nat_traversal, perhaps is some quite old code.

I already said that the two functions should be made a wrapper to a common function stored in core. That will ensure consistency. I think transport in lowrr case is the common out there.

I can take care later today or tomorrow.

lazedo:
fair enough. thank you.

PR #91, opened 2015-02-17
registrar - additional features

lazedo:
- add xavp when registered returns success
- add reg_received_avp param for strict matching the contact
- add reg_contact_avp param for strict matching the contact
- add xavp to reg_fetch_contacts

modparam("usrloc", "xavp_contact", "ulattrs")
modparam("registrar", "reg_received_avp", "$avp(chk_origin)")
modparam("registrar", "reg_contact_avp", "$avp(chk_contact)")

using xavp with registered
   $avp(chk_origin) = $su;
   $avp(chk_contact) = $(ct{nameaddr.uri});
   if (registered("location","$fu")) {
       xlog("L_INFO", "$ci|log| checking $fu - $xavp(ulattrs=>account_id)");
  }

using xavp with reg_fetch_contacts
  if(reg_fetch_contacts("location", "$fu", "caller")) {
         xlog("L_INFO", "$ci|log| checking $fu - $(ulc(caller=>ulattrs=>account_id))");
   }

### DIFF
--- a/modules/registrar/api.c
+++ b/modules/registrar/api.c
@@ -105,7 +105,7 @@ int regapi_registered(struct sip_msg *msg, char *table)
 		LM_ERR("usrloc domain [%s] not found\n", table);
 		return -1;
 	}
-	return registered(msg, d, NULL);
+	return registered(msg, d, NULL, 0);
 }
 
 /**
--- a/modules/registrar/lookup.c
+++ b/modules/registrar/lookup.c
@@ -664,16 +664,18 @@ int registered(struct sip_msg* _m, udomain_t* _d, str* _uri)
 		LM_DBG("searching with initial match flags (%d,%d)\n", match_flags, match_return_flags);
 		if(reg_xavp_cfg.s!=NULL) {
 
-			if( (vavp = xavp_get_child_with_ival(&reg_xavp_cfg, &match_flags_name)) != NULL
-					&& vavp->val.v.s.len > 0) {
-				match_flags = vavp->val.v.i;
-				LM_DBG("match flags set to %d\n", match_flags);
-			}
+			if(match_search_flags == 1) {
+				if( (vavp = xavp_get_child_with_ival(&reg_xavp_cfg, &match_flags_name)) != NULL
+						&& vavp->val.v.s.len > 0) {
+					match_flags = vavp->val.v.i;
+					LM_DBG("match flags set to %d\n", match_flags);
+				}
 
-			if( (vavp = xavp_get_child_with_ival(&reg_xavp_cfg, &match_return_flags_name)) != NULL
-					&& vavp->val.v.s.len > 0) {
-				match_return_flags = vavp->val.v.i;
-				LM_DBG("match return flags set to %d\n", match_return_flags);
+				if( (vavp = xavp_get_child_with_ival(&reg_xavp_cfg, &match_return_flags_name)) != NULL
+						&& vavp->val.v.s.len > 0) {
+					match_return_flags = vavp->val.v.i;
+					LM_DBG("match return flags set to %d\n", match_return_flags);
+				}
 			}
 
 			if((match_flags & 1)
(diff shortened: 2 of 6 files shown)


miconda:
Some of the patches are quite big and there is little description of the purpose and benefits.  But they affect a really important module, where performances really matter. Just reading the code is not easy to guess your needs and impact on existing behavior.

Send an email to sr-dev providing more details for each of the patches.

As a recommendation, before writing a lot of code changing an important module, it is better to discuss first on sr-dev, to come up with a proper solution fitting the best of common use cases. Otherwise, you may lose time writing code that will not be accepted because it affects performances or the existing behavior that is commonly needed.

miconda:
You misunderstood my proposal. I wanted a new parameter to the function: registered(talble, uri, flags), because I want to avoid searching through xavps if not needed. The parameter is read only if provided. 

If I got right the patches, you search for the flags in xavp list.

lazedo:
i don't think i did, the flags are in params.

i added the extra search for "override" to give more flexibility.
i can put that flexibility param dependent.

miconda:
I don't see any new parameter to registered(...) function. Do I miss something?

Perhaps I can change (sometime during next days) the current call-id matching to using a new flags param to registered() function and xavp. Then you can add from there what you need extra.

lazedo:
there are two params
match_flags as you proposed
1 - callid
2 - received
3 - contact

the registered() function check that flag, checks the values in xavp_cfg
and assigns the value to a temp vars, that are then compared against the
value in the record.

match_return_flags controls if the record xavp is added to the transaction.

lazedo:
i added a 3rd param to control if we search for "override" values for flags.
default is 0, don't search

miconda:
We are not on the same meaning of terms here. I refer to function parameter, not module parameter. I want to do:
```
registered("location", "$fu", "1")
```
in order to do registered() with call-id match. The new parameter can be int or pv (see fixup_igp_null() and fixup_get_ivalue() helper functions)

lazedo:
@miconda i believe this is it ?

lazedo:
@miconda can you check if this is acceptable ? thanks.

miconda:
A bit hard to follow, because now here are patches to patches to patches. Would be good to have a single patch against the master in kamailio for review.

One thing I noticed, that some flags are still a module parameter. Wouldn't it be more flexible to be part of the function parameter, different flag values (eventually starting at 8 or 16)?

lazedo:
@miconda hi, the remaining flag in the module is to control the setting of the contact xavp on return from a match. i left it here because we also want the behavior to apply to reg_fetch_contacts, is that ok ?
i can make a simple patch against master.

miconda:
But reg_fetch_contacts() is already returning the contacts in an variable structure - $ulc. I don't see a reason to add to $ulc as well as to an xavp.

Could be better to have a new function reg_fetch_contact_to_xavp() that will get the attributes in xavp instead of $ulc.

I may miss some point, in that case, provide some details when you find useful for reg_fetch_contacts() to return some data twice.

lazedo:
@miconda the $ulc structure doesn't have the contact xavp (it copies all elements but the xavp), the purpose of the flag is to add it to the $ulc structure.

miconda:
But that was a patch adding to $ulc(...), not to an $xavp(), right? Maybe seeing all the changes as a single patch will shed light on these aspects.

lazedo:
right, it doesn't add it to xavp it adds it to $ulc so if we have
modparam("usrloc", "xavp_contact", "ulattrs")
we can use the $qlc returned like this (if flag is set).
if(!reg_fetch_contacts("location", "$fu", "caller")) {
      xlog("L_INFO", "$ci|log| checking $fu - $ulc(caller=>ulattrs=>account_id)");
}

i'll put everything in one commit and create a new PR

lazedo:
https://github.com/kamailio/kamailio/pull/101

PR #92, opened 2015-02-18
db_text : missing OP_NEQ

lazedo:
this affects the presence package when it queries for active_watchers


miconda:
Can be merged and backported if needed.

PR #94, opened 2015-02-19
presence - match sip uris using alternative function

lazedo:
added param sip_uri_match
0 - case sensitive (default)
1 - case insensitive

can be extended to a more compliant version, (sensitive user, insensitive domain). 
the parameter sets the function to call to match uris.

### DIFF
--- a/modules/presence/README
+++ b/modules/presence/README
@@ -45,22 +45,26 @@ Juha Heinanen
               3.10. to_tag_pref (str)
               3.11. expires_offset (int)
               3.12. max_expires (int)
-              3.13. server_address (str)
-              3.14. subs_db_mode (int)
-              3.15. publ_cache (int)
-              3.16. subs_htable_size (int)
-              3.17. pres_htable_size (int)
-              3.18. send_fast_notify (int)
-              3.19. enable_sphere_check (int)
-              3.20. timeout_rm_subs (int)
-              3.21. fetch_rows (integer)
-              3.22. db_table_lock_type (integer)
-              3.23. local_log_level (int)
-              3.24. subs_remove_match (int)
+              3.13. min_expires (int)
+              3.14. min_expires_action (int)
+              3.15. server_address (str)
+              3.16. subs_db_mode (int)
+              3.17. publ_cache (int)
+              3.18. subs_htable_size (int)
+              3.19. pres_htable_size (int)
+              3.20. send_fast_notify (int)
+              3.21. enable_sphere_check (int)
+              3.22. timeout_rm_subs (int)
+              3.23. fetch_rows (integer)
+              3.24. db_table_lock_type (integer)
+              3.25. local_log_level (int)
+              3.26. subs_remove_match (int)
+              3.27. xavp_cfg (str)
+              3.28. retrieve_order (int)
 
         4. Functions
 
-              4.1. handle_publish(char* sender_uri)
+              4.1. handle_publish([sender_uri])
               4.2. handle_subscribe([watcher_uri])
               4.3. pres_auth_status(watcher_uri, presentity_uri)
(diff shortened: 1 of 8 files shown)


oej:
Are we still missing docs here?

Note that the param in registrar has another name. Since this is not about the full SIP uri- i hope we always are case insensitive for domains - maybe a better name would be sip_username_match

kamailio-sync:
yes, i know the docs are missing.

when is something other then a fix, i will gladly add the docs if the request is accepted.

"i hope we always are case insensitive for domains"

this implementation acts on the full sip uri. it is open to add a 3rd option that will take the full uri and matches sensitive on the username and insensitive on the domain.

before this it acts on the full sip uri as case sensitive, i give it the alternative and left it open enough (i think) to implement full RFC compliance.


oej:
If the module are comparing domains with case sensitivity today, that's a bug that needs to be fixed.

miconda:
There are helper C functions for comparing the URI, used by:

  * http://kamailio.org/docs/modules/devel/modules/siputils.html#siputils.f.cmp_uri

Not sure how this strict uri matching can be made to work with db only mode... but I guess that is a compromise.

juha-h:
lazedo writes:

> added param sip_uri_match
> 0 - case sensitive (default)
> 1 - case insensitive

should it rather be called sip_uri_user_match?

i still don't feel comfortable in starting to bloat kamailio with
non-standard behavior.  why don't you canonize uri user in config
script?

-- juha

juha-h:
Olle E. Johansson writes:

> If the module are comparing domains with case sensitivity today,
> that's a bug that needs to be fixed.

exactly. i also pointed this out earlier.  if implemented, the new
function must operate on uri user only, but i would not prefer to add
this functionality at all as i mentioned in previous message.

-- juha

lazedo:
Olle,

it acts on the sip uri not the user part of the sip uri.

juha-h:
lazedo writes:

> it acts on the sip uri not the user part of the sip uri.

and that exactly is the problem as we have tried to tell.

-- juha

miconda:
@juha-h - because the comparison is done on different URIs (sender, target, contact), I am not sure that would be easy to canonise in config.

Looking at the patch I think it is in the right direction to get this sorted out according to RFC but also for other cases if needed.

Now is case sensitive, perhaps the default should be made to use the internal uri comparison functions.

The patch makes it easy to decide on what kind of comparison to do. There will be no side effect on performances if we go for uri comparison function or allow other options via mod param.

@lazedo - can you make the patch for the docs of the module to reflect the new parameter? I am considering to merge at some point, followed by adding the option for RFC uri comparison, eventually making that default option.

lazedo:
@miconda sure. one question, shuld we also <abstract the core_hash used in the module to work the same way as the match function ? abstracting to core_hash / core_case_hash ?

miconda:
What do you mean by abstracting the core_hash / core_case_hash ? To have a function inside presence module as a wrapper?

lazedo:
yes, working the same as the uri comparison, using function pointer.

miconda:
I think here we can go with case insensitive hashing, the function is not slower and covers properly all cases. Adding a new parameter will make it more complex, exposing something that is pure technical and related only to the internals of the module.

lazedo:
ok, i wasn't thinking in adding a new param and use the previous one ( sip_uri_match ) to either call core_hash or core_case_hash.
are you proposing to change the core_hash with core_case_hash independently from sip_uri_match param?

miconda:
Yes, use core_case_hash() -- I think it is the best option: not adding any overhead and avoids extra config complexity.

lazedo:
closing here because of rebase that didn't go well.

https://github.com/kamailio/kamailio/pull/116

PR #95, opened 2015-02-19
core: add support for per-TCP connection lifetime

camilleoudot:
A per-TCP connection `lifetime` parameter is needed, to add a new functionnality in the tcpops module: the ability to dynamically modify the connection lifetime on a connection (e.g. raise the lifetime of a successfully authenticated connection).

This requires modifications in the core:
  - adding a new field in the `tcp_connection` structure
  - changing the way the `timeout` value is updated, using this new `lifetime` member instead of the core conf


miconda:
Can be merged.

PR #96, opened 2015-02-24
fix - dummy dialog rfc4235

lazedo:
grandstream validates the dialog and according to rfc4235 the id element is mandatory


linuxmaniac:
please @lazedo follow the https://www.kamailio.org/wiki/devel/git-commit-guidelines

Something like:

presence_dialoginfo: obey rfc4235 adding id element for dummy dialog

kamailio-sync:
you're right, missed the "presence_dialoginfo : "

PR #98, opened 2015-02-24
modules/sipt: fix bounds check on ACM

tsearle:


linuxmaniac:
can you please @tsearle  git rebase from origin/master and make one commit only?

miconda:
@tsearle - sipt is your module, you can just push the code to the kamailio repository directly, you have commit access.

Of course, if you want someone else to review the code for getting the safety check of another eye, you can go via pull requests, but you should somehow say that in the message of the pull request.

PR #99, opened 2015-02-24
presence_dialoginfo: rfc4235 in dummy dialog

lazedo:
according to rfc4235 the id element is mandatory.
granstream phones are affected by this.


miconda:
You can merge it.

PR #100, opened 2015-02-25
uac Adding counters for remote registrations

oej:
Adding three counters. 
* Total number of in-memory registrations
* Total number of successful registrations
* Total number of disabled registrations

This is for review. Will add documentation before commit.


miconda:
It can be documented and merged.

PR #101, opened 2015-02-25
registrar - add registered extra parameter

lazedo:
adds 3rd parameter to registered to optionally restrict the contacts

adds module parameter to optionally
 add contact xavp on successful match when calling registered
 add contact xavp to the $ulc structure

This is for review. Will add documentation before commit.

### DIFF
--- a/modules/registrar/api.c
+++ b/modules/registrar/api.c
@@ -103,7 +103,7 @@ int regapi_registered(struct sip_msg *msg, char *table)
 		LM_ERR("usrloc domain [%s] not found\n", table);
 		return -1;
 	}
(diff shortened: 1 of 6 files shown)


oej:
Docs?

miconda:
Add the patch updating the docs for the new changes: the new parameter and the the new xavp attributes for matching.

Once having the docs, I will apply the patch, but split in few parts to track the updates on different components.

The fix for iterating to get to the contact by index for $ulc() is going to be applied right now.

lazedo:
https://github.com/kamailio/kamailio/pull/115

PR #102, opened 2015-02-25
usrloc - fix contacts attrs not deleted

lazedo:
when the contact is deleted using db_delete_ucontact_ruid (default)
the uldb_attrs_delete is not called leaving the location_attrs with unwanted records

created uldb_attrs_delete_ruid to be called by db_delete_ucontact_ruid
redirected uldb_attrs_delete to uldb_attrs_delete_ruid
also changed the places inside contact where uldb_attrs_delete was called to call new uldb_attrs_delete_ruid function

### DIFF
--- a/modules/usrloc/ucontact.c
+++ b/modules/usrloc/ucontact.c
@@ -76,6 +76,8 @@ void ucontact_xavp_store(ucontact_t *_c)
 }
 #endif
 
+int uldb_delete_attrs_ruid(str* _dname, str *_ruid);
+
 /*!
  * \brief Create a new contact structure
  * \param _dom domain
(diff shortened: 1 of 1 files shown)


miconda:
I think the patch is breaking db_delete_ucontact_addr() because it removes the sip domain condition for that case.

I am talking by the next chunk in the patch:

```
@@ -1389,27 +1381,7 @@ int db_delete_ucontact_addr(ucontact_t* _c)
```

Also, use the formatting as suggested in the commit guide lines, it keeps the changelog consistent and coherent:

Respectively:

```
module_name: short description of the patch
[EMPTY LINE]
more description of the patch
```

lazedo:
isn't ruid unique per contact ?

not sure what i missed in the commit description, was it [EMPTY LINE] ?

lazedo:
@miconda please check

miconda:
The first patch seemed to break the deletion of the contact record, as the domain was no longer taken in consideration. The deletion of the contact record is done after the deletion of attrs inside db_delete_ucontact_addr().

The formatting was at least for the first line:

```
module_name: ...
```

I will check the new patch when i get a chance.

miconda:
Last commit can be merged.

Can you do it only for that one, without the previous ones and the revert? This should be backported at some point and it is good to have only one commit.

miconda:
Btw, if merging not working for a single patch, then just apply it manually to master -- you can get it from:

* https://github.com/kamailio/kamailio/commit/09c51685af36071aad6054d84571c260be2ba9e2.diff

lazedo:
merged manually

PR #105, opened 2015-03-04
Fixed path ss -> ssl

SebastianThorn:


oej:
Thank you for spotting this.

SebastianThorn:
Someone thought we should offer TLS for our sip-devices.

PR #109, opened 2015-03-11
xmlrpc: Improve doc

eloycoto:
Hi, 

I don't want to commit directly to other modules ;-) 

- Fix broken links into the xmlrpc doc. 
- Add xmlrpc_test2.py into the doc. 
- Add a small comment into the xmlrpc_test2.py. (Easy to understand how xmlrpc works.)

Regards.

### DIFF
--- a/modules/xmlrpc/doc/xmlrpc.xml
+++ b/modules/xmlrpc/doc/xmlrpc.xml
@@ -31,7 +31,7 @@
 	<section id="xmlrpc.design">
 	<title>Design Goals</title>
 	<para>
-	    
+
 	</para>
 	<itemizedlist>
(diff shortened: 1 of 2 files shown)


miconda:
You can merge it.

eloycoto:
Hi, 
1- rpm is the link from the web. So I copy&paste. (No idea if another domain is created.)
2- Yep, my fault. I'll fix ;-)

miconda:
rpm.kamailio.org was the used server by Jan when preparing the shut down old sip-router.org, to host the automatic builds of docs from xml. I made the alias build.kamailio.org, but maybe the web config was not updated.

PR #118, opened 2015-03-30
presence: add $subs pv for subscription access

lazedo:
after handle_subscribe there is no safe way in script to obtain the presently
because re-subscription RURI is ip of kamailio and To-Uri may not contain
the presently.

$subs(uri) will allow access to this value

added as a pv to allow extension to get more properties from subscription

### DIFF
--- a/modules/presence/doc/presence_admin.xml
+++ b/modules/presence/doc/presence_admin.xml
@@ -1130,6 +1130,40 @@ pres_update_watchers("sip:test@kamailio.org", "presence");
 </section>
 
 <section>
+	<title>Pseudo Variables</title>
+		
+		<section>
+			<title><varname>$subs(attr)</varname></title>
+			<para>
+				Access the attributes of handled subscription. 
+				It must be used after a call of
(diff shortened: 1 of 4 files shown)


miconda:
This can be merged.

Having references to pseudo-variables in there README/docs of the module is ok, but they must be documented in the wiki, where is the main reference documentation for variables. I added it for this pull request at:

  * http://www.kamailio.org/wiki/cookbooks/devel/pseudovariables?&#presence_module_pseudo-variables

lazedo:
@miconda thank you. for future reference, what is the proper flow to issue a pull request for documenting pseudo-variables ?

miconda:
The wiki is open for anyone, not only for developers, as we wanted the community to contribute with more examples, extra text, etc. without waiting for devs.

It requires to make an account on the wiki, given that we faced a lot of spambots in the past. But then, just edit the pages and add what you want/need. If there is something to discuss/clarify then email to sr-dev, otherwise nothing required.

PR #119, opened 2015-04-01
srv_query () function

rdboisvert:
Our team needed a new function to find SIP services using DNS SRV records but we couldn't find a Kamailio module that supported this.  After investigation we felt that the new function would fit better as an extension of the ipops module instead of creating a whole new one.

Please consider these changes to the ipops module.

Thank you,
Bob

### DIFF
--- a/modules/ipops/doc/ipops_admin.xml
+++ b/modules/ipops/doc/ipops_admin.xml
@@ -680,6 +680,91 @@ if(dns_query("test.com", "xyz"))
 
     </section>
 
+    <section id="ipops.f.srv_query">
+      <title>
+        <function moreinfo="none">srv_query(srvcname, pvid)</function>
+      </title>
+
+      <para>
+Queries DNS SRV records to resolve a service/protocol name into a list of priorities, weights, ports, and targets sorted by priority and weight as outlined in <ulink url="http://tools.ietf.org/html/rfc2782">RFC 2782</ulink>.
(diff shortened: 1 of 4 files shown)


oej:
Thank you, this can be useful in some cases.

You know that DNS SRV and NAPTR support is in the core, right?

Sorting by weight is not in RFC 2782. You sort by priority and get a set of entries for each priority. Then you select by weight - but the result should be one entry, not a sorted list by weight. Weight is for load balancing, not sorting and failover. Priority is for failover.

miconda:
This is targeting usage in config, in addition to dns_query() from same ipops module and could be useful indeed. Maybe srv_query() can be enhanced to specify (e.g., via a parameter) what to return - the first entry (only one) or all entries. I find all entries useful, as there can be a loop in config to check all of them and see what is supported on the other side.

I think the sorting needs to be fixed according to rfc on priority.

Merging should not be done automatically here -- I can do it manually, because:

  * commit messages are not reflecting what is inside the patches
  * indentation style must match the rest of the code in the module

rdboisvert:
Sorry if my documentation gave the wrong impression.  Sorting is fully compliant with RFC 2782.  Sorting is by priority, and weight is "arranged" per the algorithm outlined in the RFC using a random number.

miconda:
Pushed the patches manually with commit messages and indentation.

PR #122, opened 2015-04-13
Change log level from WARN to INFO to avoid

mikomarrache:
polluting the logs.


miconda:
Merged manually with a commit message in the recommended format and explaining the change.

PR #123, opened 2015-04-13
usrloc: change level of log

mikomarrache:
When a non local socket is found, the entry is not ignored and the message should be for debugging purpose only.


miconda:
Changing the log level is ok, but not changing the message, because is saying that the socket is ignored, helping in troubleshooting if one is not understanding why a socket is not used when it should (maybe the code on detecting the local socket gets broken at some point).

miconda:
Manual push of a commit based on above comment.

PR #125, opened 2015-04-14
acc: use appropriate types for core values

mikomarrache:
Use appropriate types for core values:
  -> start and end as dates (instead of strings)
  -> duration as double (instead of string)

### DIFF
--- a/lib/srdb1/schema/acc_cdrs.xml
+++ b/lib/srdb1/schema/acc_cdrs.xml
@@ -9,7 +9,7 @@
 
 <table id="acc_cdrs" xmlns:db="http://docbook.org/ns/docbook">
     <name>acc_cdrs</name>
-    <version>1</version>
+    <version>2</version>
     <type db="mysql">&MYSQL_TABLE_TYPE;</type>
     <description>
         <db:para>This table is used by the ACC module to report on CDRs relying on dialog callbacks. More information is available at: &KAMAILIO_MOD_DOC;acc.html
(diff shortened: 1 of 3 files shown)


linuxmaniac:
@mikomarrache just fix, amend  and push using force to your branch.

mikomarrache:
@linuxmaniac I fixed the error in the branch.
What do you mean by amend and push using force?

linuxmaniac:
I meant If you wanted to fix the original commit without having to create another one

miconda:
You can also push a new patch to fix the issue in your branch and should appear in this pull request. Then it can be merge, but also the schema has to be changed and version increased for acc_cdrs table, via lib/srdb1/schema/acc_cdrs.xml. To generate the new db/sql creation scripts, run 'make dbschema'.

mikomarrache:
I changed lib/srdb1/schema/acc_cdrs.xml as you asked and also included the fix.

I'm just a very beginner with Github, should I check out the branch, run the command and then commit?

Can you please provide me the right commands to checkout and commit?

Thanks
Mickael

miconda:
Well, you did what I said by pushing to your branch, all the patches appeared in the pull request.

miconda:
Thanks! The pull request will be merged.

linuxmaniac:
And what about upgrades from 4.2 to 4.3??
Is there any way to migrate the old values?
What is the upgrade procedure here?

miconda:
Maybe mysql can do conversion when running column change command:

```
ALTER TABLE acc_cdrs CHANGE COLUMN start_time ...
```

If not, we have to advise to backup the data, drop and recreate the table.

Issue #127, opened 2015-04-15
dispatcher "socket" feature crashes kamailio


neuhaus:
Using the new "socket" feature of the dispatcher module causes a segmentation fault.

neuhaus:
Pull request #128 fixes this bug.

miconda:
Thanks! Pushed the patch manually to have the commit message following the recommended guidelines -- allows us to get a coherent changelog for releases:

  * http://www.kamailio.org/wiki/devel/git-commit-guidelines#commit_message

PR #128, opened 2015-04-15
Fix socket feature crash bug in dispatcher module (issue #127)

neuhaus:
Fixes issue #127


miconda:
Thanks! Pushed the patch manually to have the commit message following the recommended guidelines -- allows us to get a coherent changelog for releases:

PR #129, opened 2015-04-15
dialog: add new flag for expired dialogs

mikomarrache:
The new flag will be accessible using $dlg(dflags).


mikomarrache:
Currently, I'm checking if the flag is set as follows:

if($dlg(dflags)&(1<<10)) {
...
}

Is there a cleaner solution to do it?

miconda:
Thanks, it will be merged.

PR #132, opened 2015-04-21
ERLANG module

seudin:
This module provides interact with Erlang node. The module allows sending,
receiving Erlang messages and RPC calls between each other.

### DIFF
--- a/modules/erlang/Makefile
+++ b/modules/erlang/Makefile
@@ -1,0 +1,18 @@
+# 
+# WARNING: do not run this directly, it should be run by the master Makefile
+
+include ../../Makefile.defs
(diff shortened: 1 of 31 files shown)


lemenkov:
Subscribing on this :-)

miconda:
Looks ok at the first sight. It will be merged to catch development freeze for v4.3 and a new modules group will be created for it because it has dependency for building.

You will get write access to kamailio repository so you can maintain it here in the future.

During the testing period we can refine it, if needed.

miconda:
Merged. Thank you for contributing the module!

seudin:
Thank you too.

PR #136, opened 2015-04-22
db_text : simple raw support for select

lazedo:


miconda:
How is this supposed to be used? The raw query is only the database table name?

lazedo:
yes

miconda:
Then what is the plan for the extensions to add selection of columns and where expresions? Going to build a custom query language?

lazedo:
not a custom, just a minimal sql by parsing the const str* _s parameter

SELECT FIELDS FROM TABLE WHERE CONDITIONS
UPDATE TABLE SET FIELDS=VALUES WHERE CONDITIONS
DELETE FROM TABLE WHERE CONDITIONS

1) FIELDS  => [FIELD, ...]
2) CONDITIONS [FIELD=VALUE, ...]
3) SET reuse (2)

miconda:
Then I think it is better to use directly:

```
select * from tablename
```

instead current implementation, instead of just giving the table name. It is easy the check it starts with 'select * ' and then extract the table name as last token.

Also, the documentation of db_text will need to be extended with a new section about 'Raw Query' where to present what is implemented.

lazedo:
sure, will do in the days to come just wanted to make sure this is useful for 4.3

miconda:
We are freezing the development for 4.3 at the end of today.

PR #139, opened 2015-04-23
integrate TLSF for pkg mem

camilleoudot:
following patches add TLSF support for pkg memory, and easier TLSF selection at compile time


miconda:
It can be merged, this is work to get the new memory manager coherent in behaviour with the existing ones.

PR #142, opened 2015-04-27
Added implementation for force_send_interface to also support IPv6.

smititelu:
Only IPv4 was supported before. Added socket support for IPv6 (link locals included). Also updated documentation.

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -102,6 +102,7 @@ EXTRAVERSION = -pre0
 # memory manager switcher
 # 0 - f_malloc (fast malloc)
 # 1 - q_malloc (quick malloc)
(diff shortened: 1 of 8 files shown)


miconda:
Please prefix the first line of the commit message with the name of the component that is affected by the patch. See the guidelines at:

It will be easier to track what components are affected in the changelog as well as easier to spot and react timely by the developer in charge of the component.

In this case is rtpengine and @rfuchs should review the patches.

From my point of view they are fixes, given that only ipv4 was supported, so that needs to work for ipv6 as well, so they can be merged in master.

smititelu:
Thank you for the tips.

miconda:
Going to merge this in order to get proper testing during the freezing period.

PR #146, opened 2015-05-04
db_text : add db_raw_query support

lazedo:
example
. `select watcher_username, called from active_watchers where expires <> -1`

current limitations

. only select statements
. where clause always uses "and"


miconda:
Merging this as fixing the initial patch where the raw query was table name (not like a valid query).

PR #151, opened 2015-05-07
htable: Add missing ht_slot_unlock in ht_cell_value_add

sn4kebite:
Adds a missing call to ht_slot_unlock when the init value is not an integer. This would cause processes to block on ht_slot_lock leading to a deadlock.


miconda:
Thanks for the patch!

PR #152, opened 2015-05-07
added RURI-pattern matching and priority

eschmidbauer:
Hello,
This pull request adds 2 features. 
--Prioritize the trusted records (highest to lowest priority)
--Match from-uri AND request-uri, if either is blank only one is matched.
Please review the pull request and let me know if any other changes are needed.
Thank you.
Emmanuel

### DIFF
--- a/lib/srdb1/schema/trusted.xml
+++ b/lib/srdb1/schema/trusted.xml
@@ -51,11 +51,27 @@
     </column>
 
     <column>
+        <name>ruri_pattern</name>
+        <type>string</type>
+        <size>&id_len;</size>
+        <default><null/></default>
+        <null/>
+        <description>Regular expression matches Request URI of request.</description>
+    </column>
+
+    <column>
         <name>tag</name>
         <type>string</type>
         <size>64</size>
         <null/>
         <description>Tag</description>
+    </column>
+
+    <column>
+        <name>priority</name>
+        <type>int</type>
+        <default>0</default>
+        <description>Priority of rule.</description>
     </column>
 
     <index>
--- a/modules/permissions/doc/permissions.xml
+++ b/modules/permissions/doc/permissions.xml
@@ -41,6 +41,13 @@
 			<email>jh@tutpro.com</email>
 		</address>
 		</editor>
+		<editor>
+		<firstname>Emmanuel</firstname>
+		<surname>Schmidbauer</surname>
+		<address>
+			<email>eschmidbauer@voipxswitch.com</email>
+		</address>
+		</editor>
 	</authorgroup>
 	<copyright>
 		<year>2003</year>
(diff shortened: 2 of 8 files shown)


eschmidbauer:
Hello,
Any chance this can be reviewed before the feature request window closes?
Thanks,
Emmanuel

miconda:
Master branch will be frozen again after at least 6 months, plenty of time to get new features in v4.4.

The patch got big because you changed the order of existing columns, not adding new column as last one, but inserting in between, so it was not trivial to analyze quickly when I tried a while ago as I found a bit of spare time.

Most likely I can allocate some time next week, as v4.3 will be out this week.

eschmidbauer:
Thank you for the response.
Yes, I did re-order the fields to make it more comprehensive in the long run.
I've been running this patch on my test machine for several weeks without any issues.
I hope you will consider reviewing and testing soon as we want to put this in production.
Thanks,
Emmanuel

miconda:
Some comments:

  * loading ordered on priority is using an SQL extension, making the module to fail when used with db_text, db_mongodb or other no-sql backends. This has to be reworked to get rid of the ordering extension specific to sql or made a configuration file option (module parameter)
  * is it a reason to restrict the matching only to r-uri parts of user@domain? May be useful to match also on transport or other parameters. I would live the matching against the full value of the r-uri inside the code. In the config one can set the $ru to whatever it likes to match
  * the code around the printing of sort_order is looking weird, as you add sort_order.len (which is set before to 0) to the buffer -- it is harmless but doesn't look good when someone is checking the code

miconda:
Add the changes to the db schema -- you have to update the xml file in lib/srdb1/schema/.

miconda:
And extend the documentation of the module with the appropriate description and new parameters -- changes have to be done to xml files in modules/permissions/doc/

eschmidbauer:
Hello, 
made changes as you requested.
Please see latest commit.
Thanks,
Emmanuel

miconda:
The commit message is not relevant for the all changes done by the patch, but reflects the last operation you did as per comments in this discussion. It is not suitable for the log of source code changes history.

I will apply the patch manually this time if all ok with the review. But for the future be sure you follow the guidelines for commit messages from:

Note also that you can have many commits in the same pull request. No need to rework a new patch always, you can add another patch to complete the changes done with previous patches.

eschmidbauer:
Hello,
You can download the full git diff (it applies clean to master branch) here http://paste.debian.net/download/231826
Thank you.
Emmanuel

miconda:
I applied the patch manually, split in few commits, per components.

The diff for each commit is available also on github, at:

  * https://github.com/eschmidbauer/kamailio/commit/72259b0a9b1f0f9d96e12e4bfa11ed8dbcf5dfcf.diff

Add .diff to the commit path.

I fixed the upgrade of table version for trusted.

You have to make a new pull request to update the docs regarding the priority, to specify what kind of value represents higher priority.

PR #153, opened 2015-05-07
pv: Fixed assigning $null to $vn(name) pvs

sn4kebite:
Pass NULL as value to set_var_value when assigning a NULL or PV_VAL_NULL value to a pv with the VAR_TYPE_NULL flag set.

Previously assigning $null didn't work as expected:

`$vn(foo) = $null` would result in `$vn(foo)` being assigned `0` (this can be verified by either logging the value or testing with an if statement).


miconda:
Thanks!

PR #154, opened 2015-05-09
modules/nathelper: don't include enclosing brackets in contact uri in set_contact_alias

grumvalski:
nathelper function set_contact_alias add brackets around contact's uri if the original one hadn't. This results in an invalid sip uri in the contact structure, which cause problems to other modules' functions when they try to parse the contact uri (e.g. save_pending() from ims_registrar_usrloc). This simple patch fix the issue simply mving the uri's boundaries.


miconda:
Thanks!

grumvalski:
Hi Daniel, thank you for merging. Should I backport it to stable branches?

miconda:
Yes, you can backport.

PR #161, opened 2015-05-13
Add domain to connection test query so it works with Cassandra

aphistic:
This adds the 'domain' column to the testdb_udomain db connection test in the usrloc module.  When using Cassandra 2 as the database module (a module I'm working on that I will be submitting a PR for once it's stable) doing a query by only username can cause issues if you create your Cassandra partition key as username and domain.

This shouldn't affect the query functionality in general as it's only doing a query to test the connection to the database.

### DIFF
--- a/modules/usrloc/udomain.c
+++ b/modules/usrloc/udomain.c
@@ -901,8 +901,13 @@ int db_timer_udomain(udomain_t* _d)
  */
 int testdb_udomain(db1_con_t* con, udomain_t* d)
 {
-	db_key_t key[2], col[2];
-	db_val_t val[2];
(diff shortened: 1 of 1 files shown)


miconda:
Thanks, I think it should be integrated but with the condition on domain based on **use_domain** parameter, I guess you have that set for your testing, right?

A second patch can be added to the pull request.

aphistic:
Ah yes, good point. I do have that set right now.  I'll get that updated.

miconda:
Reworked the patch because the old versions of C don't allow declarations of arrays with size specified as variable. It is now in master branch.

PR #162, opened 2015-05-14
app_lua: Fix memory leak in sr.modf

que273:
- Free pkg memory allocated to store cloned function parameters


miconda:
Thanks, you can merge it.

PR #164, opened 2015-05-19
mem: TLSF memory summary and sums implementation

camilleoudot:
TLSF now has a debug mode, that will dump every memory block when xxx_status()
is called, and a summary of still allocated blocks when xxx_sums() is called.

configure build with

    make MEMDBG=1 MEMMNG=2 cfg

to enable maximum debuging informations (source location of mallocs and frees)


miconda:
It can be merged.

PR #165, opened 2015-05-20
Add option to disable delete queries in usrloc module.

aphistic:
This adds a parameter to allow users to disable the delete queries in the usrloc module.  This is mainly for problems due to the differences between how a relational DB and Cassandra work.  If the schema for Cassandra is created in certain ways a delete query isn't required to keep the data clean.

### DIFF
--- a/modules/usrloc/ucontact.c
+++ b/modules/usrloc/ucontact.c
@@ -1431,6 +1431,10 @@ int db_update_ucontact(ucontact_t* _c)
  */
 int db_delete_ucontact_addr(ucontact_t* _c)
 {
+	if (ul_db_disable_delete) {
+		return 0;
+	}
+
 	char* dom;
 	db_key_t keys[4];
 	db_val_t vals[4];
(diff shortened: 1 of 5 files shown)


charlesrchance:
Hi,

The timer to delete records only runs if modparam "timer_interval" is greater than 0, so the new parameter is not required in this case.

Admittedly this is not documented, so the readme does need updating :)

Cheers,
Charles

aphistic:
Oh, I thought I tried that and it just ran without halting instead of not running at all. I'll need to double check that.

charlesrchance:
According to the module code, you should also ensure "timer_procs" modparam is set to 0 (default):

    /* Register cache timer */
    if(ul_timer_procs<=0)
    {
        if (timer_interval > 0)
            register_timer(ul_core_timer, 0, timer_interval);
    }
    else
        register_sync_timers(ul_timer_procs);

aphistic:
Ah, I looked through the code for usrloc and found where it registers the timer.  It looks like if you have timer_interval set to 0 but timer_procs set to something > 0 (which I did due to cruft) it will spawn a new process and use timer_interval multiplied by 1000 (resulting in 0) for the amount of time to sleep, resulting in the churn.  I should probably have looked at that first!

The relevant functions I looked at are:
https://github.com/kamailio/kamailio/blob/master/modules/usrloc/ul_mod.c#L329
https://github.com/kamailio/kamailio/blob/master/modules/usrloc/ul_mod.c#L393

PR #166, opened 2015-05-21
usrloc: run callbacks after updating the info on DB_ONLY mode

linuxmaniac:


linuxmaniac:
I didn't want to touch the flow when WRITE_THROUGH is activated

miconda:
WRITE_THROUGH keeps the records in memory and I guess the callbacks will use the records from there.

You can merged if you tested.

Issue #167, opened 2015-05-21
distclean doesn't clean at all


linuxmaniac:
seems like "pure" is the rule that actually cleans the sources. We should make distclean like pure

PR #170, opened 2015-05-22
Makefile: use distclean as shortcut of maintainer-clean

linuxmaniac:
Fixes: #167
Nowadays people expect distclean to be the rule to do the cleaning


miconda:
It looks like both should be added there, being used to detect that no dependency has to be computed for them. You can merge and push another commit afterwards with this update. Thanks.

PR #172, opened 2015-05-24
modules/tmx doc: t_reply_callid() works also for other requests than INVITE

mslehto:


mslehto:
Locally generated SUBSCRIBE, with script function uac_req_send(), seem to get replied.
Also quick look to tmx_mod.c didn't reveal anything INVITE specific.

miconda:
Thanks!

PR #175, opened 2015-05-25
debugger module: add per module facility logging + add a debug SIP message config function

smititelu:
Added 2 new params to the debugger module such as the per module facility logging is now possible. Added a new config function which prints how the SIP message looks like if it would have to be sent out at that line in the config (prints the message with the applied lumps but does _not_ apply the lumps as msg_apply changes() function).


miconda:
Thanks, it will be reviewed and merged if all is ok.

PR #177, opened 2015-05-27
Adding publish in registration functionality, sending "closed" NOTIFY when presense has expired

barabasp:
The commit contains pua_usrloc modifications, which simplifies the publish and regisration processes: in every REGISTER an implicit publish is built in. In this way, a device e.g. do not need to send two individual request towards kamailio (REGISTER, PUBLISH), it can be achieved in a single but extended REGISTER request.
Another small fix is that if the publication is expired (due to broken TCP e.g.), a "closed" NOTIFY should be sent to the contacts.

### DIFF
--- a/modules/pua_usrloc/pua_usrloc.c
+++ b/modules/pua_usrloc/pua_usrloc.c
@@ -81,7 +81,11 @@ send_subscribe_t pua_send_subscribe;
 
 static cmd_export_t cmds[]=
 {
+#ifdef REG_BY_PUBLISH
+	{"pua_set_publish", (cmd_function)pua_set_publish, 2, 0, 0, REQUEST_ROUTE},
+#else
 	{"pua_set_publish", (cmd_function)pua_set_publish, 0, 0, 0, REQUEST_ROUTE},
+#endif
 	{"bind_pua_usrloc", (cmd_function)bind_pua_usrloc, 1, 0, 0, 0},
 	{0, 0, 0, 0, 0, 0} 
 };
--- a/modules/pua_usrloc/ul_publish.c
+++ b/modules/pua_usrloc/ul_publish.c
@@ -46,10 +46,77 @@
 #include "../pua/pua.h"
 #include "pua_usrloc.h"
 
+#ifdef REG_BY_PUBLISH
+
 #define BUF_LEN   256
+
+static char* device_buffer = NULL;
+static char* state_buffer = NULL;
+static char* content_buffer = NULL;
+ 
+static void replace_string(char** dst, const char* source, int len)
+{
+	if (*dst)
+	{
+		free(*dst);
+		*dst = NULL;
+	}
+	
+	if (source)
+	{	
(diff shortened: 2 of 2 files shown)


miconda:
What do you mean by "a single but extended REGISTER request"? Can you give an example of such REGISTER request?

barabasp:
Normally, the REGISTER and PUBLISH are sent separately. In our solution, it is not needed, we should send only a „single” REGISTER with some „publish info”. A custom SIP header (PresenceInReg) is defined in REGISTER which contains the presence information. Here is an example:

REGISTER sips:example.com:443 SIP/2.0
Via: SIP/2.0/TLS 192.168.1.100;branch=z9hG4bK-524287-1---ec449f7fc5b0af2a;rport
Max-Forwards: 70
PresenceInReg: device;open;&lt;status&gt;&lt;state&gt;online&lt;/state&gt;&lt;message&gt;normal&lt;/message&gt;&lt;/status&gt;
Contact: <sip:1111111111@192.168.1.100;transport=tls>;+sip.instance="785722228"
To: "Andrea Máté"<sips:11111111@sip01.crypttalk.com:443>
From: "Andrea Máté"<sips:11111111@sip01.crypttalk.com:443>;tag=f4a82829
Call-ID: fWmDgP_8wJvQAKTM_b4ajw..
CSeq: 27 REGISTER
Expires: 800
Allow: INVITE, ACK, CANCEL, OPTIONS, BYE, UPDATE, INFO, MESSAGE, REFER, NOTIFY, SUBSCRIBE, REGISTER
Supported: norefersub, replaces, eventlist
User-Agent: Example_UA
Authorization: Digest username="1975539839",realm="example.com",nonce="VWwOB1VsDNsw7TW7jHJhrpX40TfhJA7t",uri="sips:example.com:443",response="78ab5f7a36c8ddae6f69c5262e240d8b",algorithm=MD5
X-CT-PushInfo: I;iNRvcCqBGqLeprCFaA7P+jrNwg2q+VYZ20FPw3Rv5hU=;hu;Ringing-Ringtone_long.wav,im.wav,ct_autoreject.wav
Content-Length: 0


Peter

miconda:
So this is a custom extension of you, not a standard, right?

If yes, I suggest that you change the patch to export a new function with a new name where the relevant values are given as parameter, without taking anything from the SIP request inside the function, so it becomes generic of what the custom header name is and the format of the header body.

Now you practically disable the rfc version with defines, so it is not going to work anymore for that. The RFC requirements implementation must be always there.

miconda:
Is there any interest in following up with this pull request? There was no message for a while answering to my questions and suggestions.

barabasp:
I'll check the source soon in next days and will make the proposed changes in it.

miconda:
Long time and no follow up. I am closing the pull request for now, because it replaces the standard IETF RFC requirements with some custom approach. The item can be reopen if a patch with suggestions made in comments above is provided.

PR #182, opened 2015-05-30
Allow domain_attrs 'type' field to be a mysql BIGINT

divereigh:
Allow 'type' field in domain_attrs to be a DB1_BIGINT as well as a
DB1_INT. If you are using a view with 'type' as a calculated field on
a 64 bit machine, then mysql will return the type of 'type' to be a bigint,
not an int (and no amount of CASTing or CONVERTing will fix it).


miconda:
Thanks for the submission. At the end I applied a patch with explicit casting of the big int value to be safe on all architecture.

PR #186, opened 2015-06-03
enum: fix buffer overflow and increase maximum number length

norox:
* Fix for buffer overflow in enum_pv_query_3 for char string[17] where the boundary checks for it in is_e164 uses MAX_NUM_LEN.

* Increased the size of MAX_NUM_LEN to 33 to support the maximum number length of 32, which is what we support here on our ISUP links.

Bug: The buffer overflow bug can be triggered by calling enum_pv_query with a number that is to long eg. +1234567890123456789

### DIFF
--- a/modules/enum/enum.c
+++ b/modules/enum/enum.c
@@ -34,7 +34,7 @@
  * based on the user part of the current Request-URI. These functions
  * assume that the Request URI user part consists of an international
  * phone number of the form +decimal-digits, where the number of digits is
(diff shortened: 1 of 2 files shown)


miconda:
Going to merge it.

For the future, format the commit message based on suggestions at:

Thanks

miconda:
Ahh, actually the commit message is properly formatted. Ignore my previous note.

norox:
Ok, thank you for merging the fix.

PR #194, opened 2015-06-08
Add symbols "#*+" to drouting tree

snen:
With this patch drouting can use "#*+"  in addition to the decimal digits.
#### Performance impact (kamcmd droating.reload):
decimal only: ~21sec
decimal and "#*+": ~21sec
#### Memory usage (kamcmd core.shmmem):
after start, decimal only, used: 203119976
after start, decimal and "#*+", used: 240338888

### DIFF
--- a/modules/drouting/prefix_tree.c
+++ b/modules/drouting/prefix_tree.c
@@ -119,7 +119,6 @@ get_prefix(
 {
 	rt_info_t *rt = NULL;
 	char *tmp=NULL;
-	char local=0;
 	int idx=0;
 
 	if(NULL == ptree)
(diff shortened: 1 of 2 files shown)


miconda:
The code to compute the index is repeating in 3 places. It should be made a function. You can add an addition patch to this pull request and I will merge then.

In long term, the list of characters that can be used has to be made a module parameter, for proper flexibility.

linuxmaniac:
or you can rework the patch amend and push --force

snen:
uniform code is moved to function.

miconda:
I am going to merge it.

For the future, when writing the commit messages, use the guidelines from:

PR #200, opened 2015-06-10
kazoo : add json keys function and escape character

lazedo:
json in the form of

{ "Nodes" :  {
       "mynode@my.tld.com" : { ... },
       "myothernode@my.tld.com" : { ...}
       }
}

doesn't have an easy to fetch the keys (it works already if it is an array)
kazoo_json_keys retrieves the keys to an avp

when querying inner fields in the form xx.yy.zzz
if one of xx / yy / zzz contains a dot (.) we cannot determine the field correctly.
use "json_escape_char" param value to encode the value before querying json and kazoo_json will decode using the same character


kamailio-sync:
https://kamailio.sipwise.com/job/kamailiodev-nightly-binaries/

Build broken

PR #207, opened 2015-06-13
Small accent circumflex correction.

AlessioCasco:


miconda:
This is due to different locale/what-so-ever language settings on the generating machine. The README is generated from docbook xml files in doc/ subfolder of each module.

I will merge it anyhow.

miconda:
Btw, for the future write the commit log message using the guidelines from:

PR #208, opened 2015-06-13
etc/kamailio-oob.cfg: Updated module path

tuxd00d:
- Removed "_k" on mpath as it's depreciated
- mpath full path updated to /usr/lib/x86_64-linux-gnu/ from /usr/local/lib/
- Now matches other examples


tuxd00d:
Did I follow GIT Commit Guidelines this time? :)

miconda:
Commit message is ok, but the mpath has to stay /usr/local/lib/ -- it is where shared objects are installed by source compilation. The /usr/lib/x86_64-linux-gnu/ is set by the packages, being specific to some distros (debian) and it has to stay in packaging specs.

I close this request, you can make a new one for removing the _k.

PR #214, opened 2015-06-18
P-Charging-Vector handling as RFC3455

neutrino38:
Here are some proposed modifications to handle the P-Charging-Vector header as defined in rfc3455 section 4.6 to
 - parse charging vector from SIP message
 - generate new unique charging vector
 - remove a charging vector header

a new pseudo variable

pseudo variables are exported and enable Read onlly access to charging vector fields

$pcv = whole field
$pcv.value = icid-value field (see RFC3455 section 5.6)
$pcv.genaddr = icid-generated-at field (see RFC3455 section 5.6)

a new function acc_charging_vector()

- g = generate a new charging vector or forward the orginal one 
- f = generate a new charging vector regardless of the orginal charging vector
- r = remote the P-Charging vector header from the packet


neutrino38:
Please ignore. Pull request was incomplete

PR #217, opened 2015-06-19
added doc for proirites inside permission's module

eschmidbauer:


miconda:
The commit message is not really formatted as suggested in commit guidelines.

eschmidbauer:
Please see new commit message.

PR #219, opened 2015-06-21
dispatcher: relative weight distribution

snen:
I was confused a little while using dispatcher weight alg. Smth like this If some host go away:
30/30/40 -> 70/30/0 (the third one was failed). This transformation is more correct in my mind:
30/30/40 -> 50/50/0. So I have to assign some hook to dispatcher:dst-down/dispatcher:dst-up  route for rebalancing
“rweight” (relative weight)  parameter is added in this branch. rweight is in the integer range from 1 to 100 (weight, not percentage).
Active host usage probability is  rweight/(sum of all active host rweights in destination group).
So INACTIVE/DISABLED destinations are removed from probability calculation.
let’s we have rweights 1/2/1; 100 calls will be distributed as  25/50/25
after third host failing (via ds_mark_dst([state]) or RPC dispatcher.set_state or via module pinging): 33/67/0

### DIFF
--- a/modules/dispatcher/dispatch.c
+++ b/modules/dispatcher/dispatch.c
@@ -471,10 +471,11 @@ int dp_init_relative_weights(ds_set_t *dset)
 			t++;
 		}
 	}
-	/* if the array was not completely filled (i.e., the sum of weights is
+	/* if the array was not completely filled (i.e., the sum of rweights is
 	 * less than 100 due to truncated), then use last address to fill the rest */
-	for(; t<100; t++)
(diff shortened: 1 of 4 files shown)


miconda:
Thanks for the contribution. I am going to review soon and merge if all is ok.

miconda:
I will merge it.

It is missing the patch to modules/dispatcher/doc/dispatcher_admin.xml with the documentation for the new algorithm 11 and the new special attribute rweight. Make a new pull request with it.

snen:
docs writing before approving is too hard )
added

miconda:
Thanks!

PR #220, opened 2015-06-24
modules/htable: improves error message when loading from database

mslehto:
I had trouble finding correct syntax with db_text backend, this patch helped me

### DIFF
--- a/modules/htable/ht_db.c
+++ b/modules/htable/ht_db.c
@@ -197,11 +197,17 @@ int ht_db_load_table(ht_t *ht, str *dbtable, int mode)
 	do {
 		for(i=0; i<RES_ROW_N(db_res); i++)
 		{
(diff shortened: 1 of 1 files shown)


miconda:
The check for NULL should be before the one for STRING, because the condition for STRING is printing the value and in case it is NULL will result in crash.

mslehto:
Umh, yes! Now it is other way around. I forcefully changed history of my branch, updated patch is c91116a5b253c2827f34d99a5597a14fe8f16001 (seen also in this pull request).

miconda:
One more I just noticed: if the .type!=DB1_STRING, it assume is DB1_STR, because it is printing that value, but can be any other type, like DB1_INT and then there will be a crash.

Printing type is ok, but before the second LM_ERR has to be a check on the type before printing some specific fields.

mslehto:
Thanks for patience... I removed value printing on f39736b276878f0de1e0471311119f12f1878801

PR #224, opened 2015-06-29
core: clean previous routename

linuxmaniac:
fix c44685cbcefb8f6ecfa6f11369699906db832c39

routename is not been reset when no name is use like on main or reply route


miconda:
Thanks, it can be merged.

PR #225, opened 2015-06-29
kex: per module memory debugging

smititelu:
Kex module updated to also give some details about memory used by each module.
Changed f_*, q_* and tlsf_* memory stubs to also include the module name.
Added 2 new memory API functions(for each of the memory algos) which will
iterate throughout the memory fragments and agregate the memory size used by
each function in order to print it.
Tested with MEMMNG=0,1,2 and MEMDBG=0,1. Use "**kamcmd mem.stat [mod_name]/[all] [pkg]/[shm]/[all]**" to test
Updated doku.

### DIFF
--- a/events.c
+++ b/events.c
@@ -165,6 +165,15 @@ int sr_event_register_cb(int type, sr_event_cb_f f)
 					_sr_events_list.rcv_nosip = f;
 				else return -1;
 			break;
+		case SREV_MODULE_PKG_STATS:
+				if(_sr_events_list.mod_update_pkg_stats==0)
+					_sr_events_list.mod_update_pkg_stats = f;
+				else return -1;
+		case SREV_MODULE_SHM_STATS:
+				if(_sr_events_list.mod_update_shm_stats==0)
+					_sr_events_list.mod_update_shm_stats = f;
+				else return -1;
+			break;
 		default:
 			return -1;
 	}
@@ -284,6 +293,18 @@ int sr_event_exec(int type, void *data)
 					ret = _sr_events_list.rcv_nosip(data);
 					return ret;
 				} else return 1;
+		case SREV_MODULE_PKG_STATS:
+				if(unlikely(_sr_events_list.mod_update_pkg_stats!=0))
+				{
+					ret = _sr_events_list.mod_update_pkg_stats(data);
+					return ret;
+				} else return 1;
+		case SREV_MODULE_SHM_STATS:
+				if(unlikely(_sr_events_list.mod_update_shm_stats!=0))
+				{
+					ret = _sr_events_list.mod_update_shm_stats(data);
+					return ret;
+				} else return 1;
 		default:
 			return -1;
 	}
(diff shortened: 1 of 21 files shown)


smititelu:
One approach to this feature would have been to create some callbacks to update the memory stats, called by malloc/realloc/free functions. This approach would have been questionable because processing has to be done each time somememory functions were called.
The current approach is to export some memory api functions (something similar to q_sums) which will itereate through the memory fragments and agregate their size. The kamcmd callback will call those functions and print the agregated results.

smititelu:
The new feature is under the DBG_MOD_STATS define (currently enabled in the f_,q_ and tlsf headers). While implementing this I've noticed that the structures I have to use were similar to those used when the MEMDBG=1. So I decided to "merge" some parts of the code(stubs and some structure params) with the code under the DBG_QM_MALLOC, while still maintaining log differences (this feature does not log anything).

Please tell me what do you think about it.

smititelu:
Updated pull request. Created 2 new functions (similar to qm_sums):
- qm_mod_get_stats() fills a list of _struct mem_counter_ with details
- qm_mod_free_stats() frees the above allocated list

I have a question related to compiling using different memory allocation algorithms (as seen here http://www.kamailio.org/wiki/tutorials/troubleshooting/memory):
- MEMMNG=1 MEMDBG=0/1 make cfg - will use the F_* allocation algorithm
- MEMMNG=2 MEMDBG=0/1 make cfg - will use the F_* allocation algorithm

How can I compile using Q_* and TLSF_* algorithms in order to make tests for them?

miconda:
I was just looking at the old patch to see what can be done with it, thanks for the update. I will merge it.

MEMMNG is deprecated now, all memory managers (f_malloc, q_malloc and tlsf_malloc) are compiled always -- use command line parameter -x to select the one you want (see kamailio -h).

miconda:
I merged the patch, but I split it per component so there are many commits pushed to the master branch - some remarks:

  * the patch to ctl module was changed, as it was planned to increase default values for those buffers (note that you can also change them via modparams of ctl module)
  * I noted that you changed how core is referenced in mem/src_loc.h from "<core>" to "core" -- I assumed that was to make it coherent with how module names are printed
  * I plan to add a mod param for kex to enable/disable the new feature, just in case someone needs to tune a bit more performances -- it will be enabled by default

smititelu:
Thank you for helping to organize this better:
- nice, that skipped me
- you are right, I changed the name from "\<core\>" to "core" or else someone would have to issue the command writing "\<core\>"
- an enable/disable modparam would be nice but I'm not sure if the code as it is right now affects performance in any way i.e. **only** when the "kamcmd mod.stats ..." is issued, when debugging is enabled, some processing actually happens; supposing the performance tests are done when debugging is disabled, this should not interfere in any way; please correct me if I'm wrong.

miconda:
I thought that the new events are going to be executed on operations, but tracking the code now I see that mod_update_pkg_mod_stats_list() just returns 0; same for shm.

So, what is the purpose for SREV_MODULE_PKG_STATS and SREV_MODULE_SHM_STATS, they don't seem to be used. If not and no plat to use them in the near future, then they should be removed.

smititelu:
I put them just because I thought they might be used sometimes by someone for some purpose.
They are not called right now, and no plan to used them from our side. I'll remove them.

miconda:
Then it is better to be removed.

smititelu:
Removed them in commit 53cd7df83c4065fd7a34cd3ce15b6e215ce39278.
Updated http://www.kamailio.org/wiki/tutorials/troubleshooting/memory to include mod.stats command and the "kamailio -x" info.

PR #226, opened 2015-07-01
Update INSTALL

fredposner:
Changed "flavor" to "flavour" -- maybe when more US users come we can switch to US spelling ;)

Updated bug tracker to github.


miconda:
Thanks, applied manually to reformat the commit message

PR #229, opened 2015-07-04
usrloc|nathelper: add filtering by server_id support

linuxmaniac:
Use the new field server_id on usrloc module to be able to load/get contacts filtered by server_id


miconda:
Thanks, it can be merged.

PR #238, opened 2015-07-08
cdp: Fix cdp routing for Rx interface

fedefava86:
Avoid to force vendor_id variable to zero if AVP AVP_Auth_Application_Id or AVP_Acct_Application_Id does not exist on AAR message. 
In this case, if  Vendor_id AVP exists but AVP_Auth_Application_Id or AVP_Acct_Application_Id does not exist vendor_id is forced to zero, thus cdp is not able to route the requests correctly anymore.
vendor_id and app_id variables are updated only if both AVPs (vendor_id and Acct or vendor_id and Auth exist).

### DIFF
--- a/modules/cdp/routing.c
+++ b/modules/cdp/routing.c
@@ -193,16 +193,14 @@ peer* get_routing_peer(cdp_session_t* cdp_session, AAAMessage *m) {
 
     avp_vendor = AAAFindMatchingAVP(m, 0, AVP_Vendor_Id, 0, AAA_FORWARD_SEARCH);
     avp = AAAFindMatchingAVP(m, 0, AVP_Auth_Application_Id, 0, AAA_FORWARD_SEARCH);
-    if (avp) {
-        if (avp_vendor) vendor_id = get_4bytes(avp_vendor->data.s);
-        else vendor_id = 0;
+    if (avp && avp_vendor) {
+	vendor_id = get_4bytes(avp_vendor->data.s);
         app_id = get_4bytes(avp->data.s);
     }
 
     avp = AAAFindMatchingAVP(m, 0, AVP_Acct_Application_Id, 0, AAA_FORWARD_SEARCH);
-    if (avp) {
-        if (avp_vendor) vendor_id = get_4bytes(avp_vendor->data.s);
-        else vendor_id = 0;
+    if (avp && avp_vendor) {
+	vendor_id = get_4bytes(avp_vendor->data.s);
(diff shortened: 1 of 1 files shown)


fedefava86:
Dear all,

With this mail i would like to explain better my modifications to CDP module for Rx diameter message routing. (pull request #238).
As you know, the ims_qos module create AAR Diameter message sent from P-CSCF to PCRF on REGISTER or on the first response received by the P-CSCF with a valid SDP during INVITE and STR message on BYE

After the creation of AAR message, the routing decision is made in the cdp module.

On get_routing_peer function (routing.c file), fist of all, AVP_Vendor_Specific_Application_Id is checked. If this AVP is present, AVP_Vendor_Id and AVP_Auth_Application_Id is checked together.
If they are present on the message, vendor_id and app_id are assigned with the respective values retrieved from the message.
After that, AVP_Auth_Application_Id is checked.
If this AVP is present, AVP_Vendor_Id is checked. If this is not present, vendord_id is assigned to zero.
The same behavior is implemented for AVP_Acct_Application_Id.

This way, if Auth_Application_Id is present but not AVP_Vendor_Id is not, vendor_id variable is zero and peer_handles_application function will return always zero and the module will not be able to route correctly the request (unless you by-pass the routing forcing explicitely the peer on the script).

Taking a look to the specs (TS 29.214)  AVP_Auth_Application_id is mandatory, meanwhile AVP_Vendor_Id and AVP_Acct_Application_Id is not present. For this reason, IMHO vendor_id variable cannot be forced to zero if one of them is not present.
In my patch, vendor_id variable is updated if only both AVP are present (Auth_Application_Id and Vendor_Id or
AVP_Acct_Application_Id and Vendor_Id).

In my tests, everything is working with my patch and also the other Diameter messages (e.g. from I-CSCF or S-CSCF to HSS and vice-versa). However, i would like to have your point of view on the matter.

Best regards,

Federico Favaro
R&D Department
Athonet s.r.l.

kamailio-sync:
Hi Federico,

I agree with you re. the specs. I will have a look this evening on our test
setup and commit your patch if I am happy. Thanks for your feedback!

Cheers
Jason

miconda:
@jaybeepee - any resolution to this one, can it be merged?

kamailio-sync:
Hey Daniel,

yes it can be merged. Thanks Frederico

fedefava86:
Thanks to you for the feedback!

PR #246, opened 2015-07-15
dialog: run event_route[dialog:end] in case of dialog termination by …

krieger-od:
…interface commands

### DIFF
--- a/modules/dialog/dlg_handlers.c
+++ b/modules/dialog/dlg_handlers.c
@@ -1404,8 +1404,6 @@ void dlg_ontimeout(struct dlg_tl *tl)
 		{
 			if(dlg_bye_all(dlg, NULL)<0)
 				dlg_unref(dlg, 1);
-			/* run event route for end of dlg */
-			dlg_run_event_route(dlg, NULL, dlg->state, DLG_STATE_DELETED);
 
 			dlg_unref(dlg, 1);
 			if_update_stat(dlg_enable_stats, expired_dlgs, 1);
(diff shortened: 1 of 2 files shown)


miconda:
Thanks for the patch.

At quick look, it seems that the event route will be called twice in case of a dialog lifetime timeout. At line 1408 in dlg_handlers.c, the event route code is executed again after dlg_bye_all() and dlg_bye_all() executes inside the event route code. Might be an option to remove it from that place.

krieger-od:
Thanks for this deep review. Amended that commit, it should be in this pull request now (9ccc3f510ce2d97dbe1efb690ef1d4e8f18885d5).

What about that double dlg_unref() at dlg_handlers.c near 1408 after dlg_bye_all() in case of non-zero return? Is it ok?

miconda:
The unrefs should be ok. One is done in case bye was not sent out, to destroy dialog faster, otherwise it will wait for the reply of the bye to unref and destroy dialog.

krieger-od:
> The unrefs should be ok. One is done in case bye was not sent out, to destroy dialog faster, otherwise it will wait for the reply of the bye to unref and destroy dialog.

I meant that the new code is this:
```
 			if(dlg_bye_all(dlg, NULL)<0)
 				dlg_unref(dlg, 1);
 			dlg_unref(dlg, 1);
```

miconda:
Yes, that's the code.

vinodmn:
Hi i am using kamailio 4.4.3 but still I am facing the same issue, 
**event_route[dialog:end] is not exec'd on dialog terminated by dlg.end_dlg,** ,so the CDR's are not generating properly.

PR #251, opened 2015-07-16
Update README

fredposner:
- added information about algorithms
- corrected rl typos
- corrected example (wrong algorithm)
- added note about algorithms being case sensitive
- added example using pipelimit for specific method (INVITE)


miconda:
Merging -- in the future format the commit message as per guidelines at:

Issue #253, opened 2015-07-18
dnssec: can't be loaded


linuxmaniac:
Bugreport from Debian user: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=790291

```
0(6785) DEBUG: <core> [cfg.y:1659]: yyparse(): loading module dnssec.so
 0(6785) DEBUG: <core> [sr_module.c:567]: load_module(): trying to load </usr/lib/x86_64-linux-gnu/kamailio/modules/dnssec.so>
 0(6785) ERROR: <core> [sr_module.c:574]: load_module(): could not open module </usr/lib/x86_64-linux-gnu/kamailio/modules/dnssec.so>: /usr/lib/x86_64-linux-gnu/libval-threads.so.14: undefined symbol: SSL_get_SSL_CTX
 0(6785) : <core> [cfg.y:3432]: yyerror_at(): parse error in config file /etc/kamailio/kamailio-basic.cfg, line 186, column 12-22: failed to load module
```

```
root@48977e2bd1bb:/code# ldd /usr/lib/x86_64-linux-gnu/kamailio/modules/dnssec.so 
	linux-vdso.so.1 (0x00007ffe5fcff000)
	libval-threads.so.14 => /usr/lib/x86_64-linux-gnu/libval-threads.so.14 (0x00007f242cc33000)
	libcrypto.so.1.0.0 => /usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.0 (0x00007f242c7d2000)
	libsres.so.14 => /usr/lib/x86_64-linux-gnu/libsres.so.14 (0x00007f242c5ba000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f242c39d000)
	libkmi.so.1 => /usr/lib/x86_64-linux-gnu/kamailio/libkmi.so.1 (0x00007f242c193000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f242bdea000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f242bbe6000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f242d074000)
```
Confirmed

After adding -lssl 

```
0(6807) DEBUG: <core> [cfg.y:1659]: yyparse(): loading module dnssec.so
 0(6807) DEBUG: <core> [sr_module.c:502]: load_module(): module file not found <modules/dnssec.so>
[...]

PR #254, opened 2015-07-18
dnssec: link openssl

linuxmaniac:
Fixes: #253
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=790291

### DIFF
--- a/modules/dnssec/Makefile
+++ b/modules/dnssec/Makefile
@@ -7,7 +7,7 @@
 include ../../Makefile.defs
 auto_gen=
 NAME=dnssec.so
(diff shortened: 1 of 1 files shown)


linuxmaniac:
What about licence here? Should be change the licence here as in 7771e9cdb58ea1cef008e2a70f53c5f183c26ebe ?

miconda:
Can be merged. The lib is needed by libval, no direct include of its header used in the dnssec module. Not sure if anything needs to be done in this case, would be good if libval would have a pkg-config to get its linking requirements.

Anyhow, if really needed, then it is no problem as we sorted out this once and even more, IIRC, Marius Zbihlei was directly part of conversation.

PR #270, opened 2015-07-30
Update hash.h

foucse:
presence;winfo -> presence.winfo
just a typo, imho


miconda:
Thanks, I will apply the patch manually to set the commit message according to the guidelines at:

PR #277, opened 2015-08-06
etc: simple typo fixes in kamailio.cfg

fredposner:
- fixed small typos within comments of default kamailio config


miconda:
Thanks

Issue #278, opened 2015-08-06
registrar: max_contacts logic fails when REGISTER contact has sip.instance


linuxmaniac:
```
modparam("registrar", "max_contacts", 1)
```
Two REGISTERS with the same sip.instace+reg_id should produce an update, but:
```
2015/08/06 20:26:33.903451 127.0.0.1:5061 -> 127.0.0.1:5060
REGISTER sip:127.0.0.1 SIP/2.0
Via: SIP/2.0/UDP 172.17.0.4:5061;branch=z9hG4bK-28433-1-0
From: <sip:3000@127.0.0.1>;tag=1
To: <sip:3000@127.0.0.1>
Call-ID: 1-28433@172.17.0.4
CSeq: 1 REGISTER
Contact: <sip:3000@127.0.0.1:5061>; reg-id=1;+sip.instance="<urn:uuid:C3DD6013-3000-40E3-8EA2-5849B02ED0C4>"
Max-Forwards: 10
Expires: 120
User-Agent: SIPp/Win32
Content-Length: 0


(null) (null) 127.0.0.1:5060 -> 127.0.0.1:5061
SIP/2.0 200 OK
Via: SIP/2.0/UDP 172.17.0.4:5061;branch=z9hG4bK-28433-1-0;received=127.0.0.1
From: <sip:3000@127.0.0.1>;tag=1
To: <sip:3000@127.0.0.1>;tag=b27e1a1d33761e85846fc98f5f3a7e58.2c14
Call-ID: 1-28433@172.17.0.4
CSeq: 1 REGISTER
Contact: <sip:3000@127.0.0.1:5061>;expires=30;+sip.instance="<urn:uuid:C3DD6013-3000-40E3-8EA2-5849B02ED0C4>";reg-id=1
Server: kamailio (4.3.0 (x86_64/linux))
Content-Length: 0


2015/08/06 20:26:34.315096 127.0.0.1:5061 -> 127.0.0.1:5060
REGISTER sip:127.0.0.1 SIP/2.0
Via: SIP/2.0/UDP 172.17.0.4:5061;branch=z9hG4bK-28437-1-0
From: <sip:3000@127.1.0.1>;tag=1
To: <sip:3000@127.1.0.1>
Call-ID: 1-28437@172.17.0.4
CSeq: 1 REGISTER
[...]

PR #279, opened 2015-08-06
registrar: get instance and ref_id when using msg

linuxmaniac:
needed for check max_contacts properly if sip.instance exists
as contact parameter

See: #278


miconda:
Can be merged and backported if needed. Thanks.

PR #282, opened 2015-08-11
tls: install-cfg: do not generate tls-cert by default

linuxmaniac:
In order to be able to generate reproducible builds we should not include an auto-generated cert in the default install-cfg rule at tls

### DIFF
--- a/modules/tls/Makefile
+++ b/modules/tls/Makefile
@@ -40,7 +40,10 @@ include ../../Makefile.modules
 install-tls-cert: $(cfg_prefix)/$(cfg_dir)
 	MAIN_NAME=$(MAIN_NAME) ./$(SCR_NAME)_cert.sh -d $(cfg_prefix)/$(cfg_dir)
 
-install-cfg:  install-tls-cert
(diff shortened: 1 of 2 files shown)


linuxmaniac:
FTR: https://reproducible.debian.net/rb-pkg/unstable/amd64/kamailio.html

miconda:
Thanks! I think the docs of tls module needs a note about what to do now for generating/installing the certificates:

```
make install-tls-cert
```

Then, this can be merged.

linuxmaniac:
still WIP, I will add the note here too

linuxmaniac:
uhm, there is plenty of info at tls/doc/certs_howto.xml already

miconda:
But is it about the fact that they are not generated anymore automatically at install. That's the note I would like to have, to reflect the change you do with the patch, so one has to execute manually the command for generating the certs.

linuxmaniac:
Something like that is enough?

miconda:
Should be ok for the moment.

PR #283, opened 2015-08-11
rtpengine: new read_sdp_pv parameter

camilleoudot:
this parameter allows to specify a script var or AVP for rtpengine to
get the SDP from, instead of the SIP message body.


miconda:
@rfuchs: any comments on this patch?

PR #284, opened 2015-08-11
janssonrpc-c: allow null json error property in rpc response

jnewmano:
If the JSON RPC response contains an error property, check to see if it is a json_null or not before flagging the response as an error.


miconda:
Thanks, will be merged.

PR #291, opened 2015-08-18
modules/dialog: make dialog context available in event route tm:local…

grumvalski:
…-request

Set the dialog context before calling dlg_bye_all. In this way the dialog context is the correct one in the event route tm:local-request; otherwise it could happen that, in the timer process, the context is invalid due to some previous timed-out transactions which set (and not reset) the dialog context in the dlg_onreply function (locally generated replies).

### DIFF
--- a/modules/dialog/dlg_handlers.c
+++ b/modules/dialog/dlg_handlers.c
@@ -1402,8 +1402,12 @@ void dlg_ontimeout(struct dlg_tl *tl)
 
 		if(dlg->iflags&DLG_IFLAG_TIMEOUTBYE)
 		{
+			/* set the dialog context so that it's available in
+			 * tm:local-request event route */
+			dlg_set_ctx_iuid(dlg);
 			if(dlg_bye_all(dlg, NULL)<0)
-				dlg_unref(dlg, 1);
+				LM_ERR("failed to send BYE message\n");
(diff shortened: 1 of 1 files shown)


miconda:
I see that an dlg_unref is removed and not re-added -- from the patch:

```
 			if(dlg_bye_all(dlg, NULL)<0)
-				dlg_unref(dlg, 1);
```

Any reason for that? If BYE failed to be be sent, then the dialog will hang around forever.

grumvalski:
I thought that the dlg_unref below was enough both for successful and unsuccessful case. Am I missing something?

miconda:
At that stage, dialog needs to be unref'ed two times in order to be destroyed. If BYE is sent ok, then one unref is done when processing the response of the BYE. If BYE is not sent, then the dialog is unref'ed by that IF condition, because there is no expected response.

Removing one unref is not ok and it is not 'fixed' by a next unref. This manages a reference counter, not a mutex.

grumvalski:
Thank you for the explanation Daniel. I'll fix the patch.

grumvalski:
I'm going to close this request and open a new one with a clean patch.

linuxmaniac:
FTR: no need to, If you can use push with --force on the PR branch

grumvalski:
I'll do next time :)

PR #295, opened 2015-08-24
Fix dialog module still loading DB when db_fetch_rows == 0

jaybinks:
as per http://kamailio.org/docs/modules/3.3.x/modules_k/dialog.html#idp148408 a db_fetch_row of 0 should not load data from the DB At all.

I think doing this in mod_init makes the most sense, rather than further down in select_entire_dialog_table or similar.  The docs say setting to 0 will only restrict the initial load on startup, not any subsequent loads.

### DIFF
--- a/modules/dialog/README
+++ b/modules/dialog/README
@@ -706,8 +706,8 @@ modparam("dialog", "db_update_period", 120)
    the dialog records at startup from the database. This value can be used
    to tune the load time at startup. For 1MB of private memory (default),
    it should be below 400. The database driver must support the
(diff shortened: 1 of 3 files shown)


miconda:
It is not easy to track all those patches one after the other and because the commit message is not formatted as we recommend, I would suggest:

* make a diff between your branch and kamailio main master branch
* apply it to a fresh clone of kamailio main master branch
* write the commit message as per guidelines at: http://www.kamailio.org/wiki/devel/git-commit-guidelines#commit_message
* make a new pull request -- which will be easier to review

PR #297, opened 2015-08-24
dispatcher: add new probing_mode=3 to always probe flagged gateways

evolvetsi:
- added DS_PROBE_ONLYFLAGGED probing_mode option.  If set,
     any gateways with a state of PROBING will be continually
     probed and the PROBING flag will never be cleared.  This
     allows probing specific gateways continually, keeping
     their up/down status current.

This is related to this sr-users mailing list thread:
http://lists.sip-router.org/pipermail/sr-users/2015-August/089594.html

### DIFF
--- a/modules/dispatcher/README
+++ b/modules/dispatcher/README
@@ -28,15 +28,15 @@ Alessandro Arrichiello
 
    Hewlett Packard
 
-   Copyright � 2004 FhG FOKUS
+   Copyright © 2004 FhG FOKUS
 
-   Copyright � 2005 Voice Sistem
+   Copyright © 2005 Voice Sistem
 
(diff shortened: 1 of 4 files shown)


miconda:
To change the content of README for module, you have to edit xml file inside modules/dispatcher/doc/. Then, to regenerate the readme, do:

```
make modules-readme modules=modules/dispatcher
```

You need docbook utils and xsl packages for the above command to work, but it is ok to just change the xml file, we can generate the readme and push it.

evolvetsi:
Gotcha.  I went ahead and made the changes to the xml file and regenerated the README.. Looks good to me, though I notice that the various special characters in the README file changed when I regenerated it..  You may want to regenerate the readme using your kit instead.

PR #299, opened 2015-08-24
modules/p_usrloc: fixes get_all_ucontacts() function signature

mslehto:
I am not user of p_usrloc, but accidentally compiled this and bumped into mentioned warning.
Clang 3.0 and GCC 4.7 do not complain.
Clang 3.5 and GCC 4.8 do complain.


miconda:
Thanks!

PR #300, opened 2015-08-25
modules/dialog: make dialog context available in event route tm:local…

grumvalski:
…-request

Set the dialog context before calling dlg_bye_all. In this way the dialog context is the correct one in the event route tm:local-request; otherwise it could happen that, in the timer process, the context is invalid due to some previous timed-out transaction which set (and not reset) the dialog context in the dlg_onreply function (locally generated replies).


grumvalski:
Thank you for merging it Daniel. Should we consider it a bug and backport to stable branches?

miconda:
You can go ahead and backport.

PR #302, opened 2015-08-25
db_skip_load and fixed to db_fetch_rows documentation

jaybinks:


miconda:
I think the patch is not ok. With the new parameter you skip execution of init_dlg_db() function in mod_init(). But that function does more than loading values from database, like testing table version and registering the timer function for DB_MODE_DELAYED. In this form, the patch breaks that db mode.

I guess that the db_skip_load has to be tested inside the init_dlg_db() and skip execution of:

```
	if( (load_dialog_info_from_db(dlg_hash_size, fetch_num_rows) ) !=0 ){
		LM_ERR("unable to load the dialog data\n");
		return -1;
	}
	if( (load_dialog_vars_from_db(fetch_num_rows) ) !=0 ){
		LM_ERR("unable to load the dialog data\n");
		return -1;
	}
```

For the new patch, do not forget to format the commit message as per guidelines at:

PR #305, opened 2015-08-26
modules/sipcapture: adds insertion of request, From, and To URI domain part

mslehto:
ruri_domain, from_domain and to_domain have existed since SQL schema version 3
-> no DB schema changes needed


adubovikov:
thank you

PR #306, opened 2015-08-27
usrloc: really do the ignore location record if it's a non-local socket

zxcpoiu:
* When usrloc loading location records from database, and if the socket of the record is not matched in kamailio's listening socket list, It only throws LM_ERR ignoring message only but still store into memory.
should `return 0` as well.

* And the message typo sometimes confused people to distinguish that kamailio is "sipping" the record (redirct) or "skipping" the record.

### DIFF
--- a/modules/usrloc/doc/usrloc_admin.xml
+++ b/modules/usrloc/doc/usrloc_admin.xml
@@ -1064,6 +1064,28 @@ modparam("usrloc", "db_insert_null", 1)
 		</example>
 	</section>
 
+	<section id="usrloc.p.skip_remote_socket">
+		<title><varname>skip_remote_socket</varname> (int)</title>
+		<para>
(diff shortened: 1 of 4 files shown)


miconda:
For the first patch -- the behaviour is like that on purpose, because one can have replicated location records via database and the case can be that register was processed by one server and records are loaded by another server, there being different IPs for each servers. Another situation is changing the ip address and restarting kamailio. If there is no NAT involved, the location records work fine when that socket field is ignored.

If you want to restrict this behaviour and not allow loading those records, then add a module parameter to control it.

zxcpoiu:
thanks, I see.

zxcpoiu:
I added it as a module parameter.

miconda:
Thanks, it will be merged.

PR #307, opened 2015-08-28
Small typo fix in config.

mawis:
There is a small typo in kamailio.cfg where "Rotuing" has been written instead of "Routing".


miconda:
Thanks!

PR #309, opened 2015-08-31
pv: create command to copy script vars to xavp

linuxmaniac:
pv_scriptvar2xavp(varname, xname)

Copy script variables values to a xavp.

First parameter can be '*' in order to copy all script variables.
Second parameter is the name of the destination xavp. If xavp already
exists it will be reset first.

Function can be used from ANY_ROUTE.

### DIFF
--- a/modules/pv/README
+++ b/modules/pv/README
@@ -11,11 +11,11 @@ Daniel-Constantin Mierla
 
    <daniel@asipto.com>
 
-   Copyright © 2008-2011 Daniel-Constantin Mierla (asipto.com)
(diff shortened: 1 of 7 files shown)


miconda:
Can be merged, but I would like to change the name of the new function -- it can be done in a follow up commit, after merging.

Although there few functions that uses 2 as replacement for 'to' in the name of functions, this is not the common. Then scriptvar is rather long, better use just the 'var', being the name of that variables, there is already a prefix 'pv' that points it is about pseudo-variables. For example, the new name:

```
pv_var_to_xavp(...)
```

PR #314, opened 2015-09-03
Return CSCF_RETURN_FALSE if AAR is not generated

fedefava86:
Differentiate the cases where AAR is not sent because
of an error and the cases where AAR is not sent because it is logically wrong to sent them at that point of the transaction. In this way, in the routing script it is possible to distinguish between an error and (for example) send back an error to the UE and when AAR is not generated because (for example) the answer does not contain an SDP or you are not in a SIP reply. In this case, you should not send back an error but go ahead with your transaction.
I noticed that CSCF_RETURN_FALSE (-1) is never used in the function, so it can be suitable to be returned in the cases just explained

### DIFF
--- a/modules/ims_qos/mod.c
+++ b/modules/ims_qos/mod.c
@@ -622,6 +622,7 @@ static int w_rx_aar(struct sip_msg *msg, char *route, char* dir, char *c_id, int
     //We don't ever do AAR on request for calling scenario...
     if (msg->first_line.type != SIP_REPLY) {
         LM_DBG("Can't do AAR for call session in request\n");
+		result = CSCF_RETURN_FALSE;
         return result;
     }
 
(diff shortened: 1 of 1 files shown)


miconda:
Hopefully @jaybeepee or @ngvoice should be able to review this patch, being related to ims modules.

richardgood:
Hi

This looks completely fine to me.

Regards

miconda:
@richardgood - you can just merge it if is ok.

kamailio-sync:
Hi

Ok perfect will do.

Regards


*Richard Good*
*Senior Manager: Applications & Services* *Smile Communications Pty (Ltd)*
*Mobile:* +27 (0) 72 389 8365 * Skype:* richard.a.good
richard.good@smilecoms.com <name.surname@smilecoms.com> www.smilecoms.com




On 7 September 2015 at 16:55, Daniel-Constantin Mierla <
notifications@github.com> wrote:

> @richardgood <https://github.com/richardgood> - you can just merge it if
> is ok.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/kamailio/kamailio/pull/314#issuecomment-138318109>.
>
> _______________________________________________
> sr-dev mailing list
> sr-dev@lists.sip-router.org
> http://lists.sip-router.org/cgi-bin/mailman/listinfo/sr-dev
>
>

-- 


This email is subject to the disclaimer of Smile Communications at http://www.smilecoms.com/home/email-disclaimer/ <http://www.smilecoms.com/disclaimer>

fedefava86:
Thanks everyone

PR #316, opened 2015-09-07
pv: pv_xavp_to_var()

linuxmaniac:
Reverse function of pv_var_to_xavp()


miconda:
It can be merged.

There is one thing I have to think about -- the relation with pv cache name if you add a $var(x) at runtime that is not used in config to be parsed and cached. I don't expect any issue with kamailio.cfg, only case to be revised is when trying to get it from an embedded interpreter, like lua.

PR #319, opened 2015-09-09
Update functions.c

debackel:
http_query
Remove leading empty lines in http response, in order to return the first non-empty one.


miconda:
stream.buf is dynamic allocated memory and by changing its value it will crash when freeing it, because it is no longer pointing to the start of allocated memory. If you want to strip empty chars at beginning, then do it without changing the initial value of stream.buf -- e.g., use another variable to copy the value for stream.buf and use that var for moving forward to first non-empty-char.

Also, format the commit message as per guidelines at:

debackel:
Thanks for your quick review.
Sorry about that.
As you can see, I have no C++ knowledge, and I'm not used to Github processes and procedures.
I'll try it once again.

PR #321, opened 2015-09-09
Utils module: http_request modification

debackel:
Hi,
Here is a modification of the http_request function of the utils module.
The idea is to ignore empty lines at the beginning of the http response.
This way, the response returned to Kamailio is the first non empty line of the http response.
This is to solve an issue encountered with apache2 sometimes returning empty lines.
I am no c++ developper and no acquainted to Github (double check the patch).
Thanks,
Laurent.

### DIFF
--- a/modules/utils/functions.c
+++ b/modules/utils/functions.c
@@ -86,6 +86,8 @@ int http_query(struct sip_msg* _m, char* _url, char* _dst, char* _post, char* _h
 	pv_value_t val;
 	double download_size;
 	struct curl_slist *chunk = NULL;
+	double size;
+    	char *head;
 
 	memset(&stream, 0, sizeof(http_res_stream_t));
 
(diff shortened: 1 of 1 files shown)


miconda:
I pushed a commit with a different implementation, controlled by a module parameter http_response_trim -- you have to set it to 1 to eneable the feature.

It removes all the white spaces from the beginning and end of the response, because I added another feature that allows to return the full http response instead of only first line.

Testing and feedback will be appreciated, for now I close this item.

debackel:
This is great, many thanks!
I will test in the course of the day.
BRs,
Laurent.

debackel:
Hi,
I have the following issue when making.
Do I miss something?

Thanks,
Laurent
-------


CC (gcc) [M utils.so]           utils.o
utils.c: In function ‘w_http_query’:
utils.c:469:2: error: too many arguments to function ‘http_query’
  return http_query(_m, _url, _result, NULL, NULL);
  ^
In file included from utils.c:50:0:
functions.h:43:5: note: declared here
 int http_query(struct sip_msg* _m, char* _url, char* _dst, char* _post);
     ^
utils.c: In function ‘w_http_query_post’:
utils.c:476:2: error: too many arguments to function ‘http_query’
  return http_query(_m, _url, _result, _post, NULL);
  ^
In file included from utils.c:50:0:
functions.h:43:5: note: declared here
 int http_query(struct sip_msg* _m, char* _url, char* _dst, char* _post);
     ^
utils.c: In function ‘w_http_query_post_hdr’:
utils.c:483:2: error: too many arguments to function ‘http_query’
  return http_query(_m, _url, _result, _post, _hdr);
  ^
In file included from utils.c:50:0:
functions.h:43:5: note: declared here
 int http_query(struct sip_msg* _m, char* _url, char* _dst, char* _post);
     ^
make[1]: *** [utils.o] Error 1
make: *** [install-modules] Error 1

miconda:
Is it the git master branch?

If you cherry-picked in an older branch, then you may need to pick older commits to utils in master branch. You can eventually try by copying full utils module from master to older branch, but it might not work either. Test with pure master branch first.

PR #326, opened 2015-09-15
ims_auth: fixed crash when receiving a MAA without SIP-Number-Auth-It…

grumvalski:
This fix prevent kamailio from crashing when MAA is missing SIP-Number-Auth-Items or SIP-Auth-Data-Item AVPs.


miconda:
@jaybeepee, @richardgood and @ngvoice should check this pull request.

PR #328, opened 2015-09-17
modules/path: fixes typo in outbound RFC number

mslehto:


mslehto:
huh, looks like travic ci job was started for this 1 character documentation change.
I hope it passes :)

miconda:
Thanks! (finish keyboards are having dangerous chars, so travis ci needs to validate everything in this case :-) )

PR #329, opened 2015-09-18
registrar: add contact and received info to xavp_rcd

linuxmaniac:
set xavp_rcd at registered() too


miconda:
Thanks! It can be merged.

PR #330, opened 2015-09-18
Fix crash on applications memory reservation

fedefava86:
Fix a crash on applications shared memory allocation
when Diameter CEA contains Auth_Applications or Acct_Applications
AVPs not "embedded" into Vendor_Specific grouped AVP
Before only the memory for applications found on Vendor_Specific
AVP was reserved, but if Acct_Applications or Auth_Applications
not Vendor Specific is found, you need to reserve memory for
the number of Acct_Applications or Auth_Applications AVPs found
multiplied by the number of Supported Vendor ID AVPs found.

### DIFF
--- a/modules/cdp/diameter_avp.c
+++ b/modules/cdp/diameter_avp.c
@@ -261,7 +261,6 @@ AAA_AVP  *AAAFindMatchingAVP(
 
 	/* param checking */
 	if (!msg) {
-		LM_ERR("FindMatchingAVP: param msg passed null !!\n");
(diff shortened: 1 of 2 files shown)


ngvoice:
Hi,

can you change the commit message according to
http://www.kamailio.org/wiki/devel/git-commit-guidelines#commit_message

Thanks,
Carstne

fedefava86:
Done, 

Thanks to you,

Federico Favaro

PR #334, opened 2015-09-23
tmx: safety check for bad callid header in t_precheck_tran()

zxcpoiu:
- reported by GH#331


miconda:
Thanks! It will be merged.

PR #344, opened 2015-09-24
websocket: Add support for SIP message fragmentation

que273:
- websocket connections are created with a BUF_SIZE (64K) buffer used for concatenating frames
- continuation frames (fragments) are supported for the SIP sub-protocol


miconda:
You can merge it and you can directly push the commits to websocket module, without doing push requests -- guess @pdunkley is ok with that.

que273:
Thanks!
For some features, I like to check with other devs that I'm not in conflict with the existing style or causing any unintended consequences!

miconda:
OK, that's fine, just wanted to say that if you feel it should be pushed directly, then it is ok to do it.

PR #346, opened 2015-09-28
tm: fake msg and environment to append new branches

grumvalski:
This fix tries to solve the crash reported in http://lists.sip-router.org/pipermail/sr-users/2015-September/089846.html. The cause of the crash was that the sip message used to append branch was the shmem clone of the original message. When the new branches were going through their configured  branch route, modifications done by other modules' function (like rtpproxy) were applied to this stored copy directly, causing the crash once a failure route was run.
This patch take the same approach as the failure handler one, creating a fake request from the origin message and using this one to append branches. I am not sure if this is a good approach, since the appended branch are going through the branch route and some function called there could not be compatible with a faked request.

### DIFF
--- a/modules/tm/t_append_branches.c
+++ b/modules/tm/t_append_branches.c
@@ -49,6 +49,8 @@
 int t_append_branches(void) {
 	struct cell *t = NULL;
 	struct sip_msg *orig_msg = NULL;
+	static struct sip_msg faked_req;
+	
 	short outgoings;
 
 	int success_branch;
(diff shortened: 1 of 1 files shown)


miconda:
Not much time available today, but I am going to check it sometime soon.

miconda:
Can be merged.

Anyhow, wondering if would be just enough to use a local temporary msg structure, will be simpler, but then if branch route is executed, some context attributes won't be there... I will try to analyze a bit more the whole concept with faking the environment, because sometimes could be just too complex for what exactly is needed.

grumvalski:
Thanks for looking at it Daniel. If you prefer I can wait your analysis before merging.

miconda:
If it fixes things now, merge it, refactoring can be done later.

Overall, is about not getting to fake the env too many times if same env can be re-used -- again, I need to look at other parts, it actually may not be the case for this situation.

PR #348, opened 2015-09-29
core: Fix label for RPC command ("proto" -> "port")

neuhaus:
the "port" value mistakenly had the label "proto".


miconda:
Thanks!

PR #351, opened 2015-10-02
acc_radius : more logging for acc_radius module

ycaner06:
added more log about radius result. it makes better to understand problem. i checked freeradiusclient.h and radiusclient-ng.h to have all defined variables.


miconda:
In my opinion there should be also a last 'else' branch just in case any of those library introduce a return code that you don't handle right now. You can say 'unknown response code' and print the value of the response code along with the message. You can add an additional patch for it to this pull request.

ycaner06:
Hmm. I missed that one, soryy for that . i added else branch. 
Thanks.

PR #353, opened 2015-10-02
misc_radis : more logging

ycaner06:
it helps to understand misc radius errors. i checked "REJECT_RC" never defined.


miconda:
Is it a typo in log messages "Authz ..." instead of "Auth ..."?

ycaner06:
misc radius works like "authorization" in AAA. so that abbreviation is "authz" in internet. Authentication can be shortened lie "AuthN". if you want , i can update it.

miconda:
It's ok for me, just wanted to be sure was not a type, then copy&paste propagation.

PR #355, opened 2015-10-05
acc: add acc_prepare_always parameter

linuxmaniac:
This will force the preparation of the request no matter if the acc_prepare_flag is set at the moment of the transaction creation.

Following @miconda idea at: http://lists.sip-router.org/pipermail/sr-users/2015-July/089191.html


miconda:
Thanks!

PR #357, opened 2015-10-05
misc_radius doc : more information freeradius-client

ycaner06:
added information about how to active freeradius-client library

### DIFF
--- a/modules/misc_radius/doc/misc_radius_admin.xml
+++ b/modules/misc_radius/doc/misc_radius_admin.xml
@@ -94,12 +94,20 @@
 			</para>
 			<itemizedlist>
 				<listitem>
-				<para><emphasis>radiusclient-ng</emphasis> 0.5.6 or higher -- 
+				<para><emphasis>radiusclient-ng</emphasis> 0.5.0 or freeradius-client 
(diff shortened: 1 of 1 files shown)


oej:
Shall we really remove radiusclient-ng - is it a dead project?

Would it be possible to fix this in the makefile instead of suggesting a manual change?

linuxmaniac:
fix the makefile?
we can switch the default to the newer one but we need a way to switch in order to support older distros

We already use this:
https://github.com/kamailio/kamailio/blob/master/pkg/kamailio/deb/jessie/rules#L23-L24

oej:
So why are the doc files talking about radiusclient-ng and not FreeRadius?

oej:
http://kamailio.org/docs/modules/devel/modules/auth_radius.html

miconda:
I am not familiar with the distribution of freeradius client library, I know that at some moment was only available in latest opensuse or fedora. I am fine to switch to freeradius client lib as long as the latest stable popular os distros have it.

miconda:
Merging the pull request, to get the commit in. If there is anything else to be sorted out, a new issue can be created.

PR #361, opened 2015-10-07
mem/q_malloc: log more diagnostic information in qm_debug_frag

alexdowad:
I had an instance of SER crash in production with this error message from
qm_debug_frag:

[mem/q_malloc.c:149]: BUG: qm_*: fragm. 0x4cee3b88 (address 0x4cee3ba0) end
overwritten(0, 0)!

The fragment end was been overwritten with zeroes... buffer overflow? Or an
aliasing bug? Maybe a double-free?

It would be much easier to debug problems like this if qm_debug_frag printed the
call site where qm_malloc/free was invoked from. And now it does!

### DIFF
--- a/mem/q_malloc.c
+++ b/mem/q_malloc.c
@@ -112,32 +112,37 @@ inline static unsigned long big_hash_idx(unsigned long s)
 #define END_CHECK_PATTERN2 0xabcdefed
 
 
-static  void qm_debug_frag(struct qm_block* qm, struct qm_frag* f)
+static  void qm_debug_frag(struct qm_block* qm, struct qm_frag* f,
+			      const char* file, unsigned int line)
 {
 	if (f->check!=ST_CHECK_PATTERN){
 		LOG(L_CRIT, "BUG: qm_*: fragm. %p (address %p) "
-				"beginning overwritten(%lx)!\n",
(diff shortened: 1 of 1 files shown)


miconda:
Are you running old ser version? Or you just referred to it as SER, if yes, what version are you running?

alexdowad:
Hi, @miconda. In production, I'm running Kamailio 3.2.3 (59f87e). I just called it SER because the `make_and_install` build script generates a binary called `ser`.

miconda:
If it would be a double free, then that should be caught earlier, iirc. Likely to be a buffer overflow, which may be fixed meanwhile, because 3.2.3 is rather old, you should at least upgrade to latest 3.2.x.

If you got a coredump, then print the content of the fragment to see what piece of code allocated it.

Regarding the PR, the commit message is not strictly related to the code changes, but the discussion here, so it doesn't really suits for a commit history. Once I get the time to check it and all ok, then I will apply it manually, changing the commit message to be relevant for the code changes, according to:

alexdowad:
Thanks!

miconda:
The commit was pushed manually (7c9a87948d9db5be511e94877c1eb3cdc411622c), with commit message updated to reflect only the source code changes.

alexdowad:
Thank you.

PR #375, opened 2015-10-21
acc_radius : adding radius time mode

ycaner06:
Hello;
with adding microseconds to Event-Time-Stamp , it will be more specific time. radiusclient library doesnt support double so i had to convert from double to char variable. in addition , after changing radius time mode to 1 , Event-Time-Stamp attribute needs to be switched integer/date to string in dictionary.kamailio and also radius server dictionary.
if you accept this request , i will add more information to radius docs.
Thanks.

### DIFF
--- a/modules/acc_radius/acc_radius_mod.c
+++ b/modules/acc_radius/acc_radius_mod.c
@@ -74,6 +74,7 @@ static char *radius_config = 0;
 int radius_flag = -1;
 int radius_missed_flag = -1;
 static int service_type = -1;
+int rad_time_mode=0;
 void *rh;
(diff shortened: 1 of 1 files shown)


ycaner06:
any comment please?

miconda:
The message of commit is really inappropriate:

  * https://github.com/ycaner06/kamailio/commit/065271f93947c032ddcfd48da1bf1ff28af636f9.patch

See:

Also, declare the variables at beginning of functions or blocks, to be able to compile on strict C compilers.

ycaner06:
i will close this PR and reopen again because i am not  able to change commit message.

Issue #379, opened 2015-10-23
inline function suip2a() (ip_addr.h) not working for ipv6


topotomaso:
I observed that the inline function suip2a() (ip_addr.h) contains a silly bug and therefor does not work for IPv6 conversions. This silly bug is preventing sip_trace() to work in the onsend_route block for IPv6 sip messages. 
Rem.: The function suip2a() is used only by the siptrace module.

Test background (just for understanding the use case):
- using kamailio 4.1.3 (with an additional patch to let onsend_route be called also for replies taken from kamailio 4.2.x)
- kamailio is used as loadbalancer in stateless mode (using forward() only)
- sip tracing is used by explictly calling sip_trace() in onsend_route, route and onreply routing block to be able to implement an interface specific sip tracing (no SIP transaction based sip tracing is applied by using the siptrace's trace_flag)

The silly bug:

In suip2a() the function ip6tosbuf() is called with an odd buffer length argument of value "sizeof(buf)-4" which is equal to "IP6_MAX_STR_SIZE -1". ip6tosbuf() however needs at least a buffer length of IP6_MAX_STR_SIZE (otherwise it returns immediately without writing anything into the buffer).
So regarding IPv6 adresses the ascii output of suip2a() is alway an empty IPv6 reference ([]).

```
static inline int ip6tosbuf(unsigned char* ip6, char* buff, int len)
{
	int offset;
	register unsigned char a,b,c;
	register unsigned char d;
	register unsigned short hex4;
	int r;
	#define HEXDIG(x) (((x)>=10)?(x)-10+'A':(x)+'0')
	
	
	offset=0;
	if (unlikely(len<IP6_MAX_STR_SIZE))
		return 0;
	...
}

#define SUIP2A_MAX_STR_SIZE  (IP6_MAX_STR_SIZE + 2 /* [] */ + 1 /* \0 */)
/* returns an asciiz string containing the ip
 *  (<ipv4_addr> or [<ipv6_addr>])
 */
static inline char* suip2a(union sockaddr_union* su, int su_len)
{
	static char buf[SUIP2A_MAX_STR_SIZE];
	int offs;

	if (unlikely(su->s.sa_family==AF_INET6)){
		if (unlikely(su_len<sizeof(su->sin6)))
			return "<addr. error>";
		buf[0]='[';
		offs=1+ip6tosbuf((unsigned char*)su->sin6.sin6_addr.s6_addr, &buf[1],
							sizeof(buf)-4);
[...]

PR #381, opened 2015-10-26
core: fixed passed buffer size in suip2a()

topotomaso:
- fix regarding issue #379
- it makes sense to backport this fix to 4.2 and 4.3
- Function suip2a() has never worked properly for IPv6 addresses before.
  The passed buffer size 'sizeof(buf)-4' to ip6tosbuf() was 1 byte to small.
  So ip6tosbuf() has never written any IPv6 ASCII string to the buffer and
  therefor suip2a() always returned an empty IPv6 reference string ("[]").
  Fixed this by changing passed buffer size to IP6_MAX_STR_SIZE.
  For similarity also changed uncritical passed buffer size in call to
  ip4tosbuf() to IP4_MAX_STR_SIZE.
  The function suip2a() is only used by siptrace module when sip_trace()
  script function is explicitly called on sending side e.g. in onsend_route
  block.

### DIFF
--- a/ip_addr.h
+++ b/ip_addr.h
@@ -741,14 +741,14 @@ static inline char* suip2a(union sockaddr_union* su, int su_len)
 			return "<addr. error>";
 		buf[0]='[';
 		offs=1+ip6tosbuf((unsigned char*)su->sin6.sin6_addr.s6_addr, &buf[1],
-							sizeof(buf)-4);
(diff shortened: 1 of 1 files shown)


miconda:
Applied the patch manually, the address inside it being a localhost. The commit message was simplified.

PR #382, opened 2015-10-27
permissions: possibility to specify time period in routing rules, fun…

Tetram100:
…ction to generate simple call logs

* Possibility to specify a time period following the RFC 2445 (explained in the TMREC module readme)
 at the end of a rule in the allow and deny files. The rule will be valid only during this period.
Usefull for company or people that don't want calls during night or want to receive calls from someone
only during a specific period of time. It was already possible with the TMREC module in a routing
script but this was redundant with the permission module so it simplifies the conf.

* New function log_attempt that can be use in the routing part. Usefull to log in a specific file the
calls that have been prevented to do security inspection after. Was surely possible with a script in
the routing part but it simplifies also the conf.

### DIFF
--- a/modules/tm/t_cancel.c
+++ b/modules/tm/t_cancel.c
@@ -43,8 +43,8 @@ typedef struct cancel_reason_map {
 } cancel_reason_map_t;
 
 static cancel_reason_map_t _cancel_reason_map[] = {
-	200, str_init("Answered elsewhere"),
-	0, {0, 0}
+	{ 200, str_init("Answered elsewhere") },
+	{ 0, {0, 0} }
 };
 
 /**


miconda:
Thank you for your contribution, however, there are couple of remarks I want to do:

  * patch is rather hard to follow because along with new code you reformatted the old code, resulting in bigger patch
  * reformatting the code to be more standard and remove trailing withe spaces is appreciated, but has to be done as a stand alone patch
  * log_attempt() seems quite redundant to other existing alternatives: xlog() with a particular facility directed to own file or file_write() from corex module in master branch

My suggestions:
  * remove addition of log_attempt() (it can still be discussed if you think so, but in that case make a separate pull request for it)
  * make a patch only for the new functionality with timerec
  * make a patch for reformatting the code/removing training white spaces
  * the last two patches can be in the same pull request

Tetram100:
Hi,

* I agree with your point on the log_attempt. In fact I didn't see that the corex module has now a file_write() function that can be use very easily to do the same thing.
* I will do separate patchs for the 2 things you mention

PR #383, opened 2015-10-28
tls: fixed typo param ssl_freelist_max_len mismatch with doc

zxcpoiu:
tls: fixed typo param ssl_freelist_max_len mismatch with doc


miconda:
Thanks!

PR #387, opened 2015-10-28
presence_xml: add optional param force_dummy_presence

lazedo:
- added module parameter to send empty presence notify body
  when there isn't any record in presentity


miconda:
It can be merged.

Can you extend the docs a bit and say there that 'status open' is sent with this notification? That seems to be the case based on the added XML in the code.

lazedo:
@miconda added the 'status open' to the doc. Thanks

PR #388, opened 2015-10-29
Allow sipcapture callid_aleg_header to accept a list of headers

giavac:
Headers must be separated by ';', and first match wins (so order in the list is important).
With this patch Homer's sipcapture can correlate A-leg and B-leg even when applications in the network use different custom headers to communicate on B-leg what A-leg's Call ID is.
No changes are needed on DB schema or homer API.

### DIFF
--- a/modules/sipcapture/README
+++ b/modules/sipcapture/README
@@ -45,6 +45,7 @@ Alexandr Dubovikov
               3.16. capture_node (str)
               3.17. insert_retries (integer)
               3.18. insert_retry_timeout (integer)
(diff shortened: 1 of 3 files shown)


miconda:
Added @adubovikov in the loop here.

For the future, use the suggestions for writing the commit message presented at:

It makes the changeloc coherent and easier to search changes in it for a specific component.

giavac:
> For the future, use the suggestions for writing the commit message presented at:
Sorry - sure, will do.

PR #389, opened 2015-11-04
set the log_prefix buffer size to 1024b

gled-rs:
seems to be running nicely here.

to be tested more though.


miconda:
Going to merge, for the future format the commit message as suggested at:

PR #390, opened 2015-11-05
rtpengine: hash table to keep the selected nodes

smititelu:
Shared memory hash table with global hashtable lock.
Add state maintaining the selected rtp node, for a given callid.
Hashtable entry expiration time configurable using hash_entry_tout modparam.
The actual deletion happens on the fly while insert/remove/lookup are called.
Updated doku.

### DIFF
--- a/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/modules/rtpengine/doc/rtpengine_admin.xml
@@ -73,6 +73,38 @@
 		If the set was selected using setid_avp, the avp needs to be
 		set only once before rtpengine_offer() or rtpengine_manage() call.
 	</para>
+	<para>
+		From the current implementation point of view, the sets of rtpproxy nodes
+		are shared memory(shm), so all processes can see a common list of nodes.
+		There is no locking when setting the nodes enabled/disabled (to keep the
+		memory access as fast as possible). Thus, problems related to node state
+		might appear for concurent processes that might set the nodes
+		enabled/disabled(e.g. by fifo command). This robustness problems are overcomed as follows.
+	</para>
+
+	<para>
+		If the current process sees the selected node as disabled, the node is
+		<emphasis>force tested</emphasis> before the current process actually
+		takes the disabled decision. If the test succeeds, the process will set
+		the node as enabled (but other concurrent process might still see it as disabled).
+.
+	</para>
+
+	<para>
+		If the current process sees the selected node as enabled, it does no additional checks
+		and sends the command which will fail in case the machine is actually broken.
+		The process will set the node as disabled (but other concurrent process might still see it as enabled).
+	</para>
+
+	<para>
+		The 'kamctl fifo' commands (including rtpengin ones) are executed by an exclusive
+		process which operate on the same shared memory node list.
+	</para>
+
+	<para>
+		All the nodes are pinged in the beginning by all the processes,
+		even if the node list is shared memory.
+	</para>
 	</section>
 
 	<section>
--- a/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/modules/rtpengine/doc/rtpengine_admin.xml
@@ -356,6 +356,32 @@ modparam("rtpproxy", "rtp_inst_pvar", "$avp(RTP_INSTANCE)")
 		</example>
 	</section>
 
+	<section id="rtpengine.p.hash_entry_tout">
+		<title><varname>hash_entry_tout</varname> (string)</title>
+		<para>
+			Number of seconds after an rtpengine hash table entry is marked for deletion.
+			By default, this parameter is set to 120 (seconds).
+		</para>
+		<para>
+			To maintain information about a selected rtp machine node, for a given call, entries are added in a hashtable of (callid, node) pairs.
+			When "offer" comes, insert new entry in the hastable.
+			When subsequent commands come, lookup callid and return chosen node.
+			When "delete" comes, remove old entry from hashtable.
+		</para>
+		<para>
+			NOTE: In the current implementation, the actual deletion happens <emphasis>on the fly</emphasis>,
+			while insert/remove/lookup the hastable, <emphasis>only</emphasis> for the entries in the insert/remove/lookup path.
+		</para>
+		<example>
+		<title>Set <varname>hash_entry_tout</varname> parameter</title>
+<programlisting format="linespecific">
+...
+modparam("rtpproxy", "hash_entry_tout", "300")
+...
+</programlisting>
+		</example>
+	</section>
+
 	</section>
 
 	<section>
--- a/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/modules/rtpengine/doc/rtpengine_admin.xml
@@ -388,8 +388,26 @@ modparam("rtpproxy", "rtp_inst_pvar", "$avp(RTP_INSTANCE)")
 		</example>
 	</section>
 
-	<section id="rtpengine.p.hash_entry_tout">
-		<title><varname>hash_entry_tout</varname> (string)</title>
+	<section id="rtpengine.p.hash_table_size">
+		<title><varname>hash_table_size</varname> (integer)</title>
+		<para>
+			Size of the hash table. Default value is 256.
+		</para>
+		<para>
+			NOTE: If configured size is <emphasis>less than</emphasis> 1, the size will be defaulted to 1.
+		</para>
+		<example>
+		<title>Set <varname>hash_table_size</varname> parameter</title>
+<programlisting format="linespecific">
+...
+modparam("rtpproxy", "hash_table_size", "123")
+...
+</programlisting>
+		</example>
+	</section>
+
+	<section id="rtpengine.p.hash_table_tout">
+		<title><varname>hash_table_tout</varname> (integer)</title>
 		<para>
 			Number of seconds after an rtpengine hash table entry is marked for deletion.
 			By default, this parameter is set to 120 (seconds).
@@ -405,14 +423,15 @@ modparam("rtpproxy", "rtp_inst_pvar", "$avp(RTP_INSTANCE)")
 			while insert/remove/lookup the hastable, <emphasis>only</emphasis> for the entries in the insert/remove/lookup path.
 		</para>
 		<example>
-		<title>Set <varname>hash_entry_tout</varname> parameter</title>
+		<title>Set <varname>hash_table_tout</varname> parameter</title>
 <programlisting format="linespecific">
 ...
-modparam("rtpproxy", "hash_entry_tout", "300")
+modparam("rtpproxy", "hash_table_tout", "300")
 ...
 </programlisting>
 		</example>
 	</section>
+
 
 	</section>
 
--- a/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/modules/rtpengine/doc/rtpengine_admin.xml
@@ -210,6 +210,30 @@ modparam("rtpengine", "rtpengine_tout_ms", 2000)
 </programlisting>
 		</example>
 	</section>
+	<section id="rtpengine.p.rtpengine_allow_op">
+		<title><varname>rtpengine_allow_op</varname> (integer)</title>
+		<para>
+		Enable this to allow finishing the current sessions while denying new sessions for the
+		<emphasis>manually deactivated nodes </emphasis> via kamctl command i.e. "disabled(permanent)" nodes.
+		Probably the manually deactivated machine is still running(did not crash).
+		</para>
+		<para>
+		This is <emphasis>useful</emphasis> when deactivating a node for maintanance and reject new sessions but allow current ones to finish.
+		</para>
+		<para>
+		<emphasis>
+		Default value is <quote>0</quote> to keep the current behaviour.
+		</emphasis>
+		</para>
+		<example>
+		<title>Set <varname>rtpengine_allow_op</varname> parameter</title>
+		<programlisting format="linespecific">
+...
+modparam("rtpengine", "rtpengine_allow_op", 1)
+...
+</programlisting>
+		</example>
+	</section>
 	<section id="rtpengine.p.queried_nodes_limit">
 		<title><varname>queried_nodes_limit</varname> (integer)</title>
 		<para>
--- a/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/modules/rtpengine/doc/rtpengine_admin.xml
@@ -1041,6 +1041,23 @@ $ &ctltool; fifo nh_ping_rtpp all
 			</programlisting>
 			</example>
 		</section>
+
+	    <section id="rtpengine.m.nh_show_hash_total">
+			<title><function moreinfo="none">nh_show_hash_total</function></title>
+			<para>
+				Print the total number of hash entries in the hash table at a given moment.
+			</para>
+			<example>
+			<title>
+				<function moreinfo="none">nh_show_hash_total</function> usage</title>
+			<programlisting format="linespecific">
+...
+$ &ctltool; fifo nh_show_hash_total
+...
+			</programlisting>
+			</example>
+		</section>
+
 	</section>
 
 </chapter>
--- a/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/modules/rtpengine/doc/rtpengine_admin.xml
@@ -361,7 +361,7 @@ modparam("rtpengine", "read_sdp_pv", "$var(sdp)")
 route {
 	...
 	$var(sdp) = $rb + "a=foo:bar\r\n";
-	rtpproxy_manage();
+	rtpengine_manage();
 }
 </programlisting>
 		</example>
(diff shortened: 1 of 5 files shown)


smititelu:
Basically tested it, with focus on entries expiration. Further tests will follow.

What do you think about current expired entries deletion implementation? Do you have any other advices/proposals?

smititelu:
Updated pull request. Added some modparams and kamctl for the hashtable. Added a modparam to allow session deletion to manually disabled nodes (useful when performing maintenance). Updated doku.

In my opinion tests were ok. Tested this for 100 calls with media, 10 calls/s rate. Sessions were deleted successfully upon BYE. No memory leaks found at the hash table level (used "kamcmd mod.stats rtpengine shm" for this).

miconda:
What would happen when there is a restart?

I find the feature useful overall, any limitations need to be documented when it is enabled.

Hopefully @rfuchs can comment from his point of view and merge when all is ok.

smititelu:
On a restart (i.e. rtpengine's mod_destroy() is called), rtpengine_hash_table_destroy() is called which frees the entire shm hashtable.

Possible limitations:
- overhead added by the rtpengine hashlock (i.e. decrease in calls/sec rate)
- more shm used (i.e. one has to increase the shm using kamailio's "-m" parameter)

What is the current known limitation for kamailio (calls/sec) ?

miconda:
I was looking more at the perspective of the items in hash table being lost and if the restart was in a middle of the call, then after restart, when the BYE is processed, which rtpengine is updated?

smititelu:
Hmm.. I think I get the point. If the kamailio restarts in the middle of a call, the hashtable is flushed. When the BYE comes, no entry is found for that callid and no rtpengine is updated.

smititelu:
One approach to this would be to fallback to the stateless behaviour when no entry is found in the hashtable and just select a rtpengine based on the hashed callid (with 100% matching the rtpengine machine for that call if none crashed).

modules/rtpengine/rtpengine_hash.c
```diff
@@ -0,0 +1,359 @@
...
+
+	// check rtpengine hashtable
+	if (!rtpengine_hash_table) {
+		LM_ERR("NULL rtpengine_hash_table");
+		return 0;
+	}
+
+	// get entry list
+	hash_index = str_hash(key);
+	entry = rtpengine_hash_table->entry_list[hash_index];
```
rfuchs: Don't you need to acquire the lock before accessing the start element for the list?

rfuchs: Nevermind, I see there's an empty list head. Retracted.

modules/rtpengine/rtpengine.c
```diff
@@ -2191,81 +2272,176 @@ static struct rtpp_set * select_rtpp_set(int id_set ){
...
+		LM_ERR("rtpengine hash table fail to duplicate calllen=%d callid=%.*s",
+			callid.len, callid.len, callid.s);
+	}
+	entry->node = node;
+	entry->next = NULL;
+	entry->tout = get_ticks() + hash_table_tout;
+
+	/* Insert the key<->entry from the hashtable */
+	if (!rtpengine_hash_table_insert(&callid, entry)) {
+		LM_ERR("rtpengine hash table fail to insert node=%.*s for calllen=%d callid=%.*s",
```
rfuchs: The error case needs some shm_free() I think

modules/rtpengine/rtpengine.c
```diff
@@ -2191,81 +2272,176 @@ static struct rtpp_set * select_rtpp_set(int id_set ){
...
-		return node->rn_disabled ? NULL : node;
-	}
+	unsigned i, sum, sumcut, weight_sum;
+	int was_forced = 0;
 
 	/* XXX Use quick-and-dirty hashing algo */
-	for(sum = 0; callid.len > 0; callid.len--)
-		sum += callid.s[callid.len - 1];
+	for(i = 0; i < callid.len; i++)
+		sum += callid.s[i];
```
rfuchs: "sum" is used uninitialized here

modules/rtpengine/rtpengine.c
```diff
@@ -2191,81 +2272,176 @@ static struct rtpp_set * select_rtpp_set(int id_set ){
...
+select_rtpp_node(str callid, int do_test, int op)
+{
+	if(!active_rtpp_set) {
+		LM_ERR("script error - no valid set selected\n");
+		return NULL;
+	}
+
+	// calculate and choose a node
+	if (op == OP_OFFER) {
+		// run the selection algorithm
```
rfuchs: "Offer" doesn't necessarily mean that this will be a new entry. A re-invite will also do an "offer" but needs to do a lookup into the hash table first. I think a lookup should always be done first, with fallback to creating a new entry in case of an offer.

modules/rtpengine/rtpengine.c
```diff
@@ -2191,81 +2272,176 @@ static struct rtpp_set * select_rtpp_set(int id_set ){
...
 	if (do_test) {
 		node->rn_disabled = rtpp_test(node, node->rn_disabled, 0);
 		if (node->rn_disabled)
 			goto retry;
 	}
+
+	/* build hash table entry */
+	struct rtpengine_hash_entry *entry = shm_malloc(sizeof(struct rtpp_node));
+	if (shm_str_dup(&entry->callid, &callid) < 0) {
+		LM_ERR("rtpengine hash table fail to duplicate calllen=%d callid=%.*s",
```
rfuchs: Function should abort here, otherwise risking segfault

modules/rtpengine/rtpengine.c
```diff
@@ -2191,81 +2272,176 @@ static struct rtpp_set * select_rtpp_set(int id_set ){
...
+
 found:
 	if (do_test) {
 		node->rn_disabled = rtpp_test(node, node->rn_disabled, 0);
 		if (node->rn_disabled)
 			goto retry;
 	}
+
+	/* build hash table entry */
+	struct rtpengine_hash_entry *entry = shm_malloc(sizeof(struct rtpp_node));
```
rfuchs: Return value of shm_malloc should be error checked (might be NULL)

modules/rtpengine/rtpengine_hash.c
```diff
@@ -0,0 +1,359 @@
...
+
+	// lock
+	lock_get(rtpengine_hash_lock);
+	while (entry) {
+		// if key found, return entry
+		if (str_equal(&entry->callid, (str *)key)) {
+			// unlock
+			lock_release(rtpengine_hash_lock);
+
+			return entry;
```
rfuchs: Returning the raw hash table entries straight from the hash table is dangerous as the entries are protected by the lock and so this opens the door to a race condition. A lookup might return one entry while the same entry is being freed in another process. Either return a copy of the entries or simply return its contents (the node should be enough).

modules/rtpengine/rtpengine_hash.h
```diff
@@ -0,0 +1,31 @@
...
+/* table */
+struct rtpengine_hash_table {
+	struct rtpengine_hash_entry **entry_list;	// hastable
+	unsigned int total;				// total number of entries in the hashtable
+};
+
+
+int rtpengine_hash_table_init(int size);
+int rtpengine_hash_table_destroy();
+int rtpengine_hash_table_insert(void *key, void *value);
```
rfuchs: I don't think there's a good reason to use void pointers here, since the types are known and using void pointers prevents the compiler from catching type errors.

modules/rtpengine/rtpengine_hash.c
```diff
@@ -0,0 +1,359 @@
...
+		LM_ERR("no shm left to create rtpengine_hash_table\n");
+		return 0;
+	}
+
+	// init hashtable entry_list
+	rtpengine_hash_table->entry_list = shm_malloc(hash_table_size * sizeof(struct rtpengine_hash_entry));
+
+	// init hashtable entry_list[i] (head never filled)
+	for (i = 0; i < hash_table_size; i++) {
+		rtpengine_hash_table->entry_list[i] = shm_malloc(sizeof(struct rtpengine_hash_entry));
```
rfuchs: I don't think shm_malloc() automatically zeroes the memory, and the lookup routines seem to do a strcmp on the head entries call-ID too, so the call-ID needs to be initialized as well.

modules/rtpengine/doc/rtpengine_admin.xml
```diff
@@ -356,6 +412,51 @@ modparam("rtpproxy", "rtp_inst_pvar", "$avp(RTP_INSTANCE)")
...
+			Size of the hash table. Default value is 256.
+		</para>
+		<para>
+			NOTE: If configured size is <emphasis>less than</emphasis> 1, the size will be defaulted to 1.
+		</para>
+		<example>
+		<title>Set <varname>hash_table_size</varname> parameter</title>
+<programlisting format="linespecific">
+...
+modparam("rtpproxy", "hash_table_size", "123")
```
rfuchs: Should be "rtpengine" not "rtpproxy" :)

modules/rtpengine/rtpengine.c
```diff
@@ -228,6 +235,8 @@ static pid_t mypid;
 static unsigned int myseqn = 0;
 static str extra_id_pv_param = {NULL, 0};
 static char *setid_avp_param = NULL;
+static int hash_table_tout = 120;
```
rfuchs: Is a default value of 2 minutes really all that useful? If a call lasts longer than 2 minutes, the final bye/delete won't be delivered to the correct RTP proxy any more, or do I misunderstand this value?

modules/rtpengine/rtpengine.c
```diff
@@ -1957,12 +2027,23 @@ static bencode_item_t *rtpp_function_call(bencode_buffer_t *bencbuf, struct sip_
...
-        goto error;
+		goto error;
 	}
 
 	if (body_out)
 		*body_out = body;
 
+	if (op == OP_DELETE) {
+		/* Delete the key<->value from the hashtable */
+		if (!rtpengine_hash_table_remove(&callid)) {
```
rfuchs: I don't think immediately deleting the hash table entry on "delete" is a good idea as it may only be a single branch that has been deleted and a "delete" does not preclude the possibility of more commands (offer, answer, or something else) coming for the same call-ID. Perhaps resetting the expiry value for the entry to something small might be a better idea.

gaaf:
Why not just require the dialog module and store the selected rtpproxy in the dialog?

What happens on spiraled/hairpin calls? The call-id alone does not seem specific enough.

miconda:
I would not like a dependency on dialog module, it is quite common to run rtp relay on edge proxy which is typically more or less a lightweight load balancer -- dialog management on those nodes will add a lot of overhead.

Fallback to hasing callid after a restart seams reasonable for me (giving the current behaviour).

For spiralled calls, iirc, there is an option to provide via branch to rtpproxy/rtpengine, but this is independent of selecting what rtpproxy/rtpengine to use.

smititelu:
We also thought of using dialog module for keeping the node. The main reason keeping the rtpengine module as independent as it could be. Maybe for spiraled calls this could be extended to consider also the from/to-tag when calculating the hash index ?!

juha-h:
Alex writes:

> Why not just require the dialog module and store the selected rtpproxy
> in the dialog?

dialog module is a complicated mess and its use cannot be warranted just
for enabling this feature.

-- juha

gaaf:
So, the dialog module is a no-go/

Have you considered just putting the selected rttproxy in an RR parameter? For me that has worked perfectly for years now.

juha-h:
Alex writes:

> Have you considered just putting the selected rttproxy in an RR
> parameter? For me that has worked perfectly for years now.

that is what i'm doing also:

;pm[=setid]

-- juha

miconda:
That is only the setid not the selected rtp relay within the set. Afaik, there is no way to specify the rtp relay within a set from config file.

juha-h:

> That is only the setid not the selected rtp relay within the
> set. Afaik, there is no way to specify the rtp relay within a set from
> config file.

it works if you have only one relay per set.

-- juha

miconda:
My understanding was that the patch was adding support for keeping the relation between the selected rtpengine and the callid, when there are more than one rtpengine per set. If one of the rtpengine in the set is unavailable, then the current hashing selection becomes inconsistent if the rtpengine turns back available.

If there is only one rtpengine per set, then no hashing selection is done.

modules/rtpengine/rtpengine.c
smititelu: I understand the problem that may appear in that case you said. However, I don't understand why resetting the expiry so something smaller would solve this. Proposals:
1. reset the timer to the hash_entry_tout value
2. also consider from-tag, to-tag, maybe via-branch id (add them in the hash table entry). Then match them all, when deleting the entry; thus the deletion can be done right-away.

I'd go for 2. What do you think?

modules/rtpengine/rtpengine.c
smititelu: You are right, I'll update this to default 3600 sec. Also document this idea.

modules/rtpengine/rtpengine.c
miconda: Some notes:

  * when session is created, there is no to-tag (initial invite)
  * the via-branch can be different for re-invite or bye, so it cannot be used as part of the key all the time

modules/rtpengine/rtpengine_hash.c
smititelu: True, I'll return just the node.

modules/rtpengine/rtpengine.c
rfuchs: If you can be bothered with 2, then go for it, but I think it would take a lot of work and fine-tuning to make it work just right. The proxies already do all the tag and branch handling so I see no need to duplicate it here.

Setting the expiry to the near future on "delete" would help to keep the hash table small if it was indeed the final BYE, but it would also require resetting the expiry to now+timeout on every operation, i.e. not only when the key is created but on a lookup as well.

Another option would be to extend the control protocol and have the RTP proxy tell the module whether this call is finished now or if it should be kept alive.

smititelu:
Updated pull request. Some of the most important updates:
- free previously allocated shm in case of reached memory limit and solve possible segfaults; tested this using -m 128 and a large table: kamailio didn't crashed when had no memory for the entries and still accepted calls afterwards.
- always lookup hashtable before selecting a new node; this fallbacks to the previous behaviour
- check also for viabranch when matching the call in addition to the callid; tested this for simple calls i.e. viabranch was STR_NULL; _didn't tested it_ for real-life branching scenarios as we are not using it; here I'd need some help from your side, if you use those scenarios and have time for it; IMHO the matching should be fine as I'm checking the viabranch value I get from the struct sip_msg.

smititelu:
**Accidentally** pushed all the commits when trying to "git push upstream p_usrloc_NULL_checks master"; forgot the ":".

Shall I re-push the old files? (can't push -f a reset on previous commit due to locking).

linuxmaniac:
git revert is your friend

linuxmaniac:
Done. I reverted all your commits

smititelu:
Thank you; didn't think of it :)

miconda:
@linuxmaniac -- you should add the commands you executed in the wiki:

  * https://www.kamailio.org/wiki/devel/git-commit-guidelines#useful_commands

It will be useful for devs getting into git.

linuxmaniac:
> you should add the commands you executed in the wiki:
> https://www.kamailio.org/wiki/devel/git-commit-guidelines#useful_commands

Done
https://www.kamailio.org/wiki/devel/git-commit-guidelines#revert_already_pushed_commit

smititelu:
Updated the pull request with focus on the hastable API:
- done **per row** hashtable locking instead of global hashtable locking; this should further decrease the possible limitations of using hashtable locks;
- when enabled, the allow_op modparam will allow sessions to finish for both disabled(permanent) and disabled nodes(they might be disabled due to timeout not to real machine break and one might still want the current sessions to finish);

modules/rtpengine/rtpengine_hash.c
```diff
@@ -0,0 +1,503 @@
...
+		return 0;
+	}
+	memset(rtpengine_hash_table, 0, sizeof(struct rtpengine_hash_table));
+	rtpengine_hash_table->size = hash_table_size;
+
+	// init hashtable row_entry_list
+	rtpengine_hash_table->row_entry_list = shm_malloc(rtpengine_hash_table->size * sizeof(struct rtpengine_hash_entry*));
+	if (!rtpengine_hash_table->row_entry_list) {
+		LM_ERR("no shm left to create rtpengine_hash_table->row_entry_list\n");
+		rtpengine_hash_table_destroy();
```
rfuchs: A minor nitpick: These calls to _destroy() here won't ever actually do anything, because the _sanity_checks() will always fail.

smititelu: Right. The _destroy() should do sanity checking, with memory free, when possible. Will correct this.

rfuchs:
Other than that, looks good to me.

PR #398, opened 2015-11-15
cnxcc: make cnxcc locks re-entrant

grumvalski:
Cnxcc call termination is not working anymore since commit 265dc95aac8eb4b5cee7992dde55ff08c92a7ac5. That commit introduced the call to DLGCB_TERMINATED callbacks for the locally generated BYE, which leads to a deadlock in cnxcc when it tries to end the calls because the credit (money or time) is exhausted. 
This patch introduces a new structure cnxcc_lock which implements re-entrant locks.
As a side question I'm wondering if, since other modules need re-entrant locks, this structure should be defined in lock_ops.h like the other kamailio's locks.

### DIFF
--- a/modules/cnxcc/cnxcc_check.c
+++ b/modules/cnxcc/cnxcc_check.c
@@ -24,9 +24,6 @@
 
 #include <stdio.h>
 
-#include "../../locking.h"
-#include "../../lock_ops.h"
-
 #include "cnxcc_mod.h"
 #include "cnxcc.h"
 #include "cnxcc_check.h"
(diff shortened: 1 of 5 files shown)


caruizdiaz:
Thank you for the path! I've been having no time to fix the issues that has been reported for cnxcc, so this one comes in a perfect time, thanks!

grumvalski:
You're welcome!

grumvalski:
Shouldn't the patch be backported to 4.3, since the issue is present there too?

miconda:
I had it in mind to ask in the morning, when I did some backports, but got caught in other projects. Go ahead and do it.

PR #401, opened 2015-11-18
core: NULL checks for ut.h

smititelu:
NULL checks for shm/pkg dup functions and few others.

### DIFF
--- a/ut.h
+++ b/ut.h
@@ -170,8 +170,7 @@ static char fourbits2char[16] = { '0', '1', '2', '3', '4', '5',
 /* converts a str to an u. short, returns the u. short and sets *err on
  * error and if err!=null
   */
-static inline unsigned short str2s(const char* s, unsigned int len,
-									int *err)
+static inline unsigned short str2s(const char* s, unsigned int len, int *err)
 {
 	unsigned short ret;
 	int i;
@@ -619,6 +618,10 @@ static inline void strlower(str* _s)
 {
 	int i;
 
+	if (_s == NULL) return ;
+	if (_s->len < 0) return ;
+	if (_s->s == NULL) return ;
+
 	for(i = 0; i < _s->len; i++) {
 		_s->s[i] = tolower(_s->s[i]);
 	}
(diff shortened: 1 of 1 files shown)


miconda:
Are these for safety reasons or have you found specific cases in other parts of code affected by not having this checks?

One comment I have is about the str dup functions, when the new check on len==0 results in setting dst s to NULL. Specs for malloc(0) are not very strict, saying that it can return a valid pointer for free(). However, the free(0) should not be done. With kamailio memory managers, pkg/shm malloc 0 returns a pointer of size(void*).

I think would be ok to set dst s to null if src s is null, but if src len is 0 and src s is not null, maybe is better to allocate it...

These are some thoughts, perhaps a deeper review where these str dup functions are used should be done.

smititelu:
One of the problems we met is solved by commit 44fdac69eb864125b4d02af0c650ce735dcf7aa8. Also there are many uses of shm/pkg_str_dup() where the src is not NULL checked right before doing the dup (probably NULL check is done on higher levels but some mistakes might happen).

I think you are right with setting dst->s=NULL only if src->s == NULL. Updated the pull request to respect this.

smititelu:
Considering the following _fallback case_ :
- alloc dst->s = pointer sizeof(void\*)
- dst->len = 0
- return **0**

After skimming the code a little more I came to this for the above case(i.e. src->s is NULL):
- either make dst->s = NULL; dst->len = 0 and return **-1**; this will be caught by shm_str_dup() != 0 and prevent shm_free() a NULL pointer, even though memory managers checks for this and gives L_WARN
- or keep the fallback case; further uses of the dst should be done based on the dst->len (i.e. cmp_str() or LM_ERR("%.*s"))

I'd go for the second option. What do you think?

Also I've seen that the memory manager allocate and 'unsigned int'. So, if shm/pkg_str_dup() is called with src->len < 0 it will probably fail (depending on the negative value). In this case, I'd also force the fallback case.

smititelu:
Found and interesting comment[1]  related to calling memcpy(dst->s, NULL, 0). 

[1] http://stackoverflow.com/questions/5243012/is-it-guaranteed-to-be-safe-to-perform-memcpy0-0-0

miconda:
I guess there are no cases when one would try to duplicate src->s==NULL, or at least not an intentional case. Anyhow, printing a warning instead of a segfault is more friendly for runtime. You can merge if you prefer this version.

smititelu:
Updated pull request to check for the undefined memcpy() behaviour. Will merge after travis checks.

PR #402, opened 2015-11-18
Docs change for dispatcher module, section 3.25

AlessioCasco:
Added a description for the logic used by the "ds_ping_reply_codes" where it doesn't only use client responses to set the status of the backends.

### DIFF
--- a/modules/dispatcher/README
+++ b/modules/dispatcher/README
@@ -675,6 +675,11 @@ Note
    valid response). This parameter can be modified via ser config
    framework.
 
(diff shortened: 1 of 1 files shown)


miconda:
Thanks for contribution!

The changes have to be done inside xml files from modules/dispatcher/doc/. The README file is generated from those one.

Also, the commit message must be formatted as described at:

miconda:
Adjusted the message and applied it manually to master branch.

PR #404, opened 2015-11-19
db_postgres: fix compilation, NTOHLL and NTOHLL provided by OSX 10.10

danirus:


miconda:
Thanks, the pull request was merged manually:

  * changes to git ignore made its own commit
  * patch to db_postgres had the commit message re-using the pull request title, because that is the right format for the commit message

PR #416, opened 2015-11-24
tests/travis: migrate to docker environment

linuxmaniac:
- use jessie as default distribution

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
+sudo: required
 language: c
 compiler:
   - gcc
-  - clang
+services:
+  - docker
 before_install:
-  - sudo apt-get update -qq
-  - sudo apt-get install -y gdebi-core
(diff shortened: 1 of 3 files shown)


linuxmaniac:
- clang build is broken on 4.1 right now https://s3.amazonaws.com/archive.travis-ci.org/jobs/92951963/log.txt from https://travis-ci.org/kamailio/kamailio/jobs/92951963

@miconda Should I just disable clang build for 4.1?

miconda:
That error is probably only for 4.1, because for recent versions we renamed the 'printf' member of rpc API structure. Can you disable compiling with clang only on branch 4.1?

linuxmaniac:
Yes, this last version only builds gcc

miconda:
OK -- missed your explicit note about 4.1 -- clang compilation can be disabled for it.

PR #420, opened 2015-11-27
tmx: Fix relayed_replies counter

smititelu:
Consider the simple timeout scenario:
```
                     Messages
INVITE ---------->      1             
407 <----------         1   
ACK ---------->         1              
INVITE ---------->      1           
100 <----------         1      
180 <----------         0    
183 <----------         0       
408 <----------         1
ACK ---------->         1
```   
In the current implementation, running _kamctl fifo get_statistics tmx:_ on a fresh kamailio prints:
```
...
tmx:local_replies = 2
tmx:received_replies = 1
tmx:relayed_replies = 18446744073709551615
```
After this patch:
```
...
tmx:local_replies = 2
tmx:received_replies = 1
tmx:relayed_replies = 0
```
Fix the above by counting the relayed_total and substracting the relayed_local.

### DIFF
--- a/modules/tm/t_reply.c
+++ b/modules/tm/t_reply.c
@@ -555,6 +555,7 @@ static int _reply_light( struct cell *trans, char* buf, unsigned int len,
 	update_reply_stats( code );
 	trans->relayed_reply_branch=-2;
 	t_stats_replied_locally();
+	t_stats_replied_total();
 	if (lock) UNLOCK_REPLIES( trans );
 
 	/* do UAC cleanup procedures in case we generated
@@ -1866,6 +1867,7 @@ enum rps relay_reply( struct cell *t, struct sip_msg *p_msg, int branch,
 			}
 		}
 		update_reply_stats( relayed_code );
+		t_stats_replied_total();
 		if (!buf) {
 			LOG(L_ERR, "ERROR: relay_reply: "
 				"no mem for outbound reply buffer\n");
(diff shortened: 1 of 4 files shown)


smititelu:
Consider the simple basic dialog scenario:
```
                        Messages
INVITE ---------->      1
407 <----------         1
ACK ---------->         1
INVITE ---------->      1
100 <----------         1
180 <----------         1
183 <----------         0
200 <----------         1     
ACK ---------->         1
Pause [   7000ms]       1
BYE ---------->         1             
200 <----------         1
```
In the current implementation, running _kamctl fifo get_statistics tmx:_ on a fresh kamailio prints:
```
...
tmx:local_replies = 1
tmx:received_replies = 2
tmx:relayed_replies = 1
```
After this patch:
```
...
tmx:local_replies = 1
tmx:received_replies = 2
tmx:relayed_replies = 3 (2 x 200 + 1 x 180)
```

smititelu:
IMHO, the _tmx:received_replies_ are not counded right also; right now the counter shows only completed_6xx/5xx/4xx/3xx/2xx.
I think it should be 0 for the first scenario(i.e. no received reply, only local generated ones) and 3 for the second scenario(2 x 200 + 1 x 180 received). I'm thinking of a solution for this also.

What do you think would be the right values for those scenarios?

miconda:
It can be the number of replies received and then relayed, leaving the local replies counter for the replies that are generated locally. Then the difference between received replies and relayed replies will show how many replies were absorbed by kamailio (e.g., for branch aggregation).

You can merge this pull request and if you have new updates, then you can make a new one. Thanks!

PR #422, opened 2015-11-27
Decrypt callid encrypted by topoh module before save in database

pseabra:
Homer can't relate callids encrypted and decrypted, originated by module topoh, issue https://groups.google.com/forum/#!topic/homer-discuss/1_a01BbgnD0

Issue fixed changing function sip_capture_store to decrypt callid and save decrypted callid in database.

### DIFF
--- a/modules/sipcapture/sipcapture.c
+++ b/modules/sipcapture/sipcapture.c
@@ -77,6 +77,7 @@
 #include "sipcapture.h"
 #include "hash_mode.h"
 #include "hep.h"
+#include "th_mask.h"
 
 #ifdef STATISTICS
 #include "../../lib/kcore/statistics.h"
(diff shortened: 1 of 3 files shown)


adubovikov:
thank you for your patch, but the th_ parameters should not be hardcoded, please add them to param_export_t params. As reference, use:
https://github.com/kamailio/kamailio/blob/master/modules/topoh/topoh_mod.c#L86
or
https://github.com/kamailio/kamailio/blob/master/modules/sipcapture/sipcapture.c#L278

miconda:
Few more things that need to be done:

  * the commit message must have the format as per guidelines at:
  * this patch seems to just copy a lot of code from topoh module to sipcapture. It means that if there is a change in topoh, it needs to be resync'ed in sipcapture. This is not the right way to do. Better export an intermodule API in topoh and bind to that one from sipcapture. In this way, the code and parameters from topoh will be used by sipcapture. Sipcapture should get a new module parameter to control if it should bind or not to topoh.

You can look at sl module how it exports some functions (modules/sl/sl.h, the sl_api_t structure and the sl_load_api() function) and then at registrar module how it binds to sl in mod_init() and use sl functions.

miconda:
Re-implemented via exporting topoh module api -- see topoh_unmask parameter for sipcapture module. Testing is needed.

PR #423, opened 2015-11-28
textops: Fix get_body_part() end of body headers

smititelu:
Remember the end of body headers and use it for get_body_part().


miconda:
Thanks! You can merge it.

PR #425, opened 2015-12-01
Add RR parameter on sequential requests

khoegh:
The 2 commits restores vsf=, vst= and did= RR parameters on in-dialog requests


miconda:
Thanks!

It looks like in the patch to uac needs to allocate one more char for the buffer, to store '\0' ending char. I am going to merge and then adjust the code.

Issue #432, opened 2015-12-07
crash on using http_query


jayesh1017:
tried doing a post on a https URL and kamailio crashes as soon as it receives a 200 response. Kamailio continues processing the response is error but on a successful response it crashes. Here's the backtrace from the core:
```
0x00007f62f2bf8e38 in ?? ()
#1  0x00007f62ea22251b in http_query (_m=0x7f62f2c3a038,
_url=0x7f62f2bf3548 "\330\064\277\362b\177", _dst=0x7f62f2bf9250
"\030\215\277\362b\177", _post=0x7f62f2bf35c8 "\200r\277\362b\177",
    _hdr=0x7f62f2bf91d0 "h\221\277\362b\177") at functions.c:226
#2  0x00007f62ea229e0b in w_http_query_post_hdr (_m=0x7f62f2c3a038,
_url=0x7f62f2bf3548 "\330\064\277\362b\177", _post=0x7f62f2bf35c8
"\200r\277\362b\177", _hdr=0x7f62f2bf91d0 "h\221\277\362b\177",
    _result=0x7f62f2bf9250 "\030\215\277\362b\177") at utils.c:483
#3  0x000000000041f6b5 in do_action (h=0x7ffeb128c8c0, a=0x7f62f2bf8a58,
msg=0x7f62f2c3a038) at action.c:1079
#4  0x000000000042c2ea in run_actions (h=0x7ffeb128c8c0, a=0x7f62f2bf6fc0,
msg=0x7f62f2c3a038) at action.c:1549
#5  0x000000000041f3ce in do_action (h=0x7ffeb128c8c0, a=0x7f62f2bfb9b0,
msg=0x7f62f2c3a038) at action.c:1045
#6  0x000000000042c2ea in run_actions (h=0x7ffeb128c8c0, a=0x7f62f2bb9388,
msg=0x7f62f2c3a038) at action.c:1549
#7  0x000000000041bc28 in do_action (h=0x7ffeb128c8c0, a=0x7f62f2b8f248,
msg=0x7f62f2c3a038) at action.c:678
#8  0x000000000042c2ea in run_actions (h=0x7ffeb128c8c0, a=0x7f62f2b8e268,
msg=0x7f62f2c3a038) at action.c:1549
#9  0x000000000041f3ce in do_action (h=0x7ffeb128c8c0, a=0x7f62f2b91040,
msg=0x7f62f2c3a038) at action.c:1045
[...]

Issue #400, opened 2015-11-18
http_request and curl_http_request cause crash in 4.3 and dev


oej:
Repeated use of http_request cause a crash related to the storage of the result in a PV. 

Will upload backtraces and other info to this issue. If anyone else have seen this please add info to this bug report.

miconda:
Is it latest 4.3 or master? There were couple of fixes in September, iirc.

oej:
Backtrace from the curl module in dev. Looks similar with the http_request in 4.3

#0  0x00007fabdd0a20d5 in raise () from /lib/x86_64-linux-gnu/libc.so.6
No symbol table info available.
#1  0x00007fabdd0a583b in abort () from /lib/x86_64-linux-gnu/libc.so.6
No symbol table info available.
#2  0x0000000000632ad0 in qm_debug_frag (qm=0x7fabdc86b010, f=0x7fabdc93cf78) at mem/q_malloc.c:142
        __FUNCTION__ = "qm_debug_frag"
#3  0x0000000000634069 in qm_malloc (qm=0x7fabdc86b010, size=296, file=0x7fabdc65ce70 "pv: pv_svar.c", 
    func=0x7fabdc65d01d "set_var_value", line=128) at mem/q_malloc.c:369
        f = 0x7fabdc93cf78
        hash = 168
        list_cntr = 1
        __FUNCTION__ = "qm_malloc"
#4  0x00007fabdc61483b in set_var_value (var=0x7fabdc942658, value=0x7ffd762e8070, flags=1) at pv_svar.c:128
        __FUNCTION__ = "set_var_value"
#5  0x00007fabdc5fbfa8 in pv_set_scriptvar (msg=0x7fabdc8a9048, param=0x7fabdc939d18, op=254, val=0x7ffd762e8150)
    at pv_core.c:1988
        avp_val = {n = -594293216, s = {
            s = 0x7fabdc93ce20 "{\"status\":\"OK\",\"uid\":\"device.john.smith\",\"dom\":\"example.com\",\"pwd\":\"7ymu0mAJ\",\"recfile\":\"20151103144733822~0313809101~unspecified~example.com~john.smith\",\"recdir\":\"example.com/2015/11/03\",\"rec\":true,"..., len = 295}, re = 0x7fabdc93ce20}
        flags = 1
        __FUNCTION__ = "pv_set_scriptvar"
#6  0x00007fabd2bc9f6e in curL_query_url (_m=0x7fabdc8a9048, 
[...]

PR #435, opened 2015-12-09
Fixes and improvements to curl module

que273:
- Fix invalid check of uri and postdata parameters
- Fix fixup functions
- Refactor curl function to remove dependance on pv/fixup structures
- Export curl_connect through C API
- Fix buffer overflow in curl write function

### DIFF
--- a/modules/curl/curl.c
+++ b/modules/curl/curl.c
@@ -59,10 +59,12 @@
 #include "../../rpc.h"
 #include "../../rpc_lookup.h"
 #include "../../config.h"
+#include "../../lvalue.h"
 
 #include "functions.h"
 #include "curlcon.h"
 #include "curlrpc.h"
+#include "curl_api.h"
 
 MODULE_VERSION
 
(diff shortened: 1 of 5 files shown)


oej:
Thank you! This looks great. Will test it soon and merge it.

oej:
I noted that the API function was not documented in the XML docs. Will try to fix.

oej:
This possibly handles the report in #400 - but only for the curl module. Maybe we need to look into the utils module too and backport changes there.

miconda:
Hugh (@que273) pushed already some patches to utils module as well -- see 3b6d7a28ea0b053f27cb169ac31259e20ef4ece0 .

que273:
In my testing of the curl module,before these changes, I didn't get any crashes - perhaps I was lucky.
The crash in utils/http_query (#432) is not related to these changes because the http_query_post_hdr function (which allows passing additional headers, e.g. Content-Type) has not yet been ported to the curl module. If the intention is to obsolete http_query from the utils module, this version of the command needs to be ported.
Fixing the buffer overflow may fix the crash in #400, but that was already fixed in utils (master & 4.3) so no backport is required.

oej:
The content type is already covered, but adding other headers is not part of the current curl. Maybe something I should add to the "TODO" file. Thank you for the feedback.

miconda:
Can this be merged? It would be good to get it in master for making it available for testing with common code base.

que273:
Thank you for testing @oej

PR #436, opened 2015-12-09
htable: fixing htable.stats for max slot value

snen:
add missed 'd' parameter to rpc->struct_add format string
before patch
```
[root@sw4 kamailio]# kamcmd htable.stats
{
	name: some_htable
	slots: 4096
	all: 285304
	min: 43
}
```
after patch
```
[root@sw4 kamailio]# kamcmd htable.stats
{
	name: some_htable
	slots: 4096
	all: 285304
	min: 43
	max: 103
}
```


miconda:
Thanks!

PR #442, opened 2015-12-14
Better counting of received_replies counter

smititelu:
Consider the simple basic dialog scenario:
```
                        Messages
INVITE ---------->      1
407 <----------         1
ACK ---------->         1
INVITE ---------->      1
100 <----------         1
180 <----------         1
183 <----------         0
200 <----------         1     
ACK ---------->         1
Pause [   7000ms]       1
BYE ---------->         1             
200 <----------         1
```
Before this patch:
```
tmx:local_replies = 1
tmx:received_replies = 2
tmx:relayed_replies = 3
```
After this patch:
```
tmx:received_replies = 3
tmx:received_replies_absorbed = 0
tmx:relayed_locally = 1
tmx:relayed_received = 3
tmx:relayed_total = 4
```


smititelu:
Running a cfa scenario:
```
...
tmx:received_replies = 11
tmx:received_replies_absorbed = 6
tmx:relayed_locally = 4
tmx:relayed_received = 5
tmx:relayed_total = 9
```

miconda:
Thanks for this contribution, I think the result shows better what happens with the replies handled by tm.

The patch can be merged. But afterwards, because we are already having different names and change the overall counting for those replies, I think is better to go to a more coherent naming. So, if you have time, after merging, maybe you can change the names.

I propose to use 'rpl_' prefix to suggest is about replies, so:

```
tmx:rpl_received - received replies
tmx:rpl_absorbed - absorbed replies (I think it suggests already they were received, in order to be absorbed)
tmx:rpl_generated - locally generated replies
tmx:rpl_relayed - relayed replies (I think 'relay' would be suggestive enough for 'received and sent out'
tmx:rpl_sent - all replies that were sent out (local and received)
```

If anyone has better suggestions for names, I welcome other proposals. One of the goals should be trying to keep the names rather short.

PR #443, opened 2015-12-14
modules/cfgt: debugger config test support

linuxmaniac:
generates reports of testing scenarios:

* received message
* messages sent
* flow of executed routes ( `start`, `drop`, `exit`, `return` )
* list of the variables and its value at every route executed controlled by mask
  (`CFGT_DP_NULL`, `CFGT_DP_AVP`, `CFGT_DP_SCRIPTVAR`
   `CFGT_DP_XAVP`, `CFGT_DP_OTHER`, `CFGT_DP_ALL`)

Changes in the implementation from the initial propose:

 * used a prefix on `Call-ID` instead of custom SIP headers `X-Kamailio-test` and `X-Kamailio-test-ID` allowing the ability to bypass other devices filters ( like sems whitelist SIP headers )

 * save report files as JSON at disc after the message is processed instead of using RPC commands

Initial implementation of #8

### DIFF
--- a/modules/cfgt/Makefile
+++ b/modules/cfgt/Makefile
@@ -1,0 +1,17 @@
+#
+# cfgt module makefile
+#
+# WARNING: do not run this directly, it should be run by the master Makefile
+
+include ../../Makefile.defs
+auto_gen=
+NAME=cfgt.so
+LIBS=
+
+DEFS+=-DKAMAILIO_MOD_INTERFACE
+
+SERLIBPATH=../../lib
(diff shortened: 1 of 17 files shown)


miconda:
Thanks! It can be merged.

As a suggestion for the future - changes to other modules (and core if the case) should be made own commits in order to review easier.

miconda:
Missed in first place -- I think it also needs to be listed in one of the groups in Makefile.groups for packaging.

linuxmaniac:
09e823de added cfgt to the list at Makefile.groups Thanks!

PR #447, opened 2015-12-16
core: rewrite add_interfaces() with getifaddrs() based implementation

celan69:
To make IPV6 work with Solaris, add_interfaces() had to be modified.
I decided to reimplement it with a getifaddrs() based implementation,
which is available from Solaris 11 upwards, and also on FreeBSD and
Linux. By default, the latter continues to use the netlink based
variant.


miconda:
Thanks! Merged.

PR #448, opened 2015-12-16
presence : fix notify version when subs_mode <> 3

lazedo:
when subs_mode <> 3 the notify version is always the same (first watcher) for all watchers


miconda:
Thanks, you can merge it.

PR #450, opened 2015-12-17
cfgt: second pass at documentation update

giavac:
Adds Usage section with .cfg snippet


linuxmaniac:
Just a minor change needed

giavac:
If you have a sanitised/meaningful example of the JSON in a "flow" element, it'd be worth adding it too.
Or I can add some I've generated, if you want.

PR #451, opened 2015-12-18
db_text :  column type comparison

lazedo:
compare DB1_STR, DB1_STRING, DB1_BLOB as equal


miconda:
Thanks, it can be merged.

PR #453, opened 2015-12-22
cnxcc: define inline functions in cnxcc.h

grumvalski:
On some systems compiling cnxcc we get:

cnxcc.h:36:12: warning: inline function �timestamp2isodt� declared but never defined
 inline int timestamp2isodt(str *dest, unsigned int timestamp);
            ^
cnxcc.h:35:21: warning: inline function �get_current_timestamp� declared but never defined
 inline unsigned int get_current_timestamp();
                     ^
cnxcc.h:34:13: warning: inline function �get_datetime� declared but never defined
 inline void get_datetime(str *dest);
     
and then at runtime:

ERROR: <core> [sr_module.c:576]: load_module(): could not open module </usr/lib64/kamailio/modules/cnxcc.so>: /usr/lib64/kamailio/modules/cnxcc.so: undefined symbol: get_current_timestamp

The present patch moves the definition of the inline functions into the .h file.

### DIFF
--- a/modules/cnxcc/cnxcc.c
+++ b/modules/cnxcc/cnxcc.c
@@ -28,30 +28,6 @@
 
 #include "cnxcc.h"
 
-inline void get_datetime(str *dest)
-{
-	timestamp2isodt(dest, get_current_timestamp());
-}
-
-inline unsigned int get_current_timestamp()
-{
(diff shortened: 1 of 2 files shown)


miconda:
Can be merged.

PR #454, opened 2015-12-24
tcpops: implement event_route for unexpectedly closed tcp connections

armenb:
Comments are welcome

### DIFF
--- a/events.c
+++ b/events.c
@@ -165,6 +165,11 @@ int sr_event_register_cb(int type, sr_event_cb_f f)
 					_sr_events_list.rcv_nosip = f;
 				else return -1;
 			break;
+		case SREV_TCP_CLOSED:
+				if(_sr_events_list.tcp_closed==0)
+					_sr_events_list.tcp_closed = f;
+				else return -1;
+			break;
 		default:
 			return -1;
 	}
@@ -284,6 +289,12 @@ int sr_event_exec(int type, void *data)
 					ret = _sr_events_list.rcv_nosip(data);
 					return ret;
 				} else return 1;
+		case SREV_TCP_CLOSED:
+				if(unlikely(_sr_events_list.tcp_closed!=0))
+				{
+					ret = _sr_events_list.tcp_closed(data);
+					return ret;
+				} else return 1;
 		default:
 			return -1;
 	}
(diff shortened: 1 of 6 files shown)


miconda:
Many thanks for this contribution, it was on a to-do list for quite some time. I will look over the patch and merge it if all ok.

One question for now -- given your details that the event route is executed on tcp connection kill or time out, I guess those were some examples and the event route is also executed if the client closes the connection (e.g., it goes offline/closing the sip application), isn't it?

armenb:
Hi Daniel,

I just repushed this branch. Originally, I only considered tcp connection kill via TCP RST (which results in read() == -1 && errno == ECONNRESET) and timeout events (read() == -1 && errno == ETIMEDOUT).  I have revised my branch to also execute the event_route on gracefully-closed TCP connections (i.e. via TCP FIN), which is delivered to the program as read() == 0 (i.e. EOF).

Additionally, I have restricted the execution of this event_route that have a $conid variable that does not evaluate to "<null>" in the script.  This is determined via connections that have TCP_REQ_COMPLETE(req) set to true (which is when con->id is copied to con->proto_reserved1).  Do let me know if this is not the correct strategy here.

Thanks!

camilleoudot:
Hi Armen,
thanks for your patch, I think this feature will be very useful.
I have one comment: if the tcpops module is loaded but no `tcp:closed` route is defined, a debug log line will complain about it. I think it would make sense to add a module parameter to globally enable or disable the event route lookup and execution.
Besides, as many of the tcpops function can act on a per-socket basis, it would be nice to have this event route called or not called depending on the socket (e.g. using a `tcp_connection` flag and defining a new tcpops function to set/reset this flag).
What do you think?

miconda:
I merged the patch manually, to split it in commits per component, willing to make it easier to track changes on those parts.

I also think that the suggestions by @camilleoudot are useful - a parameter for a global enable/disable of event route execution, plus the flexibility to do it per connection with a flag.

As the initial patch was merged, anyone can add those proposed enhancements.

kamailio-sync:
Camille, thanks for the feedback.  Those are great points, and I'll try to
complete work on those configuration features soon.

PR #456, opened 2015-12-29
auth_radius: support for not appending realm to the username passed to RADIUS

phil-lavin:
- Implement append_realm_to_username parameter in auth_radius which allows
  the unmolested digest username to be passed to RADIUS in the User-Name
  attribute. This is required for some RADIUS implementations.

Apologies if this doesn't belong in the 4.3 branch - this is what the commit is based off and tested on. Let me know if I should move it elsewhere...

PR #457, opened 2015-12-30
auth_radius: support for not appending realm to the username passed to RADIUS

phil-lavin:
- Implement append_realm_to_username parameter in auth_radius which allows
  the unmolested digest username to be passed to RADIUS in the User-Name
  attribute. This is required for some RADIUS implementations.

This is the same PR as #456 except based on master


miconda:
Thanks for the contribution.

PR #458, opened 2016-01-04
presence: additional info in logs

lazedo:
adds callid to NOTIFY log
adds log when multiple records are found in process_dialogs


miconda:
Can be merged.

PR #459, opened 2016-01-04
db_text: add support for replace

lazedo:
implements replace with update/insert
tries to update, inserts if not found


miconda:
Thanks, can be merged.

PR #460, opened 2016-01-04
presence_xml: fix pres_agg_nbody_empty

lazedo:
we were sending the presence element without the entity attribute.
it seems some UAs dislike this


miconda:
I am merging it, to follow up with adding some safety checks.

PR #462, opened 2016-01-05
debugger: add fifos to set/get level/facility

smititelu:


miconda:
We try to obsolete the MI interface and use only the RPC interface. While not a stric policy not to add a new MI command, but there must be an equivalent RPC command for it, so when MI is going to be disabled, then RPC has all the correspondent commands. So you have to add the rpc commands (can be separate pull request).

Also, I noticed that two commits have the same message, but one is to the core, exposing the function to get the syslog facility.

Otherwise I am fine with the pull request.

smititelu:
Added separate commit for core facility2str(). Added rpc level/facility setters/getter.

While testing this I've found issue #463 which was happening also before this pull request when issuing "kamcmd dbg.mod_level core 1" if the module level wasn't configured in cfg file: ```modparam("debugger", "mod_level", "core=3")```

miconda:
Thanks, you can merge it.

PR #467, opened 2016-01-07
kamailio sending update ccr request instead of terminate request for initial ccr response having final_unit_action flag.

DileepNunna:
If final_unit_action flag is set in initial ccr response, kamailio need to send terminate request after time out but it is sending update request.but it is fine for intermediate requests.


miconda:
Hopefully @jaybeepee, @richardgood or @ngvoice can check and confirm that the patch is good.

If all ok, I would like to merge the patch manually because the commit message is not properly formatted -- I can take care of the merge provided IMS devs are fine with the patch.

ngvoice:
@richardgood @jaybeepee : Looks good to me, what do you think?
Thanks!

kamailio-sync:
I am happy with this. you can push ;)

Cheers and thanks for the contribution!

ngvoice:
Closed through commit 581912f5ed117bb56087668aa7ac60a7b667d525.

PR #475, opened 2016-01-14
registrar: reg_xavp_cfg should not be reset when sock_flag is not set

ianlin:
- since there are functions like `registered()` would use it
- reported by Jayesh Nambiar (@jayesh1017) in the comment of GH#470


miconda:
Thanks!

PR #477, opened 2016-01-15
Use C99 boolean

mslehto:
Spotted few cases where using C99 bool can remove local typedef and corresponding macros


miconda:
Thanks!

PR #478, opened 2016-01-15
debugger: fix segfault when dbg_sip_msg() is called with no params

smititelu:


miconda:
Thanks, it can be merged.

PR #485, opened 2016-01-25
db_text: implemented raw query for UPDATE/DELETE/REPLACE

lazedo:
also fixes db_free_result for non select operations
also improves replace by exiting on first match


miconda:
It can be merged.

Not something that I can test now, but we will get in testing phase soon.

PR #486, opened 2016-01-25
presence: improvements for 4.4

lazedo:
adds presence:notify-reply event
implements $notify_reply for handling in presence:notify-reply event
adds flags, user_agent columns to active_watchers
extends $subs to other subscription properties


miconda:
Thanks!

PR #487, opened 2016-01-25
core: secondary timer interval

snen:
fix wrong function invocation period for secondary timer.
i have noticed new  secondary timer invocation period error , while looking at commit 93b5d2e
functions called more often than it should due to (c % wt->cycles==0) is TRUE.

i'm not sure this is the best possible fix, but it works

### DIFF
--- a/timer_proc.c
+++ b/timer_proc.c
@@ -284,7 +284,6 @@ typedef struct sr_wtimer_node {
 	struct sr_wtimer_node *next;
 	uint32_t interval;
 	uint32_t steps;
-	uint32_t cycles;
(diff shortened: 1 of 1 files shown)


miconda:
Thanks for heads up! I reimplemented the secondary timer as classic wheel timer -- initially I played with the idea of optimizing it for intervals fitting many times in the number of slots, but becomes too complex to deal with exceptions.

Trying the last master branch and reporting the results would be very appreciated.

snen:
it works fine now for usrloc and dispatcher secondary timers

PR #489, opened 2016-01-25
modules/ims_icscf: removes values.h dependency

mslehto:
Did not compile on FreeBSD 10.

### DIFF
--- a/modules/ims_icscf/cxdx_lir.c
+++ b/modules/ims_icscf/cxdx_lir.c
@@ -51,10 +51,8 @@
 #include "mod.h"
 #include "location.h"
 
-#if defined (__OS_freebsd)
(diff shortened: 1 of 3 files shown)


mslehto:
These are only log lines related to ims_icscf:
```
444: rm -f ims_icscf.d ims_icscf.d ims_icscf.d *~
1869: rm -f ims_icscf.d ims_icscf.d ims_icscf.d *~
```
Looks like Travis do not touch any other IMS module as well.

miconda:
Hopefully @jaybeepee, @ngvoice, @richardgood or other ims devs can review and merge.

miconda:
Merging as there are only defines cleanup. If breaks something, it can be fixed afterwards.

PR #490, opened 2016-01-26
db_text: add buffer size for reading text files

lazedo:
the static value of 4096 may not be enough when dealing with presence
tables that have xml content


miconda:
Thanks! It can be merged.

PR #491, opened 2016-01-27
Make tcpops module usable on BSD

mslehto:
Tested with FreeBSD 10 and Linux 3.2 (Debian Wheezy 7.6).
List of other platforms gathered with [FreeBSD and Linux Kernel Cross-Reference](http://fxr.watson.org/fxr/trackident?i=TCP_KEEPIDLE)

### DIFF
--- a/modules/tcpops/doc/functions.xml
+++ b/modules/tcpops/doc/functions.xml
@@ -18,7 +18,7 @@
 		<para>Meaning of the parameters is as follows:</para>
 		<itemizedlist>
 			<listitem>
-				<para><emphasis>conid</emphasis> (optionnal): the Kamailio internal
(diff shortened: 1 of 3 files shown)


camilleoudot:
Thanks for the patch.
I'm not sure where `tcp_options.h` gets the `TCP_KEEP*` macros defined from, do I miss something?

mslehto:
I have those in netinet/tcp.h.
It is included by tcp_main.c and modules/tcpops/tcpops.c
Can this still work?

camilleoudot:
then it would make sense to also include `netinet/tcp.h` in `tcp_options.h`, to avoid any issue in the future if `tcp_options.h` is re-used elsewhere

camilleoudot:
Merged, thanks for enabling the functionality on BSD

PR #492, opened 2016-01-28
pua_dialoginfo: fixed crash when loading dialogs from database on restart

phil-lavin:
- SHM pointer is freed when it is required later by dialog callbacks. Remove the free_dlginfo_cell() call.

This is causing crashing after the BYE, when a call is ongoing during a Kamailio restart

The free is actually done by the callback function `dlg_api.register_dlgcb()` when it is registered in `pua_dialoginfo.c`

### DIFF
--- a/modules/pua_dialoginfo/doc/pua_dialoginfo.xml
+++ b/modules/pua_dialoginfo/doc/pua_dialoginfo.xml
@@ -34,9 +34,14 @@
 		<affiliation><orgname>IPCom (Module implementation was partly sponsored by Silver Server (www.sil.at))</orgname></affiliation>
 	    </author>
 	    <editor>
-		<firstname>Klaus </firstname>
-		<surname>Darilion</surname>
-		<affiliation><orgname>IPCom</orgname></affiliation>
+		<firstname>Phil</firstname>
+		<surname>Lavin</surname>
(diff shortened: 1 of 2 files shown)


phil-lavin:
I will backport this to 4.3 if/when you're happy with it

miconda:
Thank you for the report of the issue and accompanied patch to fix it.

A slightly different patch was pushed, to avoid compile and sanity checks warnings that the variable was set but not used, respectively dlginfo.

The changes to the documentation are not valid, the names there refer to the authors of the documentation, not to the authors of the code. There can be code written by different persons which have no relation with the module documentation for use in kamailio.cfg (which is what docs contain). Also, there was a new other name that was not involved in any discussions.

The code ownership is reflected by copyright strings at the top of the files. But that has to reflect relevant changes and must be additions, not removals. By removing a line added in the past by someone else, there is still no code done by you that can you ask copyright for it. Your name was mentioned in the commit message, as we do for all contributors.

Again, relevant c code additions can be accompanied with copyright line at the top of changed files. You must be aware that core and main components contributions must be under BSD license, contributions to modules that depend on libssl must grant OpenSSL GPL exception.

phil-lavin:
Thanks, Daniel. All understood :)

miconda:
Btw, contributions to documentation are also very welcome and much appreciated.

PR #493, opened 2016-02-01
rtpengine: Add module

smititelu:
Add rtpengine module files from upstream to 4.1 branch. Don't do any changes to 4.1 rtpproxy-ng module.

### DIFF
--- a/modules/rtpengine/Makefile
+++ b/modules/rtpengine/Makefile
@@ -1,0 +1,15 @@
+# WARNING: do not run this directly, it should be run by the master Makefile
+
+include ../../Makefile.defs
+auto_gen=
+NAME=rtpengine.so
+LIBS=
+
+DEFS+=-DKAMAILIO_MOD_INTERFACE
+
+SERLIBPATH=../../lib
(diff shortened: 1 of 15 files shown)


linuxmaniac:
We don't introduce new features in previous versions. In my honest opinion this change must be rejected

miconda:
Indeed, we do not introduce new features in the stable branches, 4.1 is also no longer officially packaged (no new releases out of it).

Somehow related, I think we discussed with different occasions in the past, and if there are people interested in doing the work, there can be created something similar to Debian backports repository, sort of 'semi-official' branches like "backports/x.y" (x.y being an official stable branch, in this case will be backports/4.1), where people maintaining this backports branch can decide to backport whatever they like from newer versions, including development branch. I am not having much time for such effort, but I do backport from time to time new features (typically from devel to latest stable branch) for various customers and I can push them to such repository just in case others will like to use them. Maybe we can get the discussion to sr-dev/sr-users mailing lists to see if there is interest and people willing to allocate maintenance resources.

smititelu:
Closing this for now. Thanks for the suggestions.

PR #494, opened 2016-02-02
cnxcc: change some log levels from INFO to DBG

grumvalski:
When using redis as cnxcc backend, too much log is printed, specially all the logs related to the periodic user's wallet check.


miconda:
It can be merged.

grumvalski:
Thanks, can I also backport it to 4.3 branch?

PR #503, opened 2016-02-08
Add t_is_request_route() config function

smititelu:


miconda:
Even the development is frozen for 4.4, this is the missing one from the group that helps to detect the top route (along with t_is_branch_route(), t_is_failure_route(), ...) and given that is practically just a condition around route_type global variables, it might worth merging, we haven't branched yer 4.4. Other opinions?

PR #506, opened 2016-02-10
ims_registrar_scscf: fix new_notification() so it does not crash

dragos-oancea:
fix mishandling of impus (array of ptr) - avoid crash


kamailio-sync:
Thanks Dragos, this is already fixed on our side - just haven't checked in
yet.... sorry - will checkin first thing tomorrow ;)

Cheers
Jason

miconda:
Is this still actual or other commits were done instead?

ngvoice:
Issue is fixed in master.

PR #508, opened 2016-02-12
ims_dialog: fix double free (unref early dialog)

dragos-oancea:
Attempt fix for when an INVITE goes out and there is no 100 Trying reply to it , and then ims_dialog destroys the dialog twice via unref_dlg_unsafe()  .


dragos-oancea:
Hi Jason. Thanks for looking into it. Here is a log and the output of "bt full"  :
http://pastebin.com/PdXiZSV8  .  Scenario : A call comes in, then the first INVITE is relayed somewhere, and there is no reply coming to it (there are retransmissions) , kamailio is generating a 503 , and then the negative ACK comes, and then it crashes.

kamailio-sync:
Thanks Dragos. Can you check this on the new patches I gave you and post
details if you still have problems

Cheers
Jason

miconda:
Is this actual or it was worked around with other commits?

kamailio-sync:
Fixed with other commits Daniel

miconda:
OK, then closing this PR.

dragos-oancea:
Lots of bugs disappeared with the latest patches which are already in master, thank you Jason, thank you Daniel.

PR #512, opened 2016-02-16
ims_usrloc_scscf: fix link_contact_to_impu()  -  was crashing with maxcontact_behaviour == 2

dragos-oancea:
link_contact_to_impu() does not behave as intended - it will return immediately because ptr==contact .
This will lead to a crash when current contact is always different than the previous. 
To prevent this, I set ptr to 0 as it points to a chunk of freed memory anyway.
Removed call to  unlink_contact_from_impu() as I believed is redundant and it leads to double free.
This is an usual scenario and it happens with devices that reconnect on TCP , and they change the src port every time , and then the contact will be different.

### DIFF
--- a/modules/ims_usrloc_scscf/impurecord.c
+++ b/modules/ims_usrloc_scscf/impurecord.c
@@ -479,6 +479,7 @@ int insert_scontact(impurecord_t* _r, str* _contact, ucontact_info_t* _ci, ucont
                     LM_DBG("Too many contacts already registered, overwriting oldest for IMPU <%.*s>\n", _r->public_identity.len, _r->public_identity.s);
                     //we can just remove the first one seeing the contacts are ordered on insertion with newest last and oldest first
 		    mem_delete_ucontact(_r->newcontacts[0]);
+			_r->newcontacts[0] = 0;
                     break;
                 default://unknown
                     LM_ERR("unknown maxcontact behaviour..... ignoring\n");
@@ -1027,9 +1028,6 @@ int link_contact_to_impu(impurecord_t* impu, ucontact_t* contact, int write_to_d
 
     if (i < MAX_CONTACTS_PER_IMPU) {
         LM_DBG("contact [%.*s] needs to be linked to impu [%.*s] at position %d\n", contact->c.len, contact->c.s, impu->public_identity.len, impu->public_identity.s, i);
-        if (overwrite)
-            unlink_contact_from_impu(impu, impu->newcontacts[i], write_to_db, 0 /*implicit dereg of contact */); //unlink the contact we are overwriting
-
         impu->num_contacts = i + 1; //we always bump this - as unlink (in overwrite would have decremented)
         impu->newcontacts[i] = contact;
         ref_contact_unsafe(contact);


miconda:
Ping to @jaybeepee, @richardgood and @ngvoice for comments or merging.

dragos-oancea:
Hi Jason, 

Sure, send me the patch , I will test it.

Cheers,
Dragos

jaybeepee:
Here you go Dragos. This is the full patch (from modules directory). There are a few changes across various modules, mostly improvements and fixes. So you can test them all. But let me know if the usrloc fixes fix your immediate issue.

Cheers
Jason

jaybeepee:
+attachment

[kamailio-diff-dragos.zip](https://github.com/kamailio/kamailio/files/136361/kamailio-diff-dragos.zip)

dragos-oancea:
Thank you Jason, looks good so far. I compiled and ran the patches on a test machine.  I will keep on using the machine these days and report back.

dragos-oancea:
Hi Jason, I had no crash so far with your patches. 

However, I observed some strange behaviour which I have not seen before applying the patches:
1. A timeout for a reply to REGISTER.
 A 200 OK reply is expected but nothing happens. This is with the P-cscf and it does not occur all the time. I restart the P and it starts replying again. I have not noticed anything peculiar in the logs which might explain this.

2. when I stop kamailio, there is a process hanging for 1 minute (cdp) . log :
DEBUG: cdp [session.c:305]: cdp_get_session(): called get session with id pcscf.ims.mnc001.mcc001.3gppnetwork.org;2979852027;5 and hash 103
[it waits 1 minute and then shuts down: ]
CRITICAL: <core> [main.c:637]: sig_alarm_abort(): shutdown timeout triggered, dying...

Regards,
Dragos

kamailio-sync:
Hey Dragos, thanks for the feedback.

Re. the REGISTER/200OK problem. Is there any chance you can send me a debug
log and pcap? how easily recreatable?

Also can you confirm the following signalling so I understand your use case:

REG=>
<=401
REG=>
<=200 (At this stage the P-CSCF is not forwarding back to UE) - is that
correct?

Cheers
Jason

dragos-oancea:
Hi Jason, 

Actually is more strange, and I do not exclude a diameter problem or an issue with my environment now.
The  P forwards the REGISTER to the I , I to the S , S sends back 401 back to the I , and then it  arrives at the P , and then on UE.  Then UE sends REGISTER again (challenge/response) but this second REGISTER does make it past the I , nothing comes back from the I , and then the P generates a 504 Server Time-out which is being sent to UE. 
Funny thing is that when I restart the P (and not the I !) , registration starts to function normally again.
I could not reproduce today .  I will send some logs from last week to your email if it's okay with you.

Dragos

kamailio-sync:
Yes, that does sound strange. Yes, by all means, send to my email.

Cheers
Jason

miconda:
What is the status for this PR? Still needs time or already handled with other commits?

kamailio-sync:
Fixed with other commits.

miconda:
OK, closing it.

PR #513, opened 2016-02-18
srutils: change names colliding with libcrypto

camilleoudot:
Kamailio's **srutils** SHA functions have names that collide with OpenSSL's **libcrypto** SHA functions.
In practice, OpenSSL fails to validate a server cert if **srutils** is loaded before **libcrypto**, because their SHA functions are not 100% equivalent.

On my Debian, **http_async_client** was unable to issue any HTTPS request unless server certificate verification was disabled.

This patch solves the issue by renaming the SHA functions in **srutils** to non-colliding names. I think this has no side effect because they are only used internally.

### DIFF
--- a/lib/srutils/sha256.c
+++ b/lib/srutils/sha256.c
@@ -339,7 +339,7 @@ static const char *sha2_hex_digits = "0123456789abcdef";
 
 
 /*** SHA-256: *********************************************************/
-void SHA256_Init(SHA256_CTX* context) {
+void sr_SHA256_Init(SHA256_CTX* context) {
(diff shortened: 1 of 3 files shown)


miconda:
Thanks, it can be merged.

oej:
Should this be back ported?

camilleoudot:
You're right, the bug should occur every time `libsrutils.so` is loaded before `libcrypto.so` and OpenSSL is performing a SHA hash, so it's not specific to the new http modules.
I'll backport it.

PR #514, opened 2016-02-19
pua_dialoginfo: support for sending "Trying" state for callee

phil-lavin:
- pua_dialoginfo: Implement callee_trying parameter in pua_dialoginfo which causes
  a "Trying" state to be sent for both the caller and callee when
  a new dialog is created.
- presentity: Implement table locking for all routes through update_presentity in order
  to avoid a number of different race conditions which resulted in
  extraneous data being put into the presentity table.

The latter change fixes a race condition which triggers more frequently due to the first... however I'm sure there are instances when it would trigger regardless of the first change.


miconda:
Thanks!

PR #519, opened 2016-02-24
tls: enhance the logging of ssl bug #1491 workaround

phil-lavin:
- log shm_available_safe() instead of shm_available() as this is what is
  actually compared in the check
- also log the relevant threshold as it is not always obvious if the default
  value is being used


miconda:
Thanks!

PR #520, opened 2016-02-25
core/futex: fix bug in futex_try

tsearle:
atomic_cmpxchg returns the previous value.  If the previous value was 0 then it was previously unlocked and we now have acquired the lock.

Thus, it should  be c==0 that is considered as the case we enter the lock, not c!=0

As far as I can tell, lock_try (and lock_set_try) are currently no used anywhere in the code, so this fix shoudn't have any impact


miconda:
Thanks!

PR #522, opened 2016-02-29
sdpops improvements and tests

mslehto:
Complete re-implementation of sdp_remove_line_by_prefix(), maybe reviewing by diff does not make sense.
Also feedback is on unit test is welcome as I have not created such for Kamailio earlier.


miconda:
Thanks!

PR #524, opened 2016-03-01
siptrace: Use the trace_local_ip setting in sip_trace()

aphistic:
In some cases the trace_local_ip wasn't being updated when sending out sip trace messages.  This updates the cases I could find to use the trace_local_ip setting when needed.


miconda:
Thanks!

PR #534, opened 2016-03-07
siputils: tel2sip(): fix 'off by one' when allocating memory

dragos-oancea:
While building the sip uri in tel2sip() we have:
```
sip_uri.s = pkg_malloc(sip_uri.len) 
```
But later in pv_set_ruri():
```
val->rs.s[val->rs.len] = '\0';
```
It's overwriting the q_malloc control structures (a classic type of bug already).

Kamailio will log a line like this before crashing:
```
Mar  3 13:31:15 XXXXXXXX /opt/ims/sbin/kamailio[15030]: : <core> [mem/q_malloc.c:140]: qm_debug_frag(): BUG: qm_*: prev. fragm. tail overwritten(c0c0c000, abcdefed)[0x7feaf22bd558:0x7feaf22bd588]!
```
I get this crash with long "tel:" ruris, the ones that have a ";phone-context=" param . 
This patch fixes it.

### DIFF
--- a/modules/siputils/checks.c
+++ b/modules/siputils/checks.c
@@ -374,7 +374,7 @@ int tel2sip(struct sip_msg* _msg, char* _uri, char* _hostpart, char* _res)
 
 	/* reserve memory for resulting sip uri */
 	sip_uri.len = 4 + tel_uri.len - 4 + 1 + hostpart.len + 1 + 10;
-	sip_uri.s = pkg_malloc(sip_uri.len);
+	sip_uri.s = pkg_malloc(sip_uri.len+1);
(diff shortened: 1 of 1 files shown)


miconda:
What version are you using? I haven't had the time too look at the code, just that I remember there was a similar fix at some point, maybe it was a regressions afterwards ...

dragos-oancea:
I am using master. Your fix for tel2sip was in 7992a2b8d42bb7e8bcf1738cf042013ed126a47a , it's similar, but it's not the same thing.

miconda:
Ok, thanks for the clarification. So I guess it needs to be backported as well.

PR #535, opened 2016-03-09
websocket: Fix stats "connectons" typo's.

wdoekes:
The stats now list:

    ws_sip_max_concurrent_connections
    ws_msrp_max_concurrent_connections

Instead of:

    ws_sip_max_concurrent_connectons
    ws_msrp_max_concurrent_connectons

Note that the latter still exist as aliases to the corrected values for
a while.


miconda:
Thanks!

Issue #536, opened 2016-03-09
append_fromtag on upstream in-dialog requests


schoberw:
Record-Routing in in-dialog request are not needed by RFC 3261. But there are many clients out there not sending initial route-set in in-dialog requests. To provided them with the initial route-set together with UAC modified requests the ;ftag parameter must be the to-tag on upstream requests.
append_fromtag parameter added the tag value from "from" header always, even on upstream. This is fixed here.

It is not done on record_route_preset(). This should not be called on record routing anyway.

[rr_record.c.patch.txt](https://github.com/kamailio/kamailio/files/164708/rr_record.c.patch.txt)

miconda:
Can you make a pull request via github? In that way is easy to review and merge, plus Travis-CI will compile the patch to be sure it works fine on many distros.

schoberw:
Hi!

It seems I’ve done it wrong. I did a fork and then a pull request. This is obviously not seen by others. Right?
https://github.com/schoberw/kamailio/pull/1

Can you pls help me doing it right?

Thanks!
Br
Walter

Von: Daniel-Constantin Mierla [mailto:notifications@github.com]
Gesendet: Mittwoch, 09. März 2016 09:09
An: kamailio/kamailio
Cc: Walter Schober
Betreff: Re: [kamailio] append_fromtag on upstream in-dialog requests (#536)


Can you make a pull request via github? In that way is easy to review and merge, plus Travis-CI will compile the patch to be sure it works fine on many distros.

schoberw:
Thank you all! Done!

PR #537, opened 2016-03-09
append_fromtag on upstream in-dialog requests

schoberw:
Record-Routing in in-dialog request are not needed by RFC 3261. But there are many clients out there not sending initial route-set in in-dialog requests. To provided them with the initial route-set together with UAC modified requests the ;ftag parameter must be the to-tag on upstream requests.
append_fromtag parameter added the tag value from "from" header always, even on upstream. This is fixed here.

It is not done on record_route_preset().

Handles issue #536


miconda:
Thanks! I pushed the commit manually with you as author to get rid of extra merge pull request notification, but also follow the commit message guidelines for formatting (which I forgot to point to previously):

PR #539, opened 2016-03-10
rtpengine: Add timestamp column to rtpengine table

smititelu:
The rtpengine module is not affected by this change as it currently ignores the new column.
Updated db scripts and documentations.

Is it possible for the http://kamailio.org/docs/db-tables/kamailio-db-devel.html to be regenerated after this commit?

### DIFF
--- a/lib/srdb1/schema/rtpengine.xml
+++ b/lib/srdb1/schema/rtpengine.xml
@@ -50,6 +50,14 @@
         <natural/>
     </column>
 
+    <column id="stamp">
+        <name>stamp</name>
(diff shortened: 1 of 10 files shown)


smititelu:
> Then why the column is needed and at this stage of 4.4 development?

It is useful on mySQL side, to have an idea when the node was added. Maybe in the future, rtpengine module will also display this info when 'kamctl fifo nh_show_rtpp'.

miconda:
This can be merged to master, branch 4.4 was created.

I can't have any opinion myself if this worth backporting.

PR #540, opened 2016-03-10
db_mysql: change log level from WARN to INFO for ping checks

linuxmaniac:
Customers are complaining because of this kind of messages:
 > proxy[XXX]: WARNING: db_mysql [km_dbase.c:85]: db_mysql_submit_query(): driver error on ping: Lost connection to MySQL server during query

Not really and error here, so lets put it on INFO level then


miconda:
This is not a message that should appear too often and still useful to know if db connectivity doesn't stay up. Maybe warning is too much, rising various alarms of syslog monitoring tools, given that it can recover fine in such situation. However, maybe INFO should be used instead of DBG.

Issue #533, opened 2016-03-07
memory leak in permissions module


eschmidbauer:
Hello,
We've noticed a memory leak caused by reloading the trusted table data.
````
kamctl trusted reload
````
We verified this by running script to reload the data frequently and captured memory usage using kex + corex.
Thanks,
Emmanuel

**Comment by miconda** [comment]
Can you attach the stats about memory usage? You had them on irc, it will help identifying the part of the code leaking. As discussed on irc, seems to be introduced by the new feature in 4.4 related to trusted table in permissions module.

**Comment by eschmidbauer** [comment]
Hello, 
I found the memory leak in permissions.
Can you please push this patch?
```
diff --git a/modules/permissions/hash.c b/modules/permissions/hash.c
index cfc81fe..84b20e2 100644
--- a/modules/permissions/hash.c
+++ b/modules/permissions/hash.c
@@ -423,6 +423,7 @@ void empty_hash_table(struct trusted_list **table)
                while (np) {
                        if (np->src_ip.s) shm_free(np->src_ip.s);
                        if (np->pattern) shm_free(np->pattern);
+                       if (np->ruri_pattern) shm_free(np->ruri_pattern);
                        if (np->tag.s) shm_free(np->tag.s);
                        next = np->next;
                        shm_free(np);
```

miconda:
Make a pull request, then it can be easily merged. Don't forget to format the commit message as suggested by github commit guidelines from kamaikio.org wiki.

eschmidbauer:
Please see PR
https://github.com/kamailio/kamailio/pull/546

PR #546, opened 2016-03-14
Fix memory leak by freeing RURI in trusted table struct; issue #533

eschmidbauer:


miconda:
I am on small mobile device, so can't see it in full properly, but the commit message doesn't seem to be formatted properly. It doesn't have the module prefix on first line. Search on kamailio.org/wiki for the guidelines.  Iirc, there is a link to it in readme.md file in root kamailio folder.

miconda:
Thanks!

PR #550, opened 2016-03-23
Improve detection of anonymous proxies.

sokhapkin:
Anonymous proxies in mmdb2 database have no "country" attribute, but "registered_country" attribute only.


miconda:
Thanks, merged manually to prefix commit message with module name, as per commit guidelines.

PR #552, opened 2016-03-28
dialog module :  new rpc function for terminate

ycaner06:
adding new rpc function to terminate a dialog with callid, to tag and from tag


miconda:
The new rpc command for dialog module is useful. I see you closed this pull request, do you plan to make a new one for it? Thanks.

ycaner06:
travis gives error. i couldnt understand. i am testing it but couldnt find any mistake.

miconda:
If you click on the red X before the commit hash id, then it goes to travis-ci page. There you have the reports - click again on the jobs for clang or gcc to see the output of compilation.

PR #553, opened 2016-03-29
modules/sipcapture: fixed correlation_id pointer for HEPv2/v1

adubovikov:


miconda:
Should it be merged along with #554?

adubovikov:
yes, i will do it right now

PR #555, opened 2016-03-30
dialog module : new rpc function for terminate

ycaner06:
- added new rpc command for terminate a dialog with callid, fromtag and totag.
- added new rpc command to doc and serctl is removed 
- added rpc return code  to dlg_terminate_dlg function
- added rpc return code to rpc domain reload function

### DIFF
--- a/modules/dialog/dialog.c
+++ b/modules/dialog/dialog.c
@@ -1759,10 +1759,10 @@ static void rpc_dlg_terminate_dlg(rpc_t *rpc,void *c){
 
         if(ret>=0){
                 LM_WARN("Dialog is terminated callid: '%.*s' \n",callid.len, callid.s);
+                dlg_release(dlg);
                 rpc->add(c, "d", 200);
                 rpc->add(c, "s", "OK");
                 rpc->send(c);
-                dlg_release(dlg);
         }
 }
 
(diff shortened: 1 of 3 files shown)


miconda:
I am going to apply manually the patched adding the new command for dialog to terminate by callid, from/to tag. I would change the name from dlg.dld_terminate_dlg in dlg.terminate_dlg.

For the rest of the patches, as they add embedded response code/text to a successful response, not data from dialogs. I would open a discussion on sr-users to decide on it, because it changes the current behaviour that is know to work and it can break many existing tools. Maybe this new behavior can be enabled by a modparam.

ycaner06:
Thanks Daniel. it is pleasure to add a reponse code for xml rpc by me. i want to help about it.

miconda:
First two patches are pushed in master now.

I noticed that the code was indented with whitespaces. Kamailio code uses tabs -- would be easier to apply the patches if you are going to use tabs for future patches.

ycaner06:
okey, i will keep in my mind. Thanks Daniel.

miconda:
For reference - the xmlrpc specs: http://xmlrpc.scripting.com/spec.html

You should open the discussion on sr-users mailing list, because you know better what you try to solve/improve with these patches. Describe what you don't like to current behaviour and the benefits after the patches, so the other users can comment and get to a solution.

ycaner06:
after IRC meeting , when success , it returns empty string is ok.

PR #559, opened 2016-04-01
Added connection and command timeout parameters to ndb_redis module

misaksen:


miconda:
Thanks. Applied the patch manually with commit 8992021f253fc7de38359aa7e51e5d4d6c705c34, because the indentation was done with whitespaces instead of tabs. Also, for next time, format the commit message according with the suggestions from git commit guidelines (see the README.md) for more details.

PR #565, opened 2016-04-12
new function in log_systemd module to log structured events

camilleoudot:
Journald can store arbitrary **KEY=>values[]** properties for each log event. The `sd_journal_send_xavp()` function is a binding on journald's `sd_journal_sendv()` function, that allows pushing structured events in journald, using the fields from a xavp.

(use `journald --output json-pretty` to see all the fields for the events in the journal)


miconda:
Thanks, the PR can me merged.

And good to see the module is useful for further enhancements, I implemented it more like a proof of concept without much testing.

PR #567, opened 2016-04-14
kex: Call pkg_proc_update_stats only when needed

smititelu:
I created a patch for 4.1 which allows to display also the total number of pkg fragments upon calling `kamcmd pkg.stats`, inspiring myself from the latest upstream. This PR comes as a result to enabling MALLOC_STATS and noticing kamailio slowdown with my applied patch, for 4.1. The slowdown was happening because each time a memory function was called, iteration through all fragments happened upon `pkg_info`. The solution was to call the stats gathering only when kamcmd command is issued.

On master this slowdown is not happening anymore, because `qm->ffrags` variable counts the fragments "on the fly", while memory functions are called. Still, is it necessarily to call pkg_proc_update_stats() callback every time?!

Feel free to close this if not needed.

### DIFF
--- a/events.c
+++ b/events.c
@@ -125,11 +125,6 @@ int sr_event_register_cb(int type, sr_event_cb_f f)
 					_sr_events_list.run_action = f;
 				else return -1;
 			break;
-		case SREV_PKG_UPDATE_STATS:
-				if(_sr_events_list.pkg_update_stats==0)
-					_sr_events_list.pkg_update_stats = f;
-				else return -1;
-			break;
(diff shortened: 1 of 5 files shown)


miconda:
I doesn't look ok, because the stats will be built only for rpc process, not for all kamailio processes.

Iirc, the pkg stats were refactored because they were initially slow, but probably not available in 4.1. Anyhow, this solution doesn't seem ok, either a full backport or some other solution if you want to push something to 4.1.

smititelu:
Thanks for making it clear for me. Closing this for now.

PR #575, opened 2016-04-19
kex: adding total_frags to kamcmd pkg.stats

lbalaceanu:
Created a patch for 4.1 which allows to display also the total number of pkg fragments upon calling kamcmd pkg.stats. This combines part of the master logic with existing 4.1 logic (for example this version does not combine the SREV_PKG_SET_USED, SREV_PKG_SET_REAL_USED, SREV_PKG_SET_FRAGS  into one callback.)

Does this look ok?  Feel free to close this if not needed.

### DIFF
--- a/events.c
+++ b/events.c
@@ -127,6 +127,11 @@ int sr_event_register_cb(int type, sr_event_cb_f f)
 					_sr_events_list.pkg_set_real_used = f;
 				else return -1;
 			break;
+		case SREV_PKG_SET_FRAGS:
+			if(_sr_events_list.pkg_set_frags==0)
+					_sr_events_list.pkg_set_frags = f;
+				else return -1;
+			break;
 		case SREV_NET_DGRAM_IN:
 				if(_sr_events_list.net_dgram_in==0)
 					_sr_events_list.net_dgram_in = f;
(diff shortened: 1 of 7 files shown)


miconda:
Is this actually replacing the old way in 4.1 of returning total_frags (which was iterating through hash table slots) with the new way in more recent 4.x that keeps a dedicated field for this statistic and is faster?

4.1 is no longer in the focus for maintenance, no releases out of that branch. If people are still using 4.1 from git and this is a fix to the slow approach of getting the total_frags, I don't have anything against merging if nobody else finds issues with it.

lbalaceanu:
Hi Daniel,

You are right, this is meant to replace the iteration related to total_frags with the new field.

On a deeper analysis however, I see I have not fully taken in consideration (at least) the shm vs private memory distinction in relation to triggering the execution of sr_event SREV_PKG_SET_FRAGS in q/f_malloc.c (in master you introduced the qm->type==MEM_TYPE_PKG which allows you to change the position of sr_event_exec for SREV_PKG_UPDATE_STATS vs previous SREV_PKG_SET_USED ) 
This enhancement seems more complex than anticipated. I will close the pull request.

PR #583, opened 2016-04-26
modules/*/doc: reorganize “Known limitations”

mslehto:
I was reading READMEs in terminal and noticed that single sub-section can look confusing:
https://github.com/kamailio/kamailio/blob/master/modules/db_mongodb/README#L52

In comparison HTML layout does not contain such orphan section index:
http://kamailio.org/docs/modules/devel/modules/db_mongodb.html#idp19616368

This patch tries to improve confusing “Known limitations” subsections by making them independent sections.

### DIFF
--- a/modules/db_mongodb/README
+++ b/modules/db_mongodb/README
@@ -10,7 +10,7 @@ Daniel-Constantin Mierla
 
    <miconda@gmail.com>
 
-   Copyright © 2014 asipto.com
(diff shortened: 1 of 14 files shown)


mslehto:
also, textops had double “Known limitations”

miconda:
Thanks, working on docs is always very appreciated!

PR #587, opened 2016-04-28
Fix memory leak when converting body to multipart with set_body_multipart();

sergey-vb:
Hi, All.

I have some problems with pkg_memory leak.

After each call used private memory of the "udp receiver" process
increases at 2500 bytes.
On test environment I have set it to 2Mbyte, but it is full after 700 calls.

I need to modify SDP in invite and other packets.
When I do in script

```
if(has_body("application/sdp"))
   set_body_multipart();
   if (msg_apply_changes())
    {
       xlog("L_INFO", "ISUP 1 Changes Applied Succesfully");
    }
}

record_route();
dlg_manage
route(RELAY);
```

I make test this on kamailio 4.3.5 and 4.4.0 - same issue.

After discussion in mailing list with Daniel-Constantin Mierla, I have done debugging memory usage by process and found 3 points where memory hase not freeing after execution.

This patch solves it for me.

### DIFF
--- a/modules/textops/textops.c
+++ b/modules/textops/textops.c
@@ -1763,6 +1763,8 @@ int set_multibody_helper(struct sip_msg* msg, char* p1, char* p2, char* p3)
 	}
 	pkg_free(nbb);
 	if(nc.s!=NULL) pkg_free(nc.s);
+	if(convert && nb.s!=NULL) pkg_free(nb.s);
+	if(convert && oc.s!=NULL) pkg_free(oc.s);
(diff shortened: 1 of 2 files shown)


miconda:
Thanks, the patches were applied manually 2b427a33937c6a86ab231de39d5a4574309dbfc6 61435f40f0f4ca25521f7f1b8515e47954a69b2d to format properly the commit message and also use tabs for indentation of the code.

For future, it would be good to use tabs and format the commit messages like:

It will make the merge straightforward.

PR #588, opened 2016-04-28
Added: Oracle Enterprise Linux 7 kamailio.spec file

kvishnivetsky:


miconda:
Thanks!

I wonder if we should make a dedicated folder for oracle distro, like pkg/kamailio/oracle/. The pkg/kamailio/rpm/ tried to collect the generic ones, but proved over the time that each distro has its own specifics. Perhaps it needs to be split into a folder with opensuse specs, then remove. Centos has already its own folder, as well as fedora.

What do you think?

kvishnivetsky:
Agree. Let's make it this way for now.
In my own practice I use mock environment for building RPMs for different distributions from a single SPEC file with RPM macros and conditions within.
What do you think about this way, as a possiblity of multi-distribution(RPM-based certainly) environment?

miconda:
Thanks! I applied the combined patches manually in order to format the commit messages as per guidelines:

It the future would be good to format the commit message of your patches based on the guidelines to make the merge straightforward. We use the commit messages for logs and we try to keep them coherent in format and consistent in description.

Should other files from rpm folder be copied to the new oracle folder (e.g., init.d script), in order to make building the rpms without need to do manual copy at the build time?

I am not using rpm-based distro for quite many years, so I am not familiar with mock tool, but if you get the scripts/specs for mock, we can create a folder rpm-mock and host them there inside the kamailio git repo.

kvishnivetsky:
Ok. I see.

PR #589, opened 2016-04-28
lib/binrpc modules/ctl utils/kamcmd : set UNIX_PATH_MAX to lowest common value

mslehto:
sockaddr_un.sun_path[] seem to be either 104 or 108 in following Unices:
104 - NetBSD, OpenBSD, FreeBSD, DragonFly BSD, MINIX, XNU(Apple OS X)
108 - illumos, Solaris, Linux

### DIFF
--- a/lib/binrpc/binrpc_api.c
+++ b/lib/binrpc/binrpc_api.c
@@ -53,7 +53,7 @@
 #define binrpc_free internal_free
 
 #ifndef UNIX_PATH_MAX
-#define UNIX_PATH_MAX 108
+#define UNIX_PATH_MAX 104
 #endif
 
 #ifndef INT2STR_MAX_LEN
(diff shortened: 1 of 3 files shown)


miconda:
Why the lowest and not the highest? Is it used for buffers where the values are copied?

mslehto:
UNIX_PATH_MAX is used to validate string length before memcpy() to sun_path.

https://github.com/kamailio/kamailio/blob/master/lib/binrpc/binrpc_api.c#L148
https://github.com/kamailio/kamailio/blob/master/modules/ctl/init_socks.c#L94
https://github.com/kamailio/kamailio/blob/master/utils/kamcmd/kamcmd.c#L489

miconda:
Thanks for patch and pointers, helped to clarify.

PR #593, opened 2016-05-02
a single digit is a valid enum_query() request

boettner:
We use enum_query() to dispatch calls over a set of gateways. In our case single digits (for instance "8") can be a valid direct dial number which we would like to through against our enum server.

Is there any good reason to limit the request to 2 digits as minimum?


miconda:
Thanks for patches!

Can you amend the commit messages in order to be formatted as per guidelines at:

PR #594, opened 2016-05-02
a single digit is a valid enum_query() request

boettner:
We use enum_query() to dispatch calls over a set of gateways. In our case single digits (for instance "8") can be a valid direct dial number which we would like to through against our enum server.

(This is an update of #593, now with guideline formatted commit messages.)

### DIFF
--- a/modules/enum/enum.c
+++ b/modules/enum/enum.c
@@ -254,7 +254,7 @@ static inline int is_e164(str* _user)
 	int i;
 	char c;
 	
-	if ((_user->len > 2) && (_user->len < MAX_NUM_LEN) && ((_user->s)[0] == '+')) {
+	if ((_user->len > 1) && (_user->len < MAX_NUM_LEN) && ((_user->s)[0] == '+')) {
 		for (i = 1; i < _user->len; i++) {
 			c = (_user->s)[i];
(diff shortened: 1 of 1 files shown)


oej:
Have you checked what the ENUM rfc says?

miconda:
I think from e164 format point of view, single digit number is too short. But then, I think ENUM doesn't enforce a length, because it can be used out of classic telephony numbering plan.

I am fine to merge the patches, after all, the length can be checked in the config file if one wants to impose some restrictions about it.

oej:
I would suggest a config option for accepting this - some users may rely on it generating an error condition

miconda:
As a matter of fact, even the current version of allowing 2 digits is probably not an e164 number. The only country code that is one digit is US/Canada with '1', but then there is no real number in their numbering plan with just a single digit after the country code.

The patch for the code is to a part where it checks that the length is sufficient to be sure that accessing the char at position 0 (first char) is safe and it is '+'. Because after that is a FOR loop starting at positing 1 (second char) to see if there are digits.

It might be useful to add min/max_length restrictions, but is more or less not related to this patch (or better said, not only related to this part). That could be an enhancement.

miconda:
Merging this pull request. The limits can be added with a new set of patches.

PR #595, opened 2016-05-03
pkg/kamailio/deb: use libradcli-dev for radius

linuxmaniac:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=822338

kamailio-radius-modules depends on libfreeradius-client
libfreeradius-client has been deprecated in favor of radcli[1]


miconda:
So no changes required to radius modules -- only to pkg specs, right?

The PR can be merged. Thanks.

linuxmaniac:
Still under testing

linuxmaniac:
> So no changes required to radius modules -- only to pkg specs, right?

Sadly, no. The Debian package doesn't provide the compatible include files freeradius-client.h nor radiusclient-ng.h, so we have to options:

1.  remove radiusclient-ng support and use radcli instead
2. use another define to select the proper radius library ( RADCLI ?)

miconda:
A define is better because the radiusclient-ng is the only one available in some (old) OSes.

amessina:
Reporting in...  For Fedora 23, the following can be accomplished with just a spec file change:
```
%package radius
Summary: Radius AAA API for Kamailio
Group: Applications/Internet
Requires: kamailio = %{version}-%{release}
#Requires: radiusclient-ng
#BuildRequires: radiusclient-ng-devel
Requires: radcli
BuildRequires: radcli-compat-devel
```

linuxmaniac:
> A define is better because the radiusclient-ng is the only one available in some (old) OSes.

defined RADCLI

linuxmaniac:
@miconda can this changes be included in 4.4.1?

miconda:
Yes, please.

PR #597, opened 2016-05-04
Add missing modules to the spec file for CentOS 6

tdimitrov:
Some modules (auth-xkeys, dmq_userloc, rtjson, statsd, tcpops) are missing from the spec file for CentOS 6, which causes errors during rpmbuild process. These modules are added as separate packages in this branch.


miconda:
Thanks, I will merge the patches manually to format the commit messages according to:

In the future, prefix the first line of the commit message with component name, in this case can be "pkg/rpm".

miconda:
Merged manually - commit referenced above.

PR #606, opened 2016-05-09
db_unixodbc: some ODBC drivers do not support ODBC replace query

giovannimele:
unix ODBC offers the possibilty to do an update or insert in one replace query. Unfortunately, Oracle unixODBC driver doesn't support this feature. Modifications have been done in the db_unixodbc module to allow to replace the query by 2 queries : update and if not exist, insert. This can be controlled by a new parameter replace_query, set to '1' by default in order to be backward compatible.


miconda:
Thanks, it will be merged!

PR #609, opened 2016-05-10
pkg/kamailio/deb/jessie: change default location

fredposner:
- default config file is /usr/local/etc/kamailio. kamailio.init CFGFILE should match installed location.
- kamailio is installed in /usr/local/sbin. kamailio.init PATH should have /usr/local/bin:/usr/local/sbin.
- kamailio is installed in /usr/local/sbin. kamailio.init DAEMON should match installed location.
- CFGFILE and ExecStart should match installed locations of kamailio.cfg and kamailio

### DIFF
--- a/pkg/kamailio/deb/jessie/kamailio.init
+++ b/pkg/kamailio/deb/jessie/kamailio.init
@@ -14,14 +14,14 @@
 
 . /lib/lsb/init-functions
 
-PATH=/sbin:/bin:/usr/sbin:/usr/bin
-DAEMON=/usr/sbin/kamailio
+PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin:/usr/local/bin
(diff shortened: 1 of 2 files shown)


linuxmaniac:
@fredposner the installed location is /usr/sbin/kamailio, /etc/kamailio etc... in debs
why you need all of this changes?? This files are for build the debs not for manual instalations

fredposner:
@linuxmaniac I didn't realize they were for the packages. I apologize. Installing from git, and adding start-up scripts, one has to change those sections. I see, based on your comment, the value of not accepting my change.

miconda:
For init.d scripts, there is a make target, something like:

```
make install-initd-debian
```
It copies the files from pkg folder, but then does some sed replacements to adjust to the target folder, even following the PREFIX preference.

Perhaps we should add a target for systemd as well, being now the default in latest OSes.

PR #610, opened 2016-05-10
change answered elsewhere to call completed elsewhere

quentusrex:
"Call completed elsewhere", RFC3326, defines this as the string to send.

Author: Frank Carmickle <frank@silentcircle.com>

### DIFF
--- a/modules/tm/t_cancel.c
+++ b/modules/tm/t_cancel.c
@@ -43,7 +43,7 @@ typedef struct cancel_reason_map {
 } cancel_reason_map_t;
 
 static cancel_reason_map_t _cancel_reason_map[] = {
-	{ 200, str_init("Answered elsewhere") },
+	{ 200, str_init("Call completed elsewhere") },
 	{ 0, {0, 0} }
 };
 


miconda:
Thanks! I will apply the patch manually, as we do not push patched directly to stable branches. They are pushed to master branch and then backported. Also, the commit message needs to be prefixed with the component name, in this case "tm":

I noticed that github allows now templates and other tools to highlight better the contributing rules, but no time yet to work on.

oej:
The text is not normative in the RFC and it should not really matter - we should really send the text sent in the 200 OK of the answer. The "200" code is the important part - a SIP response code. the "Call completed elsewhere" text is part of the examples in the RFC.

Do you have an implementation who parses reason texts?

quentusrex:
@miconda I can update the patch. 

@oej Yes, there are a few mobile implementations that strstr match several sip messages to avoid fully parsing them.

miconda:
@quentusrex: thanks, updating the patch will make the merge easier -- don't forget to make it for master branch.

miconda:
Merged manually.

PR #615, opened 2016-05-12
modules/sdpops: remove_line_by_prefix pre-check cleanup + new test

mslehto:


mslehto:
Looks like I did not provide any description, sorry about that.

Purpose of these patches is to tidy up unnecessary checks and calculations that are already done by parse_sdp().
Previously created tests (60...) show no change in behaviour and new test confirms that null body does not cause crash.

miconda:
Thanks!

PR #623, opened 2016-05-16
pua_dialoginfo: reduce the log level of every call to dialog_publish

phil-lavin:
Change the log level of calls to `dialog_publish()` from `INFO` to `DBG`. This tends to convolute the logs and is certainly no more prevalent that many other calls to `LM_DBG` across the codebase.


miconda:
Thanks!

PR #625, opened 2016-05-17
presence: Add local_log_facility configuration parameter

phil-lavin:
Similar to local_log_level - controls the syslog facility of certain log entries


miconda:
Thanks!

PR #627, opened 2016-05-17
Add new module auth_sha256 that can be used inplace of module auth.

AndreasHuber-CH:
Providing same API and functionality as auth module, but using SHA-256 instead of MD5 for digest authentication.


sipidronov:
Can it be a part of auth module with some sort of modparam switching between algorithms?

AndreasHuber-CH:
It could. I proposed different solutions before to the mailing list (without any replies). The impact would be about:

- Add a module parameter "algorithm" that allows configuring "SHA-256" instead of "MD5"
- Add SHA-256 versions for the functions calc_HA1 and calc_response (similar to rfc2617.c/h) and add two function pointers that can either point to the MD5 or SHA256 implementations, depending on the config.
- Use the function pointers in the functions auth_s, auth_check_response and pv_authenticate where calc_HA1 or calc_response are used.
- In the function auth_check_response replace HASHHEX allocated on the stack by at least char[65] and compare the length of the credentials provided by the client dynamically depending on the config (32 for MD5, 64 for SHA-256).
- Provide the configured algorithm to the two calls to get_challenge_hf to build the challenge header field with the desired algorithm.

The defines for HASHLEN, HASHHEXLEN, HASH and HASHHEX are tricky. The are also also used in the API but must be doubled in size to account for the hash value that is twice as long in for SHA-256.

Would you prefer this solution ? If yes, I'll propose an implementation. We would certainly be happy if the SHA-256 feature is directly part of the auth module.

linuxmaniac:
> Would you prefer this solution ? If yes, I'll propose an implementation.

I would say this should to be implemented at auth module.

AndreasHuber-CH:
OK. I propose a solution which puts the new code directly in the auth module: https://github.com/kamailio/kamailio/pull/628

oej:
The problem here is migrating a user base. You want this setting to be settable on an account-level basis, both in the subscriber table and by a function or avp in the routing script (for pv-based authentication). 
We should also note that there is NO rfc supporting this - even though it is a good idea. The IETF groups are still trying to find out what to do after MD5.

AndreasHuber-CH:
Yeah, I thought that this would be nice to make render the algorithm to use more dynamic.
Maybe this could be added more easily to the other pull request I proposed: #628

For instance, the module ims_auth has a additional method "ims_www_challenge" that takes the algorithm as an additional parameter. So this would give the freedom to select the algorithm dynamically from config file.

miconda:
The PR #628 was merged.

Further enhancements are welcome by any of the developers, like those proposed here regarding specifying the algorithm via function parameters.

I close this item here, additions could be a new PR.

PR #630, opened 2016-05-20
ims_registar_scscf: Fix subscr data handling in SAA

tdimitrov:
During re-registration, Diameter SAR message is sent to the HSS. It contains User-Data-Already-Available AVP, which by specification (TS 29.228, Table 6.1.2.1) instructs the HSS not to send subscription data. However when SAA is received without User-Data an error is generated. I think this behavior is not correct.

This patch modifies SAA handling so it doesn't try to update local ims subscription information, for the case where there is none received from the HSS.


miconda:
Thanks!

Patch is slim, but maybe @jaybeepee, @richardgood or @ngvoice want to have a look and comment. If all is ok, any can do the merge.

ngvoice:
Hi,

you are right, it's not correct. I will test & review the patch later today.

Thanks,
Carsten

jaybeepee:
Thanks @tdimitrov - looks fine!

tdimitrov:
You're welcome, I'm glad I can help!

Is it possible merge this patch in 4.3 and 4.4 branches too?

PR #632, opened 2016-05-22
modules/presence_conference: Fix compilation on FreeBSD

mslehto:
Fixes compilation on FreeBSD where no special feature request is needed


miconda:
Thanks!

PR #634, opened 2016-05-24
uac: reset UAC_REG_DISABLED flag (if need) after refresh

snen:
i have noticed, it is possible set UAC_REG_DISABLED flag by uac.reg_refresh, but unpossible to reset it


miconda:
Thanks!

PR #637, opened 2016-05-25
Fix or silence warnings given by Clang

mslehto:
```
CC (clang38) [M ims_dialog.so]          dlg_hash.o
dlg_hash.c:674:37: warning: equality comparison with extraneous parentheses [-Wparentheses-equality]
            if ((d_entry_out->first == d_entry_out->last)) {
                 ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
dlg_hash.c:674:37: note: remove extraneous parentheses around the comparison to silence this warning
            if ((d_entry_out->first == d_entry_out->last)) {
                ~                   ^                   ~
dlg_hash.c:674:37: note: use '=' to turn this equality comparison into an assignment
            if ((d_entry_out->first == d_entry_out->last)) {
                                    ^~
                                    =

CC (clang38) [M ims_qos.so]             rx_avp.o
rx_avp.c:492:16: warning: comparison of unsigned expression >= 0 is always true [-Wtautological-compare]
        if (bandwidth >= 0) {
            ~~~~~~~~~ ^  ~
rx_avp.c:504:16: warning: comparison of unsigned expression >= 0 is always true [-Wtautological-compare]
        if (bandwidth >= 0) {
            ~~~~~~~~~ ^  ~

CC (clang38) [M ims_registrar_scscf.so]         save.o
save.c:114:21: warning: comparison of unsigned expression >= 0 is always true [-Wtautological-compare]
    if (expires_hdr >= 0)
        ~~~~~~~~~~~ ^  ~
```

### DIFF
--- a/lib/ims/ims_getters.c
+++ b/lib/ims/ims_getters.c
@@ -132,9 +132,9 @@ str cscf_get_private_identity(struct sip_msg *msg, str realm) {
 	}
 		
 	if (realm.len && realm.s) {
-		ret = find_credentials(msg, &realm, HDR_AUTHORIZATION_F, &h);
+		ret = find_credentials(msg, &realm, HDR_AUTHORIZATION_T, &h);
 		if (ret < 0) {
-			ret = find_credentials(msg, &realm, HDR_PROXYAUTH_F, &h);
+			ret = find_credentials(msg, &realm, HDR_PROXYAUTH_T, &h);
 			if (ret < 0) {
 				goto fallback;
 			} else {
@@ -202,7 +202,7 @@ str cscf_get_private_identity_from(struct sip_msg *msg, str realm)
 	}
         
         if (realm.len && realm.s) {
-            ret = find_credentials(msg, &realm, HDR_AUTHORIZATION_F, &h);
+            ret = find_credentials(msg, &realm, HDR_AUTHORIZATION_T, &h);
             if (ret < 0) {
                     goto fallback;
             } else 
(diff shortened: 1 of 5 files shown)


miconda:
For the commit 94b4c25, is it handling properly the value 0?

Not dug much into the code, so this is more like a safety check to see that the return code or the parameter was not supposed to be negative in case of error or missing expires header.

Given those patches are related to ims modules, maybe @jaybeepee, @richardgood or @ngvoice may want to take a look.

mslehto:
Uh, thanks. I should not have stared those warnings with too narrow view.

Maybe more proper Wtautological-compare fix is then to declare bandwidth and expires_hdr as
'int'. Type is now 'unsigned int'.

miconda:
Yes, using signed int might be a solution. Maybe you can discard/revert the patch 94b4c25, so the others can be merged.

mslehto:
Sure, reverted. Travis CI check passed, but it looks like it does not try to build any ims modules.

linuxmaniac:
Yes, on travis We build only the modules included in the group standard

miconda:
@linuxmaniac can we enable some more groups? I would be interested in the one very important such as tls, websocket, presence ...

linuxmaniac:
I'm on it. I'm going to build all modules provided by debian packages.

miconda:
Great, thanks! I will merge this pull request.

PR #638, opened 2016-05-25
Fix variable initialization and NULL pointer dereference issues

mslehto:
Most found with Clang 3.8 static analyzer.
Full results here: http://mslehto.iki.fi/clang-static-analyzer/kamailio/01/modules/


miconda:
Thanks!

PR #639, opened 2016-05-25
Add support for Authentication-Info header in auth module

AndreasHuber-CH:
The Authentication-Info header is already supported by the ims_auth module. This pull requsts proposes an implementation for auth module.
The Authentication-Info header is added in the function "post_auth" (if "add_authinfo_hdr" is configured.)
Modules depending on "auth" are adapted slightly to offer the ha1 value to the function "post_auth".


miconda:
Thanks, useful addition, so the client can now authenticate the server as well.

PR #642, opened 2016-05-26
Read all 12 columns from uacreg if the database does not have DB_CAP_…

AndyJRobinson:
When the database does not have DB_CAP_FETCH capability, only 10 columns are being read from uacreg instead of 12, so the flags and reg_delay columns are being omitted.


miconda:
Thank. I will push the patch manually to format the commit message according with the guidelines:

PR #647, opened 2016-05-30
mem: Add on-error log messages for q_/f_ functions

smititelu:
Log some errors in case some memory API functions fail.


miconda:
Thanks!

PR #654, opened 2016-06-03
presence: add new parameter `subscribe_expires_offset`

eschmidbauer:
* defaults to 0
 * adds X seconds to expires before being stored in htable or database
 * causes subscribes to not be removed prematurely

### DIFF
--- a/modules/presence/doc/presence_admin.xml
+++ b/modules/presence/doc/presence_admin.xml
@@ -341,6 +341,27 @@ modparam("presence", "expires_offset", 10)
 		</example>
 
 </section>
+
+	<section id="presence.p.add_expires_offset">
+		<title><varname>add_expires_offset</varname> (int)</title>
+		<para>
+		Add expires_offset value to expires before inserting into db/htable
+		</para>
+		<para>
+		<emphasis>Default value is <quote>0</quote>.
+		</emphasis>
+		</para>
+		<example>
+		<title>Set <varname>add_expires_offset</varname> parameter</title>
+		<programlisting format="linespecific">
+...
+modparam("presence", "add_expires_offset", 1)
+...
+</programlisting>
+		</example>
+
+</section>
+
        <section id="presence.p.max_expires">
                <title><varname>max_expires</varname> (int)</title>
                <para>
--- a/modules/presence/presence.c
+++ b/modules/presence/presence.c
@@ -145,6 +145,7 @@ char prefix='a';
 int startup_time=0;
 str db_url = {0, 0};
 int expires_offset = 0;
+uint32_t add_expires_offset = 0;
 uint32_t min_expires= 0;
 int min_expires_action= 1;
 uint32_t max_expires= 3600;
@@ -207,6 +208,7 @@ static param_export_t params[]={
 	{ "force_delete",           INT_PARAM, &pres_force_delete },
 	{ "to_tag_pref",            PARAM_STRING, &to_tag_pref },
 	{ "expires_offset",         INT_PARAM, &expires_offset },
+	{ "add_expires_offset",     INT_PARAM, &add_expires_offset },
 	{ "max_expires",            INT_PARAM, &max_expires },
 	{ "min_expires",            INT_PARAM, &min_expires },
 	{ "min_expires_action",     INT_PARAM, &min_expires_action },
--- a/modules/presence/presence.h
+++ b/modules/presence/presence.h
@@ -71,6 +71,7 @@ extern int pid;
 extern int startup_time;
 extern char *to_tag_pref;
 extern int expires_offset;
+extern uint32_t add_expires_offset;
 extern str server_address;
 extern uint32_t min_expires;
 extern int min_expires_action;
--- a/modules/presence/subscribe.c
+++ b/modules/presence/subscribe.c
@@ -617,6 +617,11 @@ int update_subscription(struct sip_msg* msg, subs_t* subs, int to_tag_gen,
 			}
 			return 1;
 		}
+
+		if (add_expires_offset > 0) {
+			subs->expires = subs->expires + expires_offset;
+		}
+
 		/* if subscriptions are stored in memory, update them */
 		if(subs_dbmode != DB_ONLY)
 		{


eschmidbauer:
@miconda 
This resolved an on-going issue we noticed with BLF line keys losing presence (on polycoms we saw the line keys turning gray). We noticed it was happening because the Polycom was not re-subscribing before the expiration and therefore, kamailio was deleting the subscription from the database (as expected).
This patch adds X seconds to the expiration to give the UA more time to re-subscribe before deleting from the database.
Please let me know if you have any questions regarding this new parameter.
Thanks,
Emmanuel

lazedo:
@eschmidbauer doesn't `expires_offset ` already addresses this ?

eschmidbauer:
@lazedo It does not. Reading the doc, `expires_offset` seems to only affect PUBLISH requests.
Testing and examining the code, I found that `expires_offset` had not affect on SUBSCRIBE requests.

lazedo:
@eschmidbauer please check again. in subscribe.c the timers for subscription updates already take `expires_offset`. maybe there's a code path that is not taking this parameter ?

eschmidbauer:
@lazedo 
According to the doc on `expires_offset`:  The value in seconds that should be subtracted from the expires value when sending a 200OK for a publish. It is used for forcing the client to send an update before the old publish expires. 

This is not what we are attempting to resolve with this PR. We only want to add X seconds to the expires before inserting into db/htable. This gives extra time for the UA to re-SUBSCRIBE before kamailio removes from db/htable.

lazedo:
@eschmidbauer please read the code related to presence timers, they make the select with the expires_offset. the change that you're proposing, it seems to me, that will have the side effect on NOTIFY messages of sending the wrong remaining period in the subscription, if the UA takes the value from the NOTIFY then you will end up with the same problem again.

if you think that you found that expires_offset has no affect on SUBSCRIBE requests, please show us where. i see it being used in `update_db_subs_timer_notifier` , `update_db_subs_timer_dbonly`, `update_db_subs_timer_dbnone`, `update_db_subs_timer`. current_time is deducted the expires_offsite and gets subscription LT that value. check here [here](https://github.com/kamailio/kamailio/blob/b0a0b3dd12b1c218fffae75b9da7d454d1a39486/modules/presence/subscribe.c#L1945-L1946)

eschmidbauer:
@lazedo reviewing the code you mentioned, we would need a negative integer value in that select.
This is because we want to select expired subscriptions *late* not early
The desired behavior of our patch is different than desired behavior of `expires_offset`

lazedo:
@eschmidbauer negative integer ? where do you see that ?
```
subscriptionTime = 10:00:00
Expires = 300
subscriptionExpiresAt 10:05:00
ExpiresOffset = 60

Time is 10:05:45
select is 10:05:15 - 60 = 10:04:45
select all that subscriptionExpiresAt  LT 10:04:15
```

eschmidbauer:
@lazedo 
You just demonstrated exactly how my PR differs from `expires_offset`
We want to *add* seconds to the expiration, not subtract
```
subscriptionTime = 10:00:00
Expires = 300
subscribe_expires_offset = 60
subscriptionExpiresAt 10:06:00
```

miconda:
Rather than introducing a new offset parameter for the module that can create confusion, given that it uses also 'offset', add a parameter that controls if the offset is subtracted or added from the expires value.

eschmidbauer:
@miconda 
I would agree with you except that my code differs greatly than that of `expires_offset`
For instance, our code does not make use of timers at all. Our only goal here was to "extend" the expiration so Kamailio does not delete it from db/htable.
I'd be happy to introduce a patch as you described, but do you think at some point it would be desirable for a user to use both `expires_offset` and `subscribe_expires_offset` ?

lazedo:
@eschmidbauer sorry if i'm not explainin my self in a way you can understand.
by adding the value to the table you will probablt mess up with UAs that adjust the subscription period when they receive a NOTIFY.

i don't believe i demonstrated what your PR is trying to solve, i believe i demonstrated that what you're trying to solve is already there without affecting other UAs that adjust the subscription on NOTIFY.

what part didn't you understand here ?
```
Time is 10:05:45
select is 10:05:15 - 60 = 10:04:45
select all that subscriptionExpiresAt  LT 10:04:15
which doesn't select the above mentioned  subscription
```

eschmidbauer:
@lazedo if you are saying that `expires_offset` should already be performing the desired behavior of my patch, then I have to disagree with you. We've been running Kamailio without my patch for several weeks with `expires_offset = 60`
We noticed during the course of that time, that several times a day the LineKeys on the Polycom phones would lose subscription. We determined that Kamailio was removing the subscriptions from the active watchers table *before* the Polycom could re-subscribe. 
This is the simple use-case that my patch addresses: Give the polycom X extra seconds to re-subscribe before deleting from db/htable. 
If you are saying `expires_offset` should be addressing the above use-case, then it is broken and perhaps a fix is needed instead of a new parameter. We've tested my parameter for over 2 weeks now and it simply fixes the issue. We've actually left the `expires_offset = 60` and it had no adverse affects to that parameter.

lazedo:
@eschmidbauer so you're actually allowing your devices to diverge for 2 minutes, maybe that's why you think its solved ? how about running with `expires_offset = 120` and your new parameter = 0 ? mayne your polycoms need extra time ?

eschmidbauer:
@lazedo 
What do you mean diverge for 2 minutes? I can remove `expires_offset` and only use the new parameter and the issue will be resolved. 
We can run `expires_offset = 60` or whatever value (120, 240, etc...) and disable new parameter, I can guarantee the issue will happen. 
Are you sure `expires_offset` is meant to resolve the issue here? I believe it is meant to address another issue unrelated to our PR. The documentation on it and the code do not demonstrate our requirements at all.

lazedo:
@eschmidbauer we also had some issues with resubscriptions that were solved with this parameter.

you said you left expires_offset = 60 and your new parameter = 60. that will add your devices 2 minutes before their removed based on initial expire value and if your devices do not update subscription period on NOTIFY.

but you didn't answer my question, if you use expires_offset  = 120 and your new parameter = 0, will your problems go away or not ? try it and show us some logs.

but, i don't want you to think that i'm blocking you, if the the default = 0 then `i'm ok to accept it` making an internal note to never use it because of those UAs that do adjust the subscription period based on the NOTIFY they receive.

linuxmaniac:
test PR comment for sr-dev, please ignore

kamailio-sync:
test comment for sr-dev mail, please ignore

linuxmaniac:
test comment for sr-dev mail, please ignore

eschmidbauer:
@lazedo 
I removed expires_offset so now i have:
```
#modparam("presence", "expires_offset", 60)
modparam("presence", "subscribe_expires_offset", 60)
```

I will test the following on our staging
```
modparam("presence", "expires_offset", 120)
#modparam("presence", "subscribe_expires_offset", 60)
```

eschmidbauer:
@lazedo 
update: our Polycom on staging is losing presence with the following configuration
```
modparam("presence", "expires_offset", 120)
//modparam("presence", "subscribe_expires_offset", 60)
```

lazedo:
@eschmidbauer any logs to sustain that ?

lazedo:
@eschmidbauer btw, which `db_mode` are you using in your config ?

eschmidbauer:
@lazedo 
```modparam("presence", "subs_db_mode", 3)```
I'll try to get logs but it's difficult as the problem is sporadic

miconda:
Still no time to look deeper at the issues, but at least the new patch c7378df is better -- no longer very intrusive, with minimal impact on performance and the new parameter reflects better the fact that it adds to the expires. If it proves to solve issues when dealing with some phones, I am ok to merge provided that default behaviour is to do no change to the expires. But I would like to see more into its description in the docs, because it seems to affect only subscriptions. Moreover, it would be good to say if it conflicts or not with the expires_offset.

eschmidbauer:
@miconda @lazedo 
I'm canceling this PR for now. 
We are testing some settings on Polycom and we want to further analyze before we are sure this code is needed, thanks.
Emmanuel

Issue #563, opened 2016-04-05
cnxcc: Kamailio crashes with a seg fault at high traffic load


lfzaracho:
After a 10 minutes benchmark at 90 CPS Kamailio 4.3.5 crashes. I'm using money based credit control. The log output is as follows: 
```
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5230]: ALERT: <core> [main.c:728]: handle_sigs(): child process 5431 exited by a signal 11
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5230]: ALERT: <core> [main.c:731]: handle_sigs(): core was generated
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5230]: INFO: <core> [main.c:743]: handle_sigs(): terminating due to SIGCHLD
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5491]: INFO: <core> [main.c:798]: sig_usr(): signal 15 received
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5488]: INFO: <core> [main.c:798]: sig_usr(): signal 15 received
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5486]: INFO: <core> [main.c:798]: sig_usr(): signal 15 received
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5492]: INFO: <core> [main.c:798]: sig_usr(): signal 15 received
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5489]: INFO: <core> [main.c:798]: sig_usr(): signal 15 received
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5493]: INFO: <core> [main.c:798]: sig_usr(): signal 15 received
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5469]: INFO: <core> [main.c:798]: sig_usr(): signal 15 received
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5451]: INFO: <core> [main.c:798]: sig_usr(): signal 15 received
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5490]: INFO: <core> [main.c:798]: sig_usr(): signal 15 received
Apr  4 17:42:32 sipserver /usr/local/sbin/kamailio[5485]: INFO: <core> [main.c:798]: sig_usr(): signal 15 received
[...]

PR #656, opened 2016-06-06
cnxcc: release credit_data in terminate_all_calls

grumvalski:
This PR is related to GH #563 and probably to some other cnxcc's crashes that have been reported.


miconda:
Thanks, it can be merged/backported and close the related bugs.

PR #659, opened 2016-06-06
ims_registrar_scscf: Remove buggy AVP from SAR

fedefava86:
Remove buggy function to add
"Call-ID" AVP which actually doesn't exists
on SAR Diameter Message
According to TS129.229, it actually exists
"Call-ID-Sip-Header" AVP witch code 643
which is grouped into "Subscription-Info"
AVP. According to the same TS, this AVP
is used for restoration procedure.

### DIFF
--- a/modules/ims_registrar_scscf/cxdx_avp.c
+++ b/modules/ims_registrar_scscf/cxdx_avp.c
@@ -142,17 +142,6 @@ static inline str cxdx_get_avp(AAAMessage *msg,int avp_code,int vendor_id,
 		return avp->data;
 }
 
-inline int cxdx_add_call_id(AAAMessage *msg, str data) 
-{
-    return 
-	cxdx_add_avp(msg,data.s,data.len,
-		AVP_Call_Id,
-		AAA_AVP_FLAG_VENDOR_SPECIFIC,
-		50,
-		AVP_DUPLICATE_DATA,
-		__FUNCTION__);
-}
-
 /**
  * Creates and adds a Destination-Realm AVP.
  * @param msg - the Diameter message to add to.
--- a/modules/ims_registrar_scscf/cxdx_avp.h
+++ b/modules/ims_registrar_scscf/cxdx_avp.h
@@ -53,7 +53,6 @@ struct AAAMessage;
 struct AAA_AVP;
 struct sip_msg;
 
-inline int cxdx_add_call_id(AAAMessage *msg, str data);
 /**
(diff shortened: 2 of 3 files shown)


miconda:
I guess @jaybeepee, @richardgood or @ngvoice want to check before merging.

kamailio-sync:
yup, checking

fedefava86:
Any feedback regarding this ?

miconda:
Any comment by @jaybeepee, @richardgood or @ngvoice? Eventually can be made optional with a modparam instead of complete remove if someone still finds it useful for some cases.

fedefava86:
Eventually, if really needed, i can try to find some time to modify the Pull Request in order to introduce a modparam. The point is that, as far as i understand, the AVP does not follow any (known to me at least) 3gpp or RFC specs...so is pointless to add it as it is, if not used against a "custom" HSS that needs this AVP

jaybeepee:
Apologies Frederico, I've been swamped. Will take a look today

fedefava86:
Still nothing?

miconda:
If you introduce a nod parameters,  then I can step in and merge. If someone needs old behaviour, nothing is lost.

fedefava86:
Sorry if i insist on the point...but really i do not see the point in leaving such AVP. To me is sufficient a feedback like "Our HSS is caring about this AVP, because we have a custom HSS that....etc...etc". I will close the PR or add a modparam and that's ok. This way we can for sure have a fallback, but basically we are mantaining an "out-of-standard" thing (that, btw, even wireshark doesn't decode) without knowing if it is really needed or not. 

Anyway, copy that, with low priority i will try to find some time to add a modparam if really needed

jaybeepee:
Frederico, the problem is that the code is already in ..... there could be
other ppl using it, so I think it would be unfair to just remove it. We
have 2 solutions here:

1. move to a modparam.
2. move it to a vendor-specific AVP

Finally, we will need to find someone who has time/priority to get this
done... if you are worried only about the std, then I think it's a low
priority. If it's affecting your ability to use Kamailio, then it's of
course higher priority.

jaybeepee:
I've added a mod-param version for now and checked-in to master. When I find some time to dig through specs, I will add the std version.

jaybeepee:
BTW, Frederico, you will see the AVP is marked as vendor-specific already... which is why wireshark will not decode...

Trace some big vendor's diameter interfaces and you will see plenty of non-parseable avps in wireshark traces ;)

PR #663, opened 2016-06-07
Fix check_route_param()

dragos-oancea:
Given that there are multiple Route sets , each with one param , check_route_param() function fails to search the proper param.s string . 
The condition params.s[0]!=';'  holds false and it stops the loop only when it encounters another ";" character which is not part of the Route header .

Eg: (these are not actual Route params, but this is what it is printing instead - this is part of my INVITE, and these headers are above Route ):
rr [loose.c:985]: check_route_param(): params are <;spi-s=424238335#015#012Require: sec-agree#015#012Proxy-Require: sec-agree#015#012Cont>

This patch is a functional workaround. 
Note I have also seen there routed_params.s not pointing to the actual params , so perhaps there might be another issue before reaching this code.

### DIFF
--- a/modules/rr/loose.c
+++ b/modules/rr/loose.c
@@ -977,7 +977,9 @@ int check_route_param(struct sip_msg * msg, regex_t* re)
 		return -1;
 
 	/* include also the first ';' */
-	for( params=routed_params ; params.s[0]!=';' ; params.s--,params.len++ );
+	params=routed_params;
+	params.s--;
+	params.len++;
 
 	/* do the well-known trick to convert to null terminted */
 	bk = params.s[params.len];


miconda:
Can you add a log message before the for block you removed, so the code becomes:

```
LM_INFO("current params are <%.*s> (%x)\n", routed_params.len, routed_params.s, *(routed_params.s-1));
for( params=routed_params ; params.s[0]!=';' ; params.s--,params.len++ );
```

and then see if the info message prints the proper route params and the previous character in hexa?

As you said, the issue might be in other place, it's good to track it down properly.

dragos-oancea:
I added that logline . Sometimes it prints strange things like this: 
 
Jun  9 12:10:09 B1-RD-IMS-2 /opt/local/ims/sbin/kamailio[23328]: INFO: rr [loose.c:980]: check_route_param(): current params are <nt> (6f)
Jun  9 12:10:09 B1-RD-IMS-2 /opt/local/ims/sbin/kamailio[23328]: DEBUG: rr [loose.c:986]: check_route_param(): params are <;spi-s=1138095176#015#012Require: sec-agree#015#012P
roxy-Require: sec-agree#015#012Cont>
Jun  9 12:10:09 B1-RD-IMS-2 /opt/local/ims/sbin/kamailio[23328]: DEBUG: siputils [checks.c:97]: has_totag(): no totag
Jun  9 12:10:09 B1-RD-IMS-2 /opt/local/ims/sbin/kamailio[23328]: DEBUG: rr [record.c:897]: add_rr_param(): rr_param_buf=<;ipsec=yes;nat=yes>

miconda:
I pushed a patch to rr on master, can you give it a try? May or may not be related, though, ...

dragos-oancea:
I applied the patch , but I still have output like this : 
 rr [loose.c:985]: check_route_param(): current params are <k.> (72)

Let me know if I can help further.

miconda:
Are you using it from failure_route or branch failure route?

dragos-oancea:
No and no . 
I am using it in my route[RTPPROXY] like this : 
if (is_request() && !isflagset(FLT_NAT)) {
   if(check_route_param("nat=yes")) {
       setflag(FLT_NAT);
   } 
}

miconda:
But maybe the route(RTPPROXY) is executed from a branch or failure route (or another route executed from those) ...

dragos-oancea:
I am pretty sure it's a request route . It's true I call route[RTPPROXY] from a reply route too , but when I have the issue is always a request_route .  It has to do with the number of params in Route , it seems the bug becomes visible only when there's only one param (;lr) and then check_route_param("nat=yes") would return false anyway .

miconda:
OK, useful hints -- I will try to reproduce when I get the first chance.

miconda:
Got back to this one and pushed a new commit which does a bit of refactoring for this function, including the ability to work better with other types of routing blocks. Can you try again and see if the issue is still there.

dragos-oancea:
Hi Daniel ,

We've cherry-picked your commit . It works. 

Jul  6 13:05:58 B1-RD-IMS-2 /opt/local/ims/sbin/kamailio[4294]: DEBUG: rr [loose.c:1066]: check_route_param(): params are <;lr;r2=on;ftag=1692917141;ipsec=yes;nat=yes>
Jul  6 13:06:06 B1-RD-IMS-2 /opt/local/ims/sbin/kamailio[4295]: DEBUG: rr [loose.c:1061]: check_route_param(): route params checking against [;lr;r2=on;ftag=1692917141;ipsec=yes;nat=yes] (orig: [lr;r2=on;ftag=1692917141;ipsec=yes;nat=yes])

Thank you.

Regards,
Dragos

miconda:
OK -- thanks for testing, commit backported to 4.4.

PR #664, opened 2016-06-09
Exporting ds_reload to cfg script

luismartingil:


modules/dispatcher/doc/dispatcher_admin.xml
```diff
@@ -1350,6 +1350,20 @@ onreply_route {
 </programlisting>
 		</example>
  	</section>
+	<section id="dispatcher.f.reload">
```
oej: dispatcher.f.ds_reload is correct

modules/dispatcher/doc/dispatcher_admin.xml
```diff
@@ -1350,6 +1350,20 @@ onreply_route {
...
+		<title>
+		<function moreinfo="none">ds_reload()</function>
+		</title>
+		<para>
+		Reloads the groups and included destinations. The command is
+		disabled for call load based dispatching (algorithm 10) since
```
oej: The code doesn't indicate that it really is disabled in this situation. Do you mean "disabled" or "should not be used".

oej:
You can already run rpc commands from the script. What does this patch add?

luismartingil:
Thanks for the input. How would you run the RPC commands from the script?

modules/dispatcher/doc/dispatcher_admin.xml
luismartingil: This is based on [this](https://github.com/kamailio/kamailio/blob/master/modules/dispatcher/dispatcher.c#L914-L924) and [this](https://github.com/kamailio/kamailio/blob/master/modules/dispatcher/doc/dispatcher_admin.xml#L1415-L1437). I copied the possible problems description using a similar function.

oej:
There is support in the jsonrpc-s module. Kind of hidden, but still there :-)

http://kamailio.org/docs/modules/4.4.x/modules/jsonrpc-s.html#jsonrpc-s.f.jsonrpc_exec

luismartingil:
Didn't see that one coming! ;-)

miconda:
It's ok to export this function directly, can be handy sometimes to use it directly, without rpc workaround.

But the description needs to be amended, because the reload is not disabled anymore for algorthm 10 (it was some time ago).

luismartingil:
Even though I see you're solution as a valid solution, I still see this exported function as valuable. Using RPC to trigger an internal mechanism can overload this procedure.

luismartingil:
@miconda I'll change the description.

oej:
I agree that it is useful as a simplification, but wanted to know if there was some additional benefits. You may want to add counters on how many dispatcher targets and groups that was loaded or something like that while you are at it - or a checksum check so I can check if the list was actually changed. Just some ideas to make this function an improvement :-)

luismartingil:
@miconda I'll just leave this _PR_ as it is, please let me know if you want to integrate this change and if you'll merge the changes and close the _PR_ or you want me to do it. Thanks!

PR #666, opened 2016-06-10
app_lua: add functions for handling script flags

jgallartm:


miconda:
These functions should be available already in master branch, but exported by the KSR module, not the old sr module:

  * https://github.com/kamailio/kamailio/blob/master/kemi.c#L484-L497

In lua script you have to use something like:

```
KSR.setflag(10);
```

The KSR is exported as part of the new KEMI, but can be used even if you don't write full routing blocks in Lua:

  * https://www.kamailio.org/wiki/devel/config-engines#lua_config_kemi_engine
  * https://github.com/kamailio/kamailio/blob/master/examples/kemi/kamailio-basic-kemi-lua.lua

jgallartm:
Hi Daniel, 

that's true, the patch makes sense for the 4.x versions. However, the flags we're trying to manage are these ones: http://kamailio.org/docs/modules/4.4.x/modules/kex.html#kex.f.setsflag, not the "regular" ones. I can't see those functions in master either.

miconda:
OK -- I wasn't expecting those "script" flags to be used much and thought it is about the message flags.

You will have to make the patch for master and use the kemi aprroach -- in this way the functions will become available to all kemi interpreters at once (right now lua and python). It should be simpler than what you did, no exporting via inter-module api and code in app_lua, it will be only some bits inside the kex module. Perhaps maxfwd is a good example to look at how a function is exported to kemi.

miconda:
Implemented via kemi - commit referenced above.

PR #670, opened 2016-06-14
utils: introduced the ability to add User-Agent header in HTTP requests.

os11k:
following commit will introduce possibility to add User-Agent header
for http_query by adding utils module parameter - user_agent,
if none is defined then no User-Agent header is added.

### DIFF
--- a/modules/utils/doc/utils_admin.xml
+++ b/modules/utils/doc/utils_admin.xml
@@ -192,6 +192,25 @@ modparam("utils", "xcap_table", "pres_xcap")
 				</programlisting>
 			</example>
 		</section>
+                <section id="utils.p.user_agent">
+                        <title><varname>user_agent</varname> (string)</title>
+                        <para>
+                        Defines User-Agent for HTTP requests.
+                        </para>
+                        <para>
+                        <emphasis>
+                                There is no default value, if none is defined then no User-Agent header is added.
+                        </emphasis>
+                        </para>
+                        <example>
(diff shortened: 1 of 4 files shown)


sipidronov:
Passing str when char* expected looks not correct for me.

os11k:
I'm very sorry, this is my first commit, so I might ask not so very clever questions. :)
I tried to use same logic for user_agent as for xcap_table, which is string.

miconda:
@sipidronov good catch, indeed. It should be user_agent.s and only set if the user_agent.len>0. Even in the current form works because .s is the firs field in the str structure, that is not safe for future.

I also think the variable and parameter should be prefixed with http_, as those related to the http query of utils are (e.g., timeout), to be more clear that it refers to the http functionality from the module.

oej:
This already exists in the http_client module. We're planning to remove http from utils, so why extend it with something you can already get in http_client?

sipidronov:
@miconda or use PARAM_STRING from the very beginning. BTW looks like passing NULL to API call is ok for curl.

os11k:
Historically we are using this module and we need possibility to add user-agent header for HTTP requests, if you plan to remove this module and you think that nobody needs such functionality, then probably we can drop this pull request. Thank you!

oej:
The functionality will only be added to coming releases and the plan is to delete the function from the utils module in favour of the new modules. The functionality already exist in 4.4 http_client module. Closing pull request.

PR #672, opened 2016-06-15
modules/tm documentation improvements

mslehto:
t_relay() return values copied from 1.5.x documentation:
http://kamailio.org/docs/modules/1.5.x/tm.html#id2514020

In route naming CAPITALS seem more common than <emphasis> (tm, tmx, sl, corex).


miconda:
Actually the tm is inherited from ser, not from kamailio 1.5, so that commit is likely not good, it has to be reverted and the source code needs to be checked for proper documentation.

mslehto:
Thanks, I'll close this for now.

PR #675, opened 2016-06-17
db_unixodbc: SQL_NO_DATA should not be considered as an error

giovannimele:
SQL_NO_DATA (100) should not be considered as an error when deletes or updates are done. currently, any return code different than 0 or 1 is considered as an error: a log is generated and the session closed.


miconda:
Thanks! Commit pushed manually to avoid the extra two merging notifications from this PR. I think you need to pull and do fast forward before making the pull request not to have those Merge ... commits as part of the PR.

kamailio-sync:
Hello,

I’ll do it next time.

Thx

Giovanni

PR #683, opened 2016-06-24
core: fix logging macros for Solaris Studio

### LINKED ISSUE #678: dprint.h logging macros are not portable
The logging macros make use of GCC (and compatible) variadic macros.  While there are ifdefs for Solaris Studio using C99 variadic macros, they don't compile and seem to be suffering from bit rot.

There are a couple ways to address this:
- rebase and fix the ```ifdef __SUNPRO_C``` portion which would introduce no changes for the GCC build but would have more longterm maintenance overhead
- replace the GCC extensions and ifdefs with C99 compliant ones which would not be as readable and require a few hacks to prevent trailing commas in the case of zero arguments.  This would remove the code maintenance burden and make the macros more portable.

I'll be happy to submit a pull request for either approach.

sjthomason:
Fix compile errors in logging macros when building with Solaris Studio

closes #678

### DIFF
--- a/dprint.h
+++ b/dprint.h
@@ -185,11 +185,11 @@ void log_prefix_init(void);
 #ifdef NO_LOG
 
 #	ifdef __SUNPRO_C
-#		define LOG__(facility, level, lname, prefix, fmt, ...)
-#		define LOG_(facility, level, prefix, fmt, ...)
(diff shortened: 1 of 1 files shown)


miconda:
Thanks!

PR #689, opened 2016-06-30
Fix Solaris Studio build

sjthomason:
This allows building with Solaris Studio, tested with 12.4.


miconda:
Analyzing the first patch (3678a66) -- it is adding a restriction on using only C99 compilers, right? Meaning that old compilers that support only c90 are no longer going to work ...

I think I am fine over all, but we should check if someone has something against as some people used to build on some old embedded systems.

sjthomason:
Hi Daniel,
I have update the PR to maintain the GCC CPP compatibility.  That was an oversight, I didn't consider embedded systems that might not support C99.

Thanks,
Spencer

miconda:
Thanks! At some point we'll ave to review and see how much of the legacy stuff we should to keep ...

PR #691, opened 2016-07-01
db_unixodbc: detect DB disconnection with generic HY000 status

giovannimele:
With Oracle DB, if the DB is restarted, db_unixodbc doesn't detect the problem as the error returned by
db_unixodbc_extract_error() is:

Jul  1 06:07:15 newvm1 /usr/sbin/kamailio[2770]: ERROR: db_unixodbc [connection.c:220]: db_unixodbc_extract_error(): unixodbc:SQLExecDirect=HY000:1:3114:[Oracle][ODBC][Ora]ORA-03114: not connected to ORACLE#012

Before, only status 08003 and 08S01 where checked to reconnect to the DB. Now, SQLSTATE HY000 (General error) is checked to trigger a reconnection.


miconda:
Thanks!

PR #692, opened 2016-07-01
Refactor cdp log

fedefava86:
This is a complete refactor of cdp log. In particular:

- It removes escape chars for colors in log: they are most of the times not interpreted so they are messing up the logs

- It refactors some log level, moving to ERROR from WARN some errors during peer connections

- In DEBUG level, cdp is too much verbose and also syslogs complains about rate-limiting. Instead of removing some logs (which is quite unfair), let's introduce a new modparam (debug_heavy) in order to activate some logs which gives some side informations about peers status. This modparam is not active by default but this can be enabled easily re-introducing previous debug logs

### DIFF
--- a/modules/cdp/doc/cdp_admin.xml
+++ b/modules/cdp/doc/cdp_admin.xml
@@ -144,6 +144,25 @@ modparam("cdp", "workerq_length_threshold_percentage", 25)
 	</programlisting>
       </example>
     </section>
+ <section>
+      <title>debug_heavy (int)</title>
+
+      <para>Flag to enable a more verbose log, which tells you
+		more details about peers. This is disabled by default
+		because it brings to a log too much verbose in DBG mode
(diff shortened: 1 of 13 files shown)


miconda:
The patch cc68bea adds a new parameter for the module -- you have to add documentation for it in the files from doc/ subfolder of the module.

fedefava86:
@miconda You are right. Completely agree with you. Done.

miconda:
Thanks, I am going to merge it.

On the other hand, if you want to clean a bit more the log messages, you can remove the name of the function from the log message, because now it is added automatically.

For example:

```
LM_INFO("peer_connect(): Trying to connect t...
```

should become:

```
LM_INFO("Trying to connect t...
```

fedefava86:
Ok @miconda , thanks for the hint. I will do it as soon as possible

Issue #655, opened 2016-06-05
TM Bus error on SPARC


sjthomason:
Calls to cfg_get(tm, tm_cfg, default_reason) and cfg_get(tm, tm_cfg, tm_auto_inv_100_r) on Solaris/SPARC cause a bus error.  Looking though some mailing list archives, it appears a similar issue was fixed in the registrar module but I couldn't find patch referenced.

miconda:
This usually happens when the field in the structure is not 32bit aligned. However, upon quick look, it seems to be in this case. It is sparc 64, right?

sjthomason:
Hi Daniel,

Yes this is sparc64. 
I did some digging and added some logging to try to find out whats going on.  The is no bus error from the main process during init.  I suspect commit 1b366aa1c6c37e2320da10fb68f0370199333f5d might have something to do with it as I reverted part of that in cfg_destroy_groups to fix another bus error during shutdown.  It looks like the pointer assignment *(char **) causes issues for some reason.

miconda:
Not sure if that is the reason, old patch, but you can try and report the result.

Maybe the pointer needs to be 8byte aligned in this case, I will try to search more on this topic.

sjthomason:
Hi Daniel,
It appears that the pointer does need to be 8 byte aligned.  Do you have any suggestions for debugging this further?  Any idea why the bus error only on the child processes?  I'm happy to provide access to hardware and any relevant backtraces.

Thanks for your help!

miconda:
Can you try with master branch? I pushed a patch to relocate the pointers at the beginning of the structure, so they should be aligned now.

sjthomason:
Hi Daniel,
Unfortunately I'm still getting the bus error in the child processes.  I added the following debug
```
diff --git a/modules/tm/tm.c b/modules/tm/tm.c
index 29a7326..cb972df 100644
--- a/modules/tm/tm.c
+++ b/modules/tm/tm.c
@@ -824,6 +824,7 @@ static int mod_init(void)
        tm_ctx_init();
 #endif
        tm_init = 1;
+       LOG(L_ERR, "tm_auto_inv_100_r: %s\n", cfg_get(tm, tm_cfg, tm_auto_inv_100_r));
        return 0;
 }
 
@@ -844,6 +845,7 @@ static int child_init(int rank)
                                " generator\n");
                return -2;
        }
[...]

PR #693, opened 2016-07-01
core: ensure cfg_block structure is 8 byte aligned

sjthomason:
- prevents bus error on stricter cpu architectures like sparc
- reported by GH #655


miconda:
Great! Thanks for pursuing and nailing it down!

PR #694, opened 2016-07-07
core: add generic byte swap macros

sjthomason:
- add bswap16() and bswap32() for 16 and 32 bit endianness swap


sjthomason:
This PR relates to issue #695 and might provide an easier fix.  I tested on LE OS X with clang and BE Solaris/SPARC with Solaris Studio and GCC.

PR #699, opened 2016-07-07
Packaging: Add new http package for Debian

SipSeb:
The new module contains the http_client and http_async_client module.

A few difficulties with this change:

1. Modules are right now grouped by lib dependencies. This caused the http_client module to end up in the kamailio-utils-modules package. No one would ever search for it there.
2. I tried to package a kamailio-http_async-modules package, but this isn't a valid Debian package name. But when renaming it into kamailio-http-modules, it felt wrong to include just the async http client and leave the http_client module stuck in the utils package, where no one would look for it. So I moved the http_client package, too.
3. The package depends on libevent, which is actually just a dependency of http_async_client. http_client would compile without. I don't know whether this breaks any policies.

### DIFF
--- a/Makefile.groups
+++ b/Makefile.groups
@@ -81,10 +81,10 @@ mod_list_carrierroute=carrierroute
 mod_list_berkeley=db_berkeley
 
 # - modules depending on curl library
-mod_list_utils=utils http_client
+mod_list_utils=utils
 
 # - modules depending on curl and libevent2 library
-mod_list_http_async=http_async_client
+mod_list_http_client=http_async_client http_client
 
 # - modules depending on purple library
 mod_list_purple=purple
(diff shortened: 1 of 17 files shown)


SipSeb:
Anything we can do for this pull request? @linuxmaniac @miconda

miconda:
IIRC @linuxmaniac added some note in the description of utils package to highlight it has the http client. I am fine with whatever, @linuxmaniac can choose what's easier from maintenance and debian project point of view.

linuxmaniac:
the description of kamailio-utils-modules package was updated at 90c90480c0de001c6c3010bae67a617b24b1ee4a and http_async was added to kamailio-extra-modules at 72daa184111009a948634c660b5c53eff0826562

grumvalski:
Thanks a lot Victor!

Cheers,

Federico

On 14 Sep 2016 6:41 pm, "Victor Seva" <notifications@github.com> wrote:

the description of kamailio-utils-modules package was updated at 90c9048
<https://github.com/kamailio/kamailio/commit/90c90480c0de001c6c3010bae67a617b24b1ee4a>
and http_async was added to kamailio-extra-modules at 72daa18
<https://github.com/kamailio/kamailio/commit/72daa184111009a948634c660b5c53eff0826562>

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/kamailio/kamailio/pull/699#issuecomment-247075839>, or mute
the thread

JonHVU:
Hi, 

Was this added to 4.4 as I cant see http_async when I install kamailio-extra-modules from;

http://deb.kamailio.org/kamailio44 jessie

Thanks

Jon

PR #701, opened 2016-07-08
Add kam_rand() macro for portability

sjthomason:
Many platforms with 4 byte integers define RAND_MAX as INT_MAX but some (Solaris and derivatives for example), RAND_MAX is only 32767 but INT_MAX is 2147483647.  This change allows falling back to the posix random() and srandom() variants with casts to int if both int and long are 4 bytes, then defining a consistent KAM_RAND_MAX of 2147483647.  If both INT_MAX and RAND_MAX are 2 bytes, then we are stuck with ANSI rand().

### DIFF
--- a/main.c
+++ b/main.c
@@ -118,6 +118,7 @@
 #include "dst_blacklist.h"
 #endif
 #include "rand/fastrand.h" /* seed */
+#include "rand/kam_rand.h"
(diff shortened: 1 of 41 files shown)


miconda:
Thanks!

PR #702, opened 2016-07-08
tm: check branch uri against current uri when deciding if append a ne…

grumvalski:
…w branch

This patch allows t_append_branches to cope with sip.instance usage, at least for the case when the contact uri has changed with regards to a previous registration.

### DIFF
--- a/modules/tm/t_append_branches.c
+++ b/modules/tm/t_append_branches.c
@@ -127,7 +127,9 @@ int t_append_branches(void) {
 		found = 0;
 		for (i=0; i<outgoings; i++) {
 			if (t->uac[i].ruid.len == ruid.len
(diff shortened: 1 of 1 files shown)


kamailio-sync:
You're right. Maybe adding also the check on the received address? I didn't
want to have a too heavy logic there, but it could make sense.

miconda:
I was thinking the same, being rather corner case.

Anyhow, you can merge this one and thing about the extra checks in a follow up commit.

Because you may want to check also the path (e.g., when other proxies were used in between, source ip is the same even if the router changed the address).

grumvalski:
Ok, I'll merge it. Should I backport it too?

miconda:
If it's a fix, then backport.

grumvalski:
Ok, thanks.

PR #703, opened 2016-07-09
core: fix regex error for modules parameters

snen:
Ambiguous was possible in params for multiple modules in one config line.
For example
```
loadmodule "presence_dialoginfo.so"
modparam("presence|dialplan", "db_url", "mysql://DBUSER:DBPASSWORD@localhost/DBNAME")
```
produced error 
`parameter <db_url> of type <1> not found in module <presence_dialoginfo>`
because of ["presence_dialoginfo" =~ /^presence|dialplan$/] => True

### DIFF
--- a/modparam.c
+++ b/modparam.c
@@ -59,15 +59,17 @@ int set_mod_param_regex(char* regex, char* name, modparam_t type, void* val)
 	}
 
 	len = strlen(regex);
-	reg = pkg_malloc(len + 2 + 1);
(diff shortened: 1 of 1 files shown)


miconda:
The size in pkg_malloc() needs to be increased, too -- from:

```
	reg = pkg_malloc(len + 2 + 1);
```

to:

```
	reg = pkg_malloc(len + 4 + 1);
```

snen:
right you are!
changed

PR #718, opened 2016-07-20
examples/[pis]cscf: adds example database structures for the P/I/S-CSCFs

vingarzan:
This helps first timers and casual users a lot, as they won't have to dig
through the history of Kamailio and various modules to find these
structures.

### DIFF
--- a/examples/icscf/icscf.mysql.sql
+++ b/examples/icscf/icscf.mysql.sql
@@ -1,0 +1,137 @@
+-- MySQL dump 10.9
+--
+-- Host: localhost    Database: icscf
+-- ------------------------------------------------------
+-- Server version	4.1.20-log
+
+/*!40101 SET @OLD_CHARACTER_SET_CLIENT=@@CHARACTER_SET_CLIENT */;
+/*!40101 SET @OLD_CHARACTER_SET_RESULTS=@@CHARACTER_SET_RESULTS */;
+/*!40101 SET @OLD_COLLATION_CONNECTION=@@COLLATION_CONNECTION */;
+/*!40101 SET NAMES utf8 */;
+/*!40014 SET @OLD_UNIQUE_CHECKS=@@UNIQUE_CHECKS, UNIQUE_CHECKS=0 */;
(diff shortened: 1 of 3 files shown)


ngvoice:
Hi Dragos,

not sure about this one:
The "proper" way would be, to add them to the regular "kamdbctl" scripts, where they are currently missing. It's likely a policy question: Should we add SQL build-scripts to the examples? Then that would apply to other examples as well.

I suggest @miconda should Merge or comment on this.

Thanks,
Carsten

miconda:
Perhaps we have to decide how to deal with the examples that need extra db tables.

I am fine having an sql file dedicated for them. Perhaps such examples can be in their own subfoder, have the cfg, sql and a short readme. Many configs related to the same topic can be in same folder, so I think the ims stuff is more or less like this.

I will not go to add them to the kamdbctl, because some of example using sqlops may be dependent on database server with some of its custom queries (e.g., if using REPLACE in cfg, then mysql is required).

If anyone has other options, let's discuss.

Otherwise I am fine with merging this one -- as said, ims examples have their own subfolder.

vingarzan:
I am fine with either or even none. 

It was though quite nasty to find a few days ago the create statements for something standard and simple like the pua module.The README said "The SQL syntax to create it can be found in presence_xml-create.sql script in the database directories in the kamailio/scripts folder." Yeah, right :). I ended up digging through very old history to find where that file went missing. Then I manually patched and extended it to match what the code was using. And now I still don't have the indices... 

In Wharf/OpenEPC we have yet another policy, which seems to work though - every module that uses a DB must have a directory "sql" inside the module that contains the said structures. I know that this combines code with SQL, bash scripts and other things, but IMO it is better to have all related things localized than spread around and lost eventually.

So while the examples dir is nice, the issues with it is that it gets too disconnected from the actual code. If things are closer-by, then devs have no excuse that they forgot to patch-in a new column in the DB, or a new parameter in the example config. Anyway, just my 2 cents, I don't want to make big waves for this little thing and waste your time.

ngvoice:
I just added a note about the database structure (as well as the I-CSCF schema), i guess we can close this topic.

miconda:
@ngvoice - ok thanks!

PR #720, opened 2016-07-20
ims_qos: create a proper flow description for the SIP signaling

vingarzan:
- before there was just an empty one in the AAR
- re-using the function typically used for the RTP media
- fixing then that function to not just hard-code protocol 17 (UDP) for everything
- now supporting also 6 (TCP) and IP (IP has no protocol number, we use just the
magic word "ip" (see RFC 3588 IPFilterRule for more details)
- adding the IP of the P-CSCF as a parameter to the module, as this is required
in the flow (UE IP <-> P-CSCF IP)
- also some inline/static fixes to function for avoiding warnings on gcc >=5


vingarzan:
Will complete with or more fix to the ims_registrar_pcscf module and then request again (sorry for the spam)

PR #725, opened 2016-07-26
registrar: Fix check of contact length and make it configurable

AndreasHuber-CH:
I split the pull request in three commits to seperate the fix from the new feature, but they go together so I propose them in the same pull request:

cc0b07d: I added some log message that would have helped me to understand a 400 Bad Request response.
c0697d7: I fixed a bug that were the contact size is not checked on the first header field and 200 OK is sent instead of 400 Bad Request for too long contact URIs but the uri ends up beein truncated by the database.
eb6b9c6: I made the contact size configurable to allow longer contacts


miconda:
Thanks!

PR #726, opened 2016-07-27
Bugfix presence race 2

phil-lavin:
presentity: don't update terminated presentity entries in the database

- This fix is pull requested based on the stuff in pull request #724. That is because the stuff done here builds on that earlier work. #724 probably wants to be merged before this one.

- Fixes a race condition caused by, for example, the call being answered at almost exactly the same time as the caller cancels. This causes a terminated state to change back to completed. The dialog is then removed from the database and the presentity entry stays in place until it expires.
- This fix explicitly prevents terminated entries being updated as the state machine in RFC 4235 prohibits this behaviour.

### DIFF
--- a/modules/presence/presentity.c
+++ b/modules/presence/presentity.c
@@ -303,6 +303,44 @@ int check_if_dialog(str body, int *is_dialog, char **dialog_id)
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


miconda:
Thanks!

PR #728, opened 2016-07-28
userblacklist: little changes in debug output

dunst0:
- Fixed confusing debug output in module for global blacklists and whitelists


miconda:
Thanks!

PR #730, opened 2016-08-01
Improvements on DB handling on S-CSCF usrloc

fedefava86:
Some improvements on the handling of contacts on S-CSCF database, namely
* Clean the contact table at startup if impu_contact table is empty
* if the contact is in state DELAYED_DELETE but still has ref_count > 0 in memory, it could mean that it has not been decremented before due to a failure to get a lock. This way we are leaking a contact on the DB and in memory. When the contact is expired and in DELAYED_DELETE, we do a query to check if it is really still referenced on impu_contact table. If not, we mark contact for deletion. 
* Add an helper function to print the actual state of a contact instead of the index of the enum

### DIFF
--- a/modules/ims_usrloc_scscf/udomain.c
+++ b/modules/ims_usrloc_scscf/udomain.c
@@ -267,6 +267,7 @@ void mem_timer_udomain(udomain_t* _d, int istart, int istep) {
     time_t now;
     int abort = 0;
     int slot;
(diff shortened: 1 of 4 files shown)


ngvoice:
Looks okay to me, if no-one objects, I will merge this pull request tomorrow.
It's my wife's birthday today, so I'll have to go now.

PR #733, opened 2016-08-02
Let multiple processes handle an rtimer

gaaf:
Extend the mode parameter to accept the number of processes that should handle the rtimer.


miconda:
Thanks!

PR #737, opened 2016-08-04
uri_db: added function to check any sip uri

dunst0:
- until now only from and to uri can be checked
- so added a function check_uri to check any uri


miconda:
Thanks!

PR #741, opened 2016-08-09
dialog: set DLG_FLAG_CHANGED for changed dialog variables

snen:
i've noticed, some dialog variables  are not saved to DB

improper flag setting has been the cause of the not DB saving in some scenarios


miconda:
Thanks!

PR #742, opened 2016-08-09
Use portable endian macros

### LINKED ISSUE #739: MD5 Broken on big endian architectures
MD5 hash functions in tree only work correctly on little endian machines.

sjthomason:
Not all platforms define WORDS_BIGENDIAN on big endian archs. This fixes #739


miconda:
Thanks!

PR #746, opened 2016-08-15
modules/siptrace: fix parsing of unsecure ws uris

denzs:
due to missing adaption of cutlen, the first decimal number of the ip address is cut off when using 'ws:' uris


adubovikov:
thank you

PR #749, opened 2016-08-17
core: fix some logging in tcp_read.c

dragos-oancea:
This patch to emphasize that ip_addr2a() uses a static buffer and because of this lots of log lines in tcp_read.c print the same IP as src and dst . It can be confusing when debugging TCP connections.
Practically LM_DBG or LOG print the same buffer twice in these cases.


miconda:
Thanks! Maybe we should consider introducing some ring buffers in such similar cases.

PR #753, opened 2016-08-19
core: fixed two double free() situations

dragos-oancea:
I ran into these while testing with SIPP  and having memlog=1 :

kamailio[21826]: : <core> [mem/q_malloc.c:453]: qm_free(): BUG: qm_free: freeing already freed pointer (0x7f24869850f8), called
 from <core>: parser/contact/parse_contact.c: free_contact(109), first free <core>: parser/contact/parse_contact.c: free_contact(109) - aborting

kamailio[21540]: : <core> [mem/q_malloc.c:453]: qm_free(): BUG: qm_free: freeing already freed pointer (0x7f4161b971c0), called
 from <core>: data_lump_rpl.c: free_lump_rpl(107), first free <core>: data_lump_rpl.c: free_lump_rpl(107) - aborting

These two patches fix these two separate issues.

### DIFF
--- a/data_lump_rpl.c
+++ b/data_lump_rpl.c
@@ -102,9 +102,12 @@ void free_lump_rpl(struct lump_rpl* lump)
 {
 	if (lump) {
 		if (!((lump->flags)&LUMP_RPL_NOFREE) && ((lump->flags)&LUMP_RPL_NODUP)
-		&& lump->text.s)
+		&& lump->text.s) {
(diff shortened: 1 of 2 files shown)


miconda:
In the patch for cleaning the hdr (	2188baf), the 'if(foo)' is not needed, because the while condition tests the same value already.

dragos-oancea:
I changed the commit (I removed the extra check).

PR #754, opened 2016-08-19
presence: fix multiple double free on shutdown caught with memlog=1

dragos-oancea:
These small changes fix the bug shown below and another potential one. 

kamailio[3062]: : <core> [mem/q_malloc.c:453]: qm_free(): BUG: qm_free: freeing already freed pointer (0x7facf958dc08), called from presence: hash.c: delete_shtable(316), first free presence: hash.c: free_subs_list(349) - aborting

It's the double free apocalypse day :)


miconda:
Thanks!

PR #756, opened 2016-08-23
examples/scscf: minor fixes (logging)

dragos-oancea:
The real return code of MAA/SAA is not shown in the logs in case of unknown error.


dragos-oancea:
Thank you Carsten. I also wish to report that this manual page is missing from all docs versions apparently :   http://www.kamailio.org/docs/modules/4.3.x/modules/ims_usrloc_scscf.html

PR #759, opened 2016-08-27
pike: fixed rpc array

kelchy:
- added array structure when returning multiple addresses instead of concatenated strings
- fixed buffer overflow on incorrect buffer size allocation


miconda:
Thanks!

Issue #580, opened 2016-04-21
sca module


oej:
The README needs to be up to date with the source code. I noted that the "Outbound proxy" modparam is not documented.

oej:
There's also eleven RPC commands, and only five in the README.

oej:
Also note that ALL error and debug messages needs to be terminated with a new line.

TheGrandWazoo:
Thank you Olle.

TheGrandWazoo:
In reference to the "Outbound Proxy" modparam, it is NOT used anywhere in the SCA module code base. I did a search for the outbound_proxy variable in the code base and it was only found in sca.c and sca.h. I commented out each reference in the sca.c and sca.h and compiled it without error. I believe it to be fictitious parameter at this time.

miconda:
It seems, most of the items were addressed. RPC commands are indexed in a separate file (see wiki) -- we need to unify that, but is global across all modules, not specific here.

If anything forgotten, then a new report would be better.

PR #762, opened 2016-08-29
sca: terminate all error and debug messages with a new line 

TheGrandWazoo:
Reported by oej GH #580. Includes whitespace cleanup and normalizing.


miconda:
I see the file utils/kamctl/mysql/ims_dialog-create.sql being also changed. Was is automatically generated?

TheGrandWazoo:
Must of been. I did not even touch that.

TheGrandWazoo:
I did a simple pull request to make sure I was conforming to the the standards. Thank you for showing me that doc.

miconda:
Committing changes that you are not aware of it's not a good thing, because it can break exiting code. Somehow you changed ims_dialog-create.sql and if we don't know why you did it, then it cannot be merged. You can use "git status" before any commit to see what files are affected by the commit.

Regarding the removal of outbound_proxy, I see that it was set in a structure. Have you checked that the structure is not passed to the dependency modules like tm, sl where the field might be used?

miconda:
Thanks! I pushed the pull request manually on your behalf to get the end result of it (commit), avoiding changing and reverting other files.

You can push the other patches you have for sca module as new pull requests.

PR #765, opened 2016-09-01
dont use url encoding for transport tag in path header

denzs:
the transport tag was the only string which used urlencoding for ';' and '=' 
see line 90 for example

this has lead (on the wire) to headers like:
Path: <sip:inbound_webrtc@10.3.66.231;lr;received=sip:127.0.0.1:41988%3Btransport%3Dws;ob>

to me this does not look like intention?!

### DIFF
--- a/modules/path/path.c
+++ b/modules/path/path.c
@@ -50,11 +50,11 @@ typedef enum {
 #define PATH_PREFIX_LEN		(sizeof(PATH_PREFIX)-1)
 
 const static char *proto_strings[] = {
-	[PROTO_TCP] = "%3Btransport%3Dtcp",
-	[PROTO_TLS] = "%3Btransport%3Dtls",
-	[PROTO_SCTP] = "%3Btransport%3Dsctp",
-	[PROTO_WS] = "%3Btransport%3Dws",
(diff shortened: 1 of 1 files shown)


miconda:
If the = and ; are not encoded, then the transport will be a parameter of the Path URI, not long seen as part of the value of the received parameter.

Were you having troubles with this? Because the module is rather old and used by many, I would expect this would have popped up as an issue if something was broken.

denzs:
thank you daniel!
i am trying to build an edge proxy for a websocket client without using nathelper
and till now i dont get the whole picture of how the edge proxy determines the correct outgoing connection to the client for invite and notifies from the backend...

denzs:
but that doesnt belong here ;)
so the pull request is obviously wrong!
thank you for your fast clarification!

miconda:
One thing to keep in mind for the future: prefix the commit message with the module name, as suggested in contribution guidelines: 

If the pull request is good, then it can be merged directly, otherwise the commit message needs to be adjusted.

You can come on sr-users mailing list with the description of what you want to do or what you don't understand properly and we will try to help.

PR #766, opened 2016-09-02
pv: add s.escape.csv transformation

camilleoudot:
escapes a string to use as a CSV field, as specified in RFC4180:
- enclose sting in double quotes
- escape double quotes with a second double quote


miconda:
Thanks! If not done yet, you have to add docs in the wiki for transformations cookbook.

PR #774, opened 2016-09-08
sca: add onhold_bflag module parameter

### LINKED ISSUE #773: sca: skip on-hold detection by parsing sdp and use bflag
Sipwise hat ON:

We have a scenario where we need to force the on-hold detection, so sca has to skip the sdp parsing and behave as previous based on a bflag

linuxmaniac:
* disable by default (-1)
* will skip parsing the sdp for on hold discovery and it will
  use the value of the bflag

Fix #773


miconda:
Thanks, it can be merged.

PR #778, opened 2016-09-13
sca: sca_handle_subscribe() get actual ruri

### LINKED ISSUE #777: sca: sca_handle_subscribe() honor the "$ru" value
Sipwise hat ON:

`sca_handle_subscribe()` does parse the message directly so if we change the value of `$ru` it doesn't use its value.

linuxmaniac:
* try to get r-uri from $ru and fallback to msg value as previously

Fix #777

### DIFF
--- a/modules/sca/sca_subscribe.c
+++ b/modules/sca/sca_subscribe.c
@@ -952,6 +952,7 @@ int sca_subscription_from_request(sca_mod *scam, sip_msg_t *msg, int event_type,
 	unsigned int expires = 0,
 	max_expires;
 	unsigned int cseq;
+	str *ruri = NULL;
 
 	assert(req_sub != NULL);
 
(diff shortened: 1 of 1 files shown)


miconda:
It can be merged, but the retrieval of current r-uri has to be changed. I don't think it is wise to use config variables names in the C code. I refer to the snippet:

```
 +	str src_uri = str_init("$ru");
...
 +	if (pv_eval_str(msg, &ruri, &src_uri) < 0) {
 +		LM_ERR("Failed to get $ru value fallback to parse msg\n");
 +		ruri.s = REQ_LINE(msg).uri.s;
 +		ruri.len = REQ_LINE(msg).uri.len;
 +	}
```

It should be as simple as:

```
if(parse_sip_msg_uri(msg)<0) { /* error handling */ }

if (msg->new_uri.s!=NULL) {
  ruri = msg->new_uri;
} else {
   ruri = REQ_LINE(msg).uri;
}

```

The name of r-uri variable (**$ru**) is set by pv module, I don't think sca (or other module) have to rely on that being forever. Also, it requires parsing the name and evaluation. In cases when there is a lot of duplicated code, better export the C function from pv module and make that a dependency, otherwise this approach can hide unpredictible behaviour in long term, without getting any errors at compilation or linking.

linuxmaniac:
What about:

```
if (parse_sip_msg_uri(msg) < 0) {
	LM_ERR("Error while parsing the Request-URI\n");
	goto error;
}
ruri = GET_URI(msg);
```

miconda:
Yep, all ok. I was looking for that macro, but didn't remember the name and no time for deeper search ...

PR #779, opened 2016-09-13
increase password length in databases

lgg:
For security reasons we can use only 32min character passwords on my work.
And when i tried to create user i got errors from DB.

So, i thought this will be useful for other kamailio users.  

Sorry for wrong commit format, but i can't understand to which part of kamailio this changes are related.


miconda:
The changes to the database schema must be done in the xml files from the lib/srdb1/schema/. The sql or other db definition files are auto-generated from xml files using **make dbschema**.

Also, it may be good that there will be a discussion on sr-users mailing list to see what would be a good size to increase to, not to jump to 64 and then see other people would like 128 or other max length.

miconda:
Size was increased to 64 using xml entities in db schema definition files.

PR #786, opened 2016-09-19
rabbitmq: Add new module

smititelu:
Add new rabbitmq module.


miconda:
It can be merged.

Then add the copyright header to the files that don't have one. Some (I guess imported from external projects) have MIT and that is compatible with gpl, so it's ok. For those that don't have, you can choose your preference between GPL, BSD (both accepted so far in the project) and MIT (given the rest of the module).

ngvoice:
Just a quick note: This Rabbit-MQ module has a different approach, than the Kazoo Modules. The intention is to make it generally available for a Rabbit-MQ infrastructure, rather than Kazoo specific.
I wrote the module together with Stefan.

smititelu:
Added MIT License for dev files in commit 3c9f88b96849e54af1559a85d61d04e2ac00cd4c.

smititelu:
@gaaf there are actually 2 conditions that consider breaking out of reconnect loop on success:

1'st one, on success, will succeed opening channel, thus will never go further:
```
// open channel - failed
if (log_ret != AMQP_RESPONSE_NORMAL) {
...
}
```

2'nd one:
```
// reconnect
if (reconnect_attempts < max_reconnect_attempts) {
...
}
```

I tried again this scenario using kamailio with 1 spawned child and I can see only 1 "LM_ERR("SUCCESS: rabbitmq_reconnect()\n");" for multiple calls, after reconnect.

The reason you may see multiple logs of that kind is because the amqp connection is **per child process**. However, the reconnect is tried only for a specific child, when it tries to _publish()/publish_consume(). 

Hope this was helpful.

PR #793, opened 2016-09-22
ipops: fix compilation under FreeBSD

dunst0:
- unable to compile detailed_ip_type.c under FreeBSD needs extra header


miconda:
Thanks!

PR #798, opened 2016-09-26
http_client Terminate exported commands list

osthyvel:
The list of exported functions was not terminated, which caused a
segfault in find_mod_export_record() when reading outside the list.


miconda:
Thanks!

oej:
Thanks for finding that out. My test module did not crash - can you elaborate how you read outside of the list? Wan't to learn so we can prevent it from happening again.

osthyvel:
I found the issue when moving from CentOS 6 -> 7. I don't know why we didn't see the issue on CentOS6 but it is probably due to placement of the global data in memory. In the CentOS6 build of the http_client (latest from master branch) we were lucky/unlucky to get a 0 in the memory just after the exported functions array when iterating in function find_mod_export_record.

PR #803, opened 2016-09-29
ims_auth: Put fresh keys from HSS on top of list

fedefava86:
If S-CSCF downloaded fresh keys from the HSS, put it
on the top of the list, as they are already been sent
to the UE via 401 message so it should be the first
to be retrived to comparison once REGISTER with challenge
comes from the UE


miconda:
@jaybeepee, @richardgood, @vingarzan or @ngvoice may have comments, if not, I am fine to merge.

PR #804, opened 2016-09-29
ims_isc: Add tel: associated URI on 3rd party REG

fedefava86:
Add also tel: as header in 3rd party REG towards TAS


miconda:
@jaybeepee, @richardgood, @vingarzan or @ngvoice may have comments, if not, I am fine to merge.

PR #812, opened 2016-10-05
sca: add "server_address" parameter

linuxmaniac:
* server address ``(sip:<host>[:port])`` which will become the value of
  Contact header filed for NOTIFY messages.


miconda:
Thanks, it can be merged.

Issue #782, opened 2016-09-14
sca: add server_id column to sca_subscriptions table


linuxmaniac:
Sipwise hat ON:

In order to be able to share the same DB and take care of the expired records only related to that server.
We need a module parameter to allow filter by server_id when loading values from DB

Related to #781

PR #814, opened 2016-10-05
sca: add server_id column to sca_subscriptions table

linuxmaniac:
In order to have more than one server with the same DB ``sca`` module has to have a ``server_id`` column to partition subscriptions

version bumped to ``2``

See #782


miconda:
Thanks, can be merged.

PR #815, opened 2016-10-07
core: added mcast to select interface for multicast

tverlaan:
You can now define a mcast parameter before a listen parameter that
contains a multicast address. After each listen parameter mcast gets
reset. mcast must contain the name of an interface, eg `eth1`.

Issue GH#813


miconda:
Thanks, you have to add some documentation about the new core parameter explaining its purpose and a small example in the core cookbook from the wiki:

  * https://www.kamailio.org/wiki/cookbooks/devel/core#mcast_loopback

tverlaan:
Done: https://www.kamailio.org/wiki/cookbooks/devel/core#mcast
I also added the feature here: https://www.kamailio.org/wiki/features/new-in-devel#parameters

Any feedback is welcome.

PR #816, opened 2016-10-07
TCP closed event route improvement

camilleoudot:
This patch for the **tcpops** module (+ some core modifications) improves the original **tcp:closed** event route handling by triggering different event routes, depending on the reason why the connection was closed:

- **tcp:closed** for "normal" closure by the other side (FIN,ACK)
- **tcp:timeout** for timeouts (unanswered keepalives)
- **tcp:reset** if reset by peer (RST)

It will break the compatibility with older scripts, as some events will require new event routes to be caught.

I'll modify the documentation accordingly if/when this is accepted.

### DIFF
--- a/modules/tcpops/doc/eventroutes.xml
+++ b/modules/tcpops/doc/eventroutes.xml
@@ -8,28 +8,73 @@
 
 ]>
 <section id="tcpops.ex.event_routes">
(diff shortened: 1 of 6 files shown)


miconda:
Thanks, changes are fine for me, it can be merged.

PR #817, opened 2016-10-07
presence: fix crush if dialog id is NULL

snen:
- check dialog_id is not NULL before call delete_presentity_if_dialog_id_exists

sometimes after  check_if_dialog(str body, int *is_dialog, char **dialog_id) it is possible
is_dialog==1
dialog_id==NULL

#0  __strcmp_sse42 () at ../sysdeps/x86_64/multiarch/strcmp.S:260
#1  0x00007feb2e446603 in delete_presentity_if_dialog_id_exists (presentity=0x7feb545c8ca8, dialog_id=0x0) at presentity.c:419


miconda:
Thanks!

PR #818, opened 2016-10-10
modules/ims_auth, modules/ims_registrar_scscf: fix condition for priv…

grumvalski:
…ate copy of the msg

I've been facing recently some inconsistent replies generated from the two modules. Those replies didn't belong to the current request, but to some old transaction. The issue is due to the check made to decide if the current private copy of the message is consistent with the current transaction: due to memory reallocation there are chances that the check gives a false negative, bringing an old message to be used as a base for the reply. Using transaction's index and label is safer IMO.

### DIFF
--- a/modules/ims_auth/pvt_message.c
+++ b/modules/ims_auth/pvt_message.c
@@ -82,7 +82,7 @@ struct sip_msg* get_request_from_tx(struct cell *t) {
         (which we cannot assume) then we would pollute the shm_msg t->uas.request if we did any parsing on it. Instead, we need to 
         make a private copy of the message and free it when we are done 
          */
-        if (_pv_treq.T != t || t->uas.request != _pv_treq.tmsgp) {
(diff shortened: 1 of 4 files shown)


miconda:
Wondering if the same should be done in tmx module for t_var.c -- if you found some issues with the current approach, then it might be the case there, as well, it looks like this code in ims is similar with the one from tmx.

grumvalski:
You're right, tmx too could be affected by the same issue. 
In my case the memory allocated for the transaction and the uas request was the same as for a recently closed transaction. And adding the check for the message id did not completely solve the issue, because of the its per-process nature I was still facing false negative in some cases.

miconda:
You can merge it. Probably we will have to do the same for the tmx.

grumvalski:
Thanks. I will try to have a look on tmx as soon as possible.

PR #823, opened 2016-10-14
Aliasdb

khoegh:
I've backported some features from opensips.
- reverse alias lookup function flag
- use_domain moved from global parameter to a function flag (defaults to yes)
- New function alias_db_find uses source and destination PV instead of ruri

### DIFF
--- a/modules/alias_db/README
+++ b/modules/alias_db/README
@@ -36,8 +36,9 @@ Daniel-Constantin Mierla
               3.3. domain_column (str)
               3.4. alias_user_column (str)
               3.5. alias_domain_column (str)
-              3.6. domain_prefix (str)
-              3.7. append_branches (int)
+              3.6. use_domain (int)
+              3.7. domain_prefix (str)
+              3.8. append_branches (int)
 
         4. Functions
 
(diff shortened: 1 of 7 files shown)


miconda:
I would like to discuss more about the removal of use_domain parameter. It is a common parameter name across many modules and it is a easy way to turn on/off multi-domain support (even done also for alias_db in default kamailio.cfg).

I would like a coherent solution, like either removing this mod param everywhere or still keep it and also allow the per-function option to enable/disable.

Moreover, changing the default in using the domain is going to break compatibility may result in unexpected behaviour without prior visible notice, as people can reuse an old config without use_domain parameter set, so no error is triggered, but behaviour is changed. So I think the default should be without domain (this is the default across kamailio modules) and a flag to enable domain based lookup.

Keeping the use_domain won't change much the patch, the flags need to be initialized based on it:

```
if(use_domain) {
  flags |= **domain_lookup_flag**
}
```

Travis reports that the inter-module API was broken -- app_lua doesn't compile as the function prototype for alias db lookup was changed. This is an easy fix, I can take care of.

khoegh:
Thanks, I will add use_domain again after the weekend.

khoegh:
I re-added use_domain module parameter and added option "u" to use domain on alias_db_lookup/alias_db_find.
Will you fix the API, thanks.

miconda:
I pushed the commit that combines this pull requests with some refactoring of the internals not to break the inter-module api as well as use straight flag meaning for domain lookup, instead of negative meaning.

It should not affect use from config, but testing would be good to do to be sure.

khoegh:
I did some testing and it seems to work as expected.
Thanks.

PR #824, opened 2016-10-17
tmx: fix condition for private copy of the message

grumvalski:
porting of the changes done for ims_auth and ims_registrar_scscf modules.


miconda:
Thanks, can be merged.

PR #825, opened 2016-10-19
modules/tmx: fix mi command t_uac_cancel

grumvalski:
fix for mi command t_uac_cancel: without calling prepare_to_cancel, the following call to cancel_uacs gives this error:

BUG: tm [t_cancel.c:327]: cancel_branch(): tm: cancel_branch: local_cancel buffer=(nil) != BUSY_BUFFER (trying to continue)

and it does not cancel the transaction.


miconda:
Thanks, it can be merged.

PR #827, opened 2016-10-24
sca: fix sca_call_info_update() [to|from] optional parameters

linuxmaniac:
* use str values
* need to build struct to_body and free memory afterwards


miconda:
Thanks, it can be merged.

PR #828, opened 2016-10-24
sca: use To header as target if SUBSCRIBE is in-dialog

linuxmaniac:


miconda:
Thanks, it can be merged.

PR #829, opened 2016-10-24
janssonrpc-c: fix memory leak when using janssonrpc_notification

eschmidbauer:


miconda:
I am fine with the patch to the code, but we do not add anymore remarks about the history of changes and contributors at the top of the c/h files. Git log does a better job at tracking changes and see who did it. Actually we are removing existing history/contributors remarks from all the files.

eschmidbauer:
@miconda removed contributor remark

PR #830, opened 2016-10-24
rr: default to port 5060 if port is omitted in route header

tverlaan:
When using double route headers the selection of the outgoing
socket is only done on IP address if the port is omitted in that
route header. This fix defaults the port to 5060 so the correct
listen socket is chosen. When no socket is found it will fallback
to the current behaviour keeping backwards compatibility.

If backwards compatibility is not required the extra 'if else' could be removed. I'm not sure if `loose_route()` is being used in this way so I opted for keeping backwards compatibility.

### DIFF
--- a/modules/rr/loose.c
+++ b/modules/rr/loose.c
@@ -827,8 +827,9 @@ static inline int after_loose(struct sip_msg* _m, int preloaded)
 			}
 
 			if (!use_ob) {
-				si = grep_sock_info( &puri.host, puri.port_no, puri.proto);
-				if (si) {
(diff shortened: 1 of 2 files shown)


miconda:
For proper clarification -- so if the port is not present, then the match is made on IP and it may select a socket that has that ip but at different port. Right?

That's indeed something to be fixed, however, i think that when protocol is tls (and maybe wss), then the 5061 port should be tried.

tverlaan:
Yes that is correct. You're right about the protocol. I'll add another check for that. I'm curious if `record_route()` adds port 5061 or omits that as well.

PR #831, opened 2016-10-25
dialog: adding dialog replication via DMQ

Soapnix:
- updating code base to actual state.
- adding repliation for DLG_STATE_EARLY to prevent hanging up unconfirmed dialogs and its infinite replication.
- fixing dialog removal on peers to prevent die on timeout and possible infinite replication.
- fixing dialog removal on peers when master node cleanly restart.

### DIFF
--- a/modules/dialog/dialog.c
+++ b/modules/dialog/dialog.c
@@ -79,6 +79,7 @@
 #include "dlg_var.h"
 #include "dlg_transfer.h"
 #include "dlg_cseq.h"
+#include "dlg_dmq.h"
 
 MODULE_VERSION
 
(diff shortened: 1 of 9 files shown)


miconda:
Thanks, useful patch!

One thing, two new global variables are defined, exported as module parameters, but they are not actually used:

```
int dlg_id_offset = 1;
int dlg_id_increment = 1;
```

By name, I guess they are intended to be used for dialog id generation, but no code or docs related to them.

Soapnix:
We are glad to help improving kamailio.
Our company is very interested in this feature (and some other - now troubleshooting dmq_usrloc).
Yeah, I'm so sorry - I missed this variables during last code clean - fuction which use them was removed.

miconda:
So, you say they are not needed for the new functionality to work properly, right?

Soapnix:
Of courese not. Do not get me wrong.
dlg_dmq works proprely, we double check they funcs and stability before I write for you.
I'm already have a commit, which removes those two unused variables.

gaaf:
I quickly scanned the current code, but fail to see how hash id collisions are avoided now that dlg_id_offset/increment are removed. Can you elaborate on that?

gaaf:
No explanation?

Soapnix:
We are working on it.

PR #832, opened 2016-10-25
pv: Use memcpy to copy len bytes instead of strcpy

zecke:
The sipcapture plugin stores bytes with VAR_VAL_STR and when
copying the data it would end up being truncated. Use memcpy
instead of strncpy and assume that the original string already
has the NUL termination (or not as with the hep plugin).

Config example:

	$var(payload) = hep(0x00f)

### DIFF
--- a/modules/pv/pv_svar.c
+++ b/modules/pv/pv_svar.c
@@ -70,7 +70,7 @@ script_var_t* add_var(str *name, int vtype)
 		return 0;
 	}
 	it->name.len = name->len;
-	strncpy(it->name.s, name->s, name->len);
+	memcpy(it->name.s, name->s, name->len);
 	it->name.s[it->name.len] = '\0';
 
 	if(vtype==VAR_TYPE_NULL) {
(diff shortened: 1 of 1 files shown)


zecke:
@adubovikov do you think this approach is good?

phil-lavin:
I'm thinking that you're going to lose the NUL byte here as presumably, if it exists, it is not included in the length of the string

zecke:
Ah okay s.len == strlen(s.s). So yes, it fixes my issue and breaks other things. Is introducing VAR_VAL_BYTES and pv_* an option for it?

phil-lavin:
I'm not entirely clear what your issue is - can you explain further. As I see it, changing from `strncpy` to `memcpy` only has benefit if the thing you're copying has a \0 in the middle of it.

zecke:
I am using the "sipcapture" module and the HEPv3 that is received by it can carry binary data inside its payload. I use hep(0x00f) to ask the sipcapture module to extract the payload of the non SIP HEP message but when trying to use it the binary data has turned into a string of 0x00 becuase the binary contains a 0x00 in the second byte...

zecke:
Anyway. len+1 is allocated so I can just not remove/re-add the NUL byte at the end.

miconda:
If binary data needs to be copied over, then memcpy should be used, indeed.

miconda:
And I see that the patch is actually for the pv module, it looks ok, I will merge it, thanks!

zecke:
Lovely! thank you.

Issue #781, opened 2016-09-14
sca: implement db_mode only


linuxmaniac:
Sipwise hat ON:

We have more than one server and we need to keep things in db in order to share sca subscription/appearance between servers

linuxmaniac:
Any thoughts @TheGrandWazoo ?

TheGrandWazoo:
I assume you mean more then one kamailio server in a load-balancing type environment?

If so, that was on my radar for future enhancements but I am glad someone is finally looking at this module and making some improvements. Thank you so much!!!! This module is one that I use at many customers and has been the bane of my existence with the hold seizes.

linuxmaniac:
@miconda I'm thinking if ``sca`` can be redesigned and integrate the subscription part of it as a module like ``presence_sca`` using ``presence`` infrastructure.

The thing is that ``presence`` has support for process only the body not for the headers and ``sca`` is adding ``Call-Info`` header to a bodyless ``NOTIFY`` reply.

Is it worth the trouble to implement that in ``presence`` or should I just keep hacking into ``sca`` directly?

PR #835, opened 2016-10-28
sca: sca_call_info.c: Remove redundant code

TheGrandWazoo:
- In reviewing the code block that I removed it is actually duplicated
in the sca_create_cononical_aor_for_ua() function which is called from
sca_create_canonical_aor() which is called inside this 'if' block, so
why do it twice. The function sca_get_msg_contact_uri() is also called
in sca_create_cononical_aor_for_ua() but it does not set or pass a
'contact' parameter so we still need it in sca_call_info_update() just
don't need to check for errors because it will error out in
sca_create_cononical_aor_for_ua() and do the same thing.
- Maybe in the long run for the 'contact_uri' parameter, create a second
function or add it as a setable paramter to the
sca_create_cononical_aor_for_ua().
- Code change has been tested and running for about 2 months without
issues.


miconda:
The PR cannot be merged because it has a conflict on modules/sca/sca_call_info.c as shown above. Can you fix it?

TheGrandWazoo:
Is there a new maintainer on the project? I has not been touched in so long that I thought I was t he only one looking at it.

miconda:
There were several commits by @linuxmaniac recently, iirc -- a git log over modules/sca should reveal more accurately who worked lately on it.

linuxmaniac:
Yes, I'm touching it and I already planned #781

TheGrandWazoo:
Ok, I will step away then and you can remove my pull request.

TheGrandWazoo:
Just an FYI that the sca.cfg needs to be updated to reflect the newer kamailio.cfg changes.
Also, this script block needs to be added to the sca.cfg at the end....

#!ifdef WITH_SCA
event_route[tm:local-request] {
    if (is_method("NOTIFY")) {
        record_route();
    }
}
#!endif

Seems to not work well without it. My experience with the module.

linuxmaniac:
> Ok, I will step away then and you can remove my pull request.

Please, keep contributing, we are more than happy to get more people involve. Take a look over the new code and feel free to close this if you think is not relevant anymore

> Just an FYI that the sca.cfg needs to be updated to reflect the newer kamailio.cfg changes.

@TheGrandWazoo Please provide a pull-request with the needed changes

TheGrandWazoo:
I did not want to step on anyone's shoes. You all have done great work with this product and I have been using it since 1.5 (I think that was it).
I do have some more ideas like using the textops module to insert the Replaces and PAI headers...just been looking and learning but not sure it's a fit yet. I just don't like those headers after the Content-Length even though it does not break any SIP rules from my research.

Thanks again.

PR #836, opened 2016-10-28
sipcapture: Terminate the log line with a \n

zecke:
Add a newline to avoid the log of other messages starting behind
the REQUEST, PRE output.


zecke:
@adubovikov please have a look.

PR #837, opened 2016-10-28
ss7: Introduce the beginning of a ss7 module

zecke:
This can be used with sipcapture and begings with M2UA and ISUP
support. More SIGTRAN (M3UA, SUA) support is planned and maybe
TCAP/MAP in the future.

The ISUP message format depends on the specific message so I have
exported a Smalltalk ITU ISUP model to C to avoid manual errors.

### DIFF
--- a/modules/ss7ops/Makefile
+++ b/modules/ss7ops/Makefile
@@ -1,0 +1,19 @@
+#
+# ss7ops module makefile
+#
+#
+# WARNING: do not run this directly, it should be run by the master Makefile
+
+include ../../Makefile.defs
+auto_gen=
+NAME=ss7ops.so
+LIBS=
+
+DEFS+=-DKAMAILIO_MOD_INTERFACE
+
+SERLIBPATH=../../lib
+SER_LIBS+=$(SERLIBPATH)/kmi/kmi
+SER_LIBS+=$(SERLIBPATH)/srdb1/srdb1
+SER_LIBS+=$(SERLIBPATH)/kcore/kcore
+SER_LIBS+=$(SERLIBPATH)/srutils/srutils
+include ../../Makefile.modules
--- a/modules/ss7ops/README
+++ b/modules/ss7ops/README
@@ -1,0 +1,73 @@
+SS7 Ops Module
+
+Holger Freyther
+
+   <help@moiji-mobile.com>
+
+Edited by
+
+Holger Freyther
+
+   <help@moiji-mobile.com>
+
+   Copyright © 2016 Strong Leaf Consultant Ltd.
+     __________________________________________________________________
+
+   Table of Contents
+
+   1. Admin Guide
+
+        1. Overview
+        2. External Libraries or Applications
+        3. Functions
+
+              3.1. isup (integer)
+
+        4. Parameters
+
+              4.1. isup_to_json (integer)
+
+Chapter 1. Admin Guide
+
+   Table of Contents
+
+   1. Overview
+   2. External Libraries or Applications
+   3. Functions
+
+        3.1. isup (integer)
+
+   4. Parameters
+
+        4.1. isup_to_json (integer)
+
+1. Overview
+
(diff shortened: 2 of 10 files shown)


adubovikov:
should we keep this module's name or change to another ?

zecke:
sipcapture is using the HEPv3 protocol and it can encapsulate non SIP payloads (e.g. M2UA+MTP+ISUP). This is a first implementation/try and only supports M2UA/ITU MTP/ITU ISUP. ISUP is converted from the wire message to a semi-parsed (some fields are extracted but not all) JSON object and then converted to a string. A parameter exists to get the string representation from C into the kamailio script. 


@adubovikov

adubovikov:
@miconda  any suggestion ?

miconda:
Trying to understand what's the module about in order to be able to make more comments on its name:

  * does it receive ss7 packet over hep3 or it can forward it to homer over hep3? In Kamailio we can also receive binary packets which are not sip (we do for stun module)
  * what is the target in long term for the module? In other words, what it tries to achieve from the point of view of sip and ss7 worlds? E.g., routing ss7 or just analyzing them?

zecke:
* It receives M2UA over hep3 (through the sipcapture module)
* Right now it is just meant to parse ISUP (and later TCAP/MAP) and store it in the home/sipcapture db.

adubovikov:
The module will be able to parse ISUP binary protocol to string serialization i.e. JSON or pseudo variables and later we can insert the parsed messages into DB and using it in Homer.

miconda:
Then maybe ss7ops is more appropriate as name, not to suggest a protocol support.

zecke:
Will rename and right now wait for more feedback.

miconda:
Some remarks:

  * there are several places where I spotted 'sipcapture' as module name
  * each .c and .h file must have at a top the license text and copyright line
  * the documentation has to be written in xml docbook format, stored inside the doc/ subfolder of the module directory (like the other modules)

zecke:
* Did the rename locally
* Will write docbook
* For license text/copyright. The ISUP table is generated and I don't think it can be copyrighted (the information is to match ITU ISUP, no creative process involved). What license text should be there?

miconda:
Ohh, I missed the fact those file were generated -- it is ok to leave them without a copyright header if the generator doesn't insert one.

zecke:
Okay, renamed directory and module name to ss7ops, instead of handwriting the README (I wondered why one would do that) added a minimal docbook (based on the sipcapture one).

oej:
Seems like "ss7capture" would be a better name if I understand this correctly. I would like to use the name to relate it to sipcapture.

adubovikov:
disagree, because we can use the module not only for "capturing scenarios", but hypothetically also for interworking and/or SIP-I/SIP-T.

oej:
Ok, I see. Then I'm ok with the name :-)

adubovikov:
don't forget also about IMS and MAP/TAP, CAMEL... this is also a huge area for ISUP

oej:
Then we need to clarify the relationship to the existing SIPT module.

adubovikov:
unfortunately existing SIP-I module has very limited functionality and couldn't be use here.

oej:
In my opinion that's fine as long as we know that these are not built to work together. We have multiple modules overlapping since before.

miconda:
Thanks for the new module, merged!

miconda:
@zecke - I got some warnings when compiling the module, so I pushed some fixes. Check the commit e0d26f780ed551bd993ca36ac125e87a16db7709 - it seemed wrong initialization of the same index, I guessed it should be the next, but it might not be the right.

miconda:
@zecke - can you also add basic examples of using the params and functions in the docbook xml. You can see in other modules, like:

  * https://github.com/kamailio/kamailio/blob/master/modules/async/doc/async_admin.xml#L155

zecke:
Thank you. 53 is indeed the right value for the UPA message.I will look into more/better documentation.

PR #838, opened 2016-10-28
dmq_usrloc: fix variable type mismatch

Soapnix:
- fix 'methods' variable type in receiving branch; by default in struct definition it is an 'unsigned int', but while receiving DMQ message value puts into 'int' and cause flags corruption in some cases

### DIFF
--- a/modules/dmq_usrloc/usrloc_sync.c
+++ b/modules/dmq_usrloc/usrloc_sync.c
@@ -299,10 +299,11 @@ int usrloc_dmq_handle_msg(struct sip_msg* msg, peer_reponse_t* resp, dmq_node_t*
 	srjson_t *it = NULL;
 	static ucontact_info_t ci;
 
-	int action, expires, cseq, flags, cflags, q, last_modified, methods, reg_id;
+	int action, expires, cseq, flags, cflags, q, last_modified, reg_id;
 	str aor, ruid, c, received, path, callid, user_agent, instance;
+	unsigned int methods;
 
-	action = expires = cseq = flags = cflags = q = last_modified = methods = reg_id = 0;
+	action = expires = cseq = flags = cflags = q = last_modified = reg_id = methods = 0;
 
 	parse_from_header(msg);
 	body = ((struct to_body*)msg->from->parsed)->uri;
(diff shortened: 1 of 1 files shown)


miconda:
I understand the need of changing local var methods to unsigned int, but why it needs to take the value from the double val field in the json? Is it printed in json as a floating point type?

```
-			methods = it->valueint;
 +			methods = it->valuedouble;
```

Soapnix:
@miconda 
Changing 'valueint' to 'valuedouble' need because 'valueint' has 'int' type  and 'methods' are sometimes not fit into them when we obtain json message. But it fits in 'valuedouble' which has type 'double' and also been set on parse.

Situation: when client supports all methods we got right sended value '4294967295' (max unsigned int), while receiving we a convert it to 'int' (valueint) and got in debug '2147483647' (got max int) and it's wrong. In more details after that 'methods' is been set to '-2147483647' and in fact we see in supported 'methods':
1111111111111111111111111111111110000000000000000000000000000000
instead of expected
1111111111111111111111111111111111111111111111111111111111111111
and client can't operate without re-reg and setting rigth 'methods' flags.

miconda:
So you say that both fields valueint and valuedouble are set in this case and you can read any of them? I expected that double is set only for numbers with a dot (floating point type).

On the other hand, the int and unsigned int have the same size. Doing a cast should work.

Soapnix:
srjson_t *srjson_CreateNumber(srjson_doc_t *doc, double num) {
	srjson_t *item = srjson_New_Item(doc);
	if (item) {
		item->type = srjson_Number;
		item->valuedouble = num; // <-- here is correct value
		item->valueint = (int) num; // <-- here is shrinked value
	} return item;
}

Correct value '4294967295' (max unsigned int), shrinked value is '2147483647' (max int).
So we need receive double and put it into unsigned int.
Possible the srjson can't correct work with unsigned int and as solution is translate unsigned int through the double.

miconda:
I think the valueint should be removed and add some helper macros to get the int/unsigned int/long/whatever number/... out of the double val. It should be more sane for the future.

miconda:
I pushed a set of commits in master removing the valueint, relying on valuedouble, plus some macros.

I will think a bit more about backporting or making a dedicated fix for 4.4 branch.

Testing and reporting back the results would be very appreciated.

Soapnix:
Ok, we check that.
Thanks for the support.

2016-10-31 23:01 GMT+03:00 Daniel-Constantin Mierla <
notifications@github.com>:

PR #845, opened 2016-11-02
Makefile.defs: clang as default and lib for 64 bit on FreeBSD

dunst0:


miconda:
Is this for all FreeBSD versions or only from a specific version on? Many server are running old versions of distros.

dunst0:
You are right. I forgot that `clang` is the default only since FreeBSD 10.
For the lib dir I am not sure how long this is done. I will do some research.

miconda:
I am closing this one, I expect there will be a new PR to cope with the concerns expressed above.

PR #847, opened 2016-11-04
nathelper: force socket if set

tverlaan:
This fixes force socket in that it will always be forced and not
only when there is no socket.


miconda:
Thanks!

I will merge, I think it is how force_socket was meant, so I will merge if nobody has complains.

On the other hand, having the option to say to use force_socket when not set, could be useful. I let that as an enhancement.

tverlaan:
I agree that it would be useful. Maybe as a default socket? Then again, the
module already has a fallback mechanism if it is not set based on the dest
uri.

I'm also planning on enhancing this modparam as in my testing (and looking
at strace) it only works when I enter an IP address without port.

PR #849, opened 2016-11-04
Updated the specs file for CentOS 7

winem:
I updated the specs file (see the changelog) to build Kamailio 5.x rpm packages for CentOS 7.2


miconda:
Thanks!

If you want to get more involved in the rpm packaging, I can grant you access to opensuse build service for kamailio project. Just let me know.

PR #850, opened 2016-11-07
P-Charging-Vector related updates

mslehto:
Original purpose was to add new $pcv(...) pseudo variables.
Ended up doing documentation stub, minor bug fix and general housekeeping.


miconda:
Thanks!

PR #853, opened 2016-11-08
Updated documentation of modules which are impacted by memory leak in…

os11k:
Updated documentation for modules who uses libcurl, regarding memory leak in Centos 6 libcurl library.


miconda:
Thanks. I think I will relocate the paragraph to its own section (named Remarks or Known Issues). This will keep the module description clean and allow adding text for other similar cases.

miconda:
Pushed remarks section to those modules.

PR #854, opened 2016-11-08
presence: add parameter `startup_mode` to disable 1dceaa24ded727aba5870f28fb63e26ed98464f3

eschmidbauer:


miconda:
I would propose to rename the new parameter, because the chosen on is misleading, given that it doesn't fix a clear issue (like a leak in a specific version of an external library), but it selects in doing or not some sort of operation at startup. I would go for **startup_mode**, if there is no other better suggestion. In the future it can be used for controlling other kinds of operations at startup for the module.

miconda:
Thanks

miconda:
I did a cleanup of extra unnecessary global variable used to implement the startup mode parameter. No testing, report if you find something working wrong.

PR #856, opened 2016-11-12
Introduce HAVE_IP_MREQN due to OpenBSD compile failing after mcast improvement

mslehto:
```
CC (gcc) [kamailio]             udp_server.o
udp_server.c: In function 'setup_mcast_rcvr':
udp_server.c:218: error: storage size of 'mreq' isn't known
udp_server.c:218: warning: unused variable 'mreq'
gmake: *** [Makefile.rules:97: udp_server.o] Error 1
```

Compile-only tested with OpenBSD 6.0, FreeBSD 10, FreeBSD 11 and Ubuntu Linux 14.04


miconda:
Thanks!

Issue #859, opened 2016-11-15
dispatcher: kamailio-devel segfaults on ping


tverlaan:
The pinging from the dispatcher module segfaults and then kamailio exits. It used to work on kamailio 4.4.3.

**dispatcher.list**
```
31 sip:10.0.20.191:5062;transport=udp 8 0 socket=udp:10.0.20.193:5021
```

**debug logging**
```
21(2188) DEBUG: dispatcher [dispatch.c:2872]: ds_ping_set(): probing set #31, URI sip:10.0.20.191:5062;transport=udp
21(2188) DEBUG: <core> [socket_info.c:559]: grep_sock_info(): checking if host==us: 11==11 && [10.0.20.193] == [10.0.20.193]
21(2188) DEBUG: <core> [socket_info.c:566]: grep_sock_info(): checking if port 5021 (advertise 0) matches port 5021
21(2188) DEBUG: tm [uac.c:411]: t_uac_prepare(): DEBUG:tm:t_uac: next_hop=<sip:10.0.20.191:5062;transport=udp>
21(2188) DEBUG: tm [uac.c:152]: dlg2hash(): DEBUG: dlg2hash: 37511
21(2188) DEBUG: tm [uac.c:616]: send_prepared_request_impl(): send_prepared_request_impl: uac: 0x7ff37f776380  branch: 0  to 10.0.20.191:5062
0(2167) ALERT: <core> [main.c:743]: handle_sigs(): child process 2188 exited by a signal 11
0(2167) INFO: <core> [main.c:759]: handle_sigs(): terminating due to SIGCHLD
```

Please let me know if you need more information.

linuxmaniac:
``kamailio -V`` and some related gdb output from the coredump would be nice

tverlaan:
```
$ kamailio -V
version: kamailio 5.0.0-dev7 (x86_64/linux)
flags: STATS: Off, USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MEM, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLACKLIST, HAVE_RESOLV_RES
[...]

PR #861, opened 2016-11-16
tm: only call onsend_route when there is an original incoming request

tverlaan:
- fix for dispatcher ping, see GH #859
- regression introduced by 2f080bd


miconda:
I pushed a different patch to do the safety checks inside the function, because it may be called from different places. Report if there are issues with this approach.

PR #862, opened 2016-11-16
Isup pare and convert more fields

zecke:


miconda:
@zecke - I sent you an invite to join the dev team so you can push commits to your ss7ops module directly. You can still use pull requests if you prefer this way and want others to look at patches. For patches to other modules or components, it's still recommended to use pull requests so original developer has a chance to check it before the merge.

zecke:
Thank you. I will go through pull requests to benefit from the travis-ci integration and not break things. :)

PR #863, opened 2016-11-16
modules/sdpops - extend sdp_remove_line_by_prefix()

mslehto:


miconda:
Is the change affecting the modules using the sdpops api -- the app_lua is using the function:

```
$ ag --cc sdp_remove_line_by_prefix
modules/app_lua/app_lua_exp.c
2186:	ret = _lua_sdpopsb.sdp_remove_line_by_prefix(env_L->msg, &media);
2279:	{"sdp_remove_line_by_prefix", lua_sr_sdpops_remove_line_by_prefix},
```

mslehto:
Didn't see that coming...
yes, I modified modules/sdpops/api.h:
```
 @@ -41,7 +41,7 @@ typedef struct sdpops_binds {
         sdp_keep_media_t     sdp_keep_codecs_by_name;
         sdp_remove_media_t   sdp_remove_media;
         sdp_remove_media_t   sdp_remove_transport;
 -       sdp_remove_media_t   sdp_remove_line_by_prefix;
 +       sdp_remove_media_type_t   sdp_remove_line_by_prefix;
```

kamailio-sync:
notifications@github.com wrote:

> @mslehto pushed 1 commit.
> 
> 18c7eaf  modules/app_lua: sdpops API changed

That seem to have fixed modules/app_lua.

miconda:
Is it now requiring two parameters from Lua?

If the second is not provided (e.g., using an old script with previous version of the function prototype), then I think it will crash, because it is doing a strlen(NULL).

mslehto:
I think I get it. I will prepare full Lua environment and test more thoroughly. Thanks for patience!

mslehto:
Now I handle original single parameter version as well as the new functionality with two parameters.

Maybe publishing functions via kemi would require less code for different language bindings, right?

miconda:
Yes, kemi should simplify and do the export only inside the sdpops module, then lua and python (at this moment, other languages to be added soon) will pick from there everything, without any change in lua/python modules.

PR #877, opened 2016-12-01
sipcapture: expose pv $hep(src_ip)

davidcsi:
- when using with Homer, on onreply_route it is not possible to
  get the source ip address sending the reply. This patch exposes
  pv $hep(src_ip), i.e.:

  onreply_route {
    ...
    xlog("[onreply_route] Source IP: $hep(src_ip)\n");
    ...
  }


adubovikov:
@miconda already wrote you in your prevision push. Please follow the guide line:

https://github.com/kamailio/kamailio/blob/master/.github/CONTRIBUTING.md

in your patch a lot changes that not needed. Please re-push.

PR #880, opened 2016-12-02
remove lib/kcore/faked_msg.h

mslehto:
Looks like lib/kcore/faked_msg.h is not needed anymore


miconda:
Thanks, very appreciated, it was for so long time in my to-do list ..

PR #888, opened 2016-12-14
dialplan: fix avp values when using PV_IDX_ITR

linuxmaniac:
* previously we were using the values from pv_printf_s() but
  that uses pv_get_buffer() so next call can interfere with
  the next iteration in the list


miconda:
It can be merged.

PR #889, opened 2016-12-15
dialog: [RPC] from-tag optional on dlg.dlg_list[_ctx]

linuxmaniac:


miconda:
Thanks, it can be merged!

PR #890, opened 2016-12-16
statsd: include usr_avp.h instead of usr_avp.c

tverlaan:
The start and stop function currently doesn't work since the AVP used by that function is not succesfully added because usr_avp is referenced wrongly. It does work when not compiled with `-Wl,-Bsymbolic-functions`, however that is the default under Ubuntu.


miconda:
Thanks!

PR #891, opened 2016-12-16
sipcapture: Fix $hep(0x007) and $hep(0x008)

zecke:
Fix $hep(0x007) and $hep(0x008) access from the cfg file. In these
two cases "i" would not be increased, ending in a infinite loop.
Modify and goto done as well.


zecke:
@adubovikov please have a look.

adubovikov:
Thank you!

PR #896, opened 2016-12-21
hep: Store ipstr in a static buffer for exporting to cfg

zecke:
When using hep(0x003) and hep(0x004) the extracted ip address
would be written to the local ipstr buffer but when returning
to the script the buffer would already be partially overwritten.

Keep the data around long enough by storing the result in a
static buffer.

### DIFF
--- a/src/modules/sipcapture/hep.c
+++ b/src/modules/sipcapture/hep.c
@@ -879,7 +879,7 @@ int hepv3_get_chunk(struct sip_msg *msg, char *buf, unsigned int len, int req_ch
         int chunk_vendor=0, chunk_type=0, chunk_length=0;
         int total_length = 0;
         int ret = 0;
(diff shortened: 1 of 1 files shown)


zecke:
Dear @adubovikov, please have a look if that is the right way to fix it and if not tell what a proper fix would look lile.

adubovikov:
for me it looks ok. But make it more sense copy all data to local pseudo variables in the kamailio.cfg and proceed it ?

zecke:
Thank you for the super quick reply. The data is already "corrupted" before it is assigned to the variable in the script.

```
$var(src_ip) = $(hep(0x003){sql.val});
```

Already causes the data to be corrupted at the 9th byte. Storing it statically seemed the most straight forward way.

adubovikov:
ok. lets do it this way! Please merge!

zecke:
Thank you!

PR #899, opened 2016-12-22
sqlops: Increase size of static escape buffer

zecke:
The code assumes that all characters require escaping and the
maximum input string length right now is 1023. When the ss7ops
module is used the length of a IAM is longer than this. Increase
the static buffer to 8k. This increases the memory usage of each
process by 7168 bytes.

### DIFF
--- a/src/modules/sqlops/sql_trans.c
+++ b/src/modules/sqlops/sql_trans.c
@@ -31,7 +31,7 @@
 
 #include "sql_trans.h"
 
-#define TR_BUFFER_SIZE 2048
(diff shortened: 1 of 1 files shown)


miconda:
Maybe making the size a modparam and allocating the buffer in mod_init will make it more flexible and easier to tune for larger buffer needs.

miconda:
Alternative implementation pushed to master branch.

zecke:
Sorry, I didn't notice the feedback and just rebased my branch. Yes. Your approach to make it runtime configurable is a lot better! Thank you.

PR #912, opened 2017-01-05
unittest updates

mslehto:
Adjust new source code paths and fix some bashism.
If sensible, next is mi_fifo to jsonrpcs fifo conversion.


miconda:
Thanks!

Issue #82, opened 2015-02-11
consistency on name of main c file for a module


linuxmaniac:
main file should be modulename.c or modulename_mod.c

oej:
Include file name should follow name of main file with module interface.

oej:
A few modules that needs renaming:
- ims_charging
- ims_qos
- ims_icscf
and friends.

oej:
Websocket main file is ws_mod.c

linuxmaniac:
[list.txt](https://github.com/kamailio/kamailio/files/63157/list.txt)

Modules without {module}.c or {module}_mod.c
* app_python
* auth_ephemeral
* auth_radius
* cdp
* cdp_avp
* cpl-c
* db2_ldap
* db_berkeley
* db_flatstore
* db_mysql
* db_postgres
* db_text
* dialog_ng
* erlang
* ims_auth
* ims_charging
* ims_icscf
* ims_isc
* ims_qos
* ims_registrar_pcscf
* ims_registrar_scscf
* ims_usrloc_pcscf
* ims_usrloc_scscf
* janssonrpc-c
* jsonrpc-c
* mohqueue
* outbound
* registrar
* uri_db
* websocket

linuxmaniac:
refreshed list:
- ims_auth KO
- ims_charging KO
- ims_icscf KO
- ims_isc KO
- ims_ocs KO
- ims_qos KO
- ims_registrar_pcscf KO
- ims_registrar_scscf KO
- ims_usrloc_pcscf KO
- ims_usrloc_scscf KO
- mi_fifo KO
- mi_rpc KO

linuxmaniac:
I think we are fine now! \o/

PR #913, opened 2017-01-05
Rename modules interface files

linuxmaniac:
rename remaining modules as describe at #82


miconda:
It can be merged as soon as Travis CI is happy with the build.

PR #914, opened 2017-01-05
presence: allow control of subscription deletion

lazedo:
use the configuration xavp to control subscription deletion after notify
searches for `delete_subscription` value in the configuration xavp


miconda:
You have to add the description in the docs for the new attribute.

lazedo:
@miconda added documentation. would liketo port this to 4.4

lazedo:
this pull requests represents a changing behaviour and in principle shouldn't be in 4.4, but we do have some contraints controlling the deletion of subscriptions. with this pull request we can manage a threshold of notify failures and delete the subscription based on that threshold.

miconda:
Thanks! I can be merged!

PR #917, opened 2017-01-07
presence: delay aux_body_processing until after subscription has been updated

charlesrchance:
Fixes out-of-sequence version when subs_db_mode <> 3 and fast_notify is enabled.

Version is incremented only when copy is updated in memory. Therefore, currently:

On SUBSCRIBE, subscription is updated first, so NOTIFY body contains newly incremented version.

But on PUBLISH, aux_body_processing is called first, so NOTIFY body contains previously stored version.

### DIFF
--- a/src/modules/presence/notify.c
+++ b/src/modules/presence/notify.c
@@ -1278,7 +1278,7 @@ error:
 
 int publ_notify(presentity_t* p, str pres_uri, str* body, str* offline_etag, str* rules_doc)
 {
-	str *notify_body = NULL, *aux_body = NULL;
+	str *notify_body = NULL;
 	subs_t* subs_array= NULL, *s= NULL;
 	int ret_code= -1;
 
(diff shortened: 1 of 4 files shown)


miconda:
Thanks -- it can be merged if nobody else has something against.

charlesrchance:
@miconda - is it OK to backport to 4.4?

miconda:
If nobody finds an issue with it, for me it's fine to backport.

charlesrchance:
Ok, I will wait another couple of weeks. We have it already in production since last week, so if no other issue reported by end of January I will assume it is safe to backport.

charlesrchance:
Hello again - given today's announcement of upcoming 4.4.5 release in the next couple of days, I would ask anyone with any issues to report it here, otherwise, I will backport tomorrow evening to ensure it is included in the release.

PR #926, opened 2017-01-13
documentation updates

mslehto:


src/modules/pdt/doc/pdt_admin.xml
```diff
@@ -501,11 +501,8 @@ pd_translate("*", "$var(a)");
 		<para>
 		RPC Command Example:
 		</para>
-		<programlisting  format="linespecific">
-...
-# &kamcmd; _sdomain_ _prefix_ _domain_
-&kamcmd; kamailio.org 123
-...
```
miconda: The dots here were on purpose to enclose the example within them in order to be more visible in README files. Also, the spaces/tabs added in front will make it show shifted a lot in the README, that's why most of the <programlisting  format="linespecific"> content is starting at the beginning of line and doesn't keep the indentation.

It can be merged, but then dots and the start at the beginning should be restored with a follow up commit.

mslehto: Thanks!
Makes sense, formatting restored in commit 06a0a1f6e91e4d1c0f87afb33519f1de4584a918

miconda:
Thanks!

PR #927, opened 2017-01-13
ims_*: fix  inline function xxx declared but never defined

linuxmaniac:
* move inline functions to bin_utils.h
* clang-format bin_utils.* files


miconda:
Does it worth moving the functions to .h file? It is a consistent size of code there, maybe just removing the 'inline' is better, probably the compiler is ignoring it anyhow and having it inside the .h means bigger object files for each .c including the .h.

PR #928, opened 2017-01-13
registrar: save() returns separate code on 503 generation

lbalaceanu:
We would like to know when the save("location") has generated a 503 message and as such we are thinking of modifying the save() function to return a code (like -2) for this error.

The default behaviour now seems to be that most errors in save() log a notice message, generate a reply via reg_send_reply() and then make sure that config doesn't go on with processing by returning 0. (tag "error:" in save() function definition)

Are you ok with such an intrusion in the code? 

Thank you

Also, do you see an alternative to determining that this 503 message has been generated?


miconda:
Thanks, it can be merged, but it needs also a note in the docs about the new return code.

More about the return code of the save() -- returning 0 seems to be a bug right now, because there is a flag to control if the reply should be sent or not from the code -- REG_SAVE_NORPL_FL.

```
error:
	update_stat(rejected_registrations, 1);
	if (is_route_type(REQUEST_ROUTE) && !is_cflag_set(REG_SAVE_NORPL_FL) )
		reg_send_reply(_m);

	return 0;
```

It should be a negative value and let the config writer to write a log, etc. and do exit. The default config file is expecting this in:

```
     if (!save("location")) {
          sl_reply_error();
      }
      exit;
```

Probably more different return codes can be used in order to differentiate between various error cases.

lbalaceanu:
Thanks for the input. I updated the pull request.

PR #929, opened 2017-01-14
rtpengine: Add kemi rtpengine_manage0/1

smititelu:
Basically tested it from python script, by grepping for incoming commands on rtpengine side.

IMHO, a nice improvement to the kemi interface would be to detect the number of input parameters given from the script and call the proper function straight ahead.

Can this be merged? Or shall we wait after 5.0 is released?

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -76,6 +76,7 @@
 #include "../../core/route.h"
 #include "../../core/rpc.h"
 #include "../../core/rpc_lookup.h"
+#include "../../core/kemi.h"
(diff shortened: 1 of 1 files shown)


miconda:
Thanks! Can be merged, but then I would rename rtpengine_manage1() to just rtpengine_manage(), to be in pair with rtpproxy kemi exports.

Variable number of arguments was in the plan, but iirc I couldn't find an easy way in python embed api to get the number of parameters (and didn't get much spare time to dig in more). Anyhow, I think in python it is always the same number of parameters given to a function call, just that those not provided explicitely have the default value specified when defining the function.

PR #930, opened 2017-01-14
kamailio-basic-kemi.cfg: add WITH_DB_TEXT

smititelu:
This is for the people out there (me included) who don't want to install mysql to play with kamailio:
- comment out ``#!define WITH_DB_MYSQL`` (still used by default)
- set the right ``DB_TEXT_DIR``, to match ``kamailio_path/misc/examples/kemi/db_text`` (absolute path is needed; I tried with relative path and it did not work)
- add users/passwords by editing ``kamailio_path/misc/examples/kemi/db_text/subscriber`` file

Hope it's helpful. Enjoy!

### DIFF
--- a/misc/examples/kemi/db_text/acc
+++ b/misc/examples/kemi/db_text/acc
@@ -1,0 +1,1 @@
+time(int) method(str) from_tag(str) to_tag(str) callid(str) sip_code(str) sip_reason(str) src_user(str) src_domain(str) src_ip(str) dst_ouser(str) dst_user(str) dst_domain(str)
--- a/misc/examples/kemi/db_text/address
+++ b/misc/examples/kemi/db_text/address
@@ -1,0 +1,1 @@
+id(int) grp(int) ip_addr(str) mask(str) port(int) tag(str)
--- a/misc/examples/kemi/db_text/bkp
+++ b/misc/examples/kemi/db_text/bkp
@@ -1,0 +1,1 @@
+src_user(str) src_domain(str) src_ip(str) dst_ouser(str) dst_user(str,null) dst_domain(str,null)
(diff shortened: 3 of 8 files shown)


miconda:
The examples in this folder are intended to reflect the equivalent of etc/kamailio-basic.cfg.

Then I think than embedding the db_text tables is a wrong approach. They can be installed using kamdbctl and that will ensure they are having the proper structure. Now, for example, a change in subscriber table structure has to be propagate manually in the new subfolder 'db_text'.

Moreover, there is a bkp table which I don't think it comes from default kamailio.

I think we should try find another alternative for hosting this, but not merge it into examples/kemi.

smititelu:
@miconda I agree, generating the db_text tables instead of manually managing them is definitely a better approach. That 'bkp' table is just a leftover file that I forgot to remove.

Maybe an approach would be to add a new kamailio-basic-dbtext.cfg in ``etc/``, with dbtext only pointing to kamdbctl regenerated tables in ``misc/scripts/dbtext/ser_db/``; also, from what I see, one could use the existing ``misc/scripts/dbtext/ser_db/ser_dbtext.sh`` script to install the dbtext table files in specific location.

miconda:
No new file should go in etc/ . The folder should be cleaned up anyhow. It is supposed to include only default config files to be used in packages.

Any other config examples have to go in misc/examples/ ... or we should create a dedicated repo not to end up with a ton of configs that need to be fetched every time with the code.

kamdbctl from utils/kamctl/ should be used for installing the db_text table files.

miconda:
I am closing this one, if there is anything else you want to discuss, then reopen or make a new pull request based on the comments above.

PR #934, opened 2017-01-17
Reuse PERLLDOPTS in db_perlvdb

ffontaine:
PERLLDOPTS is already used in app_perl module.
Reuse this variable in db_perlvdb module instead of always setting LIBS
to (shell perl -MExtUtils::Embed -e ldopts)
Indeed, this will not work when cross-compiling (the value will be
retrieved for the host perl and not the target perl)

### DIFF
--- a/src/modules/db_perlvdb/Makefile
+++ b/src/modules/db_perlvdb/Makefile
@@ -4,7 +4,11 @@
 include ../../Makefile.defs
 auto_gen=
 NAME=db_perlvdb.so
-LIBS:=$(shell perl -MExtUtils::Embed -e ldopts)
+ifeq ($(PERLLDOPTS),)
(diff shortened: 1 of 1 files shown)


miconda:
Not an app_perl user here, wrote some code for it though ...

It is not clear from the commit message: is this commit breaking the cross compilation?

What are the benefits of using the env PERLLDOPTS variable instead of computing it with `perl -MExtUtils::Embed -e ldopts`?

ffontaine:
The commit is actually fixing the cross-compilation of the db_perlvdb module, not breaking it.
Without it, this module calls perl to get the ld flags, for example -L/usr/lib which is not what you want when you cross compile as we will not find the target libraries. To find the target libraries, we have to set the PERLLDOPTS variable.

miconda:
Ok, thanks for clarification. I will merge it.

PR #940, opened 2017-01-18
db_text: implement fetch and memory constraints

lazedo:
when dealing with large db_text files, pkg_memory is not suitable for
operating the database.

implementing fetch allows modules like presence & registrar & usrloc
to query large tables without constraints on pkg_memory.

creates tmp tables in shared memory for query results

### DIFF
--- a/modules/db_text/dbt_api.c
+++ b/modules/db_text/dbt_api.c
@@ -39,7 +39,7 @@ int dbt_use_table(db1_con_t* _h, const str* _t)
 /*
  * Get and convert columns from a result
  */
-static int dbt_get_columns(db1_res_t* _r, dbt_result_p _dres)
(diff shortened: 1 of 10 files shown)


lazedo:
this pull requests represents a changing behaviour and in principle shouldn't be in 4.4, but db_text is unusable in 4.4 with large db_text files, so i consider this as a fix for the stable branch

miconda:
For clarification: with new features we mainly refer to new functions/parameters to configuration file or rpc commands as well as how the configuration file is interpreted. Reworking a part of code in a module is not always a new feature, to be discussed from case to case, evaluating the impact on the other modules.

I understand where the problem is with large dbtext tables + pkg memory and that this can be a fix. I am fine with merging and backporting.

lazedo:
@miconda awesome! thank you

PR #942, opened 2017-01-19
Create ktls_no_curl

ffontaine:
Create ktls_no_curl group to be able to compile ktls module when curl is
not available


miconda:
Nothing big, but I feel like the naming of the group is upside down. We try to use a name to reflect a bit what it contains. I see the new group name as saying what it doesn't contain. What do you think about using ktls_basic ?

ffontaine:
You're perfectly right, thanks for this suggestion. I will update the PR accordingly.

miconda:
Thanks!

PR #944, opened 2017-01-22
presence: allow use of $subs in notify process

lazedo:


miconda:
Thanks, it can be merged.

PR #945, opened 2017-01-23
presence: use sockinfo from subs to init uac

lazedo:
for deployments with multiple ports where UAs subscribe to specific port and expect the notify from that port


lazedo:
closing this in favour on #947

PR #946, opened 2017-01-24
tm: apply new updates with Via with updated request

lazedo:
this is fixed on master with the new t_run_local_req function


miconda:
Is this a backport?

lazedo:
not a backport, only a fix. its fixed in master but issue remains in 4.4 branch. to backport would have to back to https://github.com/kamailio/kamailio/pull/727. not sure how to do that.

miconda:
OK, thanks for clarification. It can be merged.

PR #955, opened 2017-01-29
tm: reset T_ASYNC_CONTINUE flag in t_suspend

grumvalski:
Related to http://lists.sip-router.org/pipermail/sr-users/2017-January/095772.html
Resetting the flag in t_suspend allows adding new branches in a failure route while suspending the transaction.

### DIFF
--- a/src/modules/tm/t_suspend.c
+++ b/src/modules/tm/t_suspend.c
@@ -143,7 +143,8 @@ int t_suspend(struct sip_msg *msg,
 	*hash_index = t->hash_index;
 	*label = t->label;
 
-
+	/* reset the continue flag to be able to suspend in a failure route */
+	t->flags &= ~T_ASYNC_CONTINUE;
 
 	/* backup some extra info that can be used in continuation logic */
 	t->async_backup.backup_route = get_route_type();


miconda:
Probably it's ok -- couldn't see a side effect right now. Is it tested/used anywhere else than src/modules/tm/t_reply.c:1123?

grumvalski:
Currently it's only used in that t_reply check.

miconda:
Unless you think it should be better to be reset when t_continue() is finished instead of inside t_suspend(), you can merge it.

grumvalski:
You're right, I'll try this 2nd approach.

kamailio-sync:
Thinking it better, wouldn't moving the flag reset at the end of t_continue
make the check in t_reply useless?
I have to admit that I don't fully understand that check anyway :)

miconda:
It's a bit foggy the purpose right now, or better said the duration when that flag is needed. I didn't have the time to dig into it too much and don't remember the full purpose for it right now. I expected to be related to the fact of choosing the final response while a suspended branch is still processed.

But I think your patch is better in this regard, because it preserves the old behaviour (which didn't have a complaint on this flag) longer, until a new suspend is done. So can be merged as it is.

grumvalski:
Thanks, should I backport it too?

PR #957, opened 2017-01-30
db_mysql: Let the database convert the timestamp

zecke:
Instead of using the time2str routines that go through timezone
conversion in kamailio, let MySQL do it. This helps with the
sipcapture module in case the system time is CET but the database
is operated in UTC. In these cases the micro_ts and date columns
would be offseted by the timezone difference and look-ups fail
to work.

### DIFF
--- a/src/modules/db_mysql/db_mysql.c
+++ b/src/modules/db_mysql/db_mysql.c
@@ -41,6 +41,7 @@ unsigned int my_connect_to = 2; /* 2 s by default */
 unsigned int my_send_to = 0; /*  enabled only for mysql >= 5.25  */
 unsigned int my_recv_to = 0; /* enabled only for mysql >= 5.25 */
 unsigned int my_retries = 1;    /* Number of retries when command fails */
+unsigned int my_server_timezone = 0; /* Use FROM_UNIXTIME() for date conversion */
 
 unsigned long my_client_ver = 0;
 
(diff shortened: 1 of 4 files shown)


zecke:
@adubovikov

adubovikov:
for me it's ok
@miconda  ?

miconda:
I think this can be added only as a modparam and disabled by default, because it can break the other modules that expect to store the server timestamp, like usrloc (location table), acc, msilo.

adubovikov:
@miconda you are right,  

https://github.com/kamailio/kamailio/blob/master/src/lib/srdb1/db_ut.c#L211-L237

db_str2time expects the "original timestamp" in the kamailio's timezone. 

@zecke can you please make it configurable ?

zecke:
Made it configurable but I have not tested/inserted data in a production system yet. So please wait for a second. I tried to update the documentation with the new parameter as well.

zecke:
Okay it seems to work (on and off) of that setting. Used hepgen.js to send some data and had a breakpoint in the code in km_val.c and it was triggered (or wasn't depending on the global setting). _s also looked sane.

juha-h:
Holger Freyther writes:

> Made it configurable but I have not tested/inserted data in a
> production system yet. So please wait for a second. I tried to update
> the documentation with the new parameter as well.

Are we in frozen state with no new features to trunk?

-- Juha

miconda:
@juha-h -- this seems to fix an issue when kamailio server (for sip routing) is different than kamailio server (for sipcapture). Overall is no change in the default behaviour, just a switch to choose the preferred date/time handling. It will be merged in master branch as the branch 5.0 was already created. But I will consider it for backporting if it looks clean and simple.

zecke:
Thank you.

PR #959, opened 2017-01-31
Update jansson_path.c

vance-od:
Hi Daniel, community!
in case of build on old debian (e.g. lenny) - statically linked libjansson has to be used, as a result build throw errors, cos libjansson has functions with the same names
proposed changes - fixed issues that I have with build on lenny


miconda:
Thanks! It will be merged. Maybe we should add also add a prefix like ksr_ just in case someone will need to use it from other file and remove the static. For now should be ok.

PR #961, opened 2017-01-31
presence: remove transaction creation from publ_notify_notifier() and mark_presentity_for_delete()

linuxmaniac:
* publ_notify_notifier() and mark_presentity_for_delete() are
  called inside a already created transaction
* adding transaction handle for msg_presentity_clean()


miconda:
If nobody else has comments about this commit, I am ok to be merged.

PR #962, opened 2017-02-01
core: avoid overrun-buffer-arg

linuxmaniac:
> Overrunning array ((struct a_rdata *)rr->rdata)->ip of 4 bytes
> by passing it to a function which accesses it at byte offset 15
> using argument len (which evaluates to 16)


miconda:
Thanks can be merged and backported.

PR #966, opened 2017-02-04
pkg/kamailio/centos: Added SPEC file for CentOS7

sergey-safarov:


sergey-safarov:
@linuxmaniac thanks for fast review.
Prefix added.

sergey-safarov:
Hello @linuxmaniac
Why not use kamailio.service and kamailio.sysconfig from debian?
This is more easy to maintain packaging.

About difference of environment file location. This is adjusted in kamailio.spec:595

https://github.com/kamailio/kamailio/pull/966/files#diff-7757a46a5c00eaeca332f4a6be46e96fR595

pkg/kamailio/centos/7/kamailio.spec
```diff
@@ -0,0 +1,1387 @@
...
+
+%description    json
+json string handling and RPC modules for Kamailio.
+
+
+%if %{with kazoo}
+%package    kazoo
+Summary:    Kazoo middle layer connector support for Kamailio
+Group:      System Environment/Daemons
+Requires:   libuuid, kamailio = %ver
```
lazedo: missing rabbitmq-c, json-c, libevent

pkg/kamailio/centos/7/kamailio.spec
```diff
@@ -0,0 +1,1387 @@
...
+BuildRequires:	rabbitmq-c-devel, json-c-devel, libuuid-devel
```
lazedo: missing libevent-devel

sergey-safarov:
@lazedo I will update master branch after this PR is merged.

linuxmaniac:
> Why not use kamailio.service and kamailio.sysconfig from debian?
> This is more easy to maintain packaging.

because is what we have in master and from my point of view is better to be independient

sergey-safarov:
Hello @linuxmaniac 
Pull request updated.

PR #968, opened 2017-02-04
pkg/kamailio/centos/7: added missing kamailio.service file

grumvalski:


sergey-safarov:
Yes. I missed add this to git.
May be add "kamailio.service" as symbolic link to "../../deb/stretch/kamailio.service"

grumvalski:
I don't think the Debian service file is fine for Centos. For example the path of the environment file is different in the two systems: /etc/default for Debian, /etc/sysconfig for Centos.

sergey-safarov:
You are correct

PR #971, opened 2017-02-07
tm: deep cloning of the request for fake environment

linuxmaniac:


grumvalski:
seems ok to me.

miconda:
@linuxmaniac, @grumvalski  - wondering if you had any time running some tests with the patch, just to have a feeling if you observed any side effect.

I didn't get to review yet, planning to do it during the next days.

linuxmaniac:
I did test with our scenarios, I didn't detect any side effect

miconda:
Can be merged and backported to 5.0, eventually later to 4.4.

Probably the deep cloning from acc makes no longer sense, I will check that myself, for now should be harmless.

grumvalski:
Sorry for the late feedback. I did try it with some of our suspend scenarios and I haven't noticed side effects. No load though.

PR #978, opened 2017-02-07
pv: add the $env() PV class to read env vars

camilleoudot:
this allows kamailio to have read-only (for now) access to environment variables

### DIFF
--- a/src/modules/pv/pv_core.c
+++ b/src/modules/pv/pv_core.c
@@ -50,6 +50,8 @@
 #include "pv_core.h"
 #include "pv_svar.h"
 
+#include <string.h>
+#include <stdlib.h>
 
 static str str_udp    = { "UDP", 3 };
 static str str_5060   = { "5060", 4 };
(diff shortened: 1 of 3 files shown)


src/modules/pv/pv_core.c
```diff
@@ -3233,3 +3235,44 @@ int pv_get_msg_attrs(sip_msg_t *msg, pv_param_t *param, pv_value_t *res)
...
+{
+	char *val;
+	char *csname = (char *) param->pvn.u.dname;
+
+	if (csname) {
+		val = getenv(csname);
+
+		if (val) {
+			return pv_get_strzval(msg, param, res, val);
+		} else {
```
linuxmaniac: what about

```
if (csname) {
  val = getenv(csname);
  if(val) return pv_get_strzval(msg, param, res, val);
}
return pv_get_null(msg, param, res);
```

src/modules/pv/pv_core.c
```diff
@@ -3233,3 +3235,44 @@ int pv_get_msg_attrs(sip_msg_t *msg, pv_param_t *param, pv_value_t *res)
 			return pv_get_null(msg, param, res);
 	}
 }
+
+int pv_parse_env_name(pv_spec_p sp, str *in)
+{
+	if(in->s==NULL || in->len<=0)
+		return -1;
+
+	char *csname;
```
linuxmaniac: can you please move the declaration to the top, please.

src/modules/pv/pv_core.c
camilleoudot: easier to read, thanks

src/modules/pv/pv_core.c
camilleoudot: sure

src/modules/pv/pv_core.c
```diff
@@ -3238,11 +3238,11 @@ int pv_get_msg_attrs(sip_msg_t *msg, pv_param_t *param, pv_value_t *res)
 
 int pv_parse_env_name(pv_spec_p sp, str *in)
 {
+	char *csname;
+
 	if(in->s==NULL || in->len<=0)
```
linuxmaniac: I would do
```
if(sp==NULL || in == NULL || in->len <=0)
``` 

as in the [kamailio-devel-guide](http://www.asipto.com/pub/kamailio-devel-guide/#c16_add_pv)

camilleoudot: it does not look like any of the parameters can be null, judging by the checks done before calling this function in https://github.com/kamailio/kamailio/blob/master/src/core/pvapi.c#L873, but I may be missing something else

linuxmaniac: and where is the documentation update?

camilleoudot: I'll document it on the wiki when/if it makes its way to master eventually (after the feature freeze?)

linuxmaniac: FTR, it works nicely 
```
route[TEST] {
	$var(tmp) = $env(FAKE);
	xlog("L_NOTICE", "PATH=$env(PATH) FAKE=$var(tmp)\n");
	return;
}
```
```
 3(1978) NOTICE: <script>: PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin FAKE=0
```

miconda:
Thanks! Can be merged into master.

PR #981, opened 2017-02-09
auth_db : more explanation for generic error

ycaner06:
added more explanation about error instead of some generic error.


miconda:
Thanks!

PR #986, opened 2017-02-13
sipcapture: Fixed missing 'break' in pv_parse_hep_name switch

davidcsi:
- Added missing 'break;' in sipcapture.c's pv_parse_hep_name function's switch
   statement.


adubovikov:
so far, it's ok. Not sure if we have permission to merge now or wait for @miconda

miconda:
@adubovikov - master branch is open for new features -- you can commit as you wish until we will freeze for 5.1 in several months.

The branch 5.0 was created last week. If it is a bug, then it can be backported, too.

PR #987, opened 2017-02-13
Set of updates for CentOS packaging

sergey-safarov:


miconda:
Thanks, I will merge it.

Issue #969, opened 2017-02-05
Packaging: Expand the list of required 3rd party libraries for main Kamailio package


oej:
(from devmeeting #Fosdem) - adding new dependencies as core dependencies 

- Curl
- libssl
- uuid
- libunistring
- libxml2 

Any other candidates? They have to be libraries that are commonly installed on most Linux/Unix system.

This would mean that a lot of more modules will be built by default, like TLS, HTTP_CLIENT, WEBSOCKET. 

We may also look over the names of the packages/groups for the build system and linux distress

This will require:
- Update of wiki and README/INSTALL
- Update and modification of debian control files and similar files for other distress
- Modification of build system (Makefile groups etc)

Anything else?

oej:
This will require:
- Update of wiki and README/INSTALL
- Update and modification of debian control files and similar files for other distress
- Modification of build system (Makefile groups etc)

Anything else?

miconda:
Following the online devel meeting discussion, this one is closed. PRs can be made based on what is considered to be adjusted in packaging.

PR #990, opened 2017-02-13
Add kjansson_basic group

ffontaine:
This group could be used to compile kjansson module when libevent is not
available

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/Makefile.groups
+++ b/src/Makefile.groups
@@ -146,8 +146,11 @@ mod_list_json=json
 # - modules depending on json (+libevent) library
 mod_list_json_event=jsonrpcc
 
+# - modules depending on jansson library
+mod_list_jansson=jansson
+
 # - modules depending on jansson (+libevent) library
-mod_list_jansson=jansson janssonrpcc
+mod_list_jansson_event=janssonrpcc
 
 # - modules depending on redis library
 mod_list_redis=ndb_redis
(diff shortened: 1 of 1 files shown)


miconda:
I wonder if makes sense to split more into groups of a single module, because it is possible to install a single module at once, which I guess it is what you are looking for.

Overall, I am not against this approach, but it may be better to understand the needs you have for packaging and other solutions may be already in place.

For example, at some point the deb packaging had the option to list the modules to be packaged separately:

  * https://github.com/kamailio/kamailio/blob/master/pkg/ser/deb/debian/rules#L81
  * https://github.com/kamailio/kamailio/blob/master/pkg/ser/deb/debian/rules#L107

I feel like if we end up having a group for each module, then the grouping makes no more sense.

Maybe other can comment as well.

There is also #969 trying to redefine the packages to have less of them based on common interest.

ffontaine:
You're completely right, having that much groups is not the best approach.

However, group_include and include_modules options can not be mixed. So I have to stick with one of these variables. I could use include_modules instead of group_include but then I would have to recopy most of what has been done in Makefile.groups that is to say trying to guess what modules should be build depending on what packages are installed on his (embedded) system.

A better approach would be to add a configure script (or Makefile variables) that would allow the user to enable/disable modules based on third-party dependencies. For example, we could add --enable-libevent, --enable-curl, --enable-ssl and so on. In case the user did not specify an enable/disable for a specific option, kamailio could use pkgconfig to try to check if the external dependency is installed on the system. Most of the open source components have this approach but it would be a bigger addition to kamailio.

For information, except kjanssson_basic, my last need would be to split kldap group as db2_ldap module depends on ldap and sasl.

miconda:
Merging and we will try to find a proper solution in the future.

PR #995, opened 2017-02-15
sipcapture: Added exported pseudo-variable $hep(dst_ip) and updated README 

davidcsi:
Added a new exported pseudo-variable to expose HEP's message's original source ip address.
Added an "Exported pseudo-variabled" to the README file to document $hep(src_ip) and $hep(dst_ip).


adubovikov:
thank you

PR #996, opened 2017-02-15
Fix execinfo error in ims_dialog module

ffontaine:
Currently, ims_dialog includes execinfo.h to use backtrace in the
backtrace2str function. However, all calls to backtrace2str are
commented so disable this function through "if 0" and comment the
include on execinfo.h to allow the cross-compilation of the ims_dialog
module as some libc's don't always have this header (uclibc-ng, musl)

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/modules/ims_dialog/dlg_hash.c
+++ b/src/modules/ims_dialog/dlg_hash.c
@@ -17,7 +17,7 @@
 #include "dlg_profile.h"
 #include "dlg_handlers.h"
 #include "dlg_db_handler.h"
-#include <execinfo.h>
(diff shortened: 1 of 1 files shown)


miconda:
Can you update the commit message to follow the recommendations from contributions guide?

  - https://www.kamailio.org/wiki/devel/github-contributions#commit_message_format

The commit messages are used to generate the changelog and it is good to have a coherence there.

miconda:
Thanks!

Issue #997, opened 2017-02-16
ucontacts in shared memory even on DB_ONLY


linuxmaniac:
I've detected that the allocated shared memory is always growing due to

```
proxy[2527]: NOTICE: fm_status:  count=376594 size= 103410648 bytes from usrloc: ucontact.c: new_ucontact(98)
proxy[2527]: NOTICE: fm_status:  count=1882939 size= 110079752 bytes from usrloc: ../../ut.h: shm_str_dup(723)
```
we do have usrloc in db_mode 3 ( DB_ONLY ).

This is kamailio 4.4.4 version

Investigation underway.

linuxmaniac:
I would say the origin of the problem is calls to get_urecord() without release_urecord()
```
/*!
 * \brief Release urecord previously obtained through get_urecord
 * \warning Failing to calls this function after get_urecord will
 * result in a memory leak when the DB_ONLY mode is used. When
 * the records is later deleted, e.g. with delete_urecord, then
 * its not necessary, as this function already releases the record.
 * \param _r released record
 */
void release_urecord(urecord_t* _r)
{
	if (db_mode==DB_ONLY) {
		free_urecord(_r);
	} else if (_r->contacts == 0) {
		mem_delete_urecord(_r->slot->d, _r);
	}
}
```

linuxmaniac:
added some debug

notice in this example that ``ucontact[0x7f7b315f1d08]`` is never freed
REGISTER
```
Feb 16 17:04:43 sp1 proxy[9687]: NOTICE: <script>: New request on proxy - M=REGISTER R=sip:192.168.1.71 F=sip:43993005@192.168.1.71 T=sip:43993005@192.168.1.71 IP=192.168.1.42:53152
 (127.0.0.1:5060) ID=8e40daa391b6e4d82b296c77eb0c6986@0.0.0.0 UA='Jitsi2.9.5521Linux'
Feb 16 17:04:43 sp1 proxy[9687]: DEBUG: usrloc [udomain.c:174]: get_static_urecord(): urecord static[0x7f7b41740200]
Feb 16 17:04:43 sp1 proxy[9687]: DEBUG: usrloc [ucontact.c:142]: new_ucontact(): new ucontact[0x7f7b315f1d08]
Feb 16 17:04:43 sp1 proxy[9687]: DEBUG: usrloc [udomain.c:1155]: get_urecord(): DB_ONLY urecord[0x7f7b41740200]
[...]

PR #998, opened 2017-02-17
pua_reginfo: fix memory leak when usrloc is DB_ONLY

linuxmaniac:
See #997 for details


miconda:
Can be merged and backported as needed. Thanks.

PR #999, opened 2017-02-17
rtpengine: really do allow unsigned setid

zxcpoiu:
## Motivation

1. DOC says `setid` in mysql should be `UNSIGNED INT`:
    http://www.kamailio.org/docs/modules/4.4.x/modules/rtpengine.html#rtpengine.p.table_name
    http://www.kamailio.org/docs/modules/4.4.x/modules/rtpengine.html#rtpengine.p.setid_col

2. kamailio did report invalid value if `setid` less than 0.

3. kamailio rtpengine module declares setid as `int setid`, thus we can not use setid more than 2147483647

This PR changes declaration of `setid` from `int` to `unsigned int`

## UNSURE (REVIEW NEEDED):

I'm not sure about:

should `int_val` in `fixup_set_ip` at [rtpengine.c#L910](https://github.com/kamailio/kamailio/blob/master/src/modules/rtpengine/rtpengine.c#L910) be declared as `unsigned int` as well?

not sure about the usage of `int pv_locate_name()` and `unsigned short str2s()` in the `fixup_set_ip()` function.

should we use `str2int` instead `str2s` ?

let me know if I'm on the wrong direction

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -140,7 +140,7 @@ static int parse_flags(struct ng_flags_parse *, struct sip_msg *, enum rtpe_oper
 static int rtpengine_offer_answer(struct sip_msg *msg, const char *flags, int op, int more);
 static int fixup_set_id(void ** param, int param_no);
 static int set_rtpengine_set_f(struct sip_msg * msg, char * str1, char * str2);
-static struct rtpp_set * select_rtpp_set(int id_set);
+static struct rtpp_set * select_rtpp_set(unsigned int id_set);
 static struct rtpp_node *select_rtpp_node_new(str, str, int, struct rtpp_node **, int);
 static struct rtpp_node *select_rtpp_node_old(str, str, int, enum rtpe_operation);
 static struct rtpp_node *select_rtpp_node(str, str, int, struct rtpp_node **, int, enum rtpe_operation);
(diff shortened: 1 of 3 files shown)


rfuchs:
Yes I think you're right. It should use `str2int` instead of `str2s`.

zxcpoiu:
thanks! updated.

src/modules/rtpengine/rtpengine.c
```diff
@@ -933,11 +934,11 @@ static int fixup_set_id(void ** param, int param_no)
...
 	} else {
-		int_val = str2s(*param, strlen(*param), &err);
-		if (err == 0) {
+		int_val = str2int(&s, &set_id);
+		if (int_val == 0) {
 			pkg_free(*param);
-			if((rtpp_list = select_rtpp_set(int_val)) ==0){
-				LM_ERR("rtpp_proxy set %i not configured\n", int_val);
+			if((rtpp_list = select_rtpp_set(set_id)) ==0){
+				LM_ERR("rtpp_proxy set %i not configured\n", set_id);
```
rfuchs: format should be %u here

src/modules/rtpengine/rtpengine.c
```diff
@@ -1442,9 +1443,9 @@ mod_init(void)
 	/* select the default set */
 	default_rtpp_set = select_rtpp_set(setid_default);
 	if (!default_rtpp_set) {
-		LM_NOTICE("Default rtpp set %d NOT found\n", setid_default);
+		LM_NOTICE("Default rtpp set %u NOT found\n", setid_default);
```
rfuchs: The type of `setid_default` itself also ought to be changed to unsigned accordingly

zxcpoiu:
@rfuchs 
thanks for review, I have it updated and rebased accordingly.

it's true, I felt a little bit awkward while I was trying to turn `int` to `unsigned`.

zxcpoiu:
just want to share my use case if anyone has similar purpose:

I was using `mediaproxy` module and have own load balancing/failover outside of `kamailio`.
the id for media server was `ip address string` which will be put into `media_relay_avp`.

so for the uniqueness, I just need to convert `ip address string` into `unsigned int` as media server id to issue command to.

`int` variable only supports ip address that the first field is less than `127`
use `unsigned int` can easily convert ip string/int back and forth while keeping uniqueness.

PR #1000, opened 2017-02-17
usrloc: fix ucontact shared leak

linuxmaniac:
keep a copy of ``urecord`` if mode is DB_ONLY as it is static preventing leaking ``ucontact`` to shared memory

See #997 for details

### DIFF
--- a/src/modules/usrloc/ucontact.c
+++ b/src/modules/usrloc/ucontact.c
@@ -1672,6 +1672,7 @@ static inline int update_contact_db(ucontact_t* _c)
  */
 int update_ucontact(struct urecord* _r, ucontact_t* _c, ucontact_info_t* _ci)
 {
+	struct urecord _ur;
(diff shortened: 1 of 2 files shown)


src/modules/usrloc/ucontact.c
```diff
@@ -1672,6 +1672,7 @@ static inline int update_contact_db(ucontact_t* _c)
  */
 int update_ucontact(struct urecord* _r, ucontact_t* _c, ucontact_info_t* _ci)
 {
+	static struct urecord _ur;
```
miconda: I don't see a reason to declare the variable _ur static here or in the other cases of the patch. It does not seem to be used outside of the function, so it can be declared locally (on the stack).

miconda:
Besides my remark with the static variable, I'm ok to get the PR merged and backported. Thanks.

PR #1007, opened 2017-02-23
siptrace: new hlog() function to send logs to homer

camilleoudot:
This adds the `hlog()` function to the siptrace module, that sends a HEP3 LOG event packet to the URI configured in the module.

Usage:
```
...
hlog("[cfg:$cfg(line)] This is a log from kamailio to Homer");
...
hlog("$hdr(P-MyID)", "Another one with a custom correlation ID");
...
```


adubovikov:
:+1:

PR #1009, opened 2017-02-24
Call-Info header support

linuxmaniac:
and use it on ``sca``

### DIFF
--- a/src/core/parser/case_c_inf.h
+++ b/src/core/parser/case_c_inf.h
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2017 Victor Seva (sipwise.com)
+ *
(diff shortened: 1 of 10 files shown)


miconda:
The license for core file has to be bsd-like (e.g., as in src/core/atomic/atomic_native.h or src/modules/topoh/topoh_mod.c), copyright stays with you, so we don't have to hunt you in few more years if we need to link to a new lib that requires an exception like we have for libssl these days.

linuxmaniac:
> The license for core file has to be bsd-like

Solved

> so we don't have to hunt you in few more years 

😨

miconda:
It can be merged, thanks!

PR #1013, opened 2017-03-03
modules/siptrace: remove duplicate HEP version check

mslehto:
I think only this check is enough:
https://github.com/kamailio/kamailio/blob/f6396033763f1d55c11744ddb9774e112865400a/src/modules/siptrace/siptrace.c#L303-L306


adubovikov:
yep. thank you!

PR #1014, opened 2017-03-03
db_text fixes & live query

lazedo:


src/modules/db_text/dbt_raw_query.c
```diff
@@ -117,15 +117,22 @@ int dbt_raw_query_select(db1_con_t* _h, str* _s, db1_res_t** _r)
 		result_cols = pkg_malloc(sizeof(db_key_t) * cols);
 		memset(result_cols, 0, sizeof(db_key_t) * cols);
 		for(n=0; n < cols; n++) {
-			result_cols[n] = &_tbc->colv[n]->name;
+			result_cols[n] = pkg_malloc(sizeof(str));
```
miconda: The result of mallocs are not checked if there is a valid pointer or NULL, which can happen on not enough memory available.

miconda:
Thanks, it can be merged!

PR #1016, opened 2017-03-03
presence: add pres_has_subscribers

lazedo:


miconda:
It needs the documentation of the new function in the docbook xml files. Then it can be merged.

PR #1017, opened 2017-03-03
presence: add a offset for allowed cseq difference

lazedo:


miconda:
The new parameter needs to be documented, then it can be merged.

PR #1020, opened 2017-03-06
app_perl: Make sv2int_str() static

jhcloos:
The function is only called from the xs file, and C11’s inline
semantics mean that gcc-5 and later miscompile this file when
no optimization is specified.

Signed-off-by: James Cloos <cloos@jhcloos.com>


miconda:
Can you update the commit message to follow the recommendations from contributions guide?

The commit messages are used to generate the changelog and it is good to have a coherence there.

jhcloos:
Re-submitting.

jhcloos:
Github didn’t bother to email me your comment; I just happened on it today.

The current tip of https://github.com/jhcloos/kamailio.git master has the updated commit.  I can’t tell whether this web ui is doing the right thing....

PR #1024, opened 2017-03-08
core: enable reuse of tcp sockets

grumvalski:
As discussed on the mailing list and at FOSDEM this is a PR to start the discussion on the support of SO_REUSEPORT.

### DIFF
--- a/src/Makefile.defs
+++ b/src/Makefile.defs
@@ -680,7 +680,7 @@ C_DEFS= $(extra_defs) \
 	 #-DSTATS \
 	 #-DNO_LOG \
 	 #-DPROFILING \
-	 #-DNO_SIG_DEBUG
+	 #-DNO_SIG_DEBUG 
 
 #PROFILE=  -pg	# set this if you want profiling
 				# you may also want to set -DPROFILING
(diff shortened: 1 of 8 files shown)


miconda:
The patch is not that complex, so it can be merged, but I think there are few things to add afterwards:

  * find a way to have the option compiled always and enabled if available. Very unlikely someone will re-compile with other define options. With a quick search, maybe uname() (https://linux.die.net/man/2/uname) can be used to detect if the kernel is at least version 3.9.0. Then still have the global kamailio.cfg parameter to enable/disable per kamailio instance.
  * make actual use of this capability -- probably that means ability to create more than one tcp manager process that can listen on the same port and accept new connections

grumvalski:
Thanks for the feedback Daniel.
About the first point, I think that we still need the #define because if SO_REUSEPORT is not defined for the running OS the compilation will fail. As far as I know the option is currently supported in FreeBSD/OpenBSD/NetBSD/Mac OSX, not in Solaris. Maybe we could conditionally add the define in Makefile.defs as it is done for other options depending from the OS?
For the second point for the moment the two concrete usages I've found is for forcing the sending socket in dispatcher and in the uac module, but for sure being able to create additional tcp manager processes is much more interesting!

miconda:
I think ifdef can be used to check if SO_REUSEPORT is defined, on macos:

```
/usr/include/sys/socket.h:136:#define	SO_REUSEPORT	0x0200		/* allow local address & port reuse */
```

On Linux:

```
/usr/include/asm-generic/socket.h:25:#define SO_REUSEPORT	15
```

So it appears it is a define when available.

The two use cases you listed are also important, I haven't thought of them, so I think that if we can get it enabled at compile time based on availability (can be a follow up commit), the PR can be merged.

grumvalski:
Just updated the PR with your suggestion.

miconda:
Thanks, it can be merged.

PR #1028, opened 2017-03-10
4.4 dmqfix backport

jchavanton:
Tested in my lab


jchavanton:
missing other modifications

PR #1033, opened 2017-03-14
utils/pdbt: FreeBSD fix, gcc is hardcoded, header is missing

dunst0:
- use CC environment variable if set
- added missing sys/stat.h header


miconda:
Thanks!

PR #1034, opened 2017-03-15
uri_db: rework the possibiltiy to use any sip uri

dunst0:
Adding the possibility to check any sip uri against the uri_db table and updates the log macros from LM_ to the new ones.

### DIFF
--- a/src/modules/uri_db/checks.c
+++ b/src/modules/uri_db/checks.c
@@ -138,7 +138,7 @@ static inline int check_username(struct sip_msg* _m, struct sip_uri *_uri,
 		 * username, the comparison is case insensitive
 		 */
 		if (_uri->user.len == username.len) {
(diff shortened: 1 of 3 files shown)


miconda:
The commits for new features should not be mixed with changes for formatting or renaming of functions/macros in existing code, because it is not easy to review what are the actual new additions. Please rework the changes and make a commit dedicated only to the new feature.

Moreover, LM_* macros are also new ones that prefix the log messages with function name, etc... no need to change those macros at all. If we decide to replace/remove them, then it will be easier to do substituttion on a LM_XYZ than on XYZ.

dunst0:
Thanks for the remarks, I will rework the changes.

dunst0:
@miconda hope it is Ok how I did the changes you asked for.

miconda:
Thanks!

PR #1036, opened 2017-03-16
Fix for no cdr after Kamailio 5.0 restart

lbalaceanu:
Hi Daniel,

This is a sketch of how I would force the execution of the acc/cdr dialog callbacks for dialogues created from the database. This solution needs more testing and I don't expect it to be accepted as is.

I think that the run_create_callbacks was omitted because at a later step, the callbacks would need to access the initial request (params->req) that originated the dialog. Looking in the generated CDRs one can also spot that certain values are left blank most probably because of the params->req message is missing.

How would you solve this problem? Is there a way to populate/ mock the request with values inside the dialog_vars table? Would you add all variables of interest inside that table or inside the dialog table? 

Thank you,
Lucian

### DIFF
--- a/src/modules/acc/acc_cdr.c
+++ b/src/modules/acc/acc_cdr.c
@@ -588,7 +588,7 @@ static void cdr_on_start( struct dlg_cell* dialog,
                           int type,
                           struct dlg_cb_params* params)
 {
-    if( !dialog || !params)
+    if( !dialog )
     {
         LM_ERR("invalid values\n!");
         return;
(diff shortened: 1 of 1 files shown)


miconda:
Maybe DLGCB_LOADED callback can be used for this situation -- I guess it is no need populate again the variables in the dlg structure as they were already done when the dialog was effectively created.

If you create a new solution, make a commit per module and prefix the first log message with the module name:

lbalaceanu:
Updated the pull request using DLGCB_LOAD which keeps changes only in acc module. This works for me in 5.0.

Part of the code was present in Diego Nadares' proposed solution too.

PR #1044, opened 2017-03-27
db_postgres insert_update and usrloc params

jchavanton:
This pull request is including two commits

1:  adding `db_postgres` insert_update support 
2:  adding new parameters to `usrloc` that are not changing the default behaviour.

I added both to the same PR to highlight that the usage of insert_update will be immediately available for `usrloc` with `db_postgres` and `db_mysql` (maybe others, the ones not supporting it will simply revert to normal insert)

### DIFF
--- a/src/modules/db_postgres/db_postgres.c
+++ b/src/modules/db_postgres/db_postgres.c
@@ -90,6 +90,7 @@ int db_postgres_bind_api(db_func_t *dbb)
 	dbb->raw_query        = db_postgres_raw_query;
 	dbb->free_result      = db_postgres_free_result;
 	dbb->insert           = db_postgres_insert;
+	dbb->insert_update    = db_postgres_insert_update;
 	dbb->delete           = db_postgres_delete; 
 	dbb->update           = db_postgres_update;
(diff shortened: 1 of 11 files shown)


miconda:
Is the new db_load parameter related to insert_update or somehow unrelated?

There is db_mode that somehow controls the relation with database and it feels it conflicts a bit with the new db_load for some of its values.

jchavanton:
`db_load` is unrelated, it is useful when using `dmq_usrloc` sync because some contacts are duplicated in memory so in such case loading from the DB is not useful.

`db_load` by default will not change anything but if set to 0 will not load registers from the DB on start up.
I did not want to add more `db_mode` because I would have needed  to double them.

I could remove `db_load` from this PR if you want but do you have an alternative ?

miconda:
I would prefer to have two pull requests, each for the two different features. In this way one or the other can be merged separately.

For example, the db_load addition may need to be adjusted to detect conflicts with db_mode, because db_mode=4 is supposed to only load from db.

jchavanton:
ok I will create two and propose a solution for db_mode=4

PR #1045, opened 2017-03-27
Keepalive module - prototype

gbour:
Hi all,   

Here is a prototype of Keepalive module I discussed on the mailing-list.
It can work standalone:
- destinations are set statically as module parameters
- *is_alive()* function allows to check destination state

It can also be used alongside with drouting module (currently only with sort_order 0)
- all destinations defined in drouting module are automatically monitored when drouting *enable_keepalive* option is set to 1
- unavailable destinations are ignored when calling do_routing() function

All comments are welcome

Thanks,
Guillaume

### DIFF
--- a/src/modules/keepalive/Makefile
+++ b/src/modules/keepalive/Makefile
@@ -1,0 +1,18 @@
+#
+# example module makefile
+#
+# 
+# WARNING: do not run this directly, it should be run by the master Makefile
+
+include ../../Makefile.defs
+
+auto_gen=
+NAME=keepalive.so
+
+LIBS=
+
+DEFS+=-DKAMAILIO_MOD_INTERFACE
+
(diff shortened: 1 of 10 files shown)


miconda:
Thanks for contributing!

Can you make a pull request only with the new module as a single commit? It will be easier to review and see if anything needs to be adjusted. Also, add xml docbook files with the documentation of the module, as the other modules have. We are not merging new modules without having the documentation written.

For drouting make another pull request because it may need dedicated review process to get the integration with keepalive.

gbour:
Hi @miconda 

I merge all commits for keepalive module, and added documentation

dragos-oancea:
squash the two commits into one :)

miconda:
Thanks! Merged manually in a single push.

Issue #822, opened 2016-10-14
Enabling DMQ cause high CPU usage


Soapnix:
Hi all!
We trying to use two instances of kamailio on FreeBSD, kamailio was build from master branch.
If I enables DMQ, 2 threads (pids) start utilize CPU, less than minute it's up to 100% (on a both servers).
At that time kamailio successful handle calls, dmq and usrloc replication wokrs fine.
While trying to understand what happens I got this debug info - something happens and K. cycles on shed_yeld() func.
All additional info I put into attachments:
[gdb_output.txt](https://github.com/kamailio/kamailio/files/528881/gdb_output.txt)
[kamailio.log.gz](https://github.com/kamailio/kamailio/files/528879/kamailio.log.gz)
[kamailio.txt](https://github.com/kamailio/kamailio/files/528883/kamailio.txt)
[kdump.txt.gz](https://github.com/kamailio/kamailio/files/528880/kdump.txt.gz)
[ktrace.out.gz](https://github.com/kamailio/kamailio/files/528878/ktrace.out.gz)
[top_output.txt](https://github.com/kamailio/kamailio/files/528882/top_output.txt)

miconda:
Kamailio is using spin locks (for mutexes) and that can cause high cpu on some systems. I am not very familair with dmq internals, but it looks like it uses mutexes to trigger the presence of a new task for its worker processes, meaning that a worker stays in mutex-get operation (spinning), until a new task arrives. Maybe @charlesrchance can shed more light here.

Of course, this high cpu can show up in case of a deadlock, but if you say all works fine whether there is traffic to handle, then is not a deadlock.

The option is to compile to use POSIX semaphores -- the Makefile.defs needs to be changed and all recompiled.

Soapnix:
On another one build I tried to enable POSIX semaphores, but problem still exists.
Compiler change also didn't help.
```
root@kamailio-c1:/var/log# kamailio -V
version: kamailio 5.0.0-dev6 (x86_64/freebsd) 580670
[...]

PR #1046, opened 2017-03-28
dmq: remove worker lock

tombeard:
- fixes a high cpu condition waiting for lock yield on FreeBSD
  reported by GH #822


charlesrchance:
Thanks - this has now been incorporated into 71a88212

PR #1047, opened 2017-03-29
db_postgress: insert_update() with DO UPDATE

jchavanton:


miconda:
Thanks, it will be merged!

PR #1052, opened 2017-04-03
topoh: add additional safety checks

claudiupb:
Hello,

I have added some extra verifications to the th_unmask* functions from topoh. The changes add statements which verify that headers contain expected prefix (including mask ip) before trying to unmask them. It seems that some of the th_unmask* functions only check that the headers exist, and then they directly call th_mask_decode on the data.

If other messages that are not encoded by topoh are processed by this function, th_mask_decode may not return an error, and instead create an unreadable string and this is added to the message. Therefore I added some additional checks to see that the message has the expected prefix (or when possible check directly that the host IP is the mask_ip set), if the header does not have the expected form then unmasking is not attempted. 

Thanks, 
Claudiu Boriga.


miconda:
Thanks!

Issue #1056, opened 2017-04-04
Segmentation fault using http_async_client on kamailio 5.0.0


davyvdm:
### Description

I'm using a few serial http_async_client functions in my config, I found one setup in which kamailio stops with a segmentation fault just doing one call. I can reproduce it in my concrete config, but can't put my finger on the bleeding wound. I do have a core file, which might expose the origin. 

#### Reproduction
```
gdb kamailio core
GNU gdb (Debian 7.7.1+dfsg-5) 7.7.1
Copyright (C) 2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from kamailio...done.
[New LWP 8187]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `kamailio -f /etc/kamailio/kamailio.cfg'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f85580c360a in t_continue () from /usr/lib/x86_64-linux-gnu/kamailio/modules/tm.so
(gdb) bt full
#0  0x00007f85580c360a in t_continue () from /usr/lib/x86_64-linux-gnu/kamailio/modules/tm.so
No symbol table info available.
#1  0x00007f8555ed3848 in async_http_cb () from /usr/lib/x86_64-linux-gnu/kamailio/modules/http_async_client.so
No symbol table info available.
#2  0x00007f8555ecda72 in check_multi_info () from /usr/lib/x86_64-linux-gnu/kamailio/modules/http_async_client.so
No symbol table info available.
#3  0x00007f8555ec56a4 in event_cb () from /usr/lib/x86_64-linux-gnu/kamailio/modules/http_async_client.so
No symbol table info available.
[...]

PR #1063, opened 2017-04-07
tm: don't reset uac reply if in a continue route

grumvalski:
Fix for https://github.com/kamailio/kamailio/issues/1056.
If the reply is continued we need the reply after sending it out, even if it's a final reply.


grumvalski:
The same change is probably also needed on line 1282 of t_reply.c, in case the transaction has failure route set.

miconda:
@grumvalski thanks for tracking this issue. You can push the patch for line 1282, I will do a review when I get the first chance.

grumvalski:
Thanks Daniel, I've updated the pull request.

miconda:
Merging it. Thanks!

PR #1064, opened 2017-04-08
ctl : add rpc buffer_size param

lazedo:


lazedo:
@miconda this is a missed commit for https://github.com/kamailio/kamailio/pull/1015

miconda:
Thanks!

PR #1068, opened 2017-04-11
modules/{acc,cdp}: allows big endian build

tmh1999:
In modules/cdp/diameter.h and modules/acc/diam_message.h, this directive still exists : 

```#error BIG endian detected!!```

Meanwhile in modules/auth_diameter/diameter_msg.h, for example, this directive was removed in 5.0 release. Is there any reason to leave this directive present in these 2 headers?

The build system is Alpine Linux s390x which is big endian.

### DIFF
--- a/src/modules/acc/diam_message.h
+++ b/src/modules/acc/diam_message.h
@@ -74,7 +74,6 @@
 	#define ST_MSG_CODE      0x13010000
 	#define MASK_MSG_CODE    0xffffff00
 #else
-	#error BIG endian detected!!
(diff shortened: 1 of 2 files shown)


miconda:
auth_diameter had a different implementation than cdp. Maybe @ngvoice, @jaybeepee or @vingarzan can comment if cdp is safe for big endian.

miconda:
Going to merge this PR, if there will be issue (as now is not allowed to run on big endian), then it can be reverted or fixed differently.

tmh1999:
Thank you, I will keep an eye on this too.

PR #1071, opened 2017-04-14
Fix memory leak in ldap module.

vitalikvoip:
There was a memory leak in the ldap module that occurred only after calling ldap_result_next().
Has run my synthetic tests and tested that it works fine.


miconda:
Thanks, I am going to merge this PR manually, because the commit message needs to follow the guidelines from:

For the future, use that format in order to be able to merge directly.

vitalikvoip:
Thanks for applying it.
Will be careful with commit messages next time.

PR #1076, opened 2017-04-18
core: dead code and ser string removal

mslehto:
While tracking 'ser' string (after incorrect use of force_send_socket script function), also found some disabled code and file nearby.

types.h introduced in 4bd1673d3decc16db0081efafaf6a4b97c58d737, process_bm_t usage removed in a76545c8aec9e9dc88facf2691e3841b4c89051d both during year 2002.


miconda:
Thanks!

PR #1078, opened 2017-04-19
db_text: change field separator from ":" to "|"

lbalaceanu:
- changed separator because if IPv6 addresses were to be saved in a db_text file, their format would upset the parser

I think that 4.4 would also benefit by back-porting this. Agree?


lazedo:
@lbalaceanu dbtext escapes the `:`,  i don't see the need for this and it will break existing deployments. you can however put the delimiter in a module parameter. thoughts ?

miconda:
I agree with @lazedo -- this should be a modparam if you prefer to use a different separator. There are some modules using `|` in some of their columns (iirc, drouting), but at the end any character can appear in the values and they must be escaped.

I understand that one could prefer a different separator to make it more convenient to work directly with the dbtext files, but this has to be configurable and still have `:` as default separator because there are many deployments out there using existing separator.

lbalaceanu:
Hi guys,

You are right, I will make the parameter configurable.

Thanks

lbalaceanu:
Updated code in the merge branch. If this also looks good to you I will close the merge.

miconda:
Thanks! It is ok to merge.

PR #1081, opened 2017-04-20
topos: fix incorrect contact in 183 and 200 OK responces

sergey-vb:
When using topos, Kamailio does not set correct contact in 183 progres and 200 OK.
But for INVITE and Bye contacts are correct.

This modification seems fix it.


miconda:
Thanks!

PR #1083, opened 2017-04-22
kazoo: Fixed connect to default vhost when used rabbitmq-c 0.8.0

sergey-safarov:


modules/kazoo/kz_amqp.c
```diff
@@ -684,8 +684,6 @@ int kz_amqp_add_connection(modparam_t type, void* val)
 
     if(newConn->info.vhost == NULL) {
     	newConn->info.vhost = KZ_URL_ROOT;
-    } else if(newConn->info.vhost[0] == '/' && strlen(newConn->info.vhost) == 1) { // bug in amqp_parse_url ?
-    	newConn->info.vhost++;
```
lazedo: this is needed for 0.5.2

pkg/kamailio/centos/7/kamailio.spec
```diff
@@ -214,8 +214,8 @@ json string handling and RPC modules for Kamailio.
 %package    kazoo
 Summary:    Kazoo middle layer connector support for Kamailio
 Group:      System Environment/Daemons
-Requires:   libuuid, rabbitmq-c, json-c, libevent, kamailio = %ver
-BuildRequires:  libuuid-devel, rabbitmq-c-devel, json-c-devel, libevent-devel
+Requires:   libuuid, librabbitmq >= 0.8.0, json-c, libevent, kamailio = %ver
+BuildRequires:  libuuid-devel, librabbitmq-devel >= 0.8.0, json-c-devel, libevent-devel
```
lazedo: you can build with 0.5.2

modules/kazoo/kz_amqp.c
sergey-safarov: But this prevents to use rabbitmq-c lib 0.8.0
With this strings kazoo module cannot connect to rabbitmq-server

pkg/kamailio/centos/7/kamailio.spec
sergey-safarov: Module can be build but later cannot be used if used rabbitmq-c lib 0.8.0
This PR is fix it

modules/kazoo/kz_amqp.c
sergey-safarov: I will rewrite this like it is done on commit https://github.com/kamailio/kamailio/commit/22e33ba64c3aad3b50e9ba649c5fa736462a30b4

lazedo:
@sergey-safarov i would prefer to backport the related commit in master. i intent to create a PR to do that for this commit commit and a lot others if that's ok with you.

sergey-safarov:
@lazedo this issue already fixed in master https://github.com/kamailio/kamailio/blob/master/src/modules/kazoo/kz_amqp.c#L92-L102

I see this PR not equal implementation in master branch.
Is required to update this PR?

sergey-safarov:
@lazedo i goes to close this PR.
Please backport required features.

PR #1084, opened 2017-04-22
kazoo: Have been declared rabbitmq-c dependency

sergey-safarov:


lazedo:
you can build with 0.5.2 too and code is handling this

sergey-safarov:
@lazedo try compile current master with rabbitmq-c lib 0.8.0 and try connect to rabbitmq server default vhost.
It is not possible.

sergey-safarov:
@lazedo think i not correct 
It may be fixed in 22e33ba64c3aad3b50e9ba649c5fa736462a30b4

PR #1089, opened 2017-04-22
pkg/kamailio: Fixed systemd unit failure on server boot

sergey-safarov:


amessina:
Same as #1090?

I'm not sure you need After=network.target and network-online.target, but based on https://fedoraproject.org/wiki/Networking/Ideas/ServiceOrdering, you will want Wants=network-online.target

sergey-safarov:
Please look https://github.com/kamailio/kamailio/pull/1104/files
Think IP_FREEBIND is preferred implementation.

miconda:
IP_FREEBIND is not going to be by default on, it is not present in many operating systems but also can expose to service failure due to unnoticed typos in ip addresses.

@amessina - are you ok merging this? I am not a centos user myself.

amessina:
@miconda @sergey-safarov, I'm also not a CentOS user, but a Fedora user :)  Still, according to https://fedoraproject.org/wiki/Networking/Ideas/ServiceOrdering, you need `Wants=network-online.target` since it is not enabled by default on RedHat based systems.  Other than that, sure.  The kamailio.service file I package is https://messinet.com/rpms/browser/kamailio/kamailio.service

sergey-safarov:
@amessina updated PR
- added "Wants"
- added "Documentation"

amessina:
I think that looks good.

miconda:
Merging, thanks!

PR #1090, opened 2017-04-22
pkg/kamailio: Fixed systemd unit failure on server boot

sergey-safarov:


amessina:
I'm not sure you need After=network.target and network-online.target, but based on https://fedoraproject.org/wiki/Networking/Ideas/ServiceOrdering, you will want Wants=network-online.target

sergey-safarov:
Please look https://github.com/kamailio/kamailio/pull/1104/files
Think IP_FREEBIND is preferred implementation.

amessina:
@sergey-safarov -- agree that I'd like Kamailio to have a way to bind to :::5060/udp, :::5061/tcp/tls for example.

That being said, I'm not sure of the Kamailio internals.  Just binding with IP_FREEBIND might not mean that functions within Kamailio detect addresses properly, etc.

miconda:
Closing this one, it looks it requires the backport of #1089 in order to be coherent.

PR #1095, opened 2017-04-27
pkg/debian: Fix installation issue by enabling mpath line in example config

SipSeb:
When installing a new Kamailio from debian packages, dpkg will fail because
Kamailio can't load any modules, resulting in a broken apt state. This is because
the "mpath" line in the example config is commented out. Enabling this line fixes
the install problem.
Tested on Debian Jessie.

### DIFF
--- a/etc/kamailio.cfg
+++ b/etc/kamailio.cfg
@@ -215,7 +215,7 @@ voicemail.srv_port = "5060" desc "VoiceMail Port"
 ####### Modules Section ########
 
 /* set paths to location of modules */
-# mpath="/usr/local/lib/kamailio/modules/"
+mpath="/usr/local/lib/kamailio/modules/"
 
 #!ifdef WITH_MYSQL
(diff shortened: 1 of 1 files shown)


miconda:
I think the problem is in other part -- the build parameters must be adjusted to set internally the proper path to modules. Explicitly setting the path in config proved to be a bit more dangerous when having more than one installation, but also because there are parts of code relying on those paths, especially for default config path -- if the build system is not setting those defines properly, there can be other side effects in long term.

If one runs:

```
kamailio -I
```

it should print the right path to modules and kamailio tries to load from there. If there is a mismatch from what `kamailio -I` prints and where the modules were installed, then it has to be fixed.

As a side note for the future, prefix the commit message with the component that is affected, in this case being kamailio.cfg, so it should look like:

```
kamailio.cfg: short commit description here
```
I am closing this one and further discussions should be on the issue you opened (#1096).

PR #1097, opened 2017-04-27
topos: fix contact for BYE before 200 OK

sergey-vb:
Currently PRACK and BYE does not handling properly, before 200 OK, because topos_d.b_contact field are filling only after receiving 200 OK.
PRACK must be send to Contact from 183 Progress response.
BYE must be send to Contact from 180 Ringing or 183 Progress if 200 OK was not received before.
So I think that we must filled topos_d.b_contact field after receiving 180 Ringing and 183 Progress.

To test this you can try to terminate call after 180 Ringing or 183 Progres before 200 OK.


miconda:
I started implementing it per-branch, can you test as described in:

   * https://github.com/kamailio/kamailio/issues/1167#issuecomment-314747014

miconda:
Closing as PRACK handling was tracked via #1167. Will check soon for BYE.

PR #1101, opened 2017-04-27
core: Don't terminate on harmless SIGCHLD.

codyherzog:
If it appears that no child process has stopped, then do not terminate
on SIGCHLD. Certain modules like app_python can run external scripts
which cause child processes to be started and stopped. That can result
in SIGCHLD being received even though there is no real problem.
Therefore, we do not terminate Kamailio unless we can find the child
process which has stopped.


miconda:
Thanks!

PR #1102, opened 2017-04-28
Fix redis pipeline

claudiupb:
Hello, I have found some bugs with the original commit that adds redis_pipe_cmd. If a command to the server fails, this may leave the redisContext in an invalid state, and this means that a reconnect to the server needs to be performed. 

The problem is that when reconnecting,  a new context is created, and all the contents of old  redisContext is lost, including the hiredis output buffer (obuf) which stored the piped commands. This means that all the commands are lost, and when a second retry is made this contains no commands and may fail, leaving the redisContext in an invalid state. This can in some cases break the REDIS pipelining feature, and cause all commands sent using redis_pipe_cmd()/redis_execute() to fail from that point onwards.

The solution is to make sure that the connection is active when adding commands to the hiredis output buffer. For this, commands are stored in a separate buffer, and are only added to the redisContext when actually sending the command (when calling redis_execute). This way it is possible to  see if any errors exist by checking the err member of the redisContext structure before appending the commands, and also commands are not lost since they are stored separately and if reconnecting the new commands can be appended to the output buffer of the new redisContext.

Commands are stored in buffers using the sds format hiredis offers (by calling redisFormatCommand  function offered by hiredis). This way they can be directly appended to the output buffer using redisAppendFormattedCommand. 

There is a problem on older hiredis versions (older than 0.12) because redisAppendFormattedCommand was not exported to the library API, so for older hiredis version the buffers are appended in our code. 
We tested the code using hiredis library version 0.10.1 and 0.12.1 and there is no problem with this approach.

In the second commit we fixed a memory leak when re-using reply ids in the same pipeline command sequence occurs.

### DIFF
--- a/src/modules/ndb_redis/redis_client.c
+++ b/src/modules/ndb_redis/redis_client.c
@@ -50,6 +50,15 @@ extern int redis_connect_timeout_param;
 extern int redis_cmd_timeout_param;
 extern int redis_cluster_param;
 
+/* backwards compatibility with hiredis < 0.12 */
+#if (HIREDIS_MAJOR == 0) && (HIREDIS_MINOR < 12)
+typedef char *sds;
+sds sdscatlen(sds s, const void *t, size_t len);
+int redis_append_formatted_command(redisContext *c, const char *cmd, size_t len);
+#else
+#define redis_append_formatted_command redisAppendFormattedCommand
+#endif
+
 /**
  *
  */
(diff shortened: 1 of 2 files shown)


miconda:
Should this be ignored or was closed by mistake?

lbalaceanu:
Hello Daniel, we actually found out we want to bring two other changes/ fixes to the module and thought it would be better to bring them all at once. The fixes here are valid though.

PR #1103, opened 2017-04-28
rtpengine: Delete flags & delete handling improvement

hdikme:
-Extending rtpengine_delete() in order to enable specifying Call-ID, To-Tag and From-Tag for timed out dialogs.
-Matching the calls to rtpengine machine correctly.

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -114,7 +114,7 @@ enum {
 #define	CPORT					"22222"
 
 struct ng_flags_parse {
-	int via, to, packetize, transport;
+	int via, to, packetize, transport, c_id, from;
 	bencode_item_t *dict, *flags, *direction, *replace, *rtcp_mux;
 };
 
@@ -1784,9 +1784,7 @@ static int parse_flags(struct ng_flags_parse *ng_flags, struct sip_msg *msg, enu
 			case 6:
 				if (str_eq(&key, "to-tag")) {
 					ng_flags->to = 1;
-					if (val.s && val.len > 0)
-						bencode_dictionary_str_add_str(ng_flags->dict,&key,&val);
-					goto next;
+					goto generic;
 				}
 				break;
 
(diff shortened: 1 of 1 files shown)


src/modules/rtpengine/rtpengine.c
```diff
@@ -1784,6 +1784,8 @@ static int parse_flags(struct ng_flags_parse *ng_flags, struct sip_msg *msg, enu
 			case 6:
 				if (str_eq(&key, "to-tag")) {
 					ng_flags->to = 1;
+					if (val.s && val.len > 0)
```
rfuchs: If a value is given, you could simply leave `ng_flags->to` alone (as zero) and `goto generic`

src/modules/rtpengine/rtpengine.c
```diff
@@ -1967,8 +1970,13 @@ static bencode_item_t *rtpp_function_call(bencode_buffer_t *bencbuf, struct sip_
 	if (ng_flags.rtcp_mux && ng_flags.rtcp_mux->child)
 		bencode_dictionary_add(ng_flags.dict, "rtcp-mux", ng_flags.rtcp_mux);
 
-	bencode_dictionary_add_str(ng_flags.dict, "call-id", &callid);
-
+        temp.s = NULL;
+        if (!bencode_dictionary_get_str(ng_flags.dict, "call-id", &temp))
```
rfuchs: You can get rid of the temp variable and the case distinction below by simply retrieving the given call ID into `callid`. Otherwise, at the very least please rename `temp` to something more meaningful, as there's quite a long distance between this and the place where it's actually being used.

Generally though, instead of doing a `_get_str` here, I would prefer `parse_flags` to take care of this and update a flag in `ng_flags` to signal that a call ID is already present. (Reason is that a dictionary lookup for a constructed dictionary reverts to a linear search, and I cringe at linear searches. :smile: ) The same applies to the fromtag/totag handling below, which can be solved in a similar way.

hdikme:
Thank you, i will make the changes by taking your comments into account.

PR #1104, opened 2017-04-29
src/core: Added IP_FREEBIND flag durring socket initialization

sergey-safarov:
Pull request allow start daemon in case not initialized network.
see https://github.com/proftpd/proftpd/issues/221

### DIFF
--- a/src/core/tcp_main.c
+++ b/src/core/tcp_main.c
@@ -2782,6 +2782,13 @@ int tcp_init(struct socket_info* sock_info)
 			/* continue since this is not critical */
 		}
 	}
+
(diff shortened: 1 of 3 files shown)


miconda:
I will merge it, but I will enclose it in defines to be sure it is enabled in the kernel, like:

```
#if defined(IP_FREEBIND)
...
```

Then I will add a config parameter to turn it on, because this can create problems when having typos in `listen` parameters, so kamailio starts but doesn't listen on proper ip address. People should be aware when enabling this feature that they have to be very careful with listen ip addresses.

miconda:
The core parameter `ip_free_bind` was introduced, by default is disabled (0).

PR #1106, opened 2017-04-29
pkg/kamailio/centos/7: Packaged sipcapture module

sergey-safarov:


miconda:
What this is actually doing? I see it removes the sipcapture module from the old packages and then creates a new packages as well as refers a sipcapture specific configuration file.

Can you describe the purpose and what is expected to happen with the new package, how is it supposed to be used?

sergey-safarov:
Purpose of this PR - simplify sipcapture installation and configuration of sipcapture daemon.
At present time to configure sipcapture required to create systemd unit file and install config file.
After this PR, sipcapture daemon will be installed using "yum install"

miconda:
I am not sure if this is the right approach, because one that wants to do an update or work on a manual installation due to other constrains in its environment, cannot get the sipcapture module without the extras you put it.

I would suggest to let sipcapture module installed as it was and make another package to install the extras you want to run a standalone sipcapture daemon. The package can be called like `kamailio-sipcapture-daemon-config` or similar.

sergey-safarov:
@miconda PR is updated.

miconda:
Back to this -- Kamailio World and traveling kept me very busy lately.

Is it really useful to have a package only for sipcapture module? I think it can stay in the main kamailio package, like it was, because that it is anyhow a dependency for the new package. So there should be only one new package comparing with what is so far, respectively the one that installs the specific configs/scripts for running sipcapture node.

sergey-safarov:
This module have small usage on current kamailio installation and needed in not many cases.
Also at present time about 50 modules have dedicated packages.
Thinkin when this module will have dedicated package, then this will be correct.

miconda:
There are companies that already installed kamailio and then need to just do upgrade. I don't see a reason to split sipcapture in its own package, it depends on kamailio core and other modules. It adds to maintenance overhead without any benefit, including a whole bunch of conflicts rules for upgrades. In many distros, when adding a new package requires a complete review process for license, etc... might not be the case here, but it is better to be coherent with packaging across distros.

I am fine to add the packages that deploys the config for sipcapture daemon, but the module is better in the main package, like so far.

sergey-safarov:
@miconda requested changes done

miconda:
Can you store the kamailio sipcapture config inside the folder with spec files or in misc/examples? Nothing new should be placed in etc/ folder, it is planned to clean it up and let there only mandatory config files, as suggested by distros and move other config examples in a docs/share folder.

sergey-safarov:
@miconda when your ready to accept PR, i will bump all commits to one.
Is required to install config file to folder like "/usr/share/kamailio/example" too?

sergey-safarov:
Daniel, let me know if required to move config file to subfolder.

miconda:
@sergey-safarov - is this patch still actual after your OBS packaging work?

sergey-safarov:
Purpose of this patch simplify homer/sipcapture installation.
OBS changes not related to this PR.

miconda:
Merged it manually with the commit referenced above. It has the config file relocated to `misc/examples/pkg/` and the install is done by `make install-cfg-pkg` (which was added in `src/Makefile` and included in spec build process).

PR #1111, opened 2017-05-02
registrar: fix xavp_rcd memory consumption

lazedo:
xavp_rcd keeps accumulating values leading to shm exaustion

### DIFF
--- a/src/modules/registrar/lookup.c
+++ b/src/modules/registrar/lookup.c
@@ -105,11 +105,13 @@ int lookup_to_dset(struct sip_msg* _m, udomain_t* _d, str* _uri)
  */
 int xavp_rcd_helper(ucontact_t* ptr)
 {
-	sr_xavp_t *xavp=NULL;
+	sr_xavp_t **xavp=NULL;
 	sr_xavp_t *list=NULL;
+	sr_xavp_t *new_xavp=NULL;
 	str xname_ruid = {"ruid", 4};
 	str xname_received = { "received", 8};
 	str xname_contact = { "contact", 7};
+	str xname_expires = {"expires", 7};
 	sr_xval_t xval;
 
 	if(ptr==NULL) return -1;
@@ -117,29 +119,37 @@ int xavp_rcd_helper(ucontact_t* ptr)
 	if(reg_xavp_rcd.s==NULL || reg_xavp_rcd.len<=0) return 0;
 
 	list = xavp_get(&reg_xavp_rcd, NULL);
-	xavp = list;
+	xavp = list ? &list->val.v.xavp : &new_xavp;
 	memset(&xval, 0, sizeof(sr_xval_t));
 	xval.type = SR_XTYPE_STR;
 	xval.v.s = ptr->ruid;
-	xavp_add_value(&xname_ruid, &xval, &xavp);
+	xavp_add_value(&xname_ruid, &xval, xavp);
 
 	if(ptr->received.len > 0) {
 		memset(&xval, 0, sizeof(sr_xval_t));
 		xval.type = SR_XTYPE_STR;
 		xval.v.s = ptr->received;
-		xavp_add_value(&xname_received, &xval, &xavp);
+		xavp_add_value(&xname_received, &xval, xavp);
 	}
 
 	memset(&xval, 0, sizeof(sr_xval_t));
 	xval.type = SR_XTYPE_STR;
 	xval.v.s = ptr->c;
-	xavp_add_value(&xname_contact, &xval, &xavp);
+	xavp_add_value(&xname_contact, &xval, xavp);
+
+	memset(&xval, 0, sizeof(sr_xval_t));
+	xval.type = SR_XTYPE_INT;
+	xval.v.i = (int) (ptr->expires - time(0));
+	xavp_add_value(&xname_expires, &xval, xavp);
 
 	if(list==NULL) {
 		/* no reg_xavp_rcd xavp in root list - add it */
 		xval.type = SR_XTYPE_XAVP;
-		xval.v.xavp = xavp;
-		xavp_add_value(&reg_xavp_rcd, &xval, NULL);
+		xval.v.xavp = *xavp;
+		if(xavp_add_value(&reg_xavp_rcd, &xval, NULL)==NULL) {
+			LM_ERR("cannot add ruid xavp to root list\n");
+			xavp_destroy_list(xavp);
+		}
 	}
 	return 0;
 }
--- a/src/modules/registrar/reply.c
+++ b/src/modules/registrar/reply.c
@@ -173,8 +173,9 @@ int build_contact(sip_msg_t *msg, ucontact_t* c, str *host)
 	unsigned int ahash;
 	unsigned short digit;
 	int mode;
-	sr_xavp_t *xavp=NULL;
+	sr_xavp_t **xavp=NULL;
 	sr_xavp_t *list=NULL;
+	sr_xavp_t *new_xavp=NULL;
 	str xname = {"ruid", 4};
 	str ename = {"expires", 7};
 	sr_xval_t xval;
@@ -213,7 +214,7 @@ int build_contact(sip_msg_t *msg, ucontact_t* c, str *host)
 	if(reg_xavp_rcd.s!=NULL)
 	{
 		list = xavp_get(&reg_xavp_rcd, NULL);
-		xavp = list;
+		xavp = list ? &list->val.v.xavp : &new_xavp;
 	}
 
 	fl = 0;
@@ -336,7 +337,7 @@ int build_contact(sip_msg_t *msg, ucontact_t* c, str *host)
 				xval.type = SR_XTYPE_STR;
 				xval.v.s = c->ruid;
 
-				if(xavp_add_value(&xname, &xval, &xavp)==NULL) {
+				if(xavp_add_value(&xname, &xval, xavp)==NULL) {
 					LM_ERR("cannot add ruid value to xavp\n");
 				}
 				/* Add contact expiry */
@@ -344,7 +345,7 @@ int build_contact(sip_msg_t *msg, ucontact_t* c, str *host)
 				xval.type = SR_XTYPE_INT;
 				xval.v.i = (int)(c->expires - act_time);
 
-				if(xavp_add_value(&ename, &xval, &xavp)==NULL) {
+				if(xavp_add_value(&ename, &xval, xavp)==NULL) {
 					LM_ERR("cannot add expires value to xavp\n");
 				}
 			}
@@ -354,16 +355,16 @@ int build_contact(sip_msg_t *msg, ucontact_t* c, str *host)
 	}
 
 	/* add xavp with details of the record (ruid, ...) */
+
 	if(reg_xavp_rcd.s!=NULL)
 	{
-		if(list==NULL && xavp!=NULL)
+		if(list==NULL && *xavp!=NULL)
 		{
-			/* no reg_xavp_rcd xavp in root list - add it */
 			xval.type = SR_XTYPE_XAVP;
-			xval.v.xavp = xavp;
+			xval.v.xavp = *xavp;
 			if(xavp_add_value(&reg_xavp_rcd, &xval, NULL)==NULL) {
 				LM_ERR("cannot add ruid xavp to root list\n");
-				xavp_destroy_list(&xavp);
+				xavp_destroy_list(xavp);
 			}
 		}
 	}


miconda:
Are not the xavps deleted automatically when the processing of the request is finished? There should not be an explicit delete from the modules, unless strictly necessary for functionality of some feature, but the xavps and avps should not leak when processing normal SIP traffic.

Or are you using them in some event routes?

lazedo:
@miconda it seems so, maybe the problem lies in the delete ? i'm using in script to create a json with the `ruid` and memory is always growing. removing just these parts, memory is back to normal.

```
modparam("registrar", "xavp_rcd", "ulrcd")
...
after successful save
$var(amqp_payload_request) = $_s({"Event-Category" : "directory", ..., "RUID" : "$xavp(ulrcd=>ruid)", ... });
....
```

miconda:
Which memory is growing? SHM, PKG or system memory?

lazedo:
shared memory tracked by
[root@dev-01 kamailio]# kamcmd mod.stats core shm
Module: core
{
*xavp_new_value(94): 4808*
}

only one client registering every 60 sec

miconda:
Can you attach log messages with debug=3 for a registration request?

lazedo:
[kamailio.log.gz](https://github.com/kamailio/kamailio/files/973067/kamailio.log.gz)
used gzip

lazedo:
Hi Daniel,
i isolated all the requests and made sure only registration was being processed.
turns out that the problem must be elsewhere as the memory is now stable.
however, the first registration uses a suspend/continue and in this case the xavp memory is not released.

[kam.log.gz](https://github.com/kamailio/kamailio/files/986341/kam.log.gz)

lazedo:
@miconda sorry, wrong report. the patch was applied, that's why memory is stable. initial registration issues still occurs (xavp not released after t_continue() )

lazedo:
@miconda this is the log without the patch applied.

[kam.log.gz](https://github.com/kamailio/kamailio/files/986380/kam.log.gz)

miconda:
Back on this after Kamailio World event -- so the issue is exposed when t_suspend()/t_continue() is used for REGISTER requests? Are you using these functions in kamailio.cfg or is via kazoo module?

lazedo:
@miconda Hi Daniel, kazoo module does the suspend/continue, but this is only for the initial registration. the 2nd registration doesn't go thru suspend/resume but we're also getting memory leaks here.

miconda:
If there is no suspend/continue, then the issue can be that the registrar module doesn't add the xavp to core list, being left unliked, otherwise the xavps (as well as avps) are destroyed at the end of executing request_route. This would have been revealed so far if there is an issue with overall avps/xavps management.

miconda:
Btw, what version you are using in this case? 5.0 or 4.4?

miconda:
While looking at the t_continue() I discovered that it may overwrite a previous set faked environment, which can happen if the function is used in failure_route, as a result xavps/avps lists can be lost without being freed. With the commit 9ff4dd36ad63f60905a49f8c0e437f997e54ae70, this situation should be solved.

I am not sure this is solving your case, but it would be good if you test with this patch.

I am also going to look at the registrar module to see if I spot something wrong there. Deleting the xavps in the module as proposed by this pull request is not the proper solution, it may be a workaround for some specific case, but not fixing the real cause.

lazedo:
using master (with your latest commits). tried the referenced commit but still have memory leak for 1st request (suspend/continue). the xavp is being created after continue (registrar reply) , it seems receive_msg cleanup is not reached (or is early reached due to suspend) in this case ?
and i agree this is not the recommended fix, its only hiding possible issues.
also, i applied the patch only to reply.c in registrar and it fixes re-register.

miconda:
Is this with the standard kazoo platform kamailio.cfg? Can you point me to the repo with the config at the line where the xavp is created? I may help seeing the config and understanding the use case.

lazedo:
https://github.com/2600hz/kazoo-configs/blob/master/kamailio/registrar-role.cfg

the xavp is not created in the script.
its created by registrar build_contact (reply.c) all cases or 
registrar registered4 (lookup.c) that calls xavp_rcd_helper.

check route `ATTEMPT_AUTHORIZATION`

for re-register, this line https://github.com/2600hz/kazoo-configs/blob/master/kamailio/registrar-role.cfg#L120 check if the device is already registered and skips to save location (no suspend/continue), leak seems to occur in registrar build_contact (reply.c).

lazedo:
@miconda about the first registration...

the `continue` executes in pid `6518` and it never calls `xavp_reset_list()`
registrar:build_contact (reply.c) is executed in this pid

```
 2(6487) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|start|received udp request REGISTER sip:teste.sip.90e9.com
 2(6487) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|source 192.168.16.12:50133 -> 192.168.16.125:8000
 2(6487) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|from sip:user_7z7b6s@teste.sip.90e9.com
 2(6487) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|to sip:user_7z7b6s@teste.sip.90e9.com
 2(6487) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|originated from external sources
 2(6487) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|correcting NATed contact in registration
 2(6487) INFO: <core> [core/xavp.c:461]: xavp_reset_list(): RESET XAVP
32(6517) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|authenticating user_7z7b6s@teste.sip.90e9.com via Kazoo query response
32(6517) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|end|issued auth challenge to failed registration attempt for user_7z7b6s@teste.sip.90e9.com 192.168.16.12:50133
 3(6488) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|start|received udp request REGISTER sip:teste.sip.90e9.com
 3(6488) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|source 192.168.16.12:50133 -> 192.168.16.125:8000
 3(6488) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|from sip:user_7z7b6s@teste.sip.90e9.com
 3(6488) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|to sip:user_7z7b6s@teste.sip.90e9.com
 3(6488) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|originated from external sources
 3(6488) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|correcting NATed contact in registration
 3(6488) INFO: <core> [core/xavp.c:461]: xavp_reset_list(): RESET XAVP
33(6518) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|authenticating user_7z7b6s@teste.sip.90e9.com via Kazoo query response
33(6518) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|log|caching sip credentials for user_7z7b6s@teste.sip.90e9.com
33(6518) INFO: <script>: 79956NjgwZTVmOTQ5Zjc0MjcxMTVjYjNlOTMxNGI5NDkwYjk|end|successful registered with contact uloc-5919b147-941-1 : <sip:user_7z7b6s@192.168.16.12:50133;rinstance=f24eba8fb96940a0>;+sip.instance="<urn:uuid:50e56137-edf1-5c8c-99ab-f5ebf869e00f>";reg-id=1
 6(6491) INFO: <core> [core/xavp.c:461]: xavp_reset_list(): RESET XAVP
 6(6491) INFO: <core> [core/xavp.c:461]: xavp_reset_list(): RESET XAVP
 8(6493) INFO: <core> [core/xavp.c:461]: xavp_reset_list(): RESET XAVP
 8(6493) INFO: <core> [core/xavp.c:461]: xavp_reset_list(): RESET XAVP
```

miconda:
In this case there should be a REGISTER transaction and the xavps should be linked to the transaction structure and destroyed when the transaction is cleaned up.

lazedo:
@miconda more info

it seems that the `faked_env( t, 0, 1);` call (restore original environent) will discard the avps xavps that can be added during processing of the script block. in this use case,  registrar save (which calls reply) is called, which creates the xavps that are not being released.

from t_suspend.c
```
		faked_env( t, faked_req, 1);

		route_type_bk = get_route_type();
		set_route_type(FAILURE_ROUTE);
		/* execute the pre/post -script callbacks based on original route block */
		if (exec_pre_script_cb(faked_req, cb_type)>0) {
			if (run_top_route(route, faked_req, 0)<0)
				LM_ERR("failure inside run_top_route\n");
			exec_post_script_cb(faked_req, cb_type);
		}
		set_route_type(route_type_bk);

		/* TODO: save_msg_lumps should clone the lumps to shm mem */

		/* restore original environment */
		faked_env( t, 0, 1);
```

miconda:
Not really -- during the execution of the route block, the x/avp lists from tm are set in the core, so the new x/avps are stored directly in the transaction. This is used for failure_route as well and any issue should have been exposed for long time. Unless it was the case of creating a faked env inside a faked env that should be fixed by my last commit.

I looked at the kazoo code and I noticed that `kz_amqp_consumer_fire_event()` is using fake sip messages to execute event routes. Are those event routes creating xavps?

miconda:
Anyhow, you can try to execute `xavp_reset_list()` inside kazoo module after t_continue(), and see if  it gets fixed. Then we know where to dig more.

lazedo:
@miconda i reverted every change and added `xavp_reset_list()` after `t_continue()` in kazoo andalso after `kz_amqp_consumer_fire_event()` and leak continues.
is there a way to lookup the memory allocated values to make sure it is coming from where we think it is ?

miconda:
Yes. Start kamailio and do some operations so you expose the leak.

Do a shm status dump:

  * http://www.kamailio.org/docs/modules/5.1.x/modules/corex.html#corex.rpc.shm_status

Then you get details of each allocated chunks -- two pointers

  * start of the memory chunk
  * start of the data inside the memory chunk

With gdb attach to any of the kamailio processes and print the content of the data pointer casted to xavp structure. Do it for several pointers so you see what is inside.

lazedo:
@miconda thanks for the info.
after digging more , i think the issue is not related to `kazoo`. can you try the below script (note that kazoo is not used) ? also maybe fake a transaction to a external auth and set the `$xavp(ulattrs=>extra_info)` on the `continue`block ?
with this configuration, the 1st registration doesn't leak but the re-registrations do leak.

```
...
modparam("usrloc", "xavp_contact", "ulattrs")
modparam("registrar", "xavp_rcd", "ulrcd")
modparam("registrar", "xavp_cfg", "regcfg")
modparam("htable", "htable", "auth_cache=>size=16;autoexpire=7200;")
...

route[REGISTRAR]
{

    $xavp(regcfg=>match_received) = $su;
    if($sht(auth_cache=>$Au) != $null && registered("location", "$rz:$Au", 2, 1) == 1) {
        xlog("L_INFO", "$ci|register|found cached registration for $rz:$Au\n");
            $var(password) = $sht(auth_cache=>$Au);
            route(SAVE_LOCATION);
    }

    if( !is_present_hf("Authorization")) {
        auth_challenge("$fd", "0");
        exit();
    }
    
    $xavp(ulattrs=>extra_info) = '{"test" : true}';
    
    route(SAVE_LOCATION);

}

route[SAVE_LOCATION]
{

    if ($sht(auth_cache=>$Au) == $null) {
        xlog("L_INFO", "$ci|log|caching sip credentials for $Au\n");
    };
    $sht(auth_cache=>$Au) = $var(password);
    $var(save_result) = save("location", "0x04");
    if($var(save_result) == -1) {
        auth_challenge("$fd", "0");
        xlog("L_INFO", "$ci|end|issued auth challenge after failed attempt to save contact for $Au $si:$sp\n");
        exit;
    } else {
        if($var(save_result) == 1) {
            $var(new_reg) = "true";
        } else {
           $var(new_reg) = "false";
        }
    }
    xlog("L_INFO", "$ci|end|save result $var(save_result) with contact $xavp(ulrcd=>ruid) : $ct\n");
}
```

lazedo:
@miconda i think i found it. no leaks on both 1st and 2nd registrations

lazedo:
@miconda so 1st registrations are ok.   i was testing with db_mode = `Write-Through` (sorry about that). with for db_mode = `db_only`, all xavps are cleared after route (with async resume).

miconda:
To summarize: all ok now with patch 3fddccd ?

Was the problem that some xavps were not linked to the root list from the core?

lazedo:
@miconda Hi, yes, all ok with https://github.com/kamailio/kamailio/commit/3fddccd0d7a3f9662699f88d1a9a708e1c1b24bc . the problem was how we obtained the existing `xavp's` and adding new values. this part `xavp = list ? &list->val.v.xavp : &new_xavp;` did the trick, we were adding values to xavp list and not to reg_xavp_rcd xavp.

miconda:
But even when adding to xavp list, they should be freed. Do you still have the xavp_reset_list() in kazoo module?

lazedo:
i didn't used kazoo for testing, but i will be adding a xavp_reset_list() to event processing (not critical).
test with this script with/without the patch.
have sip client re-register every 30 and you'll see `shm` memory consumption growing (`xavp_new_value`)
```
...
modparam("usrloc", "xavp_contact", "ulattrs")
modparam("registrar", "xavp_rcd", "ulrcd")
modparam("registrar", "xavp_cfg", "regcfg")
modparam("htable", "htable", "auth_cache=>size=16;autoexpire=7200;")
modparam("async", "workers", 2)

...

route
{
...
    if (is_method("REGISTER")) {
        route(REGISTRAR);
        exit();
    }

...
}

route[REGISTRAR]
{
    $xavp(regcfg=>match_received) = $su;
    if($sht(auth_cache=>$Au) != $null && registered("location", "$rz:$Au", 2, 1) == 1) {
        xlog("L_INFO", "$ci|register|found cached registration for $rz:$Au\n");
            $var(password) = $sht(auth_cache=>$Au);
            route(SAVE_LOCATION);
    }

    if( !is_present_hf("Authorization")) {
        auth_challenge("$fd", "0");
        exit();
    }
    
    async_route("MY_RESUME", "1");    

}

route[MY_RESUME]
{
    $xavp(ulattrs=>channel_vars) = '{"channel-vars" : true}';
    $xavp(ulattrs=>other_vars) = '{"other-vars" : false}';
    route(SAVE_LOCATION);
}

route[SAVE_LOCATION]
{
    if ($sht(auth_cache=>$Au) == $null) {
        xlog("L_INFO", "$ci|log|caching sip credentials for $Au\n");
    };
    $sht(auth_cache=>$Au) = $var(password);
    $var(save_result) = save("location", "0x04");
    if($var(save_result) == -1) {
        auth_challenge("$fd", "0");
        xlog("L_INFO", "$ci|end|issued auth challenge after failed attempt to save contact for $Au $si:$sp\n");
        exit;
    } else {
        if($var(save_result) == 1) {
            $var(new_reg) = "true";
        } else {
           $var(new_reg) = "false";
        }
    }
   xlog("L_INFO", "$ci|end|ulattrs $xavp(ulattrs=>channel_vars) , $xavp(ulattrs=>other_vars)\n");
    xlog("L_INFO", "$ci|end|save result $var(save_result) with with expires ($xavp(ulrcd=>expires)) contact $xavp(ulrcd=>ruid) : $ct\n");
}
```

miconda:
Pushed the patch split in two, as the expires part might not be needed to be backported for older releases.

If you tested and are confident in the results, then you can push any of the patches in stable branches.

lazedo:
@miconda Thanks Daniel. i've asked for more people to confirm the fix and i'm also deploying this to our sandbox environment. will cherry-pick the 1st to 4.4/5.0 and the 2nd to 5.0 if all goes well. so far my environment is stable.

lazedo:
@miconda i cherry-picked the 2 commits to 5.0 branch and the first to 4.4 branch.

PR #1112, opened 2017-05-02
Redis remove redis_execute() command with no arguments

claudiupb:
Hello, 

We realized that redis_execute should return a negative value in case of failure (just like redis_cmd), instead of always returning 1. This introduced a problem with the behavior when calling redis_execute without giving a server name. Since this function will loop through all the servers and execute the pipelined commands for each one, in case of failure the user cannot determine for what server the error occurred, by only returning a negative value. This is why we decided to remove the possibility to call redis_execute with no arguments, and have it loop through all defined servers. The user should only call redis_execute(srv_name) and check the return value for this.


miconda:
Thanks!

PR #1116, opened 2017-05-03
ndb_redis: add disable server on failure feature

claudiupb:
Hello, 

We added a new feature to the ndb_redis module, similar to how the rtpengine_disable_tout parameter of the rtpengine module works. 

If a number of consecutive commands to a redis server fail, that server is temporarily disabled. This is configurable by parameters: "allowed_timeouts" and "disable_time". 
Disabling a server means that for "disable_time" seconds any command to that sever will not send the command to the server, but an error is returned instead. 

For example if allowed_timeouts is configured to 0 and disable_time to 30, the first time redis_cmd or redis_execute functions fail the server will be disabled for 30 seconds. 

The server information is retained by each Kamailio process individually, so this means that disabling a server is done by that process only and not globally.

### DIFF
--- a/src/modules/ndb_redis/doc/ndb_redis_admin.xml
+++ b/src/modules/ndb_redis/doc/ndb_redis_admin.xml
@@ -181,6 +181,64 @@ modparam("ndb_redis", "cluster", 1)
 			</programlisting>
 		</example>
 	</section>
+	<section id="ndb_redis.p.allowed_timeouts">
+		<title><varname>allowed_timeouts</varname> (integer)</title>
+		<para>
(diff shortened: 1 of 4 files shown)


miconda:
I am ok to merge this PR, but you have to follow up with another patch to rename the global variables for the module parameters added in this PR to have `redis_` as a prefix, like for the other existing parameters. It is good in long term as it avoids collisions with other variables that could have same name with these common words.

claudiupb:
I have changed the names of the variables adding the `redis_` prefix and `_param` suffix

miconda:
Thanks!

PR #1121, opened 2017-05-05
htable: add new functions sht_rm_name_re_pv and sht_rm_value_pv to support variables

eschmidbauer:
- cleanup existing sht_rm_name_re and sht_rm_value_re functions
 - update documention

### DIFF
--- a/src/modules/htable/doc/htable.xml
+++ b/src/modules/htable/doc/htable.xml
@@ -39,6 +39,11 @@
 		<surname>Sas</surname>
 		<email>osas@voipembedded.com</email>
 	    </editor>
+	    <editor>
+		<firstname>Emmanuel</firstname>
+		<surname>Schmidbauer</surname>
+		<email>eschmidbauer@gmail.com</email>
+	    </editor>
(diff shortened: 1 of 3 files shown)


miconda:
Can you provide more details about this patch? The description in the commit message and in the pull request is not saying that much.

eschmidbauer:
problem scenario is: 
need to clear htable cache dynamically (using an event-route + nsq message)
but `sht_rm_name_re` and `sht_rm_value_re` do not support variables as parameters.
solution in patch:
allow variables by adding a 2nd parameter to each function
if called with one parameter like `sht_rm_value_re("ha=>.*");` 
traditional behavior is executed.
if passing a 2nd parameter like
 `sht_rm_value_re("$var(htable)", ".*$var(match)"); `
new behavior is executed which allows variables.
I hope that makes sense.

miconda:
I pushed a different implementation, with new functions, that allow extending it for other types of matching. For now implements the regexp - 0c69e7bc4486ff6b10da6502b72c1805ac6e5f83 .

PR #1125, opened 2017-05-09
ndb_redis: add flush_db_on_reconnect parameter

claudiupb:
Hello,
We added a new parameter to ndb_redis module. If set to 1 a "FLUSHALL" command is issued to the REDIS server when reconnecting to a server after a failed command. This way for each failed command the database is flushed. 

This is useful when a connection to the REDIS server may become unstable and some commands are issued but the replies are not received (it may happen that the command has reached the REDIS server but the reply was lost). If for example we increment and decrement counters this may leave the db in an invalid state.

This can especially happen if the allowed_timeouts/disable_time parameters are configured, since the first commands may be issued and the next ones are never tried.

### DIFF
--- a/src/modules/ndb_redis/doc/ndb_redis_admin.xml
+++ b/src/modules/ndb_redis/doc/ndb_redis_admin.xml
@@ -239,6 +239,38 @@ modparam("ndb_redis", "disable_time", 30)
 			</programlisting>
 		</example>
 	</section>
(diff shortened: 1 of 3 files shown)


miconda:
Thanks, I am going to merge it. But I will remove the db_ from the name of its new parameter be cause it is a bit long and the operations are expected to be specific to redis interaction, no need to make it very explicit.

PR #1128, opened 2017-05-12
rabbitmq: don't create reply-to queue on publish

SipSeb:
When using the rabbitmq_publish function, there is no need to create
a reply-to queue, because it will never be read. And since there is
never a real consumer, the queue will never be deleted. This
will eventually cloak up the RabbitMQ server with millions of
generic reply queues.
This bug has been fixed in master already, so this is basically a
backport.


miconda:
Thanks!

PR #1129, opened 2017-05-12
rabbitmq: create reply-to queue w/ exclusive param

SipSeb:
RabbitMQ docs say, temporary reply-to queues should be created
with an "exclusive" parameter. This allows only the current
connection to access the queue, and when the connection ends,
the queue will automatically be deleted.


miconda:
@ngvoice, @smititelu -- are you ok to merge this PR?

miconda:
Merging the PR -- if anyone has adjustments to do, then push new commits.

smititelu:
Awesome, thanks! 👍

PR #1133, opened 2017-05-19
db_sqlite: new param to open some database in read-only

jchavanton:
Using a `db_param_list` `clist` we can preset `SQLITE_OPEN_READONLY` on specific database file, this can be reused for more sqlite specific settings since we probably do not want extand `km_dbase` for db engine specific settings.

This is required when another process is writing to the db

Example: 
``modparam("db_sqlite","db_set_readonly","/var/mydb.sqlite")``


miconda:
Merging it. For the future have in mind to declare the variables at the beginning of the function or blocks, to be compatible with old C standard versions and compilers, as kamailio is used on some embedded devices with strict compilers.

PR #1139, opened 2017-05-27
dmq: update output format of dmq.list_nodes

zhangst:
change status from int to readable text


charlesrchance:
Thanks - merging now.

PR #1140, opened 2017-05-29
tsilo: fix deadlock in ts_append()

vitalikvoip:
Copy ruri parameter into pkg mem before calling ts_append().
Value saved in a static buffer from pv_get_buffer() can be corrupted if
a new branch adds an additional SIP header in its branch_route


grumvalski:
Thanks :)

PR #1142, opened 2017-05-30
htable: fix crash if deleted entry is reused

lazedo:
resets 'first' when esize == 0

### DIFF
--- a/src/modules/htable/ht_var.c
+++ b/src/modules/htable/ht_var.c
@@ -48,16 +48,16 @@ int pv_get_ht_cell(struct sip_msg *msg,  pv_param_t *param,
 		return -1;
 	}
 	htc = ht_cell_pkg_copy(hpv->ht, &htname, _htc_local);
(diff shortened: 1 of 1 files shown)


lazedo:
these crashes were caught by running with `-x tlsf`

miconda:
@lazedo - I am not sure be46742 is actually effective and fixes anything -- `first` should become NULL if the last item in the slot (entry) is removed, because it->next should be NULL at that moment.

Can you describe what was the issue you faced? Maybe the problem is in another part and it->next is wrong, but that will have bad effects in other cases.

The commit 18ba95a seems ok.

lazedo:
@miconda thanks for the comment. i  think you're right. i re-tested with only the second commit and it seems to be working.

miconda:
@lazedo - ok, thanks for the feedback and fix! You can merge that commit and backport as needed to stable branches.

lazedo:
@miconda backported to 4.4 and 5.0

PR #1144, opened 2017-06-02
core: parser reset_path_vector() remove check for msg->msg_flags

linuxmaniac:
* tm now uses shared memory for fake_req() so faked_req->path_vec
was leaking


miconda:
Thanks!

PR #1146, opened 2017-06-02
benchmark: b/f avoid undefined bytes when int and long int sizes mism…

vhernando:
…atch.

```
 void bm_rpc_enable_global(rpc_t* rpc, void* ctx)
 {
	long int v1;
 	if(rpc->scan(ctx, "d", (int*)(&v1))<1) {
```

rpc->scan does not fill all the bytes of v1 if long int and int sizes differ,
then checks like this usually fail:

```
if ((v1 < -1) || (v1 > 1)) {
		rpc->fault(ctx, 500, "Invalid Parameter Value");
		return;
}
```


miconda:
Thanks! It is going to be merged. Then we can get rid of the casts to `int*` as well.

PR #1147, opened 2017-06-02
ndb_redis: enhance access to REDIS replies

claudiupb:
Hello, 

I have enhanced the way REDIS replied are accessed in the Kamailio config, to be able to access arrays that are members of other arrays. In the original code, we could access members of an array like this: $redis(reply=>value[0]). 
But if that member was itself an array it was impossible to access it, as accessing an array directly returns (null). This case can happen if we use redis transactions with, and all commands are given between MULTI and EXEC commands. Then the EXEC command returns an array with the responses of the individual commands, so if any command returns an array, then we have an array within that array.

I have modified the syntax of the redis replies in the config, to be able to add other keys after value[n] keys. This way we can access members of the array within the first array.
For example after the following sequence:
```
redis_cmd("srvN", "MULTI", "r1");
redis_cmd("srvN", "SMEMBERS foo", "r2");
redis_cmd("srvN", "EXEC", "reply")

```
The PV to access the response will look like this

- to access the first elemenent of SMEMBERS:

   `$redis(reply=>value[0]=>value[0])` 

- to access the second element of SMEMBERS:

    `$redis(reply=>value[0]=>value[1])`

- to see how many elements SMEMBERS returned:

   `$redis(reply=>value[0]=>size)`

- to see the type the first element returned by SMEMBERS:

    `$redis(reply=>value[0]=>type[0])`

### DIFF
--- a/src/modules/ndb_redis/doc/ndb_redis_admin.xml
+++ b/src/modules/ndb_redis/doc/ndb_redis_admin.xml
@@ -319,6 +319,12 @@ modparam("ndb_redis", "flush_on_reconnect", 1)
 			</listitem>
 			</itemizedlist>
 		</para>
+		<para>
+			In case one of the members of the array is also an array (for example calling SMEMBERS
+			in a MULTI/EXEC transaction), the members of the array can be accessed by adding any of
+			the above keys, after a value[n] key. The first value[n] references the element in the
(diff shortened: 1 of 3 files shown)


miconda:
Thanks!

PR #1150, opened 2017-06-09
modules/topoh: added lookup functionality for destination IP

adubovikov:


miconda:
What is exactly the purpose of this new functionality? To allow/deny traffic from those IP addresses or for doing/skipping topology hiding?

I am somehow not very pleased with importing a lot of code similar to what permissions does in regards to address matching. I would rather do another approach that might bring more flexibility as well, respectively to run an event route where one can run functions from other modules and set some internal variable in order to control what topoh should do after execution of the event block. This should not be something big in terms of new code, I can do it in the next days if we agree on some concept that fits your needs.

adubovikov:
Yes, exactly this way, just to exclude or include destination IP for topology hiding. 
I am not sure that there is a way to do it via internal flags, because the event socket for SREV_NET_DATA_OUT will be called from retransmission  function in the tm module and in this case you don't have sip_msg structure anymore.

miconda:
We have the source code, we can change if something helps substantially -- currently, I see two options that can be controlled via a mod param (or just choose one that it is considered better):

  * give a fake request, like it is done with other event_route blocks (e.g., htable:mod-init) -- this way should be very fast
  * parse the outgoing buffer -- might be slightly slower, as it will parse even when not doing topoh

adubovikov:
so, from topoh we can call topoh-route in kamailio.cfg , there we can use allow_address from the permissions module and base on the result we can reject and accept topoh request. Should we just send empty sip_msg structure there ? Just to avoid sip message parsing in case the request will be rejected ?

miconda:
Giving a fake request is very fast, like done at:

  * https://github.com/kamailio/kamailio/blob/master/src/modules/htable/htable.c#L268

It is no parsing apart of the first time. This way is safe, because some functions need the msg structure.

adubovikov:
ok. agree.. this way will be more effective and flexible. Should I do or you wanna ? ;-)

miconda:
I did an implementation to execute an event_route before the topoh processing for messages to be sent out. Documentation will follow soon, meanwhile you can see the comments of the commit 	f7d1d23 . Basically, $sndto(ip) should point to the ip address of the destination and calling `drop` should skip topoh processing. I did no test with the new version, so feedback and improvements are welcome.

amessina:
Can this be used for topos as well as topoh?

adubovikov:
Thank you @miconda , unfortunately there is bug that I couldnt catch.

6 days ago I have implemented same way in my repository:

https://github.com/adubovikov/kamailio/commit/ae15a717e9a5849b4ac295c5864cd6e9b198b6d1

but your and my patches makes same results:

/usr/sbin/kamailio[1260]: ERROR: sl [sl_funcs.c:362]: sl_reply_error(): stateless error reply used: I'm terribly sorry, server error occurred (6/SL)
this is after execution of event route. I have tested/removed some code and found that this happens once you call " fmsg = faked_msg_next();"

also in your patch:

https://github.com/kamailio/kamailio/commit/f7d1d234a42198fc10e40c33e92cff5e8fc1af8e#diff-81d708cd01d1642b044346221062c4dbR522

should be:
if(_th_eventrt_outgoing<0 **&&** keng==NULL) {

any suggestions got to avoid this SL error ?

Wbr,
Alexandr

miconda:
Indeed, that condition should be with &&.

Is the next fake msg null, or what is failing there? Can you see what is the return code for t_relay? I guess the slow error is because of sl_send_error () in kamaulio.cfg. I am traveling, no much testing that I can do.

adubovikov:
no, the fake msg is not null, but once you call the faked_msg_next(), after you will always get "I'm terribly sorry, server error occurred (6/SL)", even if you didn't execute "run_top_route".

I will check a bit more, but the issue it can be with pointer for re-transmission.

miconda:
Might have been the execution of clear_branches() inside next faked msg function, which was resetting some global variables. I just pushed a patch in master, can you give it a try again?

adubovikov:
bingo! :-)

now another issue:

 /usr/sbin/kamailio[1798]: ERROR: topoh [topoh_mod.c:560]: th_execute_event_route(): exit due to 'drop' in event route
 /usr/sbin/kamailio[1798]: : <core> [core/mem/q_malloc.c:469]: qm_free(): BUG: qm_free: bad pointer 0x7fa3461b9000 (out of memory block!) called from core: core/parser/msg_parser.c: reset_new_uri(720) - aborting
CRITICAL: <core> [core/pass_fd.c:277]: receive_fd(): EOF on 16

and I don't have anything in core:

(gdb) bt full
#0  0x00007f69d1059067 in ?? ()
No symbol table info available.
#1  0x00007f69d105a448 in ?? ()
No symbol table info available.
#2  0x0000000000000020 in ?? ()
No symbol table info available.
#3  0x0000000000000000 in ?? ()
No symbol table info available.
(gdb) quit

adubovikov:
https://github.com/kamailio/kamailio/pull/1155

looks like it has been fixed :-)

miconda:
So #1155 is fixing also the introduced bug with bad pointer?

adubovikov:
yep. It works now! :-) thank you

miconda:
Thanks, closing it then!

PR #1153, opened 2017-06-19
db_sqlite: new param to set database journal mode

jchavanton:
Exposing journal mode as a module parameter, one great feature is to be able to use WAL mode and share a database in read / write mode with facing any locking.

https://sqlite.org/wal.html


miconda:
Thanks!

PR #1154, opened 2017-06-20
Fix compiler warnings for unused static symbols

coreyfarrell:


miconda:
Thanks for the fixes!

I merged the PR manually, mainly to format the commit message (as per https://github.com/kamailio/kamailio/blob/master/.github/CONTRIBUTING.md#commit-message-format), breaking it per modules in order to be safer when backporting to avoid conflicts.

PR #1158, opened 2017-06-20
pv: add received ip/port pvars for advertised ip/port

lazedo:
adds $RAi , $RAp
if advertise is not set, defaults to reveived ip/port


miconda:
Thanks!

PR #1162, opened 2017-06-22
app_lua: add "NOTICE" loglevel to sr.log

SipSeb:
The "notice" loglevel is missing in app_lua's sr.log function. This patch adds it.

### DIFF
--- a/src/modules/app_lua/app_lua_sr.c
+++ b/src/modules/app_lua/app_lua_sr.c
@@ -96,6 +96,8 @@ static int lua_sr_log (lua_State *L)
 				LM_DBG("%s", txt);
 			} else if(strcasecmp(level, "info")==0) {
 				LM_INFO("%s", txt);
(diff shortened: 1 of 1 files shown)


SipSeb:
@miconda Can you backport that to 5.0, too? And is there a 5.0 equivalent to this page?
 https://www.kamailio.org/wiki/embeddedapi/4.0.x/lua
I would add the new log level here: https://www.kamailio.org/wiki/embeddedapi/devel/lua

miconda:
Just backported it to 5.0.

Feel free to add wiki pages for 5.0.x version -- just take the ones from devel and make a copy. I haven't spent much on it, as the plan is to have the kemi alternative replacing it sometime in the future (once proved to be better option) and no new features should be added via `sr` lua module -- fixing and missing bits can still be added.

SipSeb:
Okay, then I'll leave it there. I added a comment providing the information about a replacement in the future by KEMI.

PR #1175, opened 2017-07-03
ndb_redis: enhance debug info related to redis replies

claudiupb:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Added a way to print the redis responses when they are arrays, since the previous log message would just print (null) instead. In case a MULTI/EXEC transaction will be used to issue redis commands, the response would be an array, making the original debug message useless. With this new debug, arrays, strings, integers, and other replies such as status or error replies are logged properly. 
Also added an error message when a reply is expected but not received.

### DIFF
--- a/src/modules/ndb_redis/redis_client.c
+++ b/src/modules/ndb_redis/redis_client.c
@@ -608,7 +608,7 @@ int redisc_exec_pipelined(redisc_server_t *rsrv)
 			goto error_exec;
 		}
 	}
-	LM_DBG("reply is [%s]",rpl->rplRedis->str);
+	LM_DBG_redis_reply(rpl->rplRedis);
 
 	/* replies are received just retrieve them */
 	for (i=1;i<rsrv->piped.pending_commands;i++)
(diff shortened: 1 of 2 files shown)


src/modules/ndb_redis/redis_client.c
```diff
@@ -1061,3 +1067,57 @@ int redis_count_err_and_disable(redisc_server_t *rsrv)
...
 }
+
+void print_redis_reply(int log_level, redisReply *rpl,int offset)
+{
+	int i;
+	char padding[MAXIMUM_NESTED_KEYS + 1];
+
+	if (redis_get_debug_level() < log_level)
+		return;
+
```
miconda: I think that instead of the above IF condition you can use:

```
if(!is_printable(log_level))
```

so redis_get_debug_level() is not needed at all.

claudiupb: Yes, this is a lot better. I will change the commit

claudiupb:
I have modified the commit to use is_printable, instead of defining a new function to get the log level for redis module.

PR #1176, opened 2017-07-03
pua_reginfo: use outbound proxy for PUBLISHes

vance-od:
Hi, guys!

Pua module provides a callback "send_publish". One of the parameters is "outbound_proxy", but pua_reginfo does not use it while PUBLISH constructing, even if the parameter is specified.

Documentation says: "The outbound_proxy uri to be used when sending Subscribe requests."
Any reason to use it only for SUBSCRIBE, and not for PUBLISH?

If PR is acceptable, do I need to correct documentation for pua_reginfo module?

I have studied the contributing gide, but if something is not OK - please point out,.

### DIFF
--- a/src/modules/enum/enum.c
+++ b/src/modules/enum/enum.c
@@ -82,7 +82,7 @@ static int cclen(const char *number)
 	d1 = number[0];
 	d2 = number[1];
 
-	if(!isdigit((int)d2))
+	if(!isdigit((int)d1) || !isdigit((int)d2))
 		return (0);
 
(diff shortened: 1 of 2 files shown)


miconda:
The commit message is not good, it is `Update usrloc_cb.c`, whichd doesn't describe what the patch does, not it follows the guidelines for contributions. Actually, the title of this pull request is suitable for commit message -- see:

Can you rework the patch so it has proper commit message as well as amend the documentation to reflect the use of outbound proxy for publish as well?

PR #1177, opened 2017-07-03
pua_reginfo: use outbound proxy for PUBLISHes

vance-od:
Hi again, duplicate of [previous pull](https://github.com/kamailio/kamailio/pull/1176), as there is some mess.

Hope this time is better.

thanks!


miconda:
The first commit had the message ok, but the second didn't -- it didn't explain what the patch does effectively: 'correct docs' or 'fixed code' are not useful at all; a useful example: 'reflect that outbound proxy is used for PUBLISH in docs'.

Anyhow, I merged the PR manually. Thanks!

vance-od:
Thanks for accepting, next time I will try to prepare all the things thoroughly!

PR #1184, opened 2017-07-11
cplc: use route names to set proxy_route modparam

oriol-c:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
After allowing routes identification using names, the integer value defined in the parameter is not triggering the route referred by the name, but to the mapped id for one of the routes defined in the script. 
The proxy_route modparam type has been changed from integer to string in order to define the route to be executed by its name.


miconda:
Thanks!

PR #1185, opened 2017-07-12
Adapt some core rpc command to json

grumvalski:
Recently playing with jsonrpcs I found an issue with core.sockets_list and core.aliases_list commands called via jsonrpc_dispatch. 
The problem for the first is that the socket's list is that the JSON produced is like
{
        "jsonrpc":      "2.0",
        "result":       {
                "socket":       {
                        "proto":        "udp",
                        "address":      "XX.XX.XX.XX",
                        "ipaddress":    "XX.XX.XX.XX",
                        "port": "5060",
                        "mcast":        "no",
                        "mhomed":       "no"
                },
                "socket":       {
                        "proto":        "udp",
                        "address":      "XX.XX.XX.XX",
                        "ipaddress":    "XX.XX.XX.XX",
                        "port": "5060",
                        "mcast":        "no",
                        "mhomed":       "no"
                },
                ....
        },
        "id":   1
 }

While this is a valid JSON document, the fact that the "socket" key is defined several time in the same object makes many parsers not to properly parse it and only keep the last value.
Solution for this is to set the RET_ARRAY flag for the rpc command core.sockets_list reply so that the jsonrpcs module will create an array for the sockets instead on an object.
Similar issue for the aliases list returned by the core.aliases_list command: in this case the proposed solution is to add an "aliases" array to the rpc reply.

### DIFF
--- a/src/core/core_cmd.c
+++ b/src/core/core_cmd.c
@@ -989,7 +989,7 @@ static rpc_export_t core_rpc_methods[] = {
 	{"core.udp4_raw_info",     core_udp4rawinfo,       core_udp4rawinfo_doc,
 		0},
 	{"core.aliases_list",      core_aliases_list,      core_aliases_list_doc,   0},
-	{"core.sockets_list",      core_sockets_list,      core_sockets_list_doc,   0},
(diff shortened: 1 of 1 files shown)


miconda:
Thanks! It can be merged and backported.

PR #1186, opened 2017-07-12
core: add support for float and bigint when checking table version number

eschmidbauer:
cockroachdb does not have support for 32-bit integers
this commit adds basic support into the version table by allowing 64-bit integers from cockroachdb
cockroachdb support can be implemented on a per module basis from here on

### DIFF
--- a/src/lib/srdb1/db.c
+++ b/src/lib/srdb1/db.c
@@ -1,7 +1,7 @@
 /*
  * Copyright (C) 2001-2003 FhG Fokus
  * Copyright (C) 2007-2008 1&1 Internet AG
- * 
+ *
  * This file is part of Kamailio, a free SIP server.
  *
  * Kamailio is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


miconda:
Did you want to check for DB1_BIGINT instead of DB1_DOUBLE?

What's the use case for this patch? Table version is cleated with int column.

eschmidbauer:
this is the table schema i loaded into cockroachdb
```
CREATE TABLE version (
    table_name text,
    table_version INT
);
```
the result type detected by kamailio for the `table_version` column matches DB1_DOUBLE.
i discovered this by outputting VAL_TYPE(ver) to the log.
if DB1_BIGINT will match, im ok with that change too.
The use case is to add support to cockroachdb (which supports postgresql-client lib)

miconda:
The description of the pull request/patch is wrong then, because DB1_DOUBLE is a floating point, not an integer.

Eventually you can add support for both DB1_BIGINT and DB1_DOUBLE. However, few remarks:

  * use if-else, to be more optimal and avoid unnecessary evaluations of conditions, with check for DB1_INT being the first
  * for DB1_BIGINT and DB1_DOUBLE do cast to (int), otherwise there will be warnings on strict compilers

miconda:
The DB1_DOUBLE type might have been the result of the bug fixed with d12f96af28e170c3197a47af5840e5e30639adf6 .

miconda:
The patch related to value type was applied manually, the commit message was not reflecting the component and the description was too vague. The one with fixing whitespaces was not applied. You can commit it if you want, but prefix the commit message with 'lib/srdb1: ...'.

PR #1187, opened 2017-07-15
core: build reply of core.sockets_list as an array

grumvalski:
I realized with more tests that my previous commit for core.sockets_list 38f514064a326f759e0e88adeffd03b32b35d6c0 is not working. The result is returned yes as an array but inside there is still the sockets list as an object (so with multiple "socket" keys). Digging more in kamailio json lib I understood that to be considered an array the rpc result must be enclosed in "[]". 
I was misled by a wrong test result, sorry for the noise.


miconda:
It can be merged and backported, thanks!

PR #1192, opened 2017-07-28
suggestions from static analyzer

mslehto:


miconda:
Thanks! I am merging, but then adjust more on the patch for db_text because there can be  memsets on null pointer.

PR #1194, opened 2017-07-31
Centos 7 specfile updates

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Update the specfile for centos 7 to:
- build rabbitmq module package
- exclude kamailio-ims package dependency when installing main kamailio package


miconda:
I am going to apply 054b865 manually, because the second patch (9010d7e) needs to be discussed. I rushed initially and merged, but then I noticed that the `lib/ims` depends on libxml2, and by moving it into the core kamailio package, that one becomes dependent of libxml2. Am I right? If yes, this needs to be sorted out in a different way (e.g., make a package kamailio-lib-ims)

PR #1200, opened 2017-08-01
kamctl: fix dangerous 'rm -rf' code in case if variables are not defined

taurus-forever:
Hi,

kamctl has a historical bomb which can be new https://github.com/MrMEEE/bumblebee-Old-and-abbandoned/issues/123 in case if variables are undefined for some reason.

Have fun! CC @linuxmaniac


miconda:
Thanks!

PR #1203, opened 2017-08-03
nathelper: correct Contact received param

SipSeb:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

- Received param was enclosed in quotes, making some UACs stumble
  upon it. This fix removes the quotes, adding the received param
  the same way as it is done for example in the path module.
- Reported by Arslan Aseed.

### DIFF
--- a/src/modules/nathelper/nathelper.c
+++ b/src/modules/nathelper/nathelper.c
@@ -2122,15 +2122,13 @@ add_rcv_param_f(struct sip_msg* msg, char* str1, char* str2)
 	}
 
 	while(c) {
-		param = (char*)pkg_malloc(RECEIVED_LEN + 2 + uri.len);
+		param = (char*)pkg_malloc(RECEIVED_LEN + uri.len);
 		if (!param) {
 			LM_ERR("no pkg memory left\n");
 			return -1;
 		}
 		memcpy(param, RECEIVED, RECEIVED_LEN);
-		param[RECEIVED_LEN] = '\"';
-		memcpy(param + RECEIVED_LEN + 1, uri.s, uri.len);
-		param[RECEIVED_LEN + 1 + uri.len] = '\"';
+		memcpy(param + RECEIVED_LEN, uri.s, uri.len);
 
 		if (hdr_param) {
 			/* add the param as header param */


miconda:
I never used the function and I am not sure about its role/where is useful.

However, the current behaviour is correct if the received is added as header parameter -- enclosing it in quotes is ok there. Without enclosing it, there can be `received=uri;transport=xyz`, so transport becomes another parameter of the header, no longer part of received value.

Then: adding `received="uri;transport=xyz"` to URI in Contact is indeed invalid, but also removing the quotes is not ok. The transport becomes parameter to URI, not part of received parameter. In this case, the `=` has to be escaped.

But again, wonder when this is useful and if escaping is fixing it overall or breaks it even more so other parts need also patches.

SipSeb:
I have never used it myself. I just wanted to help this one mailing list post, but of course, this commit is broken.

I just looked into the path module how it's done there, and there exists a struct for different transports:

```const static char *proto_strings[] = {
        [PROTO_TCP] = "%3Btransport%3Dtcp",
        [PROTO_TLS] = "%3Btransport%3Dtls",
        [PROTO_SCTP] = "%3Btransport%3Dsctp",
        [PROTO_WS] = "%3Btransport%3Dws",
        [PROTO_WSS] = "%3Btransport%3Dws",
};```

This gets appended to the received param if needed. Then a Path header looks like this:

```Path: <sip:1.2.3.4;lr;received=sip:4.3.2.1:49194%3Btransport%3Dtcp>.```

Unfortunately I'm not able to implement that for the nathelper module as my C knowledge is limited to doing such trivial stuff like the commit above.

miconda:
Even with that format, I encountered some issues with a PBX, I guess it was doing first unescape of the whole Path URI and then parsing it, because it was taking wrong transport. That's the reason the master branch has now the option in path module to add received in the form of `alias`, like `ip~port~protoid`.

Adding it to URI in escaped format might not be that difficult, the question would be if the app interpreting it is going to do it right.

SipSeb:
Okay, so should we just add a comment in the documentation warning people of known bugs? Or what would the clean solution be? Maybe just eliminate that function in master and let it die?

miconda:
Adding an warning note in the docs should be done, indeed. The issue seems only when adding the parameter to Contact URI, not to header.

mattwilliamson:
I have run into this issue too. Thanks for the patch.

mattwilliamson:
After testing, I see that the contact header is malformed-- missing the `>` at the end, e.g.:

    add_rcv_param("1");

*Contact: <sip:1000@192.168.1.116:23439;transport=tcp;avaya-sc-enabled;received=sip:24.228.40.101:23439;transport=TCP*

mattwilliamson:
The reason why I am having issues here is because on kamailio 4.0 I see these when an invite comes in: 

    ERROR: <core> [parser/parse_to.c:733]: parse_to(): ERROR: parse_to : unexpected char ["] in status 6: 

I cannot upgrade yet, but perhaps there is a commit that fixes this issue that I can backport?

miconda:
I pushed a patch to make it compliant with SIP specs by not having the value enclosed in quotes when adding it to Contact URI, but that introduced a restriction to be used only for UDP traffic, otherwise there will be two contact parameters. I added notes in the docs for it.

PR #1206, opened 2017-08-07
textops: Fix 'offset exceeds message size' crash.

codyherzog:
Ensure that buffer is null terminated when subst_run() function is used.


miconda:
Thanks!

PR #1213, opened 2017-08-11
ims_auth: fixed rare core dump, due to null ptr check

christoph-v:
The ims_auth module reads the value from the Content-Length header
of a REGISTER request, before decoding the message body.
Due to a missing null pointer check, reading this value leads
to core dump in case the Content-Length header is missing.

#### Pre-Submission Checklist

- [ x] Commit message has the format required by CONTRIBUTING guide
- [ x] Commits are split per component (core, individual modules, libs, utils, ...)

#### Type Of Change
- [ x] Small bug fix (non-breaking change which fixes an issue)

#### Checklist:

- [ x] PR should be backported to stable branches
- [ x] Tested changes locally

#### Description

Added a null pointer check in function ims_get_body() - file utils.c of
module ims_auth - to check whether Content-Length Header was present
in REGISTER message.
If Content-Length Header is not present, a Default value of 0 will be assumed.

### DIFF
--- a/src/modules/ims_auth/utils.c
+++ b/src/modules/ims_auth/utils.c
@@ -204,7 +204,9 @@ str ims_get_body(struct sip_msg * msg)
 		LM_DBG("Error parsing until header Content-Length: \n");
 		return x;
 	}
-	x.len = (int)(long)msg->content_length->parsed;
+	if (msg->content_length)
+	    // Content-Length header might be missing
+	    x.len = (int)(long)msg->content_length->parsed;
         
         if (x.len>0) 
             x.s = get_body(msg);	


ngvoice:
Merged, thanks!

christoph-v:
Thanks. What can I do to enable/support backporting to 5.0 and 4.4 branches?

miconda:
You can use `git cherry-pick -x MASTER_COMMIT_ID` in branches `5.0` and `4.4` to make the backports in your local clones, then push them here as pull requests.

christoph-v:
Thanks. We discussed and decided not to backport currently.

PR #1214, opened 2017-08-13
pkg/kamailio/centos/7: Update CentOS spec file

sergey-safarov:
update version suffix
added unpackaged `acc_diameter` files
removed dependency to not existed package `libcrypto `


miconda:
Thanks!

PR #1218, opened 2017-08-17
pv: add string corehash transform

lazedo:
#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

adds string transform with optional size parameter to return the slot if supplied

`usage`

- $(var(myvar){s.corehash}) `returns the hash`
- $(var(myvar){s.corehash, 32}) `returns the slot`

`would be great if this can be backported to branch 5.0 and included in next week release`


miconda:
Thanks!

miconda:
I merged, then pushed some adjustments.

What I noticed is that the value is returned as a string, while one may expect a number, given that's typically an hash id. Any particular reason for this?

lazedo:
@miconda thanks. the reason was for large numbers not handled properly in script (uint) afaik.

PR #1227, opened 2017-09-03
pkg/kamailio/centos: Updated RPM spec file to support build fedora, RHEL, openSUSE

sergey-safarov:
Updated RPM spec file to support build fedora, RHEL, openSUSE via openSUSE build service.
Removed unnecessary RPM spec files


sergey-safarov:
Using this RPM spec file created packages for Fedora, CentOS, RHEL, OpenSUSE distrs
[more info](https://build.opensuse.org/package/show/home:sergey-safarov:v5.0.x-rpms/kamailio50)

miconda:
Is this PR removing old specs and adds the one for OpenSuse Build Service (OBS)?

sergey-safarov:
Yes. Now one SPEC file for all RPM based dists.
Old SPEC is removed.

miconda:
As far as I understood, at least it was the case in the past, those `%if` in OBS were extensions of OBS, not working on usual rpm build. That's the reason I did OBS specs only in the OBS repo, not in Kamailio repo.

So I would rather keep the old specs as they are for those that want to build locally and have obs folder to host the specs for the suse build service. Later, if proves to be ok to use the obs spec for building locally on the rpm distros (centos, fedora, redhat, oracle, ...), then we can either make symlinks or remove the specific folder.

Therefore, you should make a PR just for inclusion of the OBS spec files. Removal of the others will be subject of further discussions in the community.

miconda:
Btw, there has to be a PR for master branch first. Then one to backport to 5.0.

sergey-safarov:
Created PR https://github.com/kamailio/kamailio/pull/1228

PR #1230, opened 2017-09-09
pkg/kamailio/obs: Updated obs SPEC file

sergey-safarov:


miconda:
`crypto` module requires `libcrypto` or `libssl` if `libcrypto` is not packaged separately. Not sure how is on centos, but there is an external library required for `crypto` module.

Then again, removing packaging specs for `centos` or other distro is not going to be done like this. You pointed to obs packaging specs, but there are custom extensions, afaik -- so one cannot build rpms locally on own centos server with obs specs. Am I right?

If you need to adjust dependencies in obs specs, do a dedicated PR.

Removing any other specs, it's a subject of a dedicated PR as well, after discussing on mailing lists.

sergey-safarov:
Hello Daniel @miconda 
This PR only contains obs SPEC update.
1. Fixed mod crypto packaging;
2. Fixed dependency declaration for OpenSUSE leap dists;
3. removed obs macro usage;
4. Added obs project config (prjconf) - %rel macro now have correct form for CentOS, RHEL, Fedora dists

miconda:
Thanks!

PR #1232, opened 2017-09-12
carrierroute: warning for the same carrier/domain having routes with only 0 probability

hdikme:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

While starting kamailio or reloading the routes, if the same carrier/domain pairs do not have
any route with a probability other than 0 (zero) then an error log will be printed on the screen.
Besides, the log "invalid dice_max value" in the cr_func.c has been made more clear.

### DIFF
--- a/src/modules/carrierroute/cr_db.c
+++ b/src/modules/carrierroute/cr_db.c
@@ -265,6 +265,7 @@ int load_user_carrier(str * user, str * domain) {
  */
 int load_route_data_db(struct route_data_t * rd) {
 	db1_res_t * res = NULL;
+	db1_res_t * prob_res = NULL;
 	db_row_t * row = NULL;
 	int i, ret;
 	struct carrier_data_t * tmp_carrier_data;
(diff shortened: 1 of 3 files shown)


src/modules/carrierroute/cr_db.c
```diff
@@ -352,6 +353,7 @@ int load_route_data_db(struct route_data_t * rd) {
 		}
 	}
 	int n = 0;
+	boolean query_done = true;
```
lbalaceanu: Please change query_done to false here and adapt the code accordingly. The name of the variable should reflect the value it holds.

hdikme:
Following the commit based on your comment, your comment "Please change query_done to false here and adapt the code accordingly. The name of the variable should reflect the value it holds." disappeared but i have done the requested changes.

miconda:
@hdikme - the comment of @lbalaceanu was on the code of the patch and github hides it, you need to press `Show outdated` in the right side of @lbalaceanu's comment above here.

PR #1239, opened 2017-09-19
Fixed tls initialization when used LibreSSL. Backport from OpenBSD

sergey-safarov:


miconda:
Thanks! Merged manually to format the commit message.

PR #1241, opened 2017-09-22
pkg/kamailio/obs: Updated obs spec

sergey-safarov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
pkg/kamailio/obs: Added support of OpenSUSE Tumbleweed
pkg/kamailio/obs Fixed build of perl module
pkg/kamailio: Added links to obs spec file for:
    - Fedora 25, 26
    - RHEL 6, 7
    - OpenSUSE Leap

### DIFF
--- a/pkg/kamailio/obs/kamailio.spec
+++ b/pkg/kamailio/obs/kamailio.spec
@@ -93,8 +93,8 @@
 %bcond_without xmlrpc
 %endif
 
-%if 0%{?leap_version}
-%define dist_name suse
+%if 0%{?suse_version}
+%define dist_name opensuse
 %define dist_version %{?suse_version}
 %bcond_without cnxcc
 %bcond_with dnssec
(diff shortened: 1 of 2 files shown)


miconda:
Thanks for the PR and your significant contributions to rpm packaging.

But again I have to say that I am not confident the (sym)linking to obs folder is the right solution for other distributions. obs itself is by design an packing system for many distros. Just storing in the repo symlinks is creating some relations that should not be placed there.

As an idea that popped up right now in my mind, maybe it is better to have the folder rhel with a README.md and a Makefile inside. The makefile will have some rules to create subfolders for rhel versions (e.g., 6, 7...) that can be symlinks to obs if that works for packaging. The readme will explain that. Then the symlinks are created on the local copy of the person in need to do the packaging.

I know that there is a symlink for one of the packaging specs, but I find it confusing and not necessary to store it in the repo, being something needed to be created temporary.

Maybe we can discuss more during the irc devel meeting:

  * https://www.kamailio.org/wiki/devel/irc-meetings/2017b

To conclude: I am fine to merge your updates, but remove those symlinking, let's discuss in the community how to do that properly.

sergey-safarov:
I removed simlinking commit.
I will try to attend a meeting

miconda:
Thanks again, this work is really appreciated!

PR #1243, opened 2017-09-22
Rich redirects

gaaf:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

If Kamailio sends a 301/302, the current dset is added as Contact header automatically. The feature in this PR will optionally extend this Contact header with more information. 

- The path vector for each contact can be added as a Contact-uri Route header (rfc 3327 conformant).
- The branch flags can be added as a Contact-header parameter (non-standard).

In addition, the get_redirects function is extended to optionally parse the new flags parameter. The Route header can be extracted from the r-uri in the script.

The new functionality must be enabled with module parameters, by default the behavior is unchanged.

This feature is based on a patchset by me as <alex@speakup.nl> published at https://lists.kamailio.org/pipermail/sr-dev/2011-August/012064.html

### DIFF
--- a/src/core/dset.c
+++ b/src/core/dset.c
@@ -39,6 +39,7 @@
 #include "dset.h"
 #include "mem/mem.h"
 #include "ip_addr.h"
+#include "strutils.h"
 
 #define CONTACT "Contact: "
 #define CONTACT_LEN (sizeof(CONTACT) - 1)
@@ -49,6 +50,8 @@
 #define Q_PARAM ";q="
 #define Q_PARAM_LEN (sizeof(Q_PARAM) - 1)
 
+#define ROUTE_PARAM "?Route="
+#define ROUTE_PARAM_LEN (sizeof(ROUTE_PARAM) - 1)
 
 /* 
  * Where we store URIs of additional transaction branches
@@ -465,6 +468,7 @@ int append_branch(struct sip_msg* msg, str* uri, str* dst_uri, str* path,
 	return 1;
 }
 
+
 /*! \brief
  * Combines the given elements into a Contact header field
  * dest = target buffer, will be updated to new position after the printed contact
@@ -472,7 +476,7 @@ int append_branch(struct sip_msg* msg, str* uri, str* dst_uri, str* path,
  * end = end of target buffer
  * Returns 0 on success or -1 on error (buffer is too short)
  */
-static int print_contact_str(char **dest, str *uri, qvalue_t q, char *end, int options)
+static int print_contact_str(char **dest, str *uri, qvalue_t q, str *path, char *end, int options)
 {
 	char *p = *dest;
 	str buf;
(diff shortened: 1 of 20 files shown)


src/modules/uac_redirect/rd_funcs.c
```diff
@@ -294,8 +295,16 @@ static int shmcontact2dset(struct sip_msg *req, struct sip_msg *sh_rpl,
...
+				if (flags_hdr_mode && scontacts[i]->flags && str2int(&(scontacts[i]->flags->body), &flags) == 0) {
+					if (flags_hdr_mode == 2)
+						flags |= bflags;
+				} else {
+					flags = bflags;
+				}
+			}
 			if(append_branch( 0, &scontacts[i]->uri, 0, 0, sqvalues[i],
-						bflags, 0, &_redirect_sruid.uid, 0,
+						flags, 0, &_redirect_sruid.uid, 0,
```
miconda: As I got it, if flags parameter in contact is set and converted successfully to integer, even `flags_hdr_mode == 2`, the bflags is not taken in consideration.

Also I don't see how `flags_hdr_mode == 0` makes use of `bflags` like it was before for backward compatibility when `flags` param in contact is set.

src/core/dset.c
```diff
@@ -595,6 +598,17 @@ static int print_contact_str(char **dest, str *uri, qvalue_t q, str *path, char
...
+
+	/* branch flags (not SIP standard conformant) */
+	buf.s = int2str(flags, &buf.len);
+	if (p + FLAGS_PARAM_LEN + buf.len > end) {
+		return -1;
+	}
+	memcpy(p, FLAGS_PARAM, FLAGS_PARAM_LEN);
+	p += FLAGS_PARAM_LEN;
+	memcpy(p, buf.s, buf.len);
+	p += buf.len;
```
miconda: This should be made somehow optional, not being standard, as well as eventually do not add the flags parameter if the value is 0 (not set).

gaaf: Patch 4 makes this optional.

miconda: Seems that follow up patches add the option. For the future, when doing the PR merge the patches done to the same component in one, makes the review simpler as one does it in the order they are listed and later discovers is no longer same code, being again changed.

src/modules/uac_redirect/rd_funcs.c
gaaf: 1. If conversion is successful, flags are OR'ed with bflags when mode == 2.

There is something wonky here, looks like an incorrectly resolved merge/rebase. I'll fix this up.

2. Good catch, i'll fix this.

gaaf:
> @gaaf - any updates on this?

I still have this on my todo list, just haven't found the time to fix it yet. Sorry for the delay.

miconda:
@gaaf - soft reminder just in case you can address this one and eventually have it merged.

src/modules/uac_redirect/rd_funcs.c
gaaf: Both issues were caused by the same broken merge/rebase. Solved in my new patchset.

gaaf:
github's broken commit sorting messes up the order of the patches, better review it as a whole.

gaaf:
Worked around github's brokenness, commits are now in the correct order.

miconda:
Thanks! Looks ok after a quick review, I will go over one more try and then merge if nothing discovered.

henningw:
@miconda - i merged it into master, if there is something else to be found we can fix it there.

Issue #1252, opened 2017-09-28
memory configuration using environment vars


sergey-safarov:
### Description
Docker container at runtime is configured using environment vars.
Kamailio now not supports configure memory using environment vars and need use workarroud like [entrypoint.sh](https://github.com/kamailio/kamailio/blob/master/pkg/kamailio/alpine_docker/entrypoint.sh).

I want remove this workarround but need to add ability to configure kamailio memory using vars `SHM_MEMORY` and `PKG_MEMORY` when kamailio started without `-m` and `-M` keys

kamailio-sync:
i think you can set shm_mem_size in script to a #DEFINE that can be overridden with -A


in script


#DEFINE SHM_MEM 64


shm_mem_size = #DEFINE



then kamailio -A SHM_MEM=32


haven't tried it, but Daniel mentioned -A in the devel meeting.


not sure if there's a pkg_mem we can also use.


________________________________

miconda:
To understand properly first, in docker one cannot pass easily arguments to an application (kamailio in this case)? Kamailio has many parameters, not only for memory, not sure it is the right approach to look for an environment variable for all of them.

sergey-safarov:
I looked to other approach to pass program keys
We can use [`ENTRYPOINT`](https://docs.docker.com/engine/reference/builder/#exec-form-entrypoint-example) like
```
ENTRYPOINT ["kamailio", "-DD", "-E"]
```
And later run container as
```
docker run -it kamailio/kamailio -m 64 -M 8
```
I create PC

miconda:
This is no longer needed, closing it.

PR #1258, opened 2017-09-30
pkg/kamailio/alpine_docker: Docker args now passed to kamailio #1252

sergey-safarov:


miconda:
Merging.

If not done yet, maybe instead of mi_fifo which was removed, you should load jsonrpcs module, so kamctl can communicate with kamailio.

PR #1276, opened 2017-10-20
Add comments to db schema and set default last_modified

imcdona:
Added comments explaining table functionality to MySQL schema. These comments can be viewed in MySQL when browsing the schema which helps admins understand what each table is used for.

Changed default value for "last_modified" in grp table from "1970-01-01 00:00:00" to CURRENT_TIMESTAMP and set to automatically update "last_modified" if the record is ever updated. This means that if a record is inserted without a "last_modified", MySQL will insert an accurate time-stamp instead of 0000-00-00 00:00:00 which breaks the SQLOPS module.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

### DIFF
--- a/misc/scripts/mysql/my_create.sql
+++ b/misc/scripts/mysql/my_create.sql
@@ -29,7 +29,7 @@ CREATE TABLE acc (
     KEY acc_cid_key (sip_callid),
     KEY acc_from_uid (from_uid),
     KEY acc_to_uid (to_uid)
-);
+) COMMENT='used by the ACC module to report on transactions - accounted calls.';
 
 CREATE TABLE missed_calls (
     id INT AUTO_INCREMENT NOT NULL,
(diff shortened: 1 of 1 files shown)


miconda:
Are you using `misc/scripts/mysql/my_create.sql`? This is for SER flavour, not used by Kamailio these days. It's there for the case someone will needed by using an old ser version.

The definition of sql tables for kamailio is done via xml files located in `src/lib/srdb1/schema/` and sql scripts are generated using `make dbschema`. From the xml files is generated only a doc describing the tables:

  * https://kamailio.org/docs/db-tables/kamailio-db-devel.html

If you want to do something similar for kamailio sql scripts, you will have to work on the flow with the xml files.

For now I am closing this one, being with a patch done to a file not really used these days. If you want to be applied there, reopen and we can discuss further.

PR #1282, opened 2017-10-25
Support gcc 7+.

oyvindkolbu:
Don't let Makefile.defs complain about old compiler when using gcc 7.

This also gives gcc 7 the same default values as the other recent gcc versions.


miconda:
Thanks!

PR #1284, opened 2017-10-25
dialog: remove replicated dialog from profile as soon as gets to term…

charlesrchance:
…inated state

- matches local (non-replicated) behaviour introduced in edf61ac


miconda:
Thanks, it can be merged and backported if needed.

PR #1285, opened 2017-10-27
pkg/kamailio/obs: git cherry-pick to 5.0 branch

sergey-safarov:
pkg/kamailio/obs: git cherry-pick to 5.0 branch


miconda:
I am merging it.

But as a curiosity, does it make sense to create those symlinks for each distro back to `../obs` instead of just using the path to obs instead of dist_name and dist_version variables?

PR #1291, opened 2017-10-31
core:removing the condition for ipv6 outbound socket to build via correctly

surendratiwari3:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/core/msg_translator.c
+++ b/src/core/msg_translator.c
@@ -2782,10 +2782,9 @@ char* via_builder( unsigned int *len,
 		pkg_free(line_buf);
 		return 0;
 	}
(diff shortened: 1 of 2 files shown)


miconda:
Thanks! I merged the patch manually for having single commit and proper commit message, the PR is listing several one and a merge operation.

However, the patch fixed some situation, but seemed incomplete to cover the situations when advertised hostname is used. I pushed another patch after yours to deal with. Can you test if all works fine for you using master (or the two related patches from core). If all fine, they will be backported.

surendratiwari3:
@miconda Thanks i will test the same.In future i will follow  your given instruction to have single commit with proper message .

PR #1293, opened 2017-11-01
Enable the same modules for RHEL 7 as for Centos 7.

oyvindkolbu:
Currently OBS build more Centos 7 modules than for RHEL 7. The modules builds just fine on RHEL7, so enable them there as well.


miconda:
Thanks! Patch applied manually to reformat commit message.

PR #1301, opened 2017-11-07
pkg/obs: enable the same modules for RHEL 7 as for Centos 7

oyvindkolbu:
Would like these modules directly from obs.


miconda:
Thanks!

PR #1303, opened 2017-11-07
core : honor parse_param separator

lazedo:
fixes the use case when spaces should not be excluded 'BS ParkingTestCase 1'
```
     xlog("L_INFO", "$ci|log exploding test\n");
     $var(test) = "Notify=false;Presence-ID=1001@kamailio.io;Account-Name=BS ParkingTestCase 1";
     xavp_params_explode($var(test), "mytest");
     xlog("L_INFO", "$ci|log exploded test\n");
```

@miconda i believe this should be included in 5.1 and backported to 5.0

### DIFF
--- a/src/core/parser/parse_param.c
+++ b/src/core/parser/parse_param.c
@@ -348,11 +348,9 @@ static inline int parse_token_param(str *_s, str *_r, char separator)
 		 * mark end of the token
 		 */
 		switch(_s->s[i]) {
-			case ' ':
 			case '\t':
 			case '\r':
 			case '\n':
-			case ',':
 				/* So if you find
 				 * any of them
 				 * stop iterating


miconda:
@lazedo - I think this breaks the rules in the sip grammar. The parameter value must be enclosed in quotes (single or double) if you have spaces inside it. Also, comma is separator for headers with multiple bodies, so again it cannot appear in parameter value which is not quoted.

Is any limitation you don't enclose the params values within quotes?

lazedo:
@miconda thanks for pointing that out. checking again, maybe jumped in here too fast.

lazedo:
@miconda enclosing the params with quotes fixes it. sorry for jumping too fast, but when i saw the `separator` and then the exceptions..., didn't think it could break sip grammar.
thanks for the explanation.

Issue #1109, opened 2017-05-02
dmq & dmq_usrloc : let server_address & notification_address use the same IP and port


zhangst:
if server_address & notification_address use same ip and different port(see below)，in dmq.list_nodes will have two node for one kamailio(one node for notify_server 5360 and one for dmq local 5260), question is when send REGISTER to port 5160, REGISTER will copy from port 5260 to 5360 on the same kamailio,this is not what i want.

dmq & dmq_usrloc cfg(**different port**):
 ```
listen=eth0:5160
listen=eth0:5260
listen=eth0:5360

modparam("dmq", "server_address", "sip:10.6.1.200:5260")
modparam("dmq", "notification_address", "sip:10.6.1.200:5360")
modparam("dmq", "multi_notify", 1)
modparam("dmq", "num_workers", 2)

modparam("dmq_usrloc", "enable", 1)
modparam("dmq_usrloc", "sync", 1)
modparam("dmq_usrloc", "batch_msg_contacts", 50)
modparam("dmq_usrloc", "batch_msg_size", 500000)
```
And when I let server_address & notification_address use same IP & Port, check in function add_server_and_notify will report error("error adding notification node"),child_init will return failed.

So I suggest that the following changes, let server_address & notification_address can use the same IP and port：
```
@@ -315,11 +315,15 @@ dmq_node_t* add_server_and_notify(str *paddr)
                for (index = 0; index < host_cnt; index++) {
                        pstr->s = puri_list [index];
                        pstr->len = strlen (puri_list [index]);
-                       if (!find_dmq_node_uri(node_list, pstr)) { // check for duplicates
+                       if (!(pnode = find_dmq_node_uri(node_list, pstr))) { // check for duplicates
                                pnode = add_dmq_node (node_list, pstr);
                                if (pnode && !pfirst)
                                        { pfirst = pnode; }
-                       }
+                       } else {
+                             if (!pfirst) {
+                                     pfirst = pnode;
+                             }
+                       }
                }
        }
```

After Such A change,  dmq.list_nodes will only have on node for (notify_server & server_addr(local)).
Is there any problem with this modification? Thanks

```
$./sbin/kamailio -v
version: kamailio 5.0.1 (x86_64/linux) 
[...]

PR #1308, opened 2017-11-08
dmq: let server_address & notification_address can use the same IP and port

zhangst:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description
issue : https://github.com/kamailio/kamailio/issues/1109

### DIFF
--- a/src/modules/dmq/notification_peer.c
+++ b/src/modules/dmq/notification_peer.c
@@ -315,10 +315,14 @@ dmq_node_t* add_server_and_notify(str *paddr)
 		for (index = 0; index < host_cnt; index++) {
 			pstr->s = puri_list [index];
 			pstr->len = strlen (puri_list [index]);
-			if (!find_dmq_node_uri(node_list, pstr)) { // check for duplicates
+			if (!(pnode = find_dmq_node_uri(node_list, pstr))) { // check for duplicates
 				pnode = add_dmq_node (node_list, pstr);
 				if (pnode && !pfirst)
 					{ pfirst = pnode; }
+			} else {
+				if (!pfirst) {
+					pfirst = pnode;
+				}
 			}
 		}
 	}


charlesrchance:
Hi - I still don't understand why this is necessary. What combination of server/notification address values causes issues for you?

zhangst:
For example, I have four kamailio instance for register, one kamailio instance recv REGISTER, it copy to other three kamailio instance by dmq_usrloc.
In this scene, I set notification_address to domain sip:ka.ns.com:5060 which resolve to all four kamailio instance(which recv REGISTER).In this way, **I do not need to build a separate kamailio cluster only for notification server.**

```
modparam("dmq", "server_address", "sip:10.6.1.200:5260")
modparam("dmq", "notification_address", "sip:ka.ns.com:5260")
# This configuration, ka.ns.com only resolve to 10.6.1.200, kamailio can not normal start.
it report : 
dmq [notification_peer.c:345]: add_server_and_notify(): error adding notification node
dmq [dmq.c:299]: child_init(): cannot retrieve initial nodelist from sip:ka.ns.com:5260
```

charlesrchance:
‘’’
modparam("dmq", "server_address", "sip:10.6.1.200:5260")
modparam("dmq", "notification_address", "sip:ka.ns.com:5260")
‘’’

^^ this will work just fine as it is, provided ka.ns.com resolves to all four nodes - or at least one other than itself.

Why, in your second example, does it only resolve to itself?

zhangst:
yes，domain only reolve to itself.

charlesrchance:
Do you have the multi_notify modparam set?

In the special case you mention, why set the notification address at all if it’s pointing to itself and you already know this will not be added as a peer?

zhangst:
already set multi_notify modparam.

In four nodes case, ka.ns.com is not resolve to four ip at once.
ka.ns.com resolve to ip1 in first visit, and resolve to ip2 in second visit, and so on.
This is DNS action. not kamailio.
Maybe it is not a general usage, DNS should resolve to all four ip.

I realized my problem,  DNS should resolve to all four ip. 
Thanks.

PR #1309, opened 2017-11-09
usrloc: copy server_id when updating contact in memory

charlesrchance:


charlesrchance:
@miconda - is it too late to backport this to 5.0 now that 5.1 branch has been started?

PR #1315, opened 2017-11-15
core: removing the condition for second record_route ipv6 header

surendratiwari3:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
when we are sending the call from ipv6 tls registered endpoint to kamailio and kamailio use the dispatcher module to forward that call to tcp sip server then second record route header have ipv6 address without [].so another sip server is saying 400 bad record route header.
before (with error):
Record-Route: <sip:2600:1F1C:74E:E100:E89F:BEB1:10AF:1CB1:5060;transport=tcp;r2=on;lr=on;ftag=msDr8yzIe5obR7fSfD-jPiJPFcJoVwwr;did=ae1.8861>

after (correct):
Record-Route: <sip:[2600:1F1C:74E:E100:E89F:BEB1:10AF:1CB1]:5060;transport=tcp;r2=on;lr=on;ftag=msDr8yzIe5obR7fSfD-jPiJPFcJoVwwr;did=ae1.8861>

### DIFF
--- a/src/core/msg_translator.c
+++ b/src/core/msg_translator.c
@@ -724,8 +724,7 @@ static inline int lumps_len(struct sip_msg* msg, struct lump* lumps,
 			case SUBST_SND_ALL: \
 				if (send_sock){ \
 					new_len+=send_address_str->len; \
-					if ((send_sock->address.af!=AF_INET) && \
-							(send_address_str==&(send_sock->address_str))) \
+					if ((send_sock->address.af!=AF_INET)) \
 						new_len+=2; \
 					if ((send_sock->port_no!=SIP_PORT) || \
 							(send_port_str!=&(send_sock->port_no_str))){ \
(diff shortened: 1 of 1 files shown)


surendratiwari3:
@miconda please review this also for record_route header

miconda:
Do you have advertise address to the listen socket?

miconda:
I just pushed another patch (referenced above), because there were other cases when the same condition was used. The new patch is taking a safer path by checking if the ipv6 address starts with `[`.

Can you test and report if ok?

surendratiwari3:
@miconda yes ,i have advertise address to the listen socket.

surendratiwari3:
@miconda i will test the same.we are expecting all ipv6 patch to be merged with stable branch so we can start using those only.

miconda:
If testing with master is ok, then I will backport. Just report if it works fine or not.

surendratiwari3:
@miconda i tested with master branch.its working fine.it will be good if you can commit all changes related to ipv6 to all stable releases specially to 5.x.x branches.reference to following pull request:
https://github.com/kamailio/kamailio/pull/1319
https://github.com/kamailio/kamailio/pull/1315
https://github.com/kamailio/kamailio/pull/1291

miconda:
Patches should be backported by now in the 5.x.x branches, if you still find an issue, open another item on the tracker.

SurendraPlivo:
@miconda thanks.

PR #1321, opened 2017-11-20
cfg_rpc: extending the functionality of cfg.get command

hdikme:
- rpc_get function can be also called with only entering the group name information.
Before it used to work only with group name and variable name and returning the value for a variable.
With the enhancement, it is also possible to retrieve all variable values belonging to a specific group.
e.g.: kamcmd cfg.get core.
Above given example will print the values of all variables of the configuration group "core".

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

rpc_get function can be also called with only entering the group name information.
Before it used to work only with group name and variable name and returning the value for a variable.
With the enhancement, it is also possible to retrieve all variable values belonging to a specific group.
e.g.: kamcmd cfg.get core.
Above given example will print the values of all variables of the configuration group "core".

### DIFF
--- a/src/modules/cfg_rpc/cfg_rpc.c
+++ b/src/modules/cfg_rpc/cfg_rpc.c
@@ -310,48 +310,88 @@ static const char* rpc_get_doc[2] = {
 
 static void rpc_get(rpc_t* rpc, void* c)
 {
-	str	group, var;
-	void	*val;
-	unsigned int	val_type;
-	int	ret;
-	unsigned int	*group_id;
+	str     group, var;
+	void    *val;
+	unsigned int    val_type;
+	int     ret, n;
+	unsigned int    *group_id;
 
-	if (rpc->scan(c, "SS", &group, &var) < 2)
-		return;
-
+	n = rpc->scan(c, "SS", &group, &var);
+	/*  2: both group and variable name are present
+	 * -1: only group is present, print all variables in the group */
+	if(n<2) {
+		if (n == -1) {
+			var.s = NULL;
+			var.len = 0;
+		}
+		else return;
+	}
 	if (get_group_id(&group, &group_id)) {
 		rpc->fault(c, 400, "Wrong group syntax. Use either \"group\", or \"group[id]\"");
 		return;
 	}
-
-	ret = cfg_get_by_name(ctx, &group, group_id, &var,
-			&val, &val_type);
-	if (ret < 0) {
-		rpc->fault(c, 400, "Failed to get the variable");
-		return;
-		
-	} else if (ret > 0) {
-		rpc->fault(c, 400, "Variable exists, but it is not readable via RPC interface");
-		return;
+	/* print value for one variable */
+	if(var.len != 0) {
+		ret = cfg_get_by_name(ctx, &group, group_id, &var,
+				&val, &val_type);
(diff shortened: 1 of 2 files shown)


src/modules/cfg_rpc/cfg_rpc.c
```diff
@@ -310,48 +310,88 @@ static const char* rpc_get_doc[2] = {
...
+	n = rpc->scan(c, "SS", &group, &var);
+	/*  2: both group and variable name are present
+	 * -1: only group is present, print all variables in the group */
+	if(n<2) {
+		if (n == -1) {
+			var.s = NULL;
+			var.len = 0;
+		}
+		else return;
+	}
```
miconda: I think that the right way is to read with optional specifier `*`:

```
n = rpc->scan(c, "S*S", &group, &var);
```

The n is two if both were read or 1 if only group is read.

When n is -1, there can be other errors.

Otherwise, the patch is useful, thanks!

Can you do a new patch based on what I suggested? It can be a follow up of this one, we can squash from the web when merging.

hdikme: Thank you for the review, sure i'll send a new patch with the changes.

miconda:
@hdikme - planning to have the new patch soon? Just to evaluate if this needs to be in 5.1 or not which is going to be out in few days...

hdikme:
@miconda - I have tested the code with the changes based on your review. First part is indeed more appropriate (S*S) but i have a problem with the value of n. If only one parameter (group name) is given , n is still assigned -1. Either was S*S or SS returns -1. What do you think?

miconda:
What do you use to send the rpc commands - `kamcmd`? If yes, can you try also with `kamctl rpc cfg.get ...`? I want to see if it is some issue in `ctl` module. For kamctl you need jsonrpcs module to be loaded.

hdikme:
Yes, I use kamcmd as I think this is the new one and this should replace kamctl someday in the future. I tried with kamctl rpc, but it just prints out the help page of kamctl. It seems kamctl does not know rpc as a parameter. Maybe you can provide me a more detailed description how to test this with kamctl? Or maybe it's even faster you test it yourself using kamctl?

miconda:
Are you testing with master branch (or 5.x)? `kamctl rpc ...` works there.

miconda:
Btw, kamcmd is not a replacement for kamctl.

kamcmd is only a binrpc client application that connects to ctl module of kamailio.

kamcli might replace kamcmd, but no timeline for that. kamctl is still very active maintained and enhanced with each release.

hdikme:
I have tested it with the version 5.0.1 using kamctl rpc ... and the result hasn't changed.
n = rpc->scan(c, "S*S", &group, &var);
n equals to -1 in case of using only the group name.

miconda:
The issue with `n = rpc->scan(c, "S*S", &group, &var);` was due to a bug in `ctl` module, using `kamcmd` revealed that. Using `kamctl` with `jsonrpcs` module was not affected. The bug is fixed now in master and it will be backported.

Besides using the `*` specifier for optional params, your patch in this PR was adjusted so that the `cfg.get` for a group returns a structure with field names and values. All are in master now.

PR #1326, opened 2017-11-23
rtpengine: fixed segfault when using read_sdp_pv

phil-lavin:
#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
This was an odd one. I can't see how `read_sdp_pv` ever worked. When the old SDP is deleted with `del_lump` from the SIP `msg`, the lump pointer is calculated assuming that `body` is a pointer relative to `msg`. When `read_sdp_pv` is set, `body` pointer is set to something entirely different and doing pointer maths relative to `msg` causes integer overflows and results in a segfault on `del_lump`.

This change explicitly gets the body pointer (`cur_body`) relative to `msg` before trying to `del_lump`. If it's preferable, I can have it only do this if `read_sdp_pv` is set but I figured it's not a heavy operation and it's safer to always do it.

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -3328,6 +3328,7 @@ rtpengine_offer_answer(struct sip_msg *msg, const char *flags, int op, int more)
 	str body, newbody;
 	struct lump *anchor;
 	pv_value_t pv_val;
+	str cur_body = {0, 0};
 
 	dict = rtpp_function_call_ok(&bencbuf, msg, op, flags, &body);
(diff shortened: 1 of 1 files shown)


miconda:
Thanks! It will be merged.

For convenience, can you make pull requests for stable branches? It makes it faster to backport by merging them via web here.

phil-lavin:
Can you remind me which versions are still being supported? I'll create backport PRs.

miconda:
Do it for 5.0 and 5.1, it should be straightforward. For 4.4 might need some more work.

PR #1332, opened 2017-11-24
Update textops.c

verticelo:
textops: fixed naming convention

- The original function is named in_ rather than is_, updating to the same standard


miconda:
Thanks! Patch pushed manually because the one from this PR doesn't have the commit message formatted as per guidelines in contributing.md

PR #1341, opened 2017-11-29
presence: update contact in update_subs_db

lazedo:
when using presence with dbmode = 3, the contact is never updated. if using dbmode != 3 the contact is updated in the hash table.

we've seen some UAs re-subscribing with a different contact port of the initial subscribe. it works with dbmode != 3 but not with dbmode == 3

this should be backported to 5.0/5.1 branch
Thanks


miconda:
Thanks! Can be merged and backported as needed to the stable branches.

PR #1344, opened 2017-11-29
nat_traversal: handle local generated transactions

lazedo:
when using the below script blocks for handling SUBSCRIBE messages, the nat_keepalive does not work as expected.
```
#!ifdef NAT_TRAVERSAL_ROLE
route[PRESENCE_NAT]
{   
    force_rport();
    if (client_nat_test("3")) {
        if(is_first_hop())
           set_contact_alias();
        nat_keepalive();
    }
}
#!endif

route[HANDLE_SUBSCRIBE]
{
    if (!is_method("SUBSCRIBE")) {
        return;
    }

    #!ifdef NAT_TRAVERSAL_ROLE
      route(PRESENCE_NAT);
    #!endif

    if (!t_newtran()) {
        sl_reply_error();
        exit;
    }
    
    if(has_totag()) {
        route(HANDLE_RESUBSCRIBE);
    } else {
        route(HANDLE_NEW_SUBSCRIBE);
    }
    t_release();    
    exit;
}
```

i believe this should be backported to stable branches
Thanks

### DIFF
--- a/src/modules/nat_traversal/nat_traversal.c
+++ b/src/modules/nat_traversal/nat_traversal.c
@@ -1358,7 +1358,7 @@ __sl_reply_out(sl_cbp_t *slcbp)
 static void
 __tm_reply_in(struct cell *trans, int type, struct tmcb_params *param)
 {
-    time_t expire;
+    time_t expire = 0;
 
     if (param->req==NULL || param->rpl==NULL)
         return;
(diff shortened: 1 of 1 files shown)


miconda:
First, it looks like `pkg_free` has to be done on `tmp`.

Then I think this needs to be tuned, otherwise can affect performances for forwarded subscribe requests, because the operation will be done twice. The execution of the callback for `TMCB_RESPONSE_SENT` should be done only when `param->rpl==FAKED_REPLY` (which is when a local response is generated).

lazedo:
@miconda thanks for the comments.
i think `tmp` will be freed with `free_sip_msg`.
will update the `FAKED_REPLY`

miconda:
`tmp` is not freed there, because DYN_BUF is not defined. You can wrap it in ifdef if you want, but that define should be removed, is inactive for long time.

lazedo:
@miconda ah, didn't know that, fixing

miconda:
I squashed and merged in a single commit. You can backport if needed.

PR #1345, opened 2017-11-29
schema: allow null in active_watchers reason

lazedo:
some code paths in presence module set the `reason` field to null causing problems updating the database.

### DIFF
--- a/src/lib/srdb1/schema/pr_active_watchers.xml
+++ b/src/lib/srdb1/schema/pr_active_watchers.xml
@@ -153,6 +153,7 @@
         <name>reason</name>
         <type>string</type>
         <size>&user_len;</size>
+        <null/>
         <description>Reason</description>
     </column>
 


miconda:
Many columns default to empty string because the code expect it and doesn't do additional checks to see if the db value is NULL. Have you checked the code for this case?

lazedo:
@miconda on new dialog subscribes the value is null. we can either add a `default` or allow null

miconda:
Have you tested with the updated schema and it works?

lazedo:
Hi Daniel,

only with sqlite which uses kamailio db routines to generate the actual sql
statement.
i can try others or, maybe i'm missing something basic in sqlite.

miconda:
I checked mysql and it doesn't allow NULL -- if some parts of the code sets it to NULL, then it should have been an error as well. Can you point to such code?

I am fine to change it to allow NULL, just to be sure there is no side effect.

lazedo:
@miconda hi Daniel,
follow the code path of handle_subscribe in presence/subscriber.c that leads to `update_subscription`
- new subscription
- pres_notifier_processes = 0
- dialog subscription => event->req_auth == 0

this code doesn't change the value its null by default, the code path with req_auth != 0, ~line 1202 does set it explicitly to null before calling the subscription handler.

PR #1354, opened 2017-12-06
presence: extra columns updated for dbmode 2

lazedo:
we noticed some phones and upstream clusters sending
diferent route / contact after initial subscription.

this commit adds contact and record_route to updated columns for dbmode = 2
also adds user_agent to initial save
and a simple fix for getting the correct message in debug

should be backported to stable/maintained to fix issue in existing deployments


miconda:
Thanks! You can merge and backport as needed to stable branches.

PR #1361, opened 2017-12-11
Set Cherry picks of OBS spec update

sergey-safarov:
Set Cherry picks of OBS spec update
Added Fedora 27 support
Removed Fedora 24 support


miconda:
I am merging this one, but there is no reason to have `sipdump` as a dedicated package. It should be moved in the main one.

Issue #1358, opened 2017-12-11
acc: save cdr entries in UTC


cruzccl:
### Description

ACC module allows setting datetime fields in UTC via [time_mode param](http://kamailio.org/docs/modules/devel/modules/acc.html#acc.p.time_mode).

However, this setting does not cause cdr entries to be saved in that format and I haven't found any equivalent parameter to achieve this.

### Troubleshooting

Analyzing the code for MySQL database storage I have found that [_db_time2str_ex_ function defined in _db_ut.c_](https://github.com/kamailio/kamailio/blob/master/src/lib/srdb1/db_ut.c#L256) uses localtime(). 

### Possible Solutions

Changing localtime() to gmtime() cdrs are saved in UTC, but it would be desirable to add a new module param that makes this configurable.

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
version: kamailio 4.4.6 (x86_64/linux) 0a8379-dirty
flags: STATS: Off, USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MEM, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLACKLIST, HAVE_RESOLV_RES
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535, DEFAULT PKG_SIZE 8MB
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
id: 0a8379 -dirty
compiled on 17:01:42 Dec  4 2017 with gcc 6.3.0

```

* **Operating System**:

```
Linux 4.9.0-3-amd64 #1 SMP Debian 4.9.30-2+deb9u3 (2017-08-06) x86_64 GNU/Linux
```

Should I proceed and try to get a working PR to add this new modparam or there is currently a way to get this that I have missed (apart from setting system time to UTC)?

Thank you and regards,

miconda:
[...]

PR #1368, opened 2017-12-14
acc: use acc.time_mode to save cdrs in gmtime

cruzccl:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #1358 

#### Description

acc.time_mode modparam allows storing acc transactions time field using the output of gmtime() when it is set to 4.

This commit aims to reuse this modparam for cdr entries (start_time, end_time). This way, if time_mode is 4, both acc and cdr entries will be stored in gmtime.

### DIFF
--- a/src/modules/acc/acc_cdr.c
+++ b/src/modules/acc/acc_cdr.c
@@ -64,6 +64,7 @@
 
 #define TIME_STR_BUFFER_SIZE 20
 #define TIME_BUFFER_LENGTH 256
+#define TIME_STRING_FORMAT "%Y-%m-%d %H:%M:%S"
 
 struct dlg_binds dlgb;
 struct acc_extra* cdr_extra = NULL;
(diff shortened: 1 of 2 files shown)


src/modules/acc/acc_cdr.c
```diff
@@ -182,13 +185,24 @@ static int db_write_cdr( struct dlg_cell* dialog,
...
-				VAL_TIME(db_cdr_vals+i) = timeval_val.tv_sec;
+				if (acc_time_mode==4) {
+					VAL_TYPE(db_cdr_vals+i)=DB1_STR;
+					t = gmtime(&timeval_val.tv_sec);
+					/* Convert time_t structure to format accepted by the database */
+					if (strftime(cdr_time_format_buf[i], TIME_STR_BUFFER_SIZE, TIME_STRING_FORMAT, t) <= 0) {
+						cdr_time_format_buf[i][0] = '\0';
+					}
+
+					VAL_STR(db_cdr_vals+i) = (str){cdr_time_format_buf[i], TIME_STR_BUFFER_SIZE-1};
```
miconda: Is this cast compliant with strict C standard? Kamailio is used on some old/embedded systems with more strict compilers.

Otherwise, I think this should just be a DB1_STRING data type, because the buffer is zero-terminated.

cruzccl: I am not sure if it's compliant, but I am sure that the way you propose is more legible and clean.

I have already changed it.

miconda:
Thanks!

PR #1371, opened 2017-12-15
Let the partition column be configurable.

oyvindkolbu:
#### Type Of Change

#### Checklist:
- [X] Tested changes locally

When using xDB Replication Server with PostgreSQL, "partition" is a reserved
name and breaks replication. Therefore let the column name be configurable.

An even better fix would be to also change the default name, as "partition" is a
SQL key word which is reserved by SQL:2011. See
https://www.postgresql.org/docs/current/static/sql-keywords-appendix.html

### DIFF
--- a/src/modules/usrloc/doc/usrloc_admin.xml
+++ b/src/modules/usrloc/doc/usrloc_admin.xml
@@ -516,6 +516,26 @@ modparam("usrloc", "keepalive_column", "kalive")
 		</example>
 	</section>
 
+	<section id="usrloc.p.partition_column">
+		<title><varname>partition_column</varname> (string)</title>
+		<para>
+		Name of database table column containing the value for partition id.
+		</para>
+		<para>
+		<emphasis>
+			Default value is <quote>partition</quote>.
+		</emphasis>
+		</para>
+		<example>
+		<title>Set <varname>partitioncolumn</varname> parameter</title>
+		<programlisting format="linespecific">
+...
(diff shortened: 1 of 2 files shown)


miconda:
I will push the patch manually, because the commit message doesn't follow the contribution guidelines:

Regarding the reserved words, I think the right solution would be to quote the column names, as mysql module does it. New SQL spec might add other reserved words. Not using postgres myself I am not sure what is the right character to quote columns and table names, if someone writes it here and also helps to test, then I can make a patch to db_postgres module to use it.

oyvindkolbu:
From https://www.postgresql.org/docs/current/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS it seems that "partition" would do the job. Quote from the page:

> There is a second kind of identifier: the delimited identifier or quoted identifier. It is formed by enclosing an arbitrary sequence of characters in double-quotes ("). A delimited identifier is always an identifier, never a key word. So "select" could be used to refer to a column or table named “select”, whereas an unquoted select would be taken as a key word and would therefore provoke a parse error when used where a table or column name is expected

miconda:
Interesting ... is the string value also in double quotes for postgres? I wonder how it makes the difference between a string value and a quoted identifier ...

oyvindkolbu:
String values are used such as: `select foo from baz where bar = 'string'`

miconda:
So, single quote for string values and double quote for identifiers...

miconda:
Merged manually.

PR #1376, opened 2017-12-19
tcp: correct queued length checking

snen:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
I can’t get big responses via jsonrpcs with default tcp params. According to log response size ~3200k,  kamailio send ~2500k of data and can’t add 700k to queue after. This patch should fix, if my understanding of “queued” value is right.

### DIFF
--- a/src/core/tcp_main.c
+++ b/src/core/tcp_main.c
@@ -611,7 +611,7 @@ inline static int _wbufq_add(struct  tcp_connection* c, const char* data,
 	
 	q=&c->wbuf_q;
 	t=get_ticks_raw();
-	if (unlikely(	((q->queued+size)>cfg_get(tcp, tcp_cfg, tcpconn_wq_max)) ||
+	if (unlikely(	(q->queued>cfg_get(tcp, tcp_cfg, tcpconn_wq_max)) ||
 					((*tcp_total_wq+size)>cfg_get(tcp, tcp_cfg, tcp_wq_max)) ||
 					(q->first &&
 					TICKS_LT(q->wr_timeout, t)) )){


miconda:
Thanks! I will look a bit at the code before merging, last weeks were too busy...

miconda:
After first review, the patch is not correct, because the size of the data to be added in the connection queue results in going over the limit of `tcpconn_wq_max` (corresponding to config parameter `tcp_conn_wq_max`).

If you have a different opinion, can you provide more details about how you concluded this would be a fix?

snen:
It woks just according to: "i can dump the same big htable via jsonrpcs/tcp  with this patch, and can't without it". 
In my mind we have to check already "queued" length, not  "queued+size", but  as mensioned before my understanding of these parameters is superficial

miconda:
`queued` gives what was written in the outbound queue already, and the `queued + size` checks that what is going to be written is not exceeding the configured limit.

Have you increased the limits with global parameters (tcp_conn_wq_max and tcp_wq_max), or just used the defaults? If you deal with large output, these values have to be adjusted.

snen:
i use default TCP params and change  _tcp_conn_wq_max_ and _tcp_wq_max_ can help of course according to if statement.

_q->queued_ is zero in my fail test,  so it is strange for me.
in docs
>tcp_conn_wq_max
> Maximum bytes queued for write allowed per connection. Attempting to queue more bytes would result in an error and in the connection being closed (**too slow**). If tcp_write_buf is not enabled, it has no effect.

So i try to add  the last 700k and it is failed always. I undestand this paramenter as not yet sent data (connection too slow), so if _q->queued<tcp_conn_wq_max_, then connection is fast and i can add more data to queue

miconda:
Resuming on this one: maybe you can describe here how this can be reproduced, eventually provide a minimal kamailio.cfg as well as some scripts to do it and we can analyse if it is expected behaviour or not.

snen:
here is minimal config for current master
[kamailio.txt](https://github.com/kamailio/kamailio/files/2170252/kamailio.txt)

here is sql script for load data
[big_htable_create.txt](https://github.com/kamailio/kamailio/files/2170253/big_htable_create.txt)

how to reproduce
```
[snen@sw5 kamailio]# cat big_htable_create.txt | mysql kamailio_dev
[snen@sw5 kamailio]# devkamctl restart
[snen@sw5 kamailio]# curl -X GET -H "Content-Type: application/json" -d '{"jsonrpc": "2.0", "method": "htable.dump", "params":["big_htable"],  "id": 1 }' http://192.168.10.190:5071/jsonrpc/ > /tmp/big_htable.data 
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
 59 4181k   59 2479k  100    79   9.7M    318 --:--:-- --:--:-- --:--:--  9.8M
curl: (18) transfer closed with 1743335 bytes remaining to read
```

miconda:
Coming back to this one after quite long time ...

So I just tried to reproduce based on your minimal config and current master branch. I had to increase the size of PKG memory (-M 12), but then all was fine -- the content of `/tmp/big_htable.data` had all 30k records (based on number of lines printed there and the closing of the json document was ok). No error was printed by kamailio.

snen:
I use "-m64 -M32" for current master and result is the same (can't read all response data).  
But then I remembered the Centos7 buffers. Current (and Centos7 default) net.ipv4.tcp_wmem is "4096 16384 4194304", after change it via sysctl to   '8194304 8194304 8194304' all works fine...

miconda:
OK, good it was sorted out and thanks for sharing the solution.

I added a section about network buffers size in the tcp tuning file:

  * https://github.com/kamailio/kamailio/commit/38a696fff66f0a453e54c92c93e8c459a1b4e77d

Closing this one.

Issue #1378, opened 2017-12-20
acc: set cdr_extra values to NULL


cruzccl:
### Description

ACC module [cdr_extra modparam](http://kamailio.org/docs/modules/devel/modules/acc.html#acc.p.cdr_extra) allows saving additional custom values in cdr entries using dlg_vars.

When you define a new value to be saved:

`modparam("acc", "cdr_extra", "c1=$dlg_var(caller)")`

and you set a $dlg_var(c1) in the logic, everything works as expected.

But if you set it to $null or you don't even define it, it is saved as '' (empty string).

Would be nice that setting to $null would store the underlying null value (NULL in SQL database) and not defining it would lead to not even including it in the final INSERT?

#### Reproduction

Define a new value to be saved:

`modparam("acc", "cdr_extra", "c1=$dlg_var(caller)")`

And in the kamailio.cfg logic:

- If you set $dlg_var(c1) to any string ('', 'foo', '2'), it is saved as expected ('', 'foo', '2').

- If you set $dlg_var(c1) to an integer different from 0 (1, 2, 3), it is saved as a string containing that number ('1', '2', '3'). This seems normal as dlg_vars can only store strings.

- If you set $dlg_var(c1) to 0, it is saved as an empty string (''). Seems normal too.

- If you set $dlg_var(c1) to $null, it is saved as an empty string ('').

- If you don't define $dlg_var(c1) at all, it is saved as an empty string ('').

#### Debugging Data

Both [_extra2strar_](https://github.com/kamailio/kamailio/blob/e101f21c37238862a474c50218d11a475d121ee0/src/modules/acc/acc_cdr.c#L223) and [_extra2strar_dlg_only_](https://github.com/kamailio/kamailio/blob/e101f21c37238862a474c50218d11a475d121ee0/src/modules/acc/acc_cdr.c#L231) functions defined in [_acc_extra.c_](https://github.com/kamailio/kamailio/blob/master/src/modules/acc/acc_extra.c) set _cdr_type_array_ entry to TYPE_NULL, but that array is not used in subsequent [for loop](https://github.com/kamailio/kamailio/blob/e101f21c37238862a474c50218d11a475d121ee0/src/modules/acc/acc_cdr.c#L239).

### Possible Solutions

[...]

PR #1385, opened 2017-12-22
acc: Allow nullable cdr extra fields

cruzccl:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #1378 

#### Description

- allow custom cdr fields defined with cdr_extra modparam to be saved
  as null. Before this change, cdr_extra variables not defined in config
  operation (or set to $null) were saved as empty string (''). This commit makes
  possible to save them as NULL value.

- new modparam to make this behaviour configurable: cdr_extra_nullable.
  Set it to 1 to enable this new behaviour (default value: 0)


miconda:
Thanks!

PR #1392, opened 2018-01-04
Minor improvements to RPM packaging

que273:
### Description
kamailio.spec: Move the ims_ocs module into the ims RPM. It has a dependancy on libkamailio_ims.so

kamailio.init: Pass the .cfg file as a parameter when starting or syntax checking. Use the pidfile to kill only the relevant instance.
This makes it easier to reuse the same script when running multiple kamailio instances on the same server. Only the environment vars at the top need to be changed.


miconda:
Thanks!

PR #1399, opened 2018-01-14
Dispatcher set_state "all"

tverlaan:
#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Adds a new option to RPC command `dispatcher.set_state` to bulk update all destinations as part of a dispatcher set. The way we use dispatcher sets this comes in handy in a few cases.


miconda:
Thanks, I am merging it, then adjusting a bit the condition to match the `all` string.

tverlaan:
great feedback! thanks!

PR #1402, opened 2018-01-15
presence: dmq integration

charlesrchance:
Enabled via module parameter.

First commit:

- initial implementation
- replication of presentity updates over DMQ
- adds ruid column for matching across cluster

Tested and in use across small production cluster (‘dialog’ and ‘message-summary’ events).

Need to update docs prior to merging but raising PR first to hear any feedback.

### DIFF
--- a/src/lib/srdb1/schema/pr_presentity.xml
+++ b/src/lib/srdb1/schema/pr_presentity.xml
@@ -9,7 +9,7 @@
 
 <table id="presentity" xmlns:db="http://docbook.org/ns/docbook">
     <name>presentity</name>
-    <version>4</version>
+    <version>5</version>
     <type db="mysql">&MYSQL_TABLE_TYPE;</type>
     <description>
 		<db:para>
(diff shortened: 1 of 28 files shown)


charlesrchance:
Also need to update database schema definition for presentity table.

eschmidbauer:
just curious, why would you want to replicate the presentity records?
Wouldn't it make more sense to replicate the active_watchers records? Much like how `usrloc_dmq`  replicates location data (which enables an INVITE to be sent to a user from any kamailio machine in the cluster), replicating the active_watchers data would allow a NOTIFY to be sent out from any kamailio machine in the cluster.

charlesrchance:
It’s a valid alternative, and one which works well if the subscribers are not in direct contact with the presence servers and instead there’s a load balancer in-between.

Where subscribers are in direct contact and also behind NAT the NOTIFYs will need to go via the server on which the subscription was received.

Also, it is assumed that there are typically more subscriptions than presentities.

This not only adds more internal traffic (proportional to the number of nodes), but also a dependency on each server for the lifetime of its ‘own’ watcher records.

If each node is responsible for its own subscribers, however, then if it disappears the subscribers simply move elsewhere (as a result of dead keepalive)


 The main purpose of the replication is to enable each

charlesrchance:
Closed in error!

miconda:
I think the new parameter is not documented, can be done in a follow up commit.

If there are no other comments from devs, then it can be merged.

charlesrchance:
Thanks - I planned to document the parameter prior to merging.

The DB schema definition for presently table needs updating too - where should this be done?

miconda:
For changes to db schema, you have to update the xml files from `src/lib/srdb1/schema/`.

The you can run `make dbschema` in the root folder of kamailio source code tree and the sql/db scripts for `kamdbctl` will be updated (they have to be committed, too).

charlesrchance:
Thanks!

charlesrchance:
Merging now, having received no further comments from devs. Any issues, please raise in the usual manner.

PR #1404, opened 2018-01-19
json: add transformation

eschmidbauer:
this was borrowed from the `kazoo` and `nsq` module json transformations
i think it makes the most sense in this module as I'd like to use the transformation without loading the `kazoo` or `nsq` modules.
It can be very useful for parsing json data from pseudo-variables like $rb.
if this can be merged, I will deprecate the transformation in `nsq` and maintain this code.

### DIFF
--- a/src/modules/json/doc/json_admin.xml
+++ b/src/modules/json/doc/json_admin.xml
@@ -74,5 +74,31 @@ xlog("foo is $var(foo)");
 	    </example>
 	</section>
 	</section>
+
+	<section>
+	<title>Transformations</title>
+	<para>You can use the transformation to extract values from the json structured pseudo-variables</para>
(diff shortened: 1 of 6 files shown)


miconda:
Is the name of the transformation the same across these modules? If yes, that creates a conflict if all modules are going to be loaded -- I understand you are going to remove from nsq, but there is also in kazoo.

eschmidbauer:
transformation is not the same -- it is entirely changed to not conflict with `kazoo` or `nsq`
example transformations:
 -  json module: `$(rb{json.parse,Event-Package}`
 - nsq module: `$(rb{nsq.json,Event-Package}`
 - kazoo module: `$(rb{kz.json,Event-Package}`

Issue #1391, opened 2018-01-04
http_async_client: coredump on 5.0.5


linuxmaniac:
### Description

I'm having crashes on the Http worker process. I didn't find a way to reproduce the crash yet. Crashes are happening on two servers after a couple of days.

#### Debugging Data

Not very useful

```
Core was generated by `/usr/sbin/kamailio -f /etc/kamailio/kamailio.cfg -P /var/run/kamailio/kamailio.'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  __GI__IO_wfile_underflow (fp=0x7f45900008c0) at wfileops.c:280
280	wfileops.c: No such file or directory.
(gdb) bt
#0  __GI__IO_wfile_underflow (fp=0x7f45900008c0) at wfileops.c:280
#1  0x00000000000004b0 in ?? ()
#2  0x0000000000000428 in ?? ()
#3  0x00007f459d9e2640 in ?? ()
#4  0x0000000000000000 in ?? ()
(gdb)
```

#### Log Messages
I have debug enabled for http_async_client module, and this is the last messages before the crash
```
Jan  3 11:02:11 vox-kamailio-1887-0 kamailio[2685]: DEBUG: http_async_client [http_multi.c:67]: timer_cb(): timeout on socket -1
Jan  3 11:02:11 vox-kamailio-1887-0 kamailio[2685]: DEBUG: http_async_client [http_multi.c:53]: multi_timer_cb(): multi_timer_cb: Setting timeout to 1 ms
Jan  3 11:02:11 vox-kamailio-1887-0 kamailio[2685]: DEBUG: http_async_client [http_multi.c:561]: check_multi_info(): REMAINING: 2
[...]
Jan  3 11:02:15 vox-kamailio-1887-0 kamailio[2702]: CRITICAL: <core> [core/pass_fd.c:277]: receive_fd(): EOF on 89
Jan  3 11:02:15 vox-kamailio-1887-0 kamailio[2618]: ALERT: <core> [main.c:743]: handle_sigs(): child process 2685 exited by a signal 11
Jan  3 11:02:15 vox-kamailio-1887-0 kamailio[2618]: ALERT: <core> [main.c:746]: handle_sigs(): core was generated
```

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
version: kamailio 5.0.5 (x86_64/linux) 
[...]

PR #1405, opened 2018-01-22
http_async_client: cleanup avps/xavps after exec route on Http worker

linuxmaniac:
when suspend_transaction is false

As suggested at #1391 to avoid shared memory leak


grumvalski:
Thanks!

Issue #1208, opened 2017-08-08
Kamailio 4.4.6 crashes in http async worker process


jonastelzio:
### Description
```
Kamailio crashes out with message: CRITICAL: <core> [pass_fd.c:277]: receive_fd(): EOF on 30
```

#### Reproduction

Happens sporadicly. Some times 6 times in a day. Some times it won't happen for a week. But on average it probably happens once per day.

#### Debugging Data

```
Reading symbols from /usr/local/sbin/kamailio...done.
[New LWP 29579]
[New LWP 9469]
[New LWP 29581]
Missing separate debuginfo for /usr/lib64/libpq.so.5
Try: yum --enablerepo='*debug*' install /usr/lib/debug/.build-id/00/595f0be1374285c79ea1d4cdcee72682108fab.debug
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Missing separate debuginfo for /usr/lib64/libstdc++.so.6
Try: yum --enablerepo='*debug*' install /usr/lib/debug/.build-id/43/4c339faa62ca4e59eb71572a60967fe54a69ad.debug
Missing separate debuginfo for /lib64/libgcc_s.so.1
Try: yum --enablerepo='*debug*' install /usr/lib/debug/.build-id/3f/d5f89de59e124ab1419a0bd16775b4096e84fd.debug
Core was generated by `/usr/local/sbin/kamailio -P /var/run/kamailio.pid -m 256 -M 8 -u kamailio -g ka'.
Program terminated with signal 11, Segmentation fault.
#0  0x00007fc9a9350521 in _int_malloc () from /lib64/libc.so.6
[...]

PR #1407, opened 2018-01-23
http_async_client: memory mismanagement

dragos-oancea:
#### Pre-Submission Checklist

#### Type Of Change
- [ X] Small bug fix (non-breaking change which fixes an issue)

#### Checklist:

- [ ] Tested changes locally
- [ ] Related to issue #1208 

#### Description
p->headers.t is dynamically allocated header_list_add() in and should be freed with free_async_query().

### DIFF
--- a/src/modules/http_async_client/async_http.c
+++ b/src/modules/http_async_client/async_http.c
@@ -631,8 +631,6 @@ void init_query_params(struct query_params *p) {
 
 void set_query_params(struct query_params *p) {
 	int len;
-	p->headers.len = 0;
-	p->headers.t = NULL;
 	p->tls_verify_host = tls_verify_host;
 	p->tls_verify_peer = tls_verify_peer;
 	p->suspend_transaction = 1;


dragos-oancea:
@grumvalski , set_query_params() is called in 3 places, for us kamailio is crashing at startup, and I suspect  header_list_add(&ah_params.headers, &tval->rs); is sometimes called before set_query_params(&ah_params) in ah_set_req() , thus leading to a kamailio crash at startup, under traffic.

grumvalski:
set_quey_params is called in mod_init by init_query_params (https://github.com/kamailio/kamailio/blob/master/src/modules/http_async_client/async_http.c#L631), so whenever ah_set_req is called the query_params struct should already have been initialized.

headers.t is a double pointer, the allocated memory is freed later by free_async_query, we are not freeing it inset_query_params. If we don't set it to NULL we end up with a growing headers list (from different request) and we go toward a crash being the memory freed elsewhere.

This being said, do you have a coredump of the crashes? Which OS/libcurl version are you using? Recently a crash at startup had been reported which was due to a combination of libcurl/OS libs: https://github.com/kamailio/kamailio/issues/1391.

dragos-oancea:
is it possible is being set to NULL before headers.t being freed in some case ? I noticed there are checks inside  set_query_params() for other already allocated memory buffers  . It's crashing randomly at startup, in this scenario, I dont have backtraces anymore.

miconda:
@dragos-oancea, @grumvalski - did this one ended in some conclusion? Merging or not?

giavac:
FYI this was kamailio `5.1.0` on debian 8, with stock libcurl (`7.38`).
I'm trying to reproduce the issue with `5.1.2` and both libcurl `7.38` and `7.59` (still on debian 8) (including with restarts under load), and I will update here with any relevant information.

miconda:
Pinging again to see if anything else was left to be addressed here.

henningw:
@giavac - any updates on this? Were you able to reproduce this with a newer 5.1.x version?

giavac:
I've been able to reproduce the crash that lead to Dragos analysis and this PR, both in debian 8 and in CentOS 7.

But I tracked it down to a different cause, which is `libcurl` crashing when using the threaded resolver for DNS lookups. This happens even with latest `libcurl`.

The work around this was building `libcurl` with `c-ares` resolver, and that proved to solve the issue. This included tests under heavy load (hundreds of requests per second).

I meant to write my findings to the libcurl mailing list and ask for corroboration or indications, but I haven't had time yet, as bringing up the failing and working scenarios in a clean way to be shared with that ml is very time consuming.

For what concerns this specific Pull Request I haven't done any test/verification. I don't have it in the environments with the c-ares fix and there are no crashes, so perhaps it's not needed.
The only thing I can suggest is that next time I do the load testing, I use the patch in this PR to verify there are no side effects.

miconda:
@grumvalski , @giavac , @dragos-oancea - can this be closed or there is still something here to work on?

PR #1408, opened 2018-01-24
json_pua: new module

eschmidbauer:
this module adds a new function `json_pua_publish()` which allows json data objects to be parsed and inserted into the presentity table. 
the function is borrowed from the nsq and kazoo modules. it allows presence to be controlled without loading the nsq or kazoo modules. 
this can be quite powerful as it will allow users to control presence by sending http json data objects to kamailio, for example:
```
event_route[xhttp:request] {
    $var(call-id) = $(rb{json.parse,Call-ID});
    if ($(rb{json.parse,Event-Package}) == "dialog") {
            json_pua_publish($rb);
            pres_refresh_watchers("$(rb{json.parse,From})", "$(rb{json.parse,Event-Package})", 1);
    }
    xhttp_reply("200", "OK", "text/html", "$rb");
}
```

example curl request to send presence update:
```
curl -d '{"Call-ID":"user01@my.domain.io","Event-Category":"presence","Event-Name":"update","Event-Package":"dialog","Expires":"3600","From":"sip:user01@my.domain.io","From-User":"user01","From-Realm":"my.domain.io","To":"sip:park+6003@my.domain.io","To-User":"park+6003","To-Realm":"my.domain.io","State":"confirmed"}' http://localhost:8080/presence/
```

### DIFF
--- a/src/modules/json/json_funcs.c
+++ b/src/modules/json/json_funcs.c
@@ -270,39 +270,6 @@ int tr_json_get_field(struct sip_msg *msg, char *json, char *field, char *dst)
 	return 1;
 }
 
-struct json_object *json_parse(const char *str)
-{
-	struct json_tokener *tok;
-	struct json_object *obj;
-
-	tok = json_tokener_new();
-	if(!tok) {
-		LM_ERR("Error parsing json: could not allocate tokener\n");
-		return NULL;
-	}
-
-	obj = json_tokener_parse_ex(tok, str, -1);
-	if(tok->err != json_tokener_success) {
-		LM_ERR("Error parsing json: %s\n", json_tokener_error_desc(tok->err));
-		LM_ERR("%s\n", str);
-		if(obj != NULL) {
-			json_object_put(obj);
-		}
-		obj = NULL;
-	}
-
-	json_tokener_free(tok);
-	return obj;
-}
-
-struct json_object *json_get_object(
-		struct json_object *jso, const char *key)
-{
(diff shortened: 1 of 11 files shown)


miconda:
Thanks, sounds interesting, but I have couple of observations:

  * the module should be named **pua_json** -- being related to presence services, we put first the services group token (see how are the other modules are named, like: presence_*, pua_*, db_*, ndb_*)
  * there are a lot of unsafe`sprintf()`, they should be replaced with `snprintf()` -- that will also give the len of the output string, `strlen()` no longer being needed
  * if I got it right, the module works directly with the database table `presentity` from presence modules, (re)defining variables with column names. This is not maintainable in long term, if someone changes the structure of the table in presence module, will have to change in other modules. I think it is better to add a function to presence module for what you need here and export it to the presence api, which in turn is going to be used in this module. The presence module is already exporting an API, so you can just add a new member there
  * for API functions, I suggest to define a structure instead of a long list of parameters, because if a new value needs to be inserted, the prototype of the API function stays unchanged, only a new member is added to the structure
  * also the way functions from json module are used inside this module are exposing a risk over the time. You declared them static inline, but if one needs to make them more complex in the future, with use of additional functions inside, then things are going to break. I suggest to do the same approach with exporting the functions in json module through an API, which is imported inside the new module

eschmidbauer:
I completely agree with all of your suggestions -- but most those changes will take me some time to implement. I'll close this PR for now

PR #1409, opened 2018-01-25
pua: removed null pointer access from subscribe_cbparam()

christoph-v:
A newly introduced condition in an if () statement leads to core
dump in some scenarios. It is removed.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Without this Change a core dump happens in the P-CSCF during handling the SIP REGISTER.
Obviously a condition was added to an if() Statement, which itself created a null pointer Access.
This additional Statement was removed.

### DIFF
--- a/src/modules/pua/send_subscribe.c
+++ b/src/modules/pua/send_subscribe.c
@@ -821,7 +821,7 @@ ua_pres_t* subscribe_cbparam(subs_info_t* subs, int ua_flag)
 	{
 		CONT_COPY(hentity, hentity->id, subs->id);
 	}
-	if(subs->extra_headers && hentity->extra_headers->s)
+	if(subs->extra_headers)
 	{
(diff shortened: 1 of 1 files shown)


miconda:
Thanks for finding this, looks like a copy&paste issue, so I think the condition should not be removed but fixed -- instead of:

```
if(subs->extra_headers && hentity->extra_headers->s)
```

There should be:

```
if(subs->extra_headers && subst->extra_headers->s)
```

christoph-v:
Hi,

Can you do it the way you proposed? I’d have to create a new branch for this anyway.

KR

miconda:
Done in master, I will backport soon.

christoph-v:
Will you backport to 5.1 ? Should I create a 5.1 pull request?

PR #1422, opened 2018-01-30
Modules: db_redis for usrloc module

SurendraPlivo:
This module provides db api interface to redis database

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally


SurendraPlivo:
@agranig @miconda

SurendraPlivo:
@Vadiml1024 first of all thank you for review .

will you please help me out by reentrancy by producing it ? because we tested it for more than 300 rps and with 70000 endpoint. we didn't face this issue.anyway its good if you can point me to specific case.

in some places i see snprintf where source string size is passed as 2 parameter (buffer size). The idea is to pass target buffer size as second parameter ???? will you please point out these places.

miconda:
The module is very tight to usrloc module, embedding even the default column names for location table. If this is going to be merge in this state, I would suggest to name it like `db_redisuloc` or something else that suggests the constraints with its usage. Later can be renamed.

Regarding the issue with snprintf size parameter reported above, I spotted like:

```
return_code = snprintf(table_name,(CON_TABLE(_h)->len+1) * sizeof(char),"%s%.*s%s",CON_TQUOTESZ(_h), CON_TABLE(_h)->len, CON_TABLE(_h)->s, CON_TQUOTESZ(_h));
```
Another aspect that needs to be take care is to check if pkg_malloc() returns NULL or not -- if there is no sufficient memory, it will cause a crash.

PR #1424, opened 2018-02-01
pua_json: new module to update presence using JSON data objects

eschmidbauer:


src/modules/pua_json/pua_json_publish.c
```diff
@@ -0,0 +1,387 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ *
+ * Contributor(s):
+ * Emmanuel Schmidbauer <eschmidbauer@gmail.com>
```
lazedo: @eschmidbauer sorry for this.. but, what % of this code is actually `only` yours ? same thing in other files on this `new` module. don't get me wrong, i approve this move from a `kazoo` centric solution to a generic one and i'm glad you found the time to make this happen.

eschmidbauer: If you look at line 4 of the files that were mostly borrowed from kazoo, i included:
``` * Copyright (C) 2010-2014 2600Hz ```

miconda:
This can be merged, but then remove the Contributors from C code files (can be a follow up commit) -- we removed history/contributors in most of the code files, with the goal to do it everywhere, so make no sense to add new ones. The git commit can be used to see the contributions over the time and explicit listing of names/companies can be done in the README file. Having in C code file can end up with a large listing over many years.

PR #1431, opened 2018-02-07
schema: allow null value for active_watchers user-agent column

lazedo:
some UAs not sending User-Agent header on SUBSCRIBE causes insert error

needs backport to 5.0/5.1


miconda:
Why the default value is no longer left to '' (empty string)? Allowing NULL should be ok also with having a default value.

Then, changing the database structure for this situation is not going to be backported, all updates of minor 5.0.x and 5.1.x will fail to just restart.

I think a better option is just to allow NULL, let the default to be '' and eventually fix in the code if there is a case in 5.0.x or 5.1.x when NULL is inserted and insert '' -- can you point in the code where the issue was exposed for you? Anyhow, this won't change the structure dramatically to require a version number update.

lazedo:
@miconda will update. so by leaving the default we can backport ?

miconda:
First let's see where is the problem in the code and if there are other solutions. Can you point in the code where you got the issue so we can analyze better?

lazedo:
@miconda here https://github.com/kamailio/kamailio/blob/master/src/modules/presence/subscribe.c#L183
UA is not sending user-agent header in subscribe and its not mandatory in [rfc3265 ](https://www.ietf.org/rfc/rfc3265.txt), the schema should follow the standard

miconda:
I pushed a commit to cope with the missing value for user_agent: 4654f8b4f341a6c9f8cf4daafc7d618871c95e1b

It should make it work even without database schema update.

I am merging the PR, we can consider even backporting the PR, but should not make an upgrade in 5.0 or 5.1 stable series to require changing the database schema.

lazedo:
@miconda great. should we backport that commit, this PR or that commit & this PR ?

lazedo:
@miconda just noticed that you already backported https://github.com/kamailio/kamailio/commit/4654f8b4f341a6c9f8cf4daafc7d618871c95e1b . i would like to backport this PR if that's ok

PR #1432, opened 2018-02-07
db_redis: Implement db_redis generic db driver

agranig:
This module implements a generic db driver for kamailio. It
requires a "schema" and "key" definition of "tables" and corresponding
keys for redis in the kamailio config file, otherwise it's supposed to
work with every module.

Implemented methods are query (w/o order-by), insert, update, delete.

Tested with usrloc and acc.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #1137 

#### Description

This module implements a generic db driver for redis, to be used by any module (with the help of some schema configuration in the config file).

### DIFF
--- a/doc/stylesheets/dbschema_k/xsl/db_redis.xsl
+++ b/doc/stylesheets/dbschema_k/xsl/db_redis.xsl
@@ -1,0 +1,126 @@
+<?xml version='1.0'?>
+<!--
+ * $Id: $
+ *
+ * XSL converter script for db_redis
+ *
+ * Copyright (C) 2001-2007 FhG Fokus
+ *
+ * This file is part of Kamailio, a free SIP server.
+ *
+ * Kamailio is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version
+ *
+ * Kamailio is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ */
+-->
+
+
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+                version='1.0'
+                xmlns:xi="http://www.w3.org/2001/XInclude">
+
+    <xsl:import href="common.xsl"/>
+    <xsl:output method="text" indent="no" omit-xml-declaration="yes"/>
+
+    <!-- Create the file for the table in dbtext subdirectory -->
+    <xsl:template match="table">
+	<xsl:variable name="name">
+	    <xsl:call-template name="get-name"/>
+	</xsl:variable>
+
+	<xsl:variable name="path" select="concat($dir, concat('/', concat($prefix, $name)))"/>
+	<xsl:document href="{$path}" method="text" indent="no" omit-xml-declaration="yes">
+		<xsl:apply-imports/>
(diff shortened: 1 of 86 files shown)


miconda:
IMO, this one can be merged.

I haven't had the time for an in depth analyzis, spotted a small type in the readme, ndb_redis can be used to send any command to redis.

Then, as you expressed in some comments, having schema definition in a file may make possible to generate it from the xml db schema definitions from src/lib/srdb1/schema, in a simialar matter done for `db_text` or `db_mongodb`. If the parameter is kept, might be good to use function callbacks for it in order to allow setting it many times without overwriting previous value, otherwise its value can become too big for a single modparm (e.g., like done for htable parameter for htable module). Anyhow, these are enhancements that can be done over the time, as I said, merging at this phase is ok from my point of view, it will make it easier for others to test and improve its stability.

agranig:
Hi @miconda, I've improved the module with the suggestions you made:

* utilize xml schema definitions and load them from files to avoid having to define them as mod params
* allow multiple keys mod params definitions to make config more readable
* reconnect improvements
* documentation fixes and improvements

miconda:
A quick note that I updated the path for including hiredis.h:

```
diff --git a/src/modules/db_redis/redis_connection.h b/src/modules/db_redis/redis_connection.h
index d6ed4e4ec3..35d58e46a4 100644
--- a/src/modules/db_redis/redis_connection.h
+++ b/src/modules/db_redis/redis_connection.h
@@ -23,7 +23,7 @@
 #ifndef _REDIS_CONNECTION_H_
 #define _REDIS_CONNECTION_H_
 
-#include <hiredis.h>
+#include <hiredis/hiredis.h>
 
 #include "db_redis_mod.h"
```

On Debian 9 is where the file is located based on pkg-config output, it doesn't compile otherwise (same on MacOS with macports).

I looked in ndb_redis module and has the same path. If you have it in your system working without `<hiredis/hiredis.h>`, then we need to do some defines to set the proper path so it works on common distros.

Thanks again for this contribution!

zenichev:
The relevant question - how to implement authentication while connecting to redis storage?
I read this document: https://kamailio.org/docs/modules/devel/modules/db_redis.html
and didn't find any reference in it.

As I understood we use following syntax to connect to the redis db:
redis://[username]@host:port/database

But authentication capability is not provided as I can see.

Of course there is an excellent module named ndb_redis.
But my main goal is to perform authentication with cnxcc module, which is not able to do that (Carlos Ruiz Díaz will add this possibility later?).

Please correct me if I'm wrong.

agranig:
Try redis://password@host:port/db and let me know if it works for you.

zenichev:
Well, for that moment it doesn't work.

My configuration is as follows:

#!define DBURL_CNXCC "redis://here_my_password@redis.domain.com:6379/1"
loadmodule "db_redis.so"
modparam("db_redis", "schema_path", "/usr/share/kamailio/db_redis/kamailio")

modparam("cnxcc", "redis", DBURL_CNXCC)

I think this is wrong way to do that.
cnxcc module uses redis db as htable and stores real-time values (like credit, call duration etc.) that exist only when call is up for certain customer, then it drops whole key.

Should I use: 'modparam("db_redis", "keys"...'  ?
Or db_redis is to be improved to work with cnxcc? (cuz it uses not "db_url" modparam but "redis" modparam).

miconda:
@zenichev - cnxcc connects directly to redis, doesn't use a DB connector at all, such as db_mysql or db_redis. In other words, there is no relation between db_redis and cnxcc.

zenichev:
@miconda - as I mentioned above, I need working authentication for cnxcc (that is not able to work with authentication).
So that, I'm trying to find a way to do that, this time by db_redis.
Do you think this is a wrong way?

miconda:
@zenichev - as said, db_redis cannot be used with cnxcc. Start a discussion on sr-users@lists.kamailio.org with what you need to achieve. The issue tracker is only for bug reports, pull requests should be discussed in the context of the committed code: to review, improve, etc... it is not how to use the new features.

zenichev:
@miconda Ok I understood it clearly. Thank you for your answer.
Will try to start discussion on sr-users list.

PR #1435, opened 2018-02-09
presence: delay freeing generated etag string

charlesrchance:
This is a minor bugfix for earlier commit (b7a5016).


miconda:
It can be merged.

PR #1436, opened 2018-02-11
ndb_redis: dynamic discovery of nodes

giavac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This change provides support in `ndb_redis` for a scenario where the Redis server is part of a cluster _and_ the IP addresses of the nodes are either not known in advance or are subject to change (for example in case the cluster sees new nodes being added).
Before this change `ndb_redis` supports the `cluster` option, so that after a `MOVED` response from a node `ndb_redis` will perform again the same query but towards the redirection target. The caveat is that `ndb_redis` must know in advance which IP addresses will be used by the nodes.
A`MOVED` response is in the form `-MOVED 1234 1.2.3.4:6379.`, indicating the hash for the request and the redirection target IP address and port.
Every node in the cluster must have a modparam entry with `name=IP_ADDRESS:PORT;addr=IP_ADDRESS;port=PORT`.
With the changes in this PR, ndb_redis doesn't need to have in advance the list of nodes but it can "discover" them at runtime.
There is a caveat to this solution: while defining the nodes in advance allows to define also `pass`, `sock` and `db`, this is currently not supported (and default values are used). If this change will be accepted though, we could add a module parameter to define `pass`, `sock` and `db` for the dynamic nodes.

### DIFF
--- a/src/modules/ndb_redis/doc/ndb_redis_admin.xml
+++ b/src/modules/ndb_redis/doc/ndb_redis_admin.xml
@@ -271,6 +271,36 @@ modparam("ndb_redis", "flush_on_reconnect", 1)
 			</programlisting>
 		</example>
 	</section>
+	<section id="ndb_redis.p.allow_dynamic_nodes">
+		<title><varname>allow_dynamic_nodes</varname> (integer)</title>
+		<para>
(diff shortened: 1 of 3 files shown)


miconda:
Thanks!

PR #1440, opened 2018-02-14
acc_json: adding module

jchavanton:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
See comments in this PR, https://github.com/kamailio/kamailio/pull/1437
new features are now implemented in the module `acc_json`

#### Load tested
```
Total 1000000 INVITE calls sent in 2294479 ms at rate of 435/sec
Total 1000000 responses received in 2295273 ms at rate of 435/sec:

Detailed responses received:
 - 200 responses:  1000000     (OK)
                    ------
 TOTAL responses:  1000000 (rate=435/sec)
```

### DIFF
--- a/src/Makefile.groups
+++ b/src/Makefile.groups
@@ -21,7 +21,7 @@ mod_list_basic=async auth benchmark blst cfg_rpc cfgutils corex counters \
 mod_list_extra=avp auth_diameter call_control call_obj dmq domainpolicy msrp \
 			     pdb qos sca seas sms sst timer tmrec uac_redirect xhttp \
 				 xhttp_rpc xprint jsonrpcs nosip dmq_usrloc statsd rtjson \
-				 log_custom keepalive ss7ops app_sqlang acc_diameter evrexec
+			  log_custom keepalive ss7ops app_sqlang acc_diameter acc_json evrexec
 
 # - common modules depending on database
 mod_list_db=acc alias_db auth_db avpops cfg_db db_text db_flatstore \
--- a/src/modules/acc_json/Makefile
+++ b/src/modules/acc_json/Makefile
@@ -1,0 +1,29 @@
+# acc module makefile
+#
+#
+# WARNING: do not run this directly, it should be run by the master Makefile
+
+include ../../Makefile.defs
+auto_gen=
+NAME=acc_json.so
+LIBS=
+
+ifeq ($(CROSS_COMPILE),)
+JSON_SUPPORTED=$(shell \
+	if pkg-config --exists jansson; then \
+		echo 'libjansson found'; \
+	fi)
+endif
+ifneq ($(JSON_SUPPORTED),)
+	DEFS+= $(shell pkg-config --cflags jansson)
+	LIBS+= $(shell pkg-config --libs jansson)
+else
+	DEFS+=-I$(LOCALBASE)/include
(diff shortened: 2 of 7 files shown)


miconda:
Thanks! I am going to merge it and do some small adjustments afterwards.

I will also grant you commit permissions to maintain this module directly, without a need for pull requests. For other modules or parts of Kamailio, pull requests should still be used, to allow the original developer as well as the other developers to review/comment.

lazedo:
this is a great addition but i think its restricted in the output.
`mqueue` and `syslog` shouldn't be here. the module should emit an event with the json payload as a new pseudo var,  and in the script we could the use mqueue, syslog or whatever the integrator/developer wishes to do with json payload.

miconda:
You can add a modparam to tune its behaviour and act as you want. Each developer is guided mainly by what he/she needs. What you suggest is also interesting to have and iirc  it was mentioned in the previous related PR, suggesting that future improvement, which can allow pushing out of kamailio via http or other channel.

Otherwise, generic operations/layers can add delays, e.g., in this case writing first in a pv buffer (private memory), then push it with another operation to mqueue (shared memory) or printed to syslog.

So, bottom line, one implementation doesn't exclude the other, they can co-exist.

jchavanton:
Hi, @lazedo, thank for the feedback, we could keep some default provided outputs as "ease of use" default solutions and add a generic PV output or another module API.

I can share my thoughts, using mqueue we can already do routing script operation without blocking the worker process. (reusing an existing pattern in Kamailio)
As you have seen I provide an example with HTTP, tests have confirmed that this is very scalable with low latency local HTTP server. (especially with connection reuse recently introduced)
I did not test http_async not sure if it can be use in this scenario (I hope), but this would provide another very flexible solution to export using HTTP to anything with fluctuating latency.

@miconda
Thank you for the review and few fixes, I will gladly maintain this module and extensively test it shortly.

quentusrex:
Great work @jchavanton Glad to see this make it upstream.

jchavanton:
Packet Type 1 - load tests

```
Total 2500000 INVITE calls sent in 2522028 ms at rate of 991/sec
Total 2500000 responses received in 2522826 ms at rate of 990/sec:

Detailed responses received:
 - 200 responses:  2500000     (OK)
                    ------
 TOTAL responses:  2500000 (rate=990/sec)
```

rsyslog
```
wc -l /var/log/json_acc.log 
7500012 /var/log/json_acc.log
```
nsqd
```
nsqd v1.0.0-compat (built w/go1.8)
start_time 2018-02-16T22:36:24Z
uptime 2h1m51.901973386s

Health: OK

   [acc            ] depth: 0     be-depth: 0     msgs: 7500012  e2e%: 
      [nsq_to_file              ] depth: 7500012 be-depth: 7490012 inflt: 0    def: 0    re-q: 0     timeout: 0     msgs: 7500012  e2e%: 
```
voip_perf client
```
./voip_perf -m INVITE -p 7075 sip:+1??????????@10.10.10.10:5060 \
            --interval=1 \                                         
            --count=2500000 \                                      
            --call-per-second=1000 \                               
            --thread-count=1 \                                     
            --timeout 17200                  
```
voip_perf server
```
./voip_perf  -p 5072 --trying --ringing --thread-count=4 -d 500 
```

Latency graph showing stable response time
image

PR #1447, opened 2018-02-17
app_python3: first release

aalba6675:
First attempt at porting app_python to app_python3.

Mostly symbols names are the same as app_python so these two
modules cannot be loaded together. Exported functions are identical so
in theory scripts need not change.

TODO: reenable thread management

#### Pre-Submission Checklist

#### Type Of Change
- [ X] New feature (non-breaking change which adds new functionality)

#### Checklist:

- [x] Tested changes locally

#### Description
- add app_python3 module
- this is a direct port of app_python and aims for feature equivalence
- script reload does not work yet
- disable thread management; fortunately kamailio does not use threads; as the Py3 API is too different;
  trying to avoid deprecated APIs like PyEval_AcquireLock(). 
- TODO: need to reenable threads: use case is python extension module that uses native threads (e.g. gRPC)

### DIFF
--- a/src/modules/app_python3/Makefile
+++ b/src/modules/app_python3/Makefile
@@ -1,0 +1,29 @@
+# 
+# WARNING: do not run this directly, it should be run by the master Makefile
+
+include ../../Makefile.defs
+auto_gen=
+NAME=app_python3.so
+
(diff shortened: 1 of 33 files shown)


aalba6675:
Squash and rebase

PR #1449, opened 2018-02-18
app_python3: first release

aalba6675:
- use same symbols names as app_python so these two modules cannot be used together
- use GIL for thread management

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
- add app_python3 module
- this is a direct port of app_python and aims for feature equivalence
- script reload does not work yet
- use PyGIL* API for thread management: we are not using sub-interpreters, so this simplifies the locking

### DIFF
--- a/src/modules/app_python3/Makefile
+++ b/src/modules/app_python3/Makefile
@@ -1,0 +1,29 @@
+# 
+# WARNING: do not run this directly, it should be run by the master Makefile
+
+include ../../Makefile.defs
+auto_gen=
+NAME=app_python3.so
+
+# If you have multiple Python versions installed make sure to modify the
+# the following to point to the correct instance. Module has been tested
+# to work with 2.6 and 2.5. Python 2.4 has been only confirmed to compile,
+# but no testing has been done with that.
+PYTHON3?=python3
+
+PYTHON3_VERSION=${shell ${PYTHON3} -c "import distutils.sysconfig;print(distutils.sysconfig.get_config_var('VERSION'))"}
+PYTHON3_LIBDIR=${shell ${PYTHON3} -c "import distutils.sysconfig;print(distutils.sysconfig.get_config_var('LIBDIR'))"}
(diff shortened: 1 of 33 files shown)


miconda:
Thanks for this work!

There are some bits that still need to be updated, as I could spot in the documentation the name of the module is app_python, but it should be app_python3.

Since we are here discussing about a new app_python implementation, I wonder if still make sense to keep the old modules like Router, Logger (not sure if there are other). The should be now redundant with the KSR, which has lot more methods exported. I am not a Python user, so I cannot assert if something in old modules is missing in KSR, but in long term, we should remove the old ones anyhow, adding whatever is missing to KSR.

As you are working on this these days, would you be able to make the RPC command to reload the script working? I did a poor attempt to implement it, but it fails badly.

Another aspect I wanted to discuss with the people being familiar with the Python and its embedded API is whether it makes sense to create a SIP msg object each time there is a SIP request/reply handled by Kamailio. It feels like adding a bit of overhead in terms of performances. The alternative is to have KSR as a static/global object, so it doesn't need a constructor to initialize for each SIP message. Then the KEMI callback functions like ksr_request_route() to be global, not part of an object. Again, just an idea, not sure it matches how Python interpreter can be embedded or how it fits with its object oriented model. But Python seems rather popular in the context of Kemi, so I am throwing some ideas that could improve the performances.

Anyhow, I think the way to go is:

  * I am going to merge this PR, so if someone wants to keep using old modules, have a chance to revert if we go in a different direction
  * you will follow up with commits that will make `app_python3` be used everywhere it should be the module name
  * based on what people find acceptable/possible to implement, next steps will be to remove old modules and switch from a dynamic object for each SIP message to use of a static Kemi/KSR package.

aalba6675:
Thanks @miconda! I'll wait until it is merged to rebase and follow-up.

The immediate tasks will be to fix documentation, and have a go at reload. 

Will gather more developer and community feedback before proposing any
refactoring.

PR #1453, opened 2018-02-23
presence: test if event is dialog to avoid unnecessary xml parsing

claudiupb:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally


miconda:
I didn't have the time to check in the code, but if that part is used only for DIALOG events, then it can be merged.

claudiupb:
Looking through the code it appears that 'check_if_dialog' should be used only for DIALOG events, but maybe someone who knows the presence module better can also take a look just to be safe.

PR #1456, opened 2018-02-26
app_python3: implement script reloading

aalba6675:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Implement script reloading by copying the method used in app_lua: the version of the script is incremented by RPC. 

For each worker, the next time it handles a method, python_exec.c:apy_exec() 
will check for a newer version and reload, if necessary.

Set a lock so the script reload only occurs at depth 0 (in the unlikely case
that apy_exec() is called recursively).

This is not thread-safe as we are using a process-wide lock: don't call back
into apy_exec() from a Python C extension that uses threads.

For Pythonistas, we are reloading the module, not reinitializing the interpreter.
Standard caveats about reloading modules apply. E.g, references in the old module
that are not redefined in the newer module remain.

### DIFF
--- a/src/modules/app_python3/app_python_mod.c
+++ b/src/modules/app_python3/app_python_mod.c
@@ -51,7 +51,7 @@ static int mod_init(void);
 static int child_init(int rank);
 static void mod_destroy(void);
 
-PyObject *_sr_apy_handler_obj;
+PyObject *_sr_apy_handler_obj = NULL;
 
 char *dname = NULL, *bname = NULL;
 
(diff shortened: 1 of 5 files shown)


miconda:
Thanks! Merging it.

I am going to invite you in the dev team so you can push commits to `app_python3` directly. For other modules or components, still use pull requests. Of course, you can still use pull requests for `app_python3` in case you want others' opinion, review or comments.

PR #1467, opened 2018-03-01
 sqlops: sqlops_do_query - Make result parameter optional in C API

que273:
Make the result string parameter optional in the sqlops_do_query C API function

### DIFF
--- a/src/modules/sqlops/sql_api.c
+++ b/src/modules/sqlops/sql_api.c
@@ -703,8 +703,7 @@ int sqlops_do_query(str *scon, str *squery, str *sres)
 		LM_ERR("invalid connection [%.*s]\n", scon->len, scon->s);
 		goto error;
 	}
-	res = sql_get_result(sres);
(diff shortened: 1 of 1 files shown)


henningw:
Hello Hugh, thank you for the pull request. Wouldn't make sense to move the null check to the beginning of the function? Otherwise sql_get_result would call with a null before your check.

henningw:
Forget my previous comment - I think we should add a null check to sql_do_query as well.

que273:
sql_get_result does not currently check the 'name' parameter and will crash in core_case_hash if this is NULL. This is also true for sql_get_connection and sql_init_con. However, these three functions are only called from within the sqlops module.
I will add a check to the C-API/KEMI wrapper functions that call them (sqlops_do_query, sqlops_get_value, sqlops_is_null, etc.)? Is it needed inside sql_get_result as well?

henningw:
Merged and closed.

PR #1471, opened 2018-03-02
app_python: don't hide the real exception on load failure

aalba6675:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Calling PyErr_Print() clears the error indicator and hides the real cause of failing to load
the module as the output is lost (module has no stderr).

The error should be left to python_handle_exception() to display via LM_ERR.


miconda:
Thanks! @aalba6675 you can commit directly to app_python as well, there is no other very active developer for the module. If you really want a review for your commit, then it is ok to do PR as well. Otherwise, from my point of view, you are more familiar with embedding python interpreter.

Issue #1474, opened 2018-03-07
KEMI onsend_route method is not called


aalba6675:
### Description

KEMI onsend_route is not called unless there is a placeholder/dummy routing block.
```
## without this block, core will never call KEMI->onsend_route
onsend_route {
 ## do nothing
}
```
#### Reproduction
1. In KEMI create ksr_onsend_route(). Observe that this function is never invoked.

2. Add a route block:
```
onsend_route {
 ## do nothing
}
```
Now observe, that KEMI ksr_onsend_route() is called.

### Possible Solutions
1. Not really a problem, just a surprise to the KEMI user. Possibly, document that in order for the KEMI function to be called, the configuration file needs a dummy routing block for onsend_route.
OR
2. Move the check for `sr_kemi_eng_get()` outside the `if  (onsend_rt.rlist[DEFAULT_RT]){` check
### Additional Information
The reason is the following logic: 
onsend.h:
```
	if (onsend_rt.rlist[DEFAULT_RT]){
		onsnd_info.to=&dst->to;
		onsnd_info.send_sock=dst->send_sock;
```

`ret=keng->froute(orig_msg, ONSEND_ROUTE, NULL, NULL);`

This line is called only if `onsend_rt.rlist[DEFAULT_RT]` is non-NULL, but this array is initialized only if the `onsend_route` block exists.

  * **Kamailio Version** - 5.1, master

* **Operating System**:
Linux CentOS 7.4

aalba6675:
In a pure KEMI script environment, printing `onsend_rt.rlist` gives
```
Breakpoint 5, run_onsend (orig_msg=0x7f622816c8b0, dst=0x7ffe78d13560, buf=0x7f622816fcf8 "ACK sip:conf-095ed1b8-a768-4af4-8ba8-a56bc139574c@10.13.20.20:5090;transport=udp SIP/2.0\r\nVia: SIP/2.0/UDP 10.13.20.15;branch=z9hG4bKa2a.0092296879fe327cb9fbf6547f44d466.0;i=f\r\nVia: SIP/2.0/TLS 10.13."..., len=587) at core/onsend.h:73
73              if (onsend_rt.rlist[DEFAULT_RT]){
$2 = (struct action *) 0x0
```

miconda:
If you have time soon, you can propose a PR with the option 2. Otherwise I will look into it whenever I get a chance, likely not today and not sure for the rest of the week.

PR #1475, opened 2018-03-07
core: invoke KEMI ksr_onsend_route() unconditionally in cfgengine

aalba6675:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #1474 (replace XXXX with an open issue number)

#### Description

Currently we need a onsend_route{} (empty is ok) block in config before KEMI ksr_onsend_route() is invoked. The latter will override the route block.

This PR will unconditonally invoke KEMI ksr_onsend_route() so we do not need a placeholder in config.

For discussion: this will unconditionally invoke ksr_onsend_route() so it might break scripts that don't have an implementation. 

I'm not too clear about the error handling in run_onsend(): it's possible that if KEMI ksr_onsend_route throws error (not implemented rather than real error) then the message might be dropped, so this might surprise script writers. Since ksr_onsend_route() is optional it's hard to tell the difference between a real error and "not implemented is not an error".

### DIFF
--- a/src/core/onsend.h
+++ b/src/core/onsend.h
@@ -70,7 +70,9 @@ static inline int run_onsend(sip_msg_t* orig_msg, dest_info_t* dst,
 		return 1;
 	}
 	ret=1;
-	if (onsend_rt.rlist[DEFAULT_RT]){
+	// do if onsend_route{} or cfgengine exists
+	keng = sr_kemi_eng_get();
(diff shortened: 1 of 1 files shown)


miconda:
Thanks! I am merging it.

However, I am thinking of adding some core params to be able to disable execution of optional core routing blocks equivalents in kemi.

aalba6675:
Yes - I think that is a good idea.

PR #1480, opened 2018-03-12
usrloc: cleanup by server_id.

tiglat:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

In usrloc module we have server_id_filter parameter which switches on filtering of location records by server id during load. This fix improves this functionality to use this filter in cleanup procedure to avoid cleaning of expired db records which are written by another kamailio instance to keep their cache correct.

### DIFF
--- a/src/modules/usrloc/README
+++ b/src/modules/usrloc/README
@@ -1001,9 +1001,7 @@ modparam("usrloc", "rm_expired_delay", 30)
 
 3.47. server_id_filter (int)
 
-   Enable (1) or disable (0) filter records by server_id on load and during cleaning of expired db records.
-   It could be usefull when you want to use the same location table for several kamailio instances which are configured to work in db_mode=1 or db_mode=2 (cache modes).
-   Otherwise one instance of proxy cleans records made by another proxy and that breaks its cache.
+   Enable (1) or disable (0) filter records by server_id on load.
 
    Default value is “0”.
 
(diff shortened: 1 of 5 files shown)


src/modules/usrloc/README
```diff
@@ -1001,7 +1001,9 @@ modparam("usrloc", "rm_expired_delay", 30)
 
 3.47. server_id_filter (int)
 
-   Enable (1) or disable (0) filter records by server_id on load.
+   Enable (1) or disable (0) filter records by server_id on load and during cleaning of expired db records.
```
linuxmaniac: You should not change README directly
https://www.kamailio.org/wiki/devel/git-commit-guidelines#generic_guidelines

src/modules/usrloc/udomain.c
```diff
@@ -909,12 +910,22 @@ int db_timer_udomain(udomain_t* _d)
...
+	vals[2].val.int_val = server_id;
+
 	if (ul_dbf.use_table(ul_dbh, _d->name) < 0) {
 		LM_ERR("use_table failed\n");
 		return -1;
 	}
+	
+	if (ul_db_srvid != 0) {
+		key_num = 3;
+	}
```
miconda: I think it is better to move this `if` bloc at line 913 above and have those assignements for keys/ops/vals with index `[2]` inside it. It is better to see have related operations grouped together, avoiding also useless operations if ul_db_srvid is 0.

tiglat: Agree. Done.

miconda:
Thanks!

PR #1485, opened 2018-03-18
textops: Fixed remove_hf_exp arguments params

sergey-safarov:
Current remove_hf_exp not works as expected. This fix it.


miconda:
Thanks!

PR #1493, opened 2018-04-03
ndb_redis: adds support for sentinel

schmurfy:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Allows kamailio to connect to a sentinel to find out which redis instance to connect to.
I modified the README in this commit but this is the generated file, I did not change it by hand.


miconda:
Thanks! One question before attempting to merge, does this require a specific version of redis server or redis client library?

schmurfy:
After a quick look on redis website the sentinel is shipped with redis since 2.8 so I suppose this is the minimum required but if the sentinel options are not used the driver should work as before with any redis version.

Issue #1507, opened 2018-04-24
Kamailio doesn't start with db_mongodb and usrloc.


tiglat:
### Description
Kamailio doesn't start with db_mongodb and usrloc.

Configuration:
kamailio 4.4.7 (tried to use db_mongodb module from 5.1.2)
usrloc is configured to use db_mode=1 and db_insert_null=1.
MongoDB 3.0.2 and 3.6.3
mongo-c-driver was tested 1.9.2 and 1.9.4.
location table contains 300 records.

#### Reproduction

1. Configure db_mode=1 (use internal cache). 
2. Make 300 successful registrations to have 300 records inside mongodb. 
3. Restart Kamailio

#### Debugging Data

db_mongodb_base.c 
function db_mongodb_convert_bson()
line 614

#### Log Messages

```

64) DEBUG: usrloc [udomain.c:469]: preload_udomain(): loading records - cycle [1]
1(13464) CRITICAL: usrloc [udomain.c:476]: preload_udomain(): empty username record in table location...skipping
1(13464) CRITICAL: usrloc [udomain.c:476]: preload_udomain(): empty username record in table location...skipping
1(13464) DEBUG: <core> [socket_info.c:564]: grep_sock_info(): checking if host==us: 1==15 && [s] == [192.168.158.178]
1(13464) DEBUG: <core> [socket_info.c:583]: grep_sock_info(): checking advertise if host==us: 1==15 && [s] == [192.168.158.178]
1(13464) DEBUG: <core> [socket_info.c:564]: grep_sock_info(): checking if host==us: 1==15 && [s] == [192.168.158.178]
1(13464) DEBUG: <core> [socket_info.c:583]: grep_sock_info(): checking advertise if host==us: 1==15 && [s] == [192.168.158.178]
1(13464) DEBUG: <core> [socket_info.c:564]: grep_sock_info(): checking if host==us: 1==15 && [s] == [192.168.158.178]
1(13464) DEBUG: <core> [socket_info.c:583]: grep_sock_info(): checking advertise if host==us: 1==15 && [s] == [192.168.158.178]
1(13464) DEBUG: <core> [socket_info.c:564]: grep_sock_info(): checking if host==us: 1==15 && [s] == [192.168.158.178]
1(13464) DEBUG: <core> [socket_info.c:583]: grep_sock_info(): checking advertise if host==us: 1==15 && [s] == [192.168.158.178]
[...]

PR #1508, opened 2018-04-24
Replaced working with bson utf8 string inside db_mongodb_convert_bson function

tiglat:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #1507 

#### Description

During query operations string fields inside result structure db1_res_t could be corrupted because of incorrect usage of bson memory. 

This issue is reproduced when we need to read more than 150 documents from the database. 

Firstly bson_iter_utf8 function returns pointer to memory which is allocated by bson library. It looks like this memory is used multiple times (some kind of circular buffer). So the data must be copied into Kamailio memory instead storing that pointer. The original pointer becomes invalid after some time.

Secondly, this pointer must not be freed by the caller of bson_iter_utf8. At the current version bson allocated pointer is freed by Kamailio function pkg_free like it is was allocated by pkg_malloc.

### DIFF
--- a/src/modules/db_mongodb/mongodb_dbase.c
+++ b/src/modules/db_mongodb/mongodb_dbase.c
@@ -631,10 +631,11 @@ static int db_mongodb_convert_bson(const db1_con_t* _h, db1_res_t* _r,
                   (uint32_t*)&VAL_BLOB(dval).len, (const uint8_t**)&VAL_BLOB(dval).s);
 				break;
 
-			case BSON_TYPE_UTF8:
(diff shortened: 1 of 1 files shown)


miconda:
Thanks!

I applied a slightly different patch to check the return code of converting the UTF8 value to internal db val with cloning.

For future PRs, format the git commit lines as per contributing guide:

It makes merging straightforward via web interface, without manual patching to adjust the commit message.

PR #1516, opened 2018-04-29
nathelper: it doesn't ping contacts if ping_nated_only=0

tiglat:
- nathelper sends ping only if nat flag is specified, but if ping_nated_only=0 it should ping all contact in any case.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description
We need to check presence of nat flag with check of ping_nated_only parameter to avoid situation when flag is not specified by user wants to ping all contacts and don't get any pings.

### DIFF
--- a/src/modules/nathelper/nathelper.c
+++ b/src/modules/nathelper/nathelper.c
@@ -2059,10 +2059,9 @@ static void nh_timer(unsigned int ticks, void *timer_idx)
 		dst.proto = PROTO_UDP;
 		dst.send_sock = send_sock;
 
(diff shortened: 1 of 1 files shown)


miconda:
You have to declare the variable at the beginning of the block/function. Otherwise, strict C compilers can throw error. You can push a new patch to this PR, we can squash from web interface before merging.

tiglat:
Fixed my mistake - moved declaration to begin of the function.

miconda:
Thanks!

PR #1533, opened 2018-05-16
dispatcher: congestion detection load balancing

jchavanton:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Presented at Kamailio World 2018, load balancing doing congestion detection
https://www.youtube.com/watch?v=AD0BGaW1g58
 
More information can be found 
http://www.castmm.com/KW2018.pdf
https://github.com/jchavanton/kam_load_balancing

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -269,6 +269,9 @@ int ds_set_attrs(ds_dest_t *dest, str *attrs)
 	for(pit = params_list; pit; pit = pit->next) {
 		if(pit->name.len == 4 && strncasecmp(pit->name.s, "duid", 4) == 0) {
 			dest->attrs.duid = pit->body;
+		} else if(pit->name.len == 2
+				  && strncasecmp(pit->name.s, "cc", 2) == 0) {
+			str2sint(&pit->body, &dest->attrs.congestion_control);
 		} else if(pit->name.len == 6
 				  && strncasecmp(pit->name.s, "weight", 6) == 0) {
 			str2sint(&pit->body, &dest->attrs.weight);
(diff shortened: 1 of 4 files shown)


jchavanton:
Small modifications in this PR, improved the precisions when all Gw are congested, compare to the version I was using before the presentation, the graph in my presentation was showing a lack of precision, better now ! 

all_gw_congested_tuning

miconda:
Couple of things to clarify:

  * this is not a new algorithm, but an option to use with relative weights algorithm
  * I noticed a mutex (lock) was added per set -- it is required only in the context of using this new feature? Does it affect using relative weights without congestion control?
  * the cc attribute needs to be set in all destination records belonging to the same set when wanting to use this new feature?

jchavanton:
Thanks for the comments and questions.

Q: this is not a new algorithm, but an option to use with relative weights algorithm
A: This is correct, I think the documentation is accurate on this point, adding new one was not needed, let me know if you think this would be more clear ?

Q: I noticed a mutex (lock) was added per set -- it is required only in the context of using this new feature? Does it affect using relative weights without congestion control?
A: Correct, it is only needed with CC, because the SIP OPTIONS ping are processed by each worker and they can endup redistributing the weight simultaneously.
Do you think we should try make the lock optional (only enforced when absolutely needed) ?
My tought was that without CC it will be used very unfrequently when a GW times out.

A: the cc attribute needs to be set in all destination records belonging to the same set when wanting to use this new feature?
- correct this is more flexible I believe since you can activate congestion detection per gateway, if CC is not set the GW will always keeps its weight.

miconda:
OK, clarified. I am going to merge it.

PR #1535, opened 2018-05-17
str_hash: Option to allocate hash in shm

Elhodred:
Added an allocator function to allocate the hash in shared memory.
Added also locking stuff

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Added a new allocator function to str_hash.h. This new function allows to allocate the hash table in shared memory.

Changes to struct str_hash_table:
- Added a gen_lock_set_t (locks) to hold the lock set
- Added an int (locks_no) to hold the number of locks
- Added an indicator (memory) to know if the table was allocated in private memory (p) or shared memory(s)

Changes to struct str_hash_head:
- Added an int(lock_idx) that hold the index of the lock when the head is locked.

Changes to struct str_hash_entry:
- Added a pointer to the head the entry belongs to. This allows to unlock the head when needed.

Added functions to lock and unlock the head when needed.

Modified functions str_hash_add and str_hash_get to lock the head when needed

Added function str_hash_destroy to remove the hash table and all its contents from memory.

### DIFF
--- a/src/core/str_hash.h
+++ b/src/core/str_hash.h
@@ -20,15 +20,19 @@
 #include "str.h"
 #include "hashes.h"
 #include "mem/mem.h"
+#include "mem/shm_mem.h"
+#include "locking.h"
 #include "clist.h"
 #include <string.h>
 
-
+#define MAX_STR_LOCKS  2048
+#define MIN_STR_LOCKS  2
 /* generic, simple str keyed hash */
 
 struct str_hash_entry{
 	struct str_hash_entry* next;
 	struct str_hash_entry* prev;
+	struct str_hash_head* head;
 	str key;
 	unsigned int flags;
 	union{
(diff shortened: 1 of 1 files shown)


henningw:
Thank you for the contribution. Just one question so far - what was your motivation in adding this functionality, where do you need this?

Elhodred:
hi @henningw, I'm developing a module that requires to have a hash in memory to keep some user information it retrieves from another system via diameter.
The use case is similar to tsilo or domain modules, the information needs to be available for all processes and accesible through a key.
However, if you look at those modules (and I think there were others that I can't remember now) they all implement their own structures to handle the hash table, the collision slots, the entries, the linked lists, the accessor functions,... when it's not needed as kamailio already has a library that contains all of that. The only missing part is that until now it only worked in private memory

miconda:
I am not sure yet about how to deal with this one.

I agree that there is space to reuse some code for hash tables in shared memory, but I feel that we should have dedicated code for it, not to mix it with the one in pkg memory. It will avoid all these if checks to lock/unlock, as we use hash tables to speed up search, it feels we need to keep it optimised.

Moreover, iirc and hope not to confuse with other api, some of the macros are even now used for hash tables in memory, just that the lock/unlock of the slot is done in the implementation code, not in the core code.

Elhodred:
I agree that another header file can be created, either leveraging the existing one for pkg or completely independent.
I will catch up with you on IRC tomorrow or next week to see how to move forward.

miconda:
Closing this one -- as discussed during IRC devel meeting, a separate implementation should be done.

PR #1537, opened 2018-05-21
tls: Fixed build tls module on Alpine dist

sergey-safarov:
`OPENSSL_cleanup` is not exist on Alpine dist ([`tls_init.c:864`](https://github.com/kamailio/kamailio/blob/master/src/modules/tls/tls_init.c#L864))

Later wanted cherry-pick to 5.1 branch


henningw:
Thank you!

Issue #1168, opened 2017-06-26
ims_isc module does not deliver to AS after asynchronous SAR to HSS (distinct PSI)


christoph-v:
### Description
This is the same test case as I used for bug Report #1160
 
I used kamailio as an scscf (based on the configuration files from kamailio/examples/scscf)
I did some changes according to my needs
I did some Basic tests - the results were satisfying --> OK
I configured a distinct PSI on the HSS (sip:04321234501@vict-ims.net) --> OK
I registered some user from a Zoiper Client --> OK
I told zoiper to call sip:04321234501@vict-ims.net --> I received 555 AS Contacting Failed - iFC terminated dialog

### Troubleshooting
For some reason, the ims_isc module cannot deliver the Initial INVITE to the AS.
First I thought, there was a bug in TM module, however there was no bug in TM module (please find the Explanation in issue #1160)

The ims_isc module issues a CRITICAL log:

File 2017_06_20_distinct_psi_03_scscf_kamailio.log:
 Line 642 - 660:
 Jun 20 16:14:12 vict-ims /usr/local/sbin/kamailio[108125]: INFO: ims_registrar_scscf [cxdx_avp.c:138]: cxdx_get_avp(): cxdx_get_experimental_result_code: Failed finding avp
 Jun 20 16:14:12 vict-ims /usr/local/sbin/kamailio[108125]: INFO: ims_registrar_scscf [cxdx_sar.c:84]: create_return_code(): created AVP successfully : [saa_return_code] - [1]
 Jun 20 16:14:12 vict-ims /usr/local/sbin/kamailio[108125]: ERROR: *** cfgtrace:failure_route=[UNREG_SAR_REPLY] c=[/usr/local/etc/kamailio-scscf/kamailio-scscf-routes.cfg] l=127 a=26 n=xlog
 Jun 20 16:14:12 vict-ims /usr/local/sbin/kamailio[108125]: ERROR: *** cfgtrace:failure_route=[UNREG_SAR_REPLY] c=[/usr/local/etc/kamailio-scscf/kamailio-scscf-routes.cfg] l=143 a=21 n=switch
 Jun 20 16:14:12 vict-ims /usr/local/sbin/kamailio[108125]: ERROR: *** cfgtrace:failure_route=[UNREG_SAR_REPLY] c=[/usr/local/etc/kamailio-scscf/kamailio-scscf-routes.cfg] l=130 a=26 n=xlog
[...]

PR #1538, opened 2018-05-22
Bugfix issue 1168 - Unregistered Terminating Services (PSI) Configured at HSS

christoph-v:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally against Version 5.1.0
- [x] Related to issue #1168 (replace XXXX with an open issue number)

#### Description

The ims_isc module cannot cope with the fact that asynchronous Response from HSS (Diameter SAA message) is received as "FAILURE_ROUTE". ims_isc module erroneously assumes, it is an answer from the AS. Change: added a "firstflag", which makes a difference between first call (SAA from HSS) and subsequent call (FAILURE ROUTE from AS).

### DIFF
--- a/src/modules/ims_isc/ims_isc_mod.c
+++ b/src/modules/ims_isc/ims_isc_mod.c
@@ -224,6 +224,8 @@ int isc_match_filter(struct sip_msg *msg, char *str1, udomain_t* d) {
     //sometimes s is populated by an ims_getter method cscf_get_terminating_user that alloc memory that must be free-ed at the end
     int free_s = 0;
 
+    //the callback from the Cx interface in case of unreg terminating initial message is a FAILURE_ROUTE. Hence we need an addl. flag
+    int firstflag = 0;
 
     int ret = ISC_RETURN_FALSE;
     isc_mark new_mark, old_mark;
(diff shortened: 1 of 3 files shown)


christoph-v:
Should I also send this request as one commit, as it was requested for PR 1539

henningw:
As described in the pull template: " Each component has a single commit (if not, squash them into one commit)" - so yes this is generally good.
It makes review easier, we also don't want to include your local development "history" in our master branch.

christoph-v:
Thank you for the Explanation. Now it should be one single commit in the PR, @henningw

christoph-v:
Hi @uts09 ,
I think there are several misunderstandings.
1) scscf.xml is read by the cdp module, not by the ims_isc module
1a) The SAR message to the HSS is not triggered by the ims_isc module, but by the ims_registrar_scscf module, afaik
2) The HSS does not process SIP requests, but it processes DIAMETER
3) Are you sure, you have a Scenario, where you send an Initial INVITE with a PSI in the Request Line?
4) Have you configured the PSI at the HSS?

miconda:
Merging this one, thanks!

christoph-v:
I see P-Served-User: <sip:alice@net1.test>;**sescase=term;regstate=reg** --> this is not unregistered terminating case and has nothing to do with this pull request. If you need further support, please write to the mailing list :-)

However: in the logs I see that the ims_isc module has found a MATCH -> sip:hss.net1.test:5065 () handling 0.

In my humble opinion this means the FQDN hss.net1.test has been sent by the HSS in the IMS Subscription Info of Alice.

We need a tshark trace from the registration of Alice, when the HSS sends the SAA message (this should contain the value sip:hss.net1.test:5065 as an address of an application server). Then we can be sure there is a misconfiguration at the HSS. Now this is only an assumption.

christoph-v:
Thank you @miconda for accepting this change. This bug has hindered us to use kamailio correctly for a long time :-)

PR #1560, opened 2018-06-11
src/Makefile: fix source code revision output

mslehto:
I noticed that ```kamailio -I``` and ```kamailio -V``` do not give git version hash anymore.

This commit fixes at least kamailio -I and -V output, I think.

I don't know / remember if further adjustment is needed for other REPO_ constants.

# Before

```
#define REPO_VER ""
#define REPO_HASH "unknown"
#define REPO_STATE ""
```
```
$ src/kamailio -I|grep Source
Source code revision ID: unknown
```

# After
```
#define REPO_VER "5c48b2"
#define REPO_HASH "5c48b2"
#define REPO_STATE ""
```
```
$ src/kamailio -I|grep Source
Source code revision ID: 5c48b2
```

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -348,7 +348,7 @@ modules-cfg modules-list modules-lst:
 	rm -f modules.lst
 	$(MAKE) modules.lst
 
-ifneq ($(wildcard .git),)
+ifneq ($(wildcard ../.git),)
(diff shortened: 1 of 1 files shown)


henningw:
Thank you!

PR #1566, opened 2018-06-15
[dialog] stats incremented twice

jchavanton:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description
As you can see looking at the modifications stats counters where increment twice.

Hi @oej, If you find a few minutes  you could review this MR. 
Not sure if you remember, this was introduced in a bug fix (see below)
I kept some of your modifications, they seem more strait forward anyway.

```
commit b0cd09d2b451005a736396e6f38eac139ec31301
Author: Olle E. Johansson <oej@edvina.net>
Date:   Wed May 11 21:59:44 2016 +0200

    dialog Make sure statistics are updated when initializing from database
    
    Issue #424
```

### DIFF
--- a/src/modules/dialog/dialog.c
+++ b/src/modules/dialog/dialog.c
@@ -130,8 +130,6 @@ int dlg_h_id_step = 1;
 
 /* statistic variables */
 int dlg_enable_stats = 1;
-int active_dlgs_cnt = 0;
-int early_dlgs_cnt = 0;
 int detect_spirals = 1;
 int dlg_send_bye = 0;
 int dlg_timeout_noreset = 0;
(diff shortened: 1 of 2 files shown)


jchavanton:
update_stat is optimized `inline static void update_stat(stat_var* v, int n)` so the extra code to call `if_update_stat` only once was indeed not needed

henningw:
Thank you for the patch. It seems that there was a change in the functionality regarding the stats counter implementation.
I don't understand your comment regarding the update_stats method though. In your patch the if_update_stat macro is called, which then calls update_stat.?

jchavanton:
Hi Henning, in the end we where calling  `if_update_stat` twice.
This code is only called when retrieving dialog from the DB.

Yes my comment was about removing the optimized version and keeping the simpler one, `if_update_stat` is is a macro calling several inline functions.
One way of calling it was to using a counter when in order to call `if_update_stat` only once.
Then as second call to  `if_update_stat` was added in `b0cd09` probably because the optimization with the counter was obfuscating things slightly.

I think the optimization are not needed the cost per dialog seems very low.
But we could still select to keep the optimized version if we want, knowing that since 2016 both version where used anyway

henningw:
Thank you

PR #1585, opened 2018-07-09
tls: much faster startup using shared SSL_CTX

jchavanton:
#### Pre-Submission Checklist

#### Type Of Change
- [x] Optimization

#### Checklist:

- [x] Tested changes locally

#### Description
Kamailio
TLS mod_init is creating one SSL_CTX, per process, some of the fonctions are taking between 1-3 seconds to execute, this is slowing down the startup sequence greatly.
```
SSL_CTX_load_verify_locations
SSL_CTX_set_client_CA_list  // list sent to the client
SSL_load_client_CA_file
SSL_CTX_get_client_CA_list
```

In fact it is safe to share the SSL_CTX since it is only used to store settings that will be used to internalize new structure, see the documentation reference :
```
    tls: faster startup using shared SSL_CTX
    
    https://www.openssl.org/docs/man1.1.1/man7/ssl.html
    SSL_CTX (SSL Context)
    This is the global context structure which is created by a server or
    client once per program life-time and which holds mainly default values
    for the SSL structures which are later created for the connections.
```

I load tested this with 1000 TLS connections.
We could push the refactoring further, this simple modification as a huge impact since the functions are now called only once per SSL / SNI, not per process ...

### DIFF
--- a/src/modules/tls/tls_domain.c
+++ b/src/modules/tls/tls_domain.c
@@ -561,18 +561,23 @@ static int load_ca_list(tls_domain_t* d)
 		return -1;
 	procs_no=get_max_procs();
 	for(i = 0; i < procs_no; i++) {
-		if (SSL_CTX_load_verify_locations(d->ctx[i], d->ca_file.s, 0) != 1) {
-			ERR("%s: Unable to load CA list '%s'\n", tls_domain_str(d),
-					d->ca_file.s);
-			TLS_ERR("load_ca_list:");
-			return -1;
-		}
-		SSL_CTX_set_client_CA_list(d->ctx[i],
-				SSL_load_client_CA_file(d->ca_file.s));
-		if (SSL_CTX_get_client_CA_list(d->ctx[i]) == 0) {
-			ERR("%s: Error while setting client CA list\n", tls_domain_str(d));
-			TLS_ERR("load_ca_list:");
-			return -1;
+		if (i > 0) {
+			LM_DBG("[SSL_CTX_set_cert_store][%s]\n", tls_domain_str(d));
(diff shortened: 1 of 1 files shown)


miconda:
Is this impacting in any way the reload commands that can be done at runtime?

I didn't have the time to look properly at libssl api on this topic, if you want a merge on fast lane, I am open to do it if this is made a config option via modparam for tls module, with the default value that keeps the behaviour like it is now.

jchavanton:
Q: Is this impacting in any way the reload commands that can be done at runtime? 
A: not tested, based on the assumption that SSL_CTX is only a storage for "default values", this should not be a problem, I will test it in case the documentations and other sources of information may be wrong (or study the openssl source code further) to be absolutely certain.

In the mean time, as you proposed we could make it optional, in the end this is becoming a problem when you have multiple "config" and many processes because this delay is multiplied. (# process * # config)
So this optimisation, may not be needed by everyone.
Main value is to have faster kamailio restart.

jchavanton:
After SSL_CTX is not only containing default settings but hash of sessions, this seems to be thread safe using a locks.

```
SSL_CTX_add_session
    CRYPTO_THREAD_write_lock(ctx->lock);
    ...
    CRYPTO_THREAD_unlock(ctx->lock);
```

OpenSSL can be safely used in multi-threaded applications provided that support for the underlying OS threading API is built-in. Currently, OpenSSL supports the pthread and Windows APIs. OpenSSL can also be built without any multi-threading support, for example on platforms that don't provide any threading support or that provide a threading API that is not yet supported by OpenSSL.

Not sure if there could be a pitfall for forking process, better keep the default value as it is.

It seems another option, could be to copy only the X509_store

miconda:
I understand that the conclusion at this point is not to merge it (at least, yet).

jchavanton:
Thanks for the review, I will revisit and re open later

PR #1596, opened 2018-07-17
topos: adding param contact_hostname

jchavanton:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

When using TLS  with TOPOS the remote UAS could open a new TLS socket to the contact header.
The contact header must contain a domain name with a signed certitificate.

Adding the parameter to TOPOS module to control the host part of the contact headers
```
modparam("topos", "contact_hostname", "proxy.domain.com")
```

Default behavior is left unchanged

I will update documentation it his modification is confirmed  to be valuable.
Else I would be curious to know how this should be handled differently.

### DIFF
--- a/src/modules/topos/doc/topos_admin.xml
+++ b/src/modules/topos/doc/topos_admin.xml
@@ -278,6 +278,31 @@ modparam("topos", "event_mode", 2)
 </programlisting>
 		</example>
 	</section>
+	<section id="topos.p.contact_host">
+		<title><varname>contact_host</varname> (str)</title>
+		<para>
+			You may need to control the host part of the Contact header added
+			by topos.
+
+			For example when using TLS with TOPOS the remote UAS must be able to open
+			a new TLS socket to the contact header.
+			In this case, the contact header must contain a domain name with a trusted CA
+			signed certitificate.
+		</para>
+		<para>
+		<emphasis>
+			Default value is taken from the Record-Route URI.
+		</emphasis>
+		</para>
+		<example>
+		<title>Set <varname>contact_host</varname> parameter</title>
+		<programlisting format="linespecific">
+...
+modparam("topos", "contact_host", "proxy.domain.com")
+...
+</programlisting>
+		</example>
+	</section>
 	</section>
 	<section>
 	<title>Event Routes</title>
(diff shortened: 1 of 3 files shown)


miconda:
I am fine to add this parameter, but the patch lacks checking if the size is exceeding the available space in the buffer -- it checks only sv.len, which is the uri from where to extract ip, port and transport, not it should also check for param len. If this and the docs are added, then can be merged.

There is an alternative to use record_route_preset(), because the contact address is built from the Record-Route URI added by kamailio.

But again, I am fine to add a parameter if it is found to be more convenient and simpler for config.

jchavanton:
Hi, reading the code I found it was taken from the RR header, I missed `record_route_preset()`

This param was not absolutely necessary but I guess it is making things more clear for someone wanting to use TOPOS.
And maybe simpler as well.

I have tested the code "back ported" only for now, but I believe this should be enough to make sure it is working fine.

Thanks for the review

miconda:
Just to clarify: do you want that the port and transport from Record-Route to be kept in the Contact when using the parameter with the domain, right? That practically results only in doing an A/AAA DNS query on the domain, if you want SRV or NAPTR to be done, then this way is not working, but I see also the benefits of preserving the port and transport from R-R.

jchavanton:
I think it is better to use the contact header param only taking optional port and transport from it.
And not use the Record route information at all this way we support NAPTR/SRV and things are less confusing for the user.

Thanks for the pointing this out, it was not my intent to use the port + transport from the RR.

I will : 
- update the PR and the doc too explicitly state that the host part can also contain params.
- parse the module contact param in mod_init and fail if is it invalid.
- make sure we are not using puri.port and puri.transport from the RR
- do a few tests

jchavanton:
I just made more tests trying to control the transport from the param.
I then realized that support for NAPTR in the Contact is not making sense because we should use the transport used by the received request.
Remote UACs should be doing SRV lookups even if the transport is specified, unless you know otherwise.  
I have reached the conclusion that the best option is  to use the port and transport from RR.
At this point I do not see any limitations with this way of doing things.

jchavanton:
If we conclude that the value of this new param is too low, we can drop this MR and add more documentation on how Topos is using RR and how you end-up controlling the contact header instead.

PR #1605, opened 2018-07-27
ims_ipsec_pcscf: sec-agree implementation for IMS

tdimitrov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

This is an implementation of sec-agree used in IMS with IPSec. It's not a complete sec-agree implementation, only the flows used by IMS. The code is usable, but needs a few improvements, which I plan to push in the near future. **My work is based on the implementation in OpenIMSCore.**

For IPSec implementation the XFRM framework from the Linux kernel is used. Security association (SA) and Policies creation/removal is performed via netlink messages. For this reason the module depends on libmnl (a minimalistic netlink library).

As XFRM is Linux specific, the code is not portable and can't be used on operating system different from Linux. The code will not compile on *BSDs too. However all platform specific code resides in ipsec.c so support for other OSes/IPSec implementations can be added relatively easy.

The README file, which is commited is generated from docs dir with xsltproc.

Issues I still work on:
- Kamailio must be run as root in order to be able to send netlink messages and create XFRM SAs and Policies.
- SAs and Policies are not deleted on Kamailio startup and shutdown.
- According to the current contact implementation in the PCSCF modules (adn the 3GPP specs) the IPSec tunnel should be created on two steps. Initial parameters should be saved in security_tmp and on confirmation - in security. At the moment everything remains in security.

As this is my first more serious contribution to the project, all kinds of feedback is highly appreciated :)

### DIFF
--- a/src/modules/ims_ipsec_pcscf/Makefile
+++ b/src/modules/ims_ipsec_pcscf/Makefile
@@ -1,0 +1,20 @@
+#
+# registrar module makefile
+#
+#
+# WARNING: do not run this directly, it should be run by the master Makefile
+
+include ../../Makefile.defs
+auto_gen=
+NAME=ims_ipsec_pcscf.so
+
+ifneq ($(OS),darwin)
+	LIBS += -lpthread
+	LIBS += -lmnl
+endif
+
+DEFS+=-DOPENSER_MOD_INTERFACE
+
+SERLIBPATH=../../lib
+SER_LIBS+=$(SERLIBPATH)/ims/kamailio_ims
+include ../../Makefile.modules
--- a/src/modules/ims_ipsec_pcscf/cmd.c
+++ b/src/modules/ims_ipsec_pcscf/cmd.c
@@ -1,0 +1,628 @@
+/*
+ * Copyright (C) 2012 Smile Communications, jason.penton@smilecoms.com
+ * Copyright (C) 2012 Smile Communications, richard.good@smilecoms.com
+ *
+ * The initial version of this code was written by Dragos Vingarzan
+ * (dragos(dot)vingarzan(at)fokus(dot)fraunhofer(dot)de and the
+ * Fruanhofer Institute. It was and still is maintained in a separate
+ * branch of the original SER. We are therefore migrating it to
+ * Kamailio/SR and look forward to maintaining it from here on out.
+ * 2011/2012 Smile Communications, Pty. Ltd.
+ * ported/maintained/improved by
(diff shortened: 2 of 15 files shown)


henningw:
Hello,
thank you for your contribution! I did a quick look to the code and noticed a few simple areas that should refined before going looking deeper in the functionality.

* Please remove the README file from the commit. This is auto-generated and should not be commited to git
* You mentioned that your work is based on OpenIMScore. Do you actually use code from this project in your implementation? Then you need to add proper copyright notices that refer to this code.
* With regards to copyright - you need to also choose a copyright to your code. Most modules choose "GPLv2 and later", but BSD would be also ok if you prefer this. Have a look to other modules for an  example copyright header.

Best regards,
Henning

tdimitrov:
Hello Henning,

Thanks for your comments. I've removed the README, however, the copyright/licensing part is a bit complicated for me.

I have copied code from ims_registrar_pcscf for things like module structure and finding the contact in memory, etc.
I've used the implementation in OpenIMS mainly as a reference - what functions they expose to the routing logic and how they use internal Kamailio structures to redirect SIP messages.
The IPSec handling itself is written from scratch because they used shell scripts, instead of netlink sockets.

For there reasons I copied the Copyright notice from ims_registrar_pcscf (which is actually the same as in OpenIMSCore) and put it in the sources, based on other's work. Everything else is in GPLv2.

Do you feel this is correct or it's better to put OpenIMSCore's copyright everywhere? 

In general, all I want is to contribute the code back to the project and not to abuse the licensing. I don't care if my name will stay anywhere or not.

Best regards,
Tsvetomir

miconda:
@tdimitrov - going to merge it and then add you to the dev team so you can maintain the new module.

@henningw - for the future, when a new module is added, the README must be committed, either by the developer in the initial patch or by someone else after the merge. I will do it after the merge. For safety, the automatic script is not adding new files, only updates exiting ones -- maybe it can be enhanced, but initially I thought that running `git add` can create troubles if some files are generated or relocated by mistake.

henningw:
Tsvetomir - thank you for the clarification about the licensing. If the code in question is GPLv2, then all is fine if you keep it GPLv2 also in your license for new files. Regarding the copyright notices you choose the right approach, add the notice from the "original" file to your file if you add some parts from it.

Thank you Daniel for the clarification - I did not know this about the README for new modules.

tdimitrov:
@miconda @henningw Thank you!

PR #1609, opened 2018-08-01
presence: limit the number of subscriptions handled in timer_dbonly

lazedo:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

due to the way update_db_subs_timer_dbonly handles the query to the
database (getting all records), if for some reason a burst of
terminating subscriptions occurs, most likely there will be no package
memory to process all expiring subscriptions.

this commit uses the same pattern as other routines in presence by using
db_fetch_query with fetch_rows parameter module.

because we create the subs in the loop and then call handle_expired_subs
to avoid locking issues the subscription should already be deleted from
the database when it returns from handle_expired_subs, there's no reason
to issue the last delete, and that was removed.

### DIFF
--- a/src/modules/presence/subscribe.c
+++ b/src/modules/presence/subscribe.c
@@ -2026,14 +2026,14 @@ void update_db_subs_timer_dbonly(void)
 	db_val_t qvals[1];
 	db_key_t result_cols[18];
 	int pres_uri_col, to_user_col, to_domain_col, from_user_col, from_domain_col,
-		callid_col, totag_col, fromtag_col, event_col, event_id_col,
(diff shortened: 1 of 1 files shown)


lazedo:
@miconda yes it was, the callback of the notify deletes the subscription. it actually is required because in the case where we have `fetch_rows=500` and the entire result set is `5000` rows then we would loose `4500` rows because they would be deleted.

lazedo:
this is causing problems in some `production` servers and i would like to backport to stable branches if that's ok

miconda:
Ok to backport.

PR #1611, opened 2018-08-02
core: update first line parser to recognize new method IDs

charlesrchance:
- bring up to date with new methods added in 4fe95ab


miconda:
Thanks, can be merged.

PR #1621, opened 2018-08-14
permissions: check for db_url being set on reload

armenb:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

### DIFF
--- a/src/modules/permissions/address.c
+++ b/src/modules/permissions/address.c
@@ -255,6 +255,11 @@ dberror:
  */
 int reload_address_table_cmd(void)
 {
+	if(!db_url.s) {
+		LM_ERR("db_url not set\n");
+		return -1;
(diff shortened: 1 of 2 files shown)


armenb:
This fixes a null pointer exception in a config that loads permission.so but does not set its db_url.

To repro, create a config with loadmodule "permissions.so" but NOT modparam("permissions", "db_url", "foo"), and then run "kamctl trusted reload".  Kamailio will crash without this patch.  With this patch, kamctl will return with "Reload failed" and a message will be printed in kamailio logs.

henningw:
Hi Armen,
thank you for the pull request. I applied it to master and also back-port  to 5.0 and 5.1 branches.
Best regards,
Henning

PR #1626, opened 2018-08-30
dmq: add optional parameter to dmq_handle_message()

eschmidbauer:


eschmidbauer:
some context for this PR...
we are testing using the `tsilo` module for mobile push notifications while maintaining multiple registrars with `dmq_usrloc`
if an `INVITE` goes to a registrar "A" and then the UA registers to registrar "B", the dmq message is sent to registrar "A" with UA location info,and the INVITE can be resumed successfully. 

```
    if (is_method("KDMQ")) {
        dmq_handle_message("1");
        if ($tU == "usrloc") {
            $var(a) = "";
            json_get_field("$rb", "aor", "$var(a)");
            $var(a) = $(var(a){s.strip,1});     #strip quotes
            $var(a) = $(var(a){s.striptail,1}); #strip quotes
            $var(p) = "sip:" + $var(a);
            route(PUSHJOIN);
        }
        exit;
   }
```

charlesrchance:
Very cool!

It is definitely useful. However, I have one very small comment.

I'm not sure what value it adds allowing the config to determine the returned value - only that the function returns without exiting. Possibly just semantics right now, but it could potentially restrict future module development - or not, but better to be safe.

Would it still serve its purpose if the parameter was renamed to something like "return" or "continue" and only return 1 if set or 0 if not?

Again, nice use case :)

PR #1636, opened 2018-09-05
dispatcher: add ping_from attribute

giavac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

I was looking for a solution to set a dedicated From URI in OPTIONS
requests to a specific dispatcher target.

ds_ping_from (
https://www.kamailio.org/docs/modules/devel/modules/dispatcher.html#dispatcher.p.ds_ping_from)
applies globally, while I needed the granularity of a single target.

I can just add an attribute to the others, e.g. instead of

```
1 sip:127.0.0.1:7070 <http://127.0.0.1:7070> 0 1
```

something like:

```
1 sip:127.0.0.1:7070 <http://127.0.0.1:7070> 0 1 ping_from=sip.custom.com
```

When the ping_from attribute is not specified, the general ds_ping_from
applies and is used.

This would save me from a solution like intercepting the outgoing OPTIONS
and manipulating the From header.

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -300,6 +300,9 @@ int ds_set_attrs(ds_dest_t *dest, str *vattrs)
 			} else {
 				LM_ERR("rweight %d not in 1-100 range; skipped", tmp_rweight);
 			}
+		} else if(pit->name.len == 9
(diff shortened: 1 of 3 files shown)


miconda:
Thanks!

Issue #1429, opened 2018-02-07
[master] jsonrpcs doesn't work on AARCH64 ?!


BroWater:
### Description

I upgraded from 4.2 to latest commit on master branch.

During migrating, I found mi was replaced by RPC, 
(https://www.kamailio.org/wiki/install/upgrade/4.4.x-to-5.0.0)
and when I tested with command like `sh kamctl ps`,
it showed:

```
# sh kamctl ps
awk: bad regex '^{.+"id"[       ]*:[    ]*[0-9]+[       ]*}$': Invalid preceding regular expression
```
I also tested it in standard configuration on my PC according to the following steps:
```
make cfg
make all
make install
```

and it works

```
# sh kamctl ps
{
  "jsonrpc":  "2.0",
  "result": [
    {
      "IDX":  0,
      "PID":  30916,
      "DSC":  "main process - attendant"
    }, {
      "IDX":  1,
      "PID":  30917,
      "DSC":  "udp receiver child=0 sock=127.0.0.1:5060"
    }, {
      "IDX":  2,
      "PID":  30918,
      "DSC":  "udp receiver child=1 sock=127.0.0.1:5060"
    }, {
      "IDX":  3,
      "PID":  30919,
      "DSC":  "udp receiver child=2 sock=127.0.0.1:5060"
    ....
   
```

My configuration settings and install steps are like followings:

CC=/opt/toolchains/aarch64-thunderx/bin/aarch64-thunderx-linux-gnu-gcc make cfg ARCH=aarch64 LOCALBASE=/mnt/test/kamailio/
make all
make install

Any suggestion??

#### Debugging Data

```
(paste your debugging data here)
```

#### Log Messages

```
(paste your log messages here)
```

#### SIP Traffic

```
(paste your sip traffic here)
```

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
# ./kamailio -v
version: kamailio 5.2.0-dev3 (aarch64/linux)
[...]

PR #1669, opened 2018-10-07
kamctl: make jsonrpc filter portable

micmac1:
The filter has a regex looking for a literal '{' in the beginning of a
line. Some awk implementations interpret this as a meta character, so
the regex is deemed broken. Example with busybox awk (POSIX):

`root@hank2:~# kamctl ps`
`awk: bad regex '^{.+"id"[ 	]*:[ 	]*[0-9]+[ 	]*}$': Invalid contents of {}`
`root@hank2:~#`

To fix this enclose the character in square brackets. This always
matches for a literal '{' and is portable.

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #1429 

#### Description

Hi all,

I was trying to run kamctl on OpenWrt and the regex failed. I found a (closed) issue and the remedy was to use gawk. But it's actually easy to fix in the regex and if I can use busybox's awk I don't need to install gawk, so it saves some space. OpenWrt is usually run on embedded so space is often an issue :)

Thanks for looking at this!

Kind regards,
Seb

### DIFF
--- a/utils/kamctl/kamctl.base
+++ b/utils/kamctl/kamctl.base
@@ -715,7 +715,7 @@ filter_json()
 	$AWK 'function ltrim(s) { sub(/^[ \t\r\n]+/, "", s); return s }
 		BEGIN { line=0; IGNORECASE=1; }
 		{ line++; }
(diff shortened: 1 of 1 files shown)


miconda:
Thanks!

Issue #1673, opened 2018-10-10
acc: generating duplicates on missed / failed calls


jchavanton:
### Description
I found confusion about using the integer value of the flags when checking them.
I think this as been creating little confusing problems for a long time (not 100% sure about the exact impact).
I believe users where finding work around solutions.

acc_run_engines (integrer value of the flag)
```
  if((type==1) && (msg->flags&(e->missed_flag))) {
```
and is_eng_acc_on (integer value of the flag)
```
 if(msg->flags & e->missed_flag) {
```
and (integer values if the flag)
```
flags_to_reset |= log_missed_flag
```

However when we reset it we use the (bit position value)
```
reset_acc_flag( req, flags_to_reset ); 
#define reset_acc_flag(_rq,_flag)   (resetflag((_rq), (_flag))) 
```

Everywhere else in the application (I have seen)
We use the helper fuctions like `isflagset`, `resetflag`, etc. and use integer representing the bit position
This is why I am proposing a patch to reuse this pattern.

jchavanton:
just a little update, I this should be handled properly since the  flags should be kept even if modified by a transaction callback, good now I need to find out why this is not taking place.
```
/* if failure handler changed flag, update transaction context */
shmem_msg->flags = faked_req->flags;
```

jchavanton:
I found consufion about using the integer value of the flag when checking it.

acc_run_engines
```
  if((type==1) && (msg->flags&(e->missed_flag))) {
```
and 
is_eng_acc_on 
```
 if(msg->flags & e->missed_flag) {
```

However when we reset it we use the bit position value
```
reset_acc_flag( req, flags_to_reset ); 
#define reset_acc_flag(_rq,_flag)   (resetflag((_rq), (_flag))) 
```

Looking at this further

henningw:
Commenting about the third topic directly in this issue. I don't understand this remark:
"However when we reset it we use the (bit position value)"
```
reset_acc_flag( req, flags_to_reset ); 
#define reset_acc_flag(_rq,_flag)   (resetflag((_rq), (_flag))) 
```
The function resetflag is defined in a similar logic as setflag: struct sip_msg* msg, flag_t flag.
And flag_t is defined (as you correctly identified) as unsigned int.

henningw:
Should be fixed by pull request #1674

PR #1674, opened 2018-10-11
acc: generating duplicates

jchavanton:
on missed / failed calls issue #1673

#### Pre-Submission Checklist

#### Type Of Change
- [x] Breaking change (not tested with all scenarios, review could confirm that this is not introducing regressions)

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #1673 

#### Description
problem with confusion when using  flags (integer value vs bit position value)
proposed solution is using the flags helpers functions `isflagset`, etc. where possible.

### DIFF
--- a/src/modules/acc/acc_logic.c
+++ b/src/modules/acc/acc_logic.c
@@ -463,15 +463,6 @@ static inline void acc_onreply_in(struct cell *t, struct sip_msg *req,
 	}
 }
 
-static void reset_acc_flags(struct sip_msg *req, int flags_to_reset) {
-	int x;
-	for (x=0;x<32;x++) {
-		if (flags_to_reset & 1<<x) {
-			resetflag(req, x);
-			LM_DBG("reset[%d]\n", x);
-		}
-	}
-}
 
 /* initiate a report if we previously enabled MC accounting for this t */
 static inline void on_missed(struct cell *t, struct sip_msg *req,
@@ -529,7 +520,7 @@ static inline void on_missed(struct cell *t, struct sip_msg *req,
 	 * These can't be reset in the blocks above, because
 	 * it would skip accounting if the flags are identical
 	 */
-	reset_acc_flags(req, flags_to_reset);
+	resetflags(req, flags_to_reset);
 
 	if (new_uri_bk.len>=0) {
 		req->new_uri = new_uri_bk;
--- a/src/modules/acc/acc_logic.c
+++ b/src/modules/acc/acc_logic.c
@@ -55,7 +55,6 @@ struct acc_enviroment acc_env;
 
 
 #define is_acc_flag_set(_rq,_flag)  (((_flag) != -1) && (isflagset((_rq), (_flag)) == 1))
-#define reset_acc_flag(_rq,_flag)   (resetflag((_rq), (_flag)))
 
 #define is_failed_acc_on(_rq)  is_acc_flag_set(_rq,failed_transaction_flag)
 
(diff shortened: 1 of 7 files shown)


src/modules/acc/acc_logic.c
```diff
@@ -501,19 +501,18 @@ static inline void on_missed(struct cell *t, struct sip_msg *req,
 	 * forwarding attempt fails; we do not wish to
 	 * report on every attempt; so we clear the flags;
 	 */
-
 	if (is_log_mc_on(req)) {
 		env_set_text( ACC_MISSED, ACC_MISSED_LEN);
 		acc_log_request( req );
-		flags_to_reset |= log_missed_flag;
+		flags_to_reset |= 1 << log_missed_flag;
```
jchavanton: I wonder why this was not creating more problem in the past the flags variable where containing "bit position", this is one point that the reviewer can help to confirm. (I may triple check that this is always true)

henningw:
Hello Julien, thank you for this request.

So did I understood it correctly, you want to unify the flag manipulation and also fix some bugs this way? It is probably better to not combine this two changes. If you change different modules, these changes should be also in individual commits, if they are not interdependent on each other.

I had a quick look to the patch.

-			if((type==1) && (msg->flags&(e->missed_flag))) {
+			if((type==1) && isflagset(msg, e->missed_flag) == 1) {

This should be identically, according to the definition of isflagset. I don't see any big advantage of one format to the other, or I am wrong?

One question about this change (it is represented several times):

-		_acc_diameter_engine.missed_flag = 1<<diameter_missed_flag;
+		_acc_diameter_engine.missed_flag = diameter_missed_flag;

This will not produce identical results, I think. The first one is a left shift of diameter_missed_flag length, and the second one an assignment. Example:

diameter_missed_flag                      = 00000000000000000000000000001000 (8)
_acc_diameter_engine.missed_flag = 00000000000000000000000010000000 (256)

What is your reasoning behind this change?

jchavanton:
Hi Hennig, thanks for the review so far.

Q: "you want to unify the flag manipulation and also fix some bugs this way?"
A: Correct, to fix the bug we need to change inconsitant flag manipulation creating the bug 

Q: This should be identically, according to the definition of isflagset. I don't see any big advantage of one format to the other, or I am wrong?
A: They are not identical :

The first one is expecting an input we the integer contain the bit mask
`if((type==1) && (msg->flags&(e->missed_flag))) {` 
will with with the input of : 1,2,4,8,...,4294967296

The second one is expecting an input where integer contain the bit position
`if((type==1) && isflagset(msg, e->missed_flag) == 1) {`
will do the bit shifting and will work with the input : 1,2,3,..,32

```
int isflagset( struct sip_msg* msg, flag_t flag ) {
        return (msg->flags & (1<<flag)) ? 1 : -1; // << notice the bit shifting
}
```

`resetflag` and `setflag` are also doing the bit shifthing, seems like at one point someone got confused and introduce a mix which endedup doing resetflag on an alredy bit shifted input.
I guess this is also why acc_mod is not doing bit shifting on the module param while acc_diameter, acc_raduis and acc_json are (I am changing this in the patch)

henningw:
Hello Julien, you are right - this is indeed confusing. So you are also saying (regarding my second remark) that the other acc modules are doing it wrong?

jchavanton:
Just pushed a complete working solution, 

tested with `acc` 
```
modparam("acc", "log_missed_flag", 9)
modparam("acc", "db_missed_flag", 10)
```
and with `acc_json`
```
modparam("acc_json", "acc_flag", 2)
modparam("acc_json", "acc_missed_flag", 7)
```
`acc_radius` and `acc_diameter` using the exact same integration

jchavanton:
I believe this fix is ready to be merged

henningw:
Thank you Julien, I will merge this later today. I will need to do it manually, in order to split the commits in modules and core.

jchavanton:
I see, should have known ... thanks

henningw:
Merged manually into git master

PR #1683, opened 2018-10-22
usrloc: call contact expired call back for a domain with db_mode: DB_ONLY

adil-mafzool:
usrloc: call contact expired call back for a domain with db_mode: DB_ONLY; database rows are removed by the timer function: db_timer_udomain

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ X] PR should be backported to stable branches
- [X ] Tested changes locally

#### Description

Currently call contact expired call back for a domain with db_mode: DB_ONLY doesn't callback function for "usrloc:contact-expired" to carry out call back application actions since database rows are removed by the timer function: db_timer_udomain. There is addition of new function  udomain_contact_expired_cb which will only call the call back function when contact are being removed from the database.

### DIFF
--- a/src/modules/usrloc/udomain.c
+++ b/src/modules/usrloc/udomain.c
@@ -911,6 +911,10 @@ int udomain_contact_expired_cb(db1_con_t* _c, udomain_t* _d)
 	urecord_t* r;
 	ucontact_t* c;
 
+	if (db_mode!=DB_ONLY) {
+		return 0;
+	}
+
 	columns[0] = &user_col;
 	columns[1] = &contact_col;
 	columns[2] = &expires_col;
--- a/src/modules/usrloc/udomain.c
+++ b/src/modules/usrloc/udomain.c
@@ -40,6 +40,7 @@
 #include "usrloc.h"
 #include "utime.h"
 #include "usrloc.h"
+#include "ul_callback.h"
 #include "urecord.h"
 
 extern int ul_rm_expired_delay;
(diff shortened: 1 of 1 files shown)


miconda:
Thanks for the contribution!

This will bring coherence with the other values for db_mode, so if the other devels have nothing against, it can be merged, even if we are in the freezing phase for v5.2.0.

The only adjustment I would make is to add a mod param to control the execution of the new function, because it can be a lengthly execution when there are many records in the database and this functionality is not needed always.

henningw:
@miconda: It is not strictly a bugfix - but as you said it brings coherence to the module with the other values for db_mode. So I have nothing against the exception to merge it now in the current phase of development. To addea module parameter is a good idea as well.

@adil-mafzool: I have one question about your patch: you create both a new urecord and a ucontact in the timer callback function, and I think both are allocated in new memory. But you only cleanup the ucontact, not the urecord after the callback processing. Is this not necessary or it is cleanup at another place?

adil-mafzool:
@miconda: Adding db_mode would be required since db_timer_udomain is now being called for different db_modes.
@henningw: Had seen that but since the concept was duplicated from ``preload_udomain``` where ucontact was being freed where as urecord was not. It looks like urecord is using static memory function get_static_urecord and doesn't use shm_malloc.

miconda:
This is not a pure fix to a bug -- I said I am fine to merge it in this phase of testing the upcoming 5.2.0 release because it brings coherence with the other db modes. With git is easy to clone the repo and maintain back ports as you need.

adil-mafzool:
@miconda thats ok!!

henningw:
Thank you, merged into master branch.

PR #1726, opened 2018-11-16
tmx: add t_drop

lazedo:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

we have a use case where we need to drop the request within a transaction, and that is not possible because tm module will always send the final reply.

```
route[MYREQ]
{
    ## protect from retrans
    if (!t_newtran()) {
        xlog("L_ERROR", "$ci|log|failed to create transaction\n");
        drop;
    }

    if(some early business use case that requires drop) {
        xlog("L_INFO", "$ci|log|dropping request\n");
        t_drop();
    }

    async_func_that_suspends_and_continues("$anyparam", "TR_OK", "TR_ERROR");

}

failure_route[TR_ERROR]
{
    xlog("L_INFO", "$ci|log|failed $T_reply_code $T_reply_reason\n");
    t_drop();
}

onreply_route[TR_OK]
{
    xlog("L_INFO", "$ci|log|checking transaction result\n");
   if(some business use case that requires dropping the request) {
      t_drop();
   }
...
}
```
if we use `drop` instead of `t_drop` , we always get a final reply sent to the requestor

not sure if this is the right approach to the problem but at least fixes for us and can start the discussion.

### DIFF
--- a/src/modules/tmx/doc/tmx_admin.xml
+++ b/src/modules/tmx/doc/tmx_admin.xml
@@ -432,6 +432,64 @@ route[MYROUTE] {
 </programlisting>
 		</example>
 	</section>
+    <section id="tmx.f.t_drop">
+        <title>
+            <function moreinfo="none">t_drop([Code])
+            </function>
+        </title>
+        <para>
+        Drops the transaction with error code (500 default).
+        </para>
+        <para>
+        Parameters:.
+        </para>
+        <itemizedlist>
+            <listitem><para>
+                <emphasis>Code</emphasis> - error code to set in uas status.
+            </para></listitem>
+        </itemizedlist>
+        <para>
+        This function can be used in ANY_ROUTE.
+        </para>
+        <example>
+        <title><function>t_drop</function> usage</title>
+        <programlisting format="linespecific">
+...
+route[MYREQ]
+{
+    ## protect from retrans
+    if (!t_newtran()) {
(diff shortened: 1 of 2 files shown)


lazedo:
there seems to be an unrelated? error in the build
` Compiler: clang C, DIST=sid`

linuxmaniac:
yes, that's unrelated. Already pushed https://github.com/kamailio/pkg-kamailio-docker/commit/facc89d0bad0801bdd117061ac5fd85179fc824c to fix that issue. Waiting for the images to be updated https://hub.docker.com/r/kamailio/pkg-kamailio-docker/builds/

miconda:
@lazedo - the PR does not have documentation for the new function.

Also, I wonder if makes sense to have a parameter for the function in order to set the status code of uas, now it is hardcoded to 500. It can be helpful in case of some tm callbacks (e.g., for accounting or dialog) to be able to have a specific code chosen from config.

Then, I am not sure if just setting the uas response code is enough overall, more or less in the context of tm callbacks being executed when transaction is destroyed. Here I need to look a bit more at what is needed from the UAS structure of tm.

lazedo:
@miconda i updated the documentation and added parameter to set uas.status code

miconda:
Pushed it manually with some changes: use directly the core fixup function, adjusted the docs with existing functions in examples.

lazedo:
@miconda thanks

PR #1732, opened 2018-11-20
Module: rtpengine timeout log fixed

surendratiwari3:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
Previously for timeout reply kamailio was not printing the command for which we received the timeout, Also if kamailio not able to read the response then also command was not getting printed. some indention fix is also done.


miconda:
Thanks, I pushed the changes from PR manually, to have a single commit and a better commit message, because it is not a fix, but adding to a log message. Also, the indentation adjustments were done with whitespaces, not using the same style from the existing code.

PR #1742, opened 2018-11-29
rtpengine : crash is fixed and codec flags added to doc

ycaner06:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Kamailio crashes when codec-flags value are not setted with codecs. it is fixed.
codec-flags information added to document.
An example is added to document.

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -218,7 +218,7 @@ static int pv_parse_var(str *inp, pv_elem_t **outp, int *got_any);
 static int mos_label_stats_parse(struct minmax_mos_label_stats *mmls);
 static void parse_call_stats(bencode_item_t *, struct sip_msg *);
 
-static int control_cmd_tos = -1; 
+static int control_cmd_tos = -1;
 static int rtpengine_allow_op = 0;
 static struct rtpp_node **queried_nodes_ptr = NULL;
 static pid_t mypid;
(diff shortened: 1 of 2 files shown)


miconda:
@rfuchs - does the patch to the rtpengine module code look ok for you?

src/modules/rtpengine/rtpengine.c
```diff
@@ -1992,6 +1992,10 @@ static int parse_flags(struct ng_flags_parse *ng_flags, struct sip_msg *msg, enu
 		if (str_key_val_prefix(&key, "transcode", &val, &s)
 				|| str_key_val_prefix(&key, "codec-transcode", &val, &s))
 		{
+			if(s.len<=0 && s.s==NULL){
```
rfuchs: There's actually a much better solution to this, which is to change `str_key_val_prefix()`, which is where the bug actually is.

The function should read:

```
static inline int str_key_val_prefix(const str *p, const char *q, const str *v, str *out) {
	if (str_eq(p, q)) {
		if (!v->s || !v->len)
			return 0;
		*out = *v;
		return 1;
	}
...
`

Then all the explicit null checks here and below can be removed.

ycaner06: I avoid changing function because i afraid that breaking other parts. At monday , i can change it as you wish.

rfuchs:
Looks good to me now. @miconda do you prefer to have these commits squashed or merge as they are now?

miconda:
@rfuchs - up to you, do you it as you prefer -- usually we squash if a commit is changing or is an improvement of the previous commit.

PR #1756, opened 2018-12-05
pkg/kamailio/obs: Fixed warning about not existen kamailio group

sergey-safarov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
Kamailio is installed to systemd as member of "daemon" group.
Fixed not correct group usage in spec file.

### DIFF
--- a/pkg/kamailio/obs/kamailio.service
+++ b/pkg/kamailio/obs/kamailio.service
@@ -6,7 +6,7 @@ After=network-online.target
 [Service]
 Type=simple
 User=kamailio
-Group=daemon
+Group=kamailio
 Environment='CFGFILE=/etc/kamailio/kamailio.cfg'
 Environment='SHM_MEMORY=64'
 Environment='PKG_MEMORY=4'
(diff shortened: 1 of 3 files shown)


miconda:
Isn't the group `kamailio` created with the install scripts from rpm? I think on debian, the group is created. It would be good to have same approach on both types of packaging, makes it easier for docs and being coherent no matter of operating system, as systemd is supposed to be same configs everywhere...

sergey-safarov:
I looked into [kamailio.postinst](https://github.com/kamailio/kamailio/blob/master/pkg/kamailio/deb/debian/kamailio.postinst) and can find add user command but cannot see group addition command.
@linuxmaniac could you point me to group addition on debian packaging script.

sergey-safarov:
Looks as here options `--system --group` is create group `kamailio`

linuxmaniac:
From the manpage of adduser:
>  If adduser is called with the --group option and without the --system option, or addgroup is called respectively, a user group will be added.

sergey-safarov:
PR is updated, but not tested locally.
If now of, then i will test locally and squash commits.

miconda:
The diff of the PR looks ok, so can be squashed and committed from my point of view.

sergey-safarov:
PR is squashed and locally tested.
Now ready to merge.

miconda:
@sergey-safarov - you can merge and backport if needed.

PR #1760, opened 2018-12-07
mqueue: new RPC command mqueue.fetch

jchavanton:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

using this new RPC method we can fetch elements key/value from any queue from an external application. This could be useful in many cases, one use case is with acc_json

```
curl --data-binary '{"jsonrpc":"2.0","id":"1","method":"mqueue.fetch","params": ["MY_QUEUE"]}' -H 'content-type:text/plain;' http://1270.0.1:9999:jsonrpc/ 
{
	"jsonrpc":	"2.0",
	"result":	{
		"key":	"my_key",
		"val":	"{\"time\": 1544217754, \"method\": \"INVITE\","\"}"
	},
	"id":	"1"
}
```

```
kamcmd mqueue.fetch my_queue
error: 404 - Empty queue
```

### DIFF
--- a/src/modules/mqueue/doc/mqueue.xml
+++ b/src/modules/mqueue/doc/mqueue.xml
@@ -35,10 +35,19 @@
 		<email>osas@voipembedded.com</email>
 		<affiliation><orgname>VoIP Embedded, Inc.</orgname></affiliation>
 	    </editor>
+	    <editor>
+		<firstname>Julien</firstname>
+		<surname>Chavanton</surname>
+		<email>jchavanton@gmail.com</email>
+	    </editor>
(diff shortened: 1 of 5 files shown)


miconda:
Is this fetching from the queue -- get the first item and remove it from queue? Looks like only getting the item (key, value).

miconda:
Actually looks ok, with the first look somehow I missed the use of mq_head_fetch() in the rpc command c code ...

jchavanton:
yes, it is equivalent of scripting command `mq_fetch(queue)`

PR #1763, opened 2018-12-11
http_client: http_connect_raw

jchavanton:
similar to http_connect for http/post, however the data parameter is not PV parsed
this way json and any other content can be used without any escaping

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description
I have ran into the issue, where a dollar sign was present in the data  : 

https://lists.kamailio.org/pipermail/sr-dev/2009-March/000898.html
```
Try to replace $ with $$ inside the $var(data), something like:

$var(m) = "$";
$var(r) = "$$";
$var(newdata) = $(var(data){s.replace,$var(m),$var(r)});
```
This fix is working, but I thought others could fell into this pitfall and have builtin way to send raw data over http could be best.

By adding a module command to send `http_connect_raw()`

I wonder if this should be called something else ...

### DIFF
--- a/src/modules/http_client/doc/http_client_admin.xml
+++ b/src/modules/http_client/doc/http_client_admin.xml
@@ -646,7 +646,7 @@ http_follow_redirect = no
 					</para>
 					<para>
 						<emphasis>data</emphasis> - Data or a pseudo variable holding
-						data to be posted.
+						data to be posted. (may contain pseudo variable)
 					</para>
 					<para>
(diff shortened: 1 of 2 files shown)


miconda:
I merged the patch manually, the first line of the commit message was not that explanatory -- it is the message printed by git short log that I use for backports or extracting new features for releases.

Commit: 0d99ddb6856176aa8feacb17254ac0c0da48ad21

Issue #1757, opened 2018-12-06
Proxy Protocol support


teotwaki:
### Description

I would like to implement support for the Proxy Protocol as implemented by HAProxy and AWS ELBs. The main advantage this brings is better connection information, even behind multiple layers of NAT or LBs. This feature was already discussed a couple of years ago, and it appears the project was not against receiving a patch that adds support for this protocol: https://lists.kamailio.org/pipermail/sr-users/2016-June/093497.html

For example, at the moment, my Kamailio deployment is "blind" when it comes to the origin of inbound TCP connections, because it is only able to report the internal IP address of our ELBs, rather than the actual IP address of the client, and IP address of the LB the connection was received on. Dixa is offering to sponsor work (my own hours) to implement this, and we would also be a production user of this feature.

The Proxy Protocol is fairly well documented here: https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

As far as I know, AWS ELBs only implement version 1 of the protocol, but I believe there is value in implementing both (while we're at it), and the second version of the protocol is a lot easier to parse, anyway.

I'm opening this feature request to a/ track progress, b/ discuss implementation details.

henningw:
Your idea sounds interesting, and I can see a benefit as well especially for the mentioned use cases (AWS and haproxy setups). Daniel has probably also some comments on this, but in my opinion your approach sounds reasonable.

   * Add a core variable, e.g. ksr_tcp_accept_proxy
   * add a function to parse the proxy protocol(s), like tcp_read_proxy_protocol(..)
   * probably it makes sense to add a own flag for this as well to check the proper parsing later on (similar to the hep3 execution path)
   * call the tcp_read_proxy_protocol() in tcp_read_req(..)
   * add also a proxy_process_msg(..) function to actually use the data to set the internal connection information(s)
  * then check the flag in receive_tcp_msg(..) and call this function, after your special handling related to the proxy protocol call the normal receive_msg(..) function to process the payload

Just to show you one possible approach (obviously from a high level view).

miconda:
@teotwaki - following the hep3 handling should be a good approach, thanks for considering to contribute and working on this feature!

teotwaki:
@miconda & @henningw: Could I ask for some early input on my current progress?

https://github.com/kamailio/kamailio/compare/master...teotwaki:sl/implement-proxy-protocol

[...]

PR #1765, opened 2018-12-13
core: Add PROXY protocol implementation

teotwaki:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #1757

#### Description

Note: This PR follows the discussion in #1757.

- introduce new global variable `ksr_tcp_accept_proxy`.
- this variable can be modified by using the `tcp_accept_proxy=yes`
  configuration parameter.
- when active, inbound TCP connections are expected to behave according
  to the PROXY protocol[1].
- Both the v1 (human-readable) and v2 (binary) versions of the protocol
  are supported.

[1]: https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

### DIFF
--- a/src/core/tcp_main.c
+++ b/src/core/tcp_main.c
@@ -46,6 +46,7 @@
 #define BSD_COMP  /* needed on older solaris for FIONREAD */
 #endif /* HAVE_FILIO_H / __OS_solaris */
 #include <sys/ioctl.h>  /* ioctl() used on write error */
+#include <arpa/inet.h>  /* for inet_pton() */
 #include <netinet/in.h>
 #include <netinet/in_systm.h>
 #include <netinet/ip.h>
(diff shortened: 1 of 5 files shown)


miconda:
At the first sight, it looks good overall, just few changes that I would suggest:

  * I think it would be better to name the new core parameter `ksr_tcp_accept_haproxy` instead of `ksr_tcp_accept_proxy`. The `proxy` term is also used in SIP a lot (Kamailio being a proxy itself), so I think using `haproxy` is more suggestive for the purpose of the parameter.

  * there are couple of local variables when parsing the PROXY header which are not declared at the beginning of a block (or at the beginning of a function). Latest C compilers do not complain about, but we prefer to stay a bit more strict compliant with older specs in order to have it working on some embedded devices or old distros.

lazedo:
@miconda the `proxy` protocol was first defined by haproxy but there may be other programs that implement the protocol (not `haproxy` exclusive). i suggest `accept_proxy_protocol`

miconda:
@lazedo - even implemented in other applications I still find that `tcp_accept_haproxy` is more suggestive and directs to what it is supposed to do. It is where it started and the specs are still hosted by haproxy.org.

src/core/tcp_main.c
```diff
@@ -985,16 +1196,29 @@ struct tcp_connection* tcpconn_new(int sock, union sockaddr_union* su,
...
+	if (unlikely(ksr_tcp_accept_haproxy && state == S_CONN_ACCEPT)) {
+		ret = tcpconn_read_haproxy(c);
+
+		if (ret == -1) {
+			LM_ERR("invalid PROXY protocol header\n");
+			goto error;
+		} else if (ret == 1) {
+			LM_DBG("PROXY protocol did not override IP addresses\n");
+			goto read_ip_info;
+		}
```
henningw: Small suggestion, add a LM_DBG output here, that the tcpconn_read_haproxy call was successful.

teotwaki:
> related to error cases, what would happen if you e.g. set in the IPv6 case 0x21 the dst_ip->af = AF_INET6, but later one run unto an error in the parsing. Then you would return an error to the callee, and would go to the read_ip_info path. Here the AF_INET6 would be still be set, or I am wrong?

I believe you might be misreading the code here. I've added a comment to clarify things.

There is only a single error case after the `0x21` case, and that's the inability to read from the socket. In this case, we return `-1`, which in the calling function would result in a `goto error;`, which terminates the connection. We only set the `*_ip->af` after we've done most of the v2 parsing. Same thing is true for the v1 parsing: once we get to the `TCP{4,6}` part of the line, only `0` or `-1` are possible outputs for the function.

Your comment did help me discover that I was missing the `LOCAL` case (line 1075), and mistakenly returned `0` in that case instead of `1`, so thanks for that.

> You work on a local pointer src_ip and dst_ip in the parsing part, but directly on the c->rcv.src_port/c->rcv.dst_port - is there a particular reasons for it?

Convenience. There is only one pointer access for `c->rcv.src_port` and `c->rcv.dst_port`, whereas there's two pointer accesses in `c->rcv.src_ip->...`, and multiple fields are accessed in `{src,dst}_ip` (`af`, `len`, `u`, etc). I'm happy to use a local pointer if requested.

src/core/tcp_main.c
teotwaki: There is an `LM_DBG()` indicating whether we received a PROXY v1 or PROXY v2 header. Do you feel that a parsing-success message is strictly required here (as opposed to a lack of parsing-failure message)?

miconda:
From my point of view, it can be merged.

henningw:
Thank you for the comments and extensions. About this question:

> related to error cases, what would happen if you e.g. set in the IPv6 case 0x21 the dst_ip->af = 
> AF_INET6, but later one run unto an error in the parsing. Then you would return an error to the 
> callee, and would go to the read_ip_info path. Here the AF_INET6 would be still be set, or I am 
> wrong?

I probably should formulated it better. My more general questions would be: are there error conditions that could occur during the proxy header parsing that could leave the Kamailio core with some (socket) data from the proxy protocol, and some data from the normal parsing? I am asking especially as this would be a classical border condition that could lead to e.g. some security issue.

Otherwise the code also for me looks ready to merge. Please add some documentation about this new core parameter in the wiki core cookbook (https://www.kamailio.org/wiki/cookbooks/devel/core).

PR #1787, opened 2019-01-07
secfilter: changed char* by str in RPC function params

Pepelux:
#### Description

Changed char* by str in RPC function params to avoid memory problems

### DIFF
--- a/src/modules/secfilter/secfilter_rpc.c
+++ b/src/modules/secfilter/secfilter_rpc.c
@@ -30,25 +30,27 @@
 /* RPC commands */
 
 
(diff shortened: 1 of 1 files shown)


miconda:
Just a side note for the future: passing structure variables as function parameters is not optimized if the structure is large (some compilers and static analyzers will complain). `str` is a small structure and it is fine, so not something that must to change here. Otherwise is recommended to pass the address (the pointer). What I mean is to have:

```
static int get_type(str *ctype)
```

instead of:

```
static int get_type(str ctype)
```

Issue #1783, opened 2019-01-03
dialog: Error registering a dialog that causes the cnxcc module doesn't work correctly


Pepelux:
### Description
I have been checking the reason why the cnxcc module doesn't work and I have seen that the problem is in the dialog module. Cnxcc uses dialog callback functions to obtain the data of the new dialogs. However there is a check that always gives incorrect and the creation of the new dialog is never executed. The check is this:

```
if((req->flags&dlg_flag_mask)!=dlg_flag_mask)
    return;

LM_DBG("dialog creation on config flag\n");
dlg_new_dialog(req, t, 1);
dlg = dlg_get_ctx_dialog();
```

I don't know is this check is important and I don't know if this is the best solution but removing this check the cnxcc module works fine for me.
```
diff --git a/src/modules/dialog/dlg_handlers.c b/src/modules/dialog/dlg_handlers.c
index 96406bde3..ddfdffc27 100644
--- a/src/modules/dialog/dlg_handlers.c
+++ b/src/modules/dialog/dlg_handlers.c
@@ -761,8 +761,6 @@ void dlg_onreq(struct cell* t, int type, struct tmcb_params *param)
                }
        }
        if (dlg==NULL) {
-               if((req->flags&dlg_flag_mask)!=dlg_flag_mask)
-                       return;
                LM_DBG("dialog creation on config flag\n");
                dlg_new_dialog(req, t, 1);
                dlg = dlg_get_ctx_dialog();
```

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
version: kamailio 5.3.0-dev1 (x86_64/linux) eb3fd4-dirty
flags: STATS: Off, USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLACKLIST, HAVE_RESOLV_RES
ADAPTIVE_WAIT_LOOPS 1024, MAX_RECV_BUFFER_SIZE 262144, MAX_URI_SIZE 1024, BUF_SIZE 65535, DEFAULT PKG_SIZE 8MB
[...]

PR #1791, opened 2019-01-07
dialog: in dlg_onreq function, the result of this check is always false

Pepelux:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #1783

#### Description

The result of this check is always false (or I think so because I always obtain 0!=8) and the function dlg_new_dialog() is never executed. 

The function dlg_new_dialog() must be executed to save the new dialog. In other case the cnxcc module (I don know if there are more modules affected) doesn't work.

Can anyone review if this check is important, please?

### DIFF
--- a/src/modules/dialog/dlg_handlers.c
+++ b/src/modules/dialog/dlg_handlers.c
@@ -761,8 +761,6 @@ void dlg_onreq(struct cell* t, int type, struct tmcb_params *param)
 		}
 	}
 	if (dlg==NULL) {
-		if((req->flags&dlg_flag_mask)!=dlg_flag_mask)
-			return;
 		LM_DBG("dialog creation on config flag\n");
 		dlg_new_dialog(req, t, 1);
 		dlg = dlg_get_ctx_dialog();


miconda:
Did you set the `dlg_flag` parameter for dialog module?

  * https://www.kamailio.org/docs/modules/stable/modules/dialog.html#dialog.p.flag

If yes, that is not required if you use `dlg_manage()`. If you set the modparam, then you have to do setflag(X) where X is the value for dlg_flag modparam.

Otherwise, if I didn't miss something, the condition there should to be ok to execute the code - dlg_flag_mask is 0 by default (with no dlg_flag set) and the expression evaluates to if(0!=0) return:

```
if((req->flags&dlg_flag_mask)!=dlg_flag_mask) return;
``

**Comment by Pepelux** [comment]
You are right, sorry for the noise. But I have some doubts.

You said that dlg_flag parameter is not required but if I don't set the dlg_flag, kamailio gives me an error when load dialog module:
```
Jan  9 10:53:08 pepelux /usr/local/sbin/kamailio[27387]: ERROR: dialog [dialog.c:519]: mod_init(): invalid dlg flag -1!!
Jan  9 10:53:08 pepelux /usr/local/sbin/kamailio[27387]: ERROR: <core> [core/sr_module.c:911]: init_mod(): Error while initializing module dialog (/usr/local/lib64/kamailio/modules/dialog.so)
```

It worked for me as follow:
```
#!define DLG_FLAG 3
modparam("dialog", "dlg_flag", DLG_FLAG)
route(CNXCC);
dlg_manage();
```

My problem was that I called route(CNXCC) after dlg_manage() and it seems that it is necessary to execute cnxcc functions before the dlg_manage().

Thanks for the help

miconda:
Hmm, I thought that requiring to set dlg_flag was removed, but apparently it is not. We should get rid of that, needs to be fixed somehow.

miconda:
I just pushed a patch (404f403786478a917637186446bb285e8e32cc59) that should remove the requirement of setting the dlg_flag parameter. Can you give it some tests and confirm if all works fine or not? When all ok, it can be backported.

Pepelux:
All works fine for me. Thanks for the fix

PR #1815, opened 2019-01-20
kamctl: keep it POSIX

micmac1:
Hi all,

I was trying to add entries to a remote database on OpenWrt and kamctl (address add ...) seemed to just hang there. A bit of digging showed that it prompted for a password, but I couldn't see it because it redirected the prompt to /dev/stderr, which doesn't exist on every system. This patch addresses this.

Use portable "#!/bin/sh" shebang. This may still be a symlink to bash
(or some other shell).

Also replace the redirection to "/dev/stderr" with a POSIX redirection
to stderr. "/dev/stderr" is not available on all systems and is also not
recognized by all shells (example: ash). On systems without a working
"/dev/stderr" the user will not see the prompts etc.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/utils/kamctl/kamctl
+++ b/utils/kamctl/kamctl
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 #
 # control tool for maintaining Kamailio
 #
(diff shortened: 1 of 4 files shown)


henningw:
Thank you, merged after small adaption to the log message

PR #1816, opened 2019-01-20
modules: fix hiredis handling

micmac1:
The define WITH_HIREDIS_PATH causes the source files to include
"<hiredis/hiredis.h>". But the Makefiles check HIREDISDEFS for the
string "hiredis" and set WITH_HIREDIS_PATH if not found. Which is
exactly the opposite of what is needed. This commit corrects the logic.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/modules/db_redis/Makefile
+++ b/src/modules/db_redis/Makefile
@@ -19,7 +19,7 @@ else
 	HIREDISDEFS = $(shell $(HIREDIS_BUILDER) --cflags)
 	HIREDISLIBS = $(shell $(HIREDIS_BUILDER) --libs)
 
-ifeq (,$(findstring hiredis,$(HIREDISDEFS)))
+ifneq (,$(findstring hiredis,$(HIREDISDEFS)))
 	DEFS+=-DWITH_HIREDIS_PATH
 endif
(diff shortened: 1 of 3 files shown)


miconda:
Did you have issues compiling?

Iirc, we tested to get this solution and at least the deb packages built fine for the past months since this change, after having couple of builds failed back then.

What it tries to do is that if pkg-config for cflags is not giving a path that has `hiredis` in it, then include `<hiredis/hiredis.h>`. If what pkg-config returns has `hiredis`, then just include `<hiredis.h>`.

I got on different systems and versions, either (1):

```
-I/opt/local/include/hiredis
```

or (2):

```
-I/opt/local/include
```

So for (1), include `<hiredis.h>` and for (2), include `<hiredis/hiredis.h>`.

Am I wrong in what is doing it right now?

Maybe there is something else to fix, but I find the current behaviour as expected.

micmac1:
Hello Daniel,

Sorry, my mistake. The current handling is correct.

I'll close this pull request.

With kind regards,
Seb

PR #1786, opened 2019-01-04
utils/kamctl/kamdbctl: Merged patch from alpine packaging

sergey-safarov:
#### Pre-Submission Checklist
- [x] Commit message has the format required by CONTRIBUTING guide
- [ ] Commits are split per component (core, individual modules, libs, utils, ...)
- [ ] Each component has a single commit (if not, squash them into one commit)
- [x] No commits to README files for modules (changes must be done to docbook files
in `doc/` subfolder, the README file is autogenerated)

#### Type Of Change
- [ ] Small bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would change existing functionality)

#### Checklist:
- [ ] PR should be backported to stable branches
- [ ] Tested changes locally
- [ ] Related to issue #XXXX (replace XXXX with an open issue number)

#### Description
This also fix docker packaging on alpine dist after eb3fd4c2672a18f3851ca255b9c6cb40cb0829ad

PR #1826, opened 2019-01-31
[5.1] set of cherry-pick to properly package on alpine dists

sergey-safarov:


miconda:
The patch for 78e948d does not seem to touch kamdbctl at all, so the commit message is not matching what the patch does.

sergey-safarov:
At https://github.com/kamailio/kamailio/pull/1786 commit partially (but not all) is back ported to 5.1 branch.
Remaining part is important for `kamailio:kamailio-ci` docker image and here is cherry-picked remaining strings.

miconda:
For the future, a patch that touches different components must be split in commits per component. Right now that commit has nothing to do with kamdbctl. Overall, it is not something critical, but is good to keep the commit message coherent with what it does, because the commits history is used quite often to try to track back changes.
I am going to merge it, thanks.

PR #1832, opened 2019-02-01
dialog: Add setting to loop BYE through proxy

gaaf:
The setting "keep_proxy_rr" will add the Record-Route headers added by the
proxy to the route_set stored in the dialog. When in use, sending locally
generated in-dialog requests will loop back to the proxy with a proper
Record-Route header, including any parameters.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally on 5.1, compile tested on master

### DIFF
--- a/src/modules/dialog/dlg_handlers.c
+++ b/src/modules/dialog/dlg_handlers.c
@@ -62,6 +62,7 @@ static int       dlg_flag_mask=0;	/*!< flag for dialog tracking */
 static pv_spec_t *timeout_avp;		/*!< AVP for timeout setting */
 static int       default_timeout;	/*!< default dialog timeout */
 static int       seq_match_mode;	/*!< dlg_match mode */
+static int       keep_proxy_rr;		/*!< keep the proxy's record-route in both route-sets */
 static int       shutdown_done = 0;	/*!< 1 when destroy_dlg_handlers was called */
 extern int       detect_spirals;
 extern int       dlg_timeout_noreset;
@@ -103,7 +104,7 @@ int dlg_set_tm_waitack(tm_cell_t *t, dlg_cell_t *dlg);
  */
 void init_dlg_handlers(char *rr_param_p, int dlg_flag_p,
 		pv_spec_t *timeout_avp_p ,int default_timeout_p,
-		int seq_match_mode_p)
+		int seq_match_mode_p, int keep_proxy_rr_p)
 {
 	rr_param.s = rr_param_p;
 	rr_param.len = strlen(rr_param.s);
@@ -113,6 +114,7 @@ void init_dlg_handlers(char *rr_param_p, int dlg_flag_p,
 	timeout_avp = timeout_avp_p;
 	default_timeout = default_timeout_p;
 	seq_match_mode = seq_match_mode_p;
+	keep_proxy_rr = keep_proxy_rr_p;
 }
 
 
@@ -184,7 +186,7 @@ static inline int add_dlg_rr_param(struct sip_msg *req, unsigned int entry,
 int populate_leg_info( struct dlg_cell *dlg, struct sip_msg *msg,
 	struct cell* t, unsigned int leg, str *tag)
 {
-	unsigned int skip_recs;
+	unsigned int skip_recs, own_rr = 0;
 	str cseq;
 	str contact;
 	str rr_set;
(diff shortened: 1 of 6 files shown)


henningw:
@gaaf Thank you, can you add a bit here about the targeted use case for this functionality?

miconda:
@gaaf - I see that dlg_update_rr_set() is used only for DLG_CALLER_LEG, no need to do it for CALLEE as well?

Having an option to update Record-Route sets within dialog can be useful, I think at this moment they are stored only during the call setup.

On the other hand, for what you try to achieve, I wonder if it would not be simpler to use somehow the caller/callee sockets fields, they are already stored inside the dialog record and they should overlap in address, port and transport with local record-route. So based on modparam, the local requests are sent to the caller/callee socket, with top record route built at storage time or at send time (store vs cpu question here, like none that relevant to decide for one or the other option) -- this just as an idea, I haven't really analyzed if it is a simpler solution, I just thought that those addresses are already stored.

gaaf:
@henningw: In the current situation, the local Record-Route headers are stripped before they are stored in the dialog module. If a request is sent by the dialog module (forced BYE for example) it completely bypasses the routing script as it will be sent to the remote parties directly. With this PR the local Record-Route headers will not be stripped so the request will be looped back to the proxy as the first Route header is now the proxy itself. This causes the request to follow the routing script as if it was sent by one of the parties in the call and all processing that would normally be done for an "external" request would also be done for this request.

@miconda:
- The callee is handled only when the dialog is confirmed. As there is no previous RR set, no update is needed. The caller's leg is preliminary set when the request is handled, but the final RR set is only known when the dialog gets confirmed (dependent on the winning branch), hence the update is needed.
- Changing the RR in-dialog is not the intention of this PR.
- The sockets don't store (record-)route parameters so this would not be enough. As alternative, maybe using local_route is possible, but it still needs access to RR parameters to be complete. As this patchset predates the local_route functionality I have not investigated much into that direction.


In short, this PR makes the dialog module to not strip the local record-route headers, something I have always considered to be a bug.

henningw:
Thank you for the explanation, understand it now. I will have a brief look to the patch and comment there if I spot something.

src/modules/dialog/dlg_hash.c
```diff
@@ -528,7 +533,7 @@ int dlg_set_leg_info(struct dlg_cell *dlg, str* tag, str *rr, str *contact,
 
 	if(dlg->tag[leg].s)
 		shm_free(dlg->tag[leg].s);
-	dlg->tag[leg].s = (char*)shm_malloc( tag->len + rr->len );
```
henningw: Why you decrease here the shm malloc'ed area, do you change something in the tag as well?

src/modules/dialog/dlg_handlers.c
```diff
@@ -269,6 +271,25 @@ int populate_leg_info( struct dlg_cell *dlg, struct sip_msg *msg,
 
 	if (rr_set.s) pkg_free(rr_set.s);
```
henningw: This free is in current git master at the end of this function and seems redundant now. Maybe it make sense to combine it with the other free in line 290 and move it to the end of the function?

src/modules/dialog/dlg_handlers.c
```diff
@@ -269,6 +271,25 @@ int populate_leg_info( struct dlg_cell *dlg, struct sip_msg *msg,
...
+			goto error0;
+		}
+		LM_DBG("updating caller route_set %.*s\n",
+			rr_set.len, rr_set.s);
+		if (dlg_update_rr_set( dlg, DLG_CALLER_LEG, &rr_set)!=0) {
+			LM_ERR("dlg_update_rr_set failed\n");
+			if (rr_set.s) pkg_free(rr_set.s);
+			goto error0;
+		}
+		if (rr_set.s) pkg_free(rr_set.s);
```
henningw: see above

henningw:
@gaaf This is apparently still open, I would like to resolve this for the upcoming 5.3 to actually merge it. I've had a minor few remarks/questions to the code, would be great if you can have a look to that.

src/modules/dialog/dlg_handlers.c
gaaf: print_rr_body() allocates the memory for rrset.s, it is up to the caller to free it. As it may be allocated twice, it must be freed twice and that must be done before the next call to print_rr_body(). Each free is in the same scope as the call to print_rr_body(). I fail to see how these can be combined into one free.

src/modules/dialog/dlg_handlers.c
gaaf: see above

src/modules/dialog/dlg_hash.c
gaaf: It is not necessarily decreased, it is just realloc'ed to the size of the new tag, which may be bigger or smaller.

The tag is not "changed" it is stored untouched from the incoming response.

There may be multiple to-tag's in a single dialog as the call may have been forked. The dialog module does not track individual branches separately, it just stores 1 of them, we just hope it is the correct one. When storing the tag of the 200 OK, it may be overwriting a previous tag (from a 1XX response).

henningw: Thank you, understood.

src/modules/dialog/dlg_handlers.c
henningw: Thanks for the clarification

henningw:
Thank you!

Issue #1848, opened 2019-02-11
TOPOS/REDIS : returning unexpected 481 not found after 4 minutes


jchavanton:
### Description

BYE requests received after 4 minutes are failing with 481

#### Debugging Data

Seems like REDIS is set to expire the leg infomartion after 3 minutes.
Not sure where / why this is the case.

```
T 2019/02/11 16:49:55.818935 137.54.60.160:58580 -> 137.54.2.161:6379 [AP]
  *3..$6..EXPIRE..$26..d:z:atpsh-5c3d1562-16-cf25..$5..10800..
#
T 2019/02/11 16:50:21.325761 137.54.60.160:58580 -> 137.54.2.161:6379 [AP]
  *3..$6..EXPIRE..$26..d:z:atpsh-5c3d1562-16-df25..$5..10800..
#
T 2019/02/11 16:50:26.582118 137.54.60.160:58574 -> 137.54.2.161:6379 [AP]
  *3..$6..EXPIRE..$26..d:z:atpsh-5c3d1562-16-df25..$3..180..
#
T 2019/02/11 16:50:43.299621 137.54.60.160:58574 -> 137.54.2.161:6379 [AP]
  *3..$6..EXPIRE..$26..d:z:atpsh-5c3d1562-16-df25..$3..180..
#
```

I will troubleshoot this further.

jchavanton:
https://lists.kamailio.org/pipermail/sr-users/2018-March/100835.html

henningw:
Do you get the same error messages as in the linked mailing list thread?

miconda:
@jchavanton - the expires of 180 is set on dialog key when the BYE is processed.

Can you reproduce the case? Can you look at the sip traffic?

What is your kamailio version?

jchavanton:
Thanks for the hint about the BYE/EXPIRE 180s, good to know about this expected behavior.

I will get back with more debugging information/messages when I can grab them.

The version I am using in this case is a little bit forked , I did look and but the module TOPOS and TOPOS_REDIS are the same as latest master.

Hope to gather more information soon.

jchavanton:
```
Feb 12 16:18:33 localhost 11f127b633b3[15693]:  2(18) NOTICE: <script>: [MAIN][INVITE][2017998478_92254488@74.120.93.200] from[+12063889971] to[+12012673277]
[...]

PR #1849, opened 2019-02-12
topos_redis: fix branch insert expire duration

jchavanton:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [x] Related to issue #1848

### DIFF
--- a/src/modules/topos/tps_msg.c
+++ b/src/modules/topos/tps_msg.c
@@ -1003,7 +1003,8 @@ int tps_request_sent(sip_msg_t *msg, int dialog, int local)
 	}
 
 	if(dialog!=0) {
-		tps_storage_end_dialog(msg, &mtsd, ptsd);
+		if(get_cseq(msg)->method_id==METHOD_BYE)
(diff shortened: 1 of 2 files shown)


miconda:
The commit 728bbc4 does not seem right, it is setting branch expire to the dialog value, which is not correct.

As for commit cf70716 -- the tps_storage_end_dialog() has a check inside to be executed only for BYE:

  * https://github.com/kamailio/kamailio/blob/master/src/modules/topos_redis/topos_redis_storage.c#L1322

Can you check if you have the same code base, give that you said you run some custom version.

If you have it, then maybe the issue is somewhere else, not setting properly the method id in the structure.

jchavanton:
After double checking the code base I am missing a few old commits ...

Sorry, I made a mistake when checking initially, I will test with the right source code.

And update the issue.

jchavanton:
These modifications are not needed, the related issue does not exist on the latest master branch.

PR #1852, opened 2019-02-15
 kamdbctl: added verification before reinit 

fredposner:
- add get_answer and warning that reinit will drop exisiting database
- even seasoned users forget and theres no backing out

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

If one _accidentally_ runs kamdbctl reinit, there is no easy way to abort without dropping your existing database. Adding the get_answer provides a warning that the database will drop and a method of exiting without damage.

### DIFF
--- a/utils/kamctl/kamdbctl
+++ b/utils/kamctl/kamdbctl
@@ -386,6 +386,13 @@ case $1 in
 	reinit)
 		# delete database and create a new one
 		# create new database structures
+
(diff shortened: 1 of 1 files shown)


henningw:
Thank you! Maybe it would be also good to add the same logic for the drop command as well.

PR #1869, opened 2019-02-26
General improvements to imc module

janakj:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This patch set contains a bunch of improvements to the imc module. Most of the changes are backwards-compatible, but some change the semantics of individual imc commands slightly. I have tested the changes with the Zoiper client on Android and with a custom WebRTC client. The motivation behind these changes was to make the imc module a bit more consistent, predictable, and usable.

### DIFF
--- a/src/modules/imc/README
+++ b/src/modules/imc/README
@@ -363,7 +363,7 @@ kamcmd imc_list_members _room_
   -eg: #remove sip:john@kamailio.org, sent to sip:chat-000@kamailio.org
   -error case: return codes: -80 -- -89
 
-7.exit
+7.leave
   -leaving a room
   -takes one optional parameter - the address of the room - if not
     present it will be considered to be the address in the To header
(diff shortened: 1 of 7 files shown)


miconda:
Thanks! I applied you commits manually with `git am ...` after editing the first line of each commit message to follow the contributing guidelines:

Respectively, I added the `imc: ` prefix there.

We use git log to generate the changelog for each release and it is useful to spot quickly which component was affected by that commit.

Another remark for the future: you do not need to generate the README locally and then create a commit for it. You can generate it locally to check if all is ok with the docbook syntax, but then you can delete the file and do git checkout path/to/the/README. This is because we generate the readme files for modules via a cron.d job on kamailio server in order to have coherent format of the output. In the past, a small change to the docbook file could end up in a completely different style of whitespacing in the README, depending on the system where the README was generated.

Last, you have git write privileges, so you can push commits directly without making pull requests. If you want someone to review the patches, then of course you can do pull requests (btw, I haven't really reviewd the code of your commits in this pull request at this moment, only edited the patch set for proper commit message format). The imc module doesn't have an active maintainer, you can step in if you plan to work more on this module, so there is no need for additional review unless you want explicitly.

PR #1877, opened 2019-03-03
dispatcher: remove destination from in-memory dispatcher set

tverlaan:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

- add dispatcher.remove call to remove entry from in-memory dispatcher
list
- allow arguments to callback in ds_iter_set()

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -176,7 +176,7 @@ int ds_hash_load_destroy(void)
 /**
  * Recursivly iterate over ds_set and execute callback
  */
-void ds_iter_set(ds_set_t *node, void (*ds_action_cb)(ds_set_t *node, int i))
+void ds_iter_set(ds_set_t *node, void (*ds_action_cb)(ds_set_t *node, int i, void *arg), void *ds_action_arg)
 {
 	if(!node)
 		return;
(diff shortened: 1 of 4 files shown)


henningw:
Hello, thank you for the pull request. I had a quick look to the diff - looks fine for me. One question though, I noticed that you changed also the attr str handling in ds_add_dst() function, why was this changed as well?

tverlaan:
Hi, thanks for taking a look. I changed it since I saw that pack_dest checks if attr is not NULL and now removing and adding through rpc both set it to NULL instead of creating a zeroed string when calling that function. It looks a bit more uniform to me.

miconda:
Iirc, the attrs set to empty string was an workaround to be sure the AVPs with attrs value do not get shifted (associated) with a wrong destination. The AVP cannot be null, so if among selected destinations some of them were with attrs and some without, then the number of AVPs with attrs was lower than the number of destinations.

Maybe is no longer needed with xavps, as the fields related to a destination are kept in an XAVP sublist. But that has to be reviewed to see if there is any side effect. Might be better to split this chance in its own commit.

tverlaan:
I removed the change from this commit and focused this PR entirely on the feature addition.

I'd have to do some more investigation to understand full scope. I'll reconsider a PR for that when I better understand. Thanks!

miconda:
OK, merging, thanks!

PR #1882, opened 2019-03-05
fixes & improvements

lazedo:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

- added a worker_id for timer routines. when using rtimer in script with mode=4 there's no way to know which of the workers is in use. this is useful for db queries based on slot
- fixed {uri.tosocket} transformation
- added t_release_transaction to tm api
- use t_release_transaction in tmx `t_drop`
- fixed xavp_clone_level_nodata
- add sockinfo to $subs
- increase `MAX_DEFINES` as we run out of  `defines`

Note: please use rebase instead of merge for this PR

to backport:
- https://github.com/kamailio/kamailio/commit/eea9c5f72b73df9c3d898cff18dced6b7f72c7d4
- https://github.com/kamailio/kamailio/commit/e3ed896d80171feba78d3292484a09f0016e68e0
- https://github.com/kamailio/kamailio/commit/b4696169e8b464da01774b1e40af677f941c803b

### DIFF
--- a/src/core/cfg.lex
+++ b/src/core/cfg.lex
@@ -1718,7 +1718,7 @@ static int sr_pop_yy_state()
 
 /* define/ifdef support */
 
-#define MAX_DEFINES    256
+#define MAX_DEFINES    512
 static ksr_ppdefine_t pp_defines[MAX_DEFINES];
 static int pp_num_defines = 0;
 static int pp_define_type = 0;
@@ -1728,7 +1728,7 @@ static int pp_define_index = -1;
  * ifdef(defined), ifndef(undefined), or the opposite of these
  * two, but in an else branch
  */
-#define MAX_IFDEFS    256
+#define MAX_IFDEFS    512
 static int pp_ifdef_stack[MAX_IFDEFS];
 static int pp_sptr = 0; /* stack pointer */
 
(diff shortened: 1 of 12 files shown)


henningw:
Hi, thank you for the request. I guess you mean "rebase and merge" instead of "squash and merge", understood. I will look tomorrow to it that we can get the backported fixes into the upcoming maintenance release. One question: can you explain a bit the reason for change e3ed896, especially why this needs to be backported?

lazedo:
i see it as a bug since `tosocket` is using `msg` to determine the transport instead of parsed uri from input, try to run it in a timer route and you'll see but i guess that it will also give wrong result if you receive a `udp` message and then try to apply the transform to "sip:xxx@yyy;transport=tcp". as for the backport, i think this is in 5.2 but not 5.1.

henningw:
Thank you for the explanation, request was merged. The necessary backports (three to 5.2, xavp clone issue to 5.1) were done as well.

henningw:
Could you add some documentation for 186d200 in the module README section 7 as well? Does the change in 20fc14f maybe needs to be documented in the wiki or module README as well?

Issue #1827, opened 2019-01-31
bug: RPCFIFOPATH / DEFINE_FIFO_NAME invalid acces due changed to root at run


mckaygerhard:
### Description

the /var/run/kamailio directory for the fifo and crtl files are  denied for acces to the service
as i can see directory are owed byt root and only ower and group have
access 

kamailio 4.4 backported by me for Debian jessie have this:
`drwxr-xr-x 2 kamailio kamailio 100 Aug 15 17:33 kamailio`
kamailio 5.2 installed on buster/testing build from sources:
`drwxr-x--- 2 root root 120 Jan 30 09:35 kamailio`

**NOTE** no definiton are setup, and surprise if i defined some, the only working dir are /tmp !!!

### Troubleshooting

kamailio 4.4 backported by me for Debian jessie have this:
`drwxr-xr-x 2 kamailio kamailio 100 Aug 15 17:33 kamailio`
kamailio 5.2 installed on buster/testing build from sources:
`drwxr-x--- 2 root root 120 Jan 30 09:35 kamailio`

due that several issues are happened if we started using some modules
due unable to manage the socket control:

```
Jan 31 15:35:51 venenux-debian kamailio[16861]: 17(16889) ERROR:
<core> [core/io_wait.h:1040]: io_wait_loop_epoll(): epoll_wait(7,
0x7f6ed6f88d88, 0, 5000): Invalid argument [22]
Jan 31 15:35:51 venenux-debian kamailio[16861]: 0(16863) ERROR: ctl
[ctl.c:390]: mod_destroy(): ERROR: ctl: could not delete unix socket
/var/run/kamailio//kamailio_ctl: Permission denied (13)
Jan 31 15:35:51 venenux-debian kamailio[16861]: 0(16863) DEBUG: <core>
[db_pool.c:100]: pool_remove(): removing connection from the pool
Jan 31 15:35:51 venenux-debian kamailio[16861]: 0(16863) DEBUG: tm
[t_funcs.c:85]: tm_shutdown(): start
Jan 31 15:35:51 venenux-debian kamailio[16861]: 0(16863) DEBUG: tm
[t_funcs.c:88]: tm_shutdown(): emptying hash table
Jan 31 15:35:51 venenux-debian kamailio[16861]: 0(16863) DEBUG: tm
[t_funcs.c:90]: tm_shutdown(): removing semaphores
Jan 31 15:35:51 venenux-debian kamailio[16861]: 0(16863) DEBUG: tm
[...]

PR #1889, opened 2019-03-12
Systemd unit update

sergey-safarov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [ ] Tested changes locally
- [ ] Related to issue #1827

#### Description
1) tmpfiles creation is implemented via systemd
2) owner of kamailio process implemented via systemd drop-in
3) kamailio started as simple process, that allow systemd check kamailio process without PID file.

### DIFF
--- a/pkg/kamailio/deb/buster/kamailio.service
+++ b/pkg/kamailio/deb/buster/kamailio.service
@@ -5,17 +5,17 @@ After=network-online.target
 
 [Service]
 Type=forking
+User=kamailio
+Group=kamailio
 Environment='CFGFILE=/etc/kamailio/kamailio.cfg'
 Environment='SHM_MEMORY=64'
 Environment='PKG_MEMORY=8'
-Environment='USER=kamailio'
-Environment='GROUP=kamailio'
 EnvironmentFile=-/etc/default/kamailio
 EnvironmentFile=-/etc/default/kamailio.d/*
 # PIDFile requires a full absolute path
 PIDFile=/var/run/kamailio/kamailio.pid
 # ExecStart requires a full absolute path
-ExecStart=/usr/sbin/kamailio -P /var/run/kamailio/kamailio.pid -f $CFGFILE -m $SHM_MEMORY -M $PKG_MEMORY -u $USER -g $GROUP
+ExecStart=/usr/sbin/kamailio -P /var/run/kamailio/kamailio.pid -f $CFGFILE -m $SHM_MEMORY -M $PKG_MEMORY
 Restart=on-failure
 # /var/run/kamailio in tmpfs
 RuntimeDirectory=kamailio
--- a/pkg/kamailio/deb/buster/kamailio.service
+++ b/pkg/kamailio/deb/buster/kamailio.service
@@ -4,7 +4,7 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-Type=forking
+Type=simple
 User=kamailio
 Group=kamailio
 Environment='CFGFILE=/etc/kamailio/kamailio.cfg'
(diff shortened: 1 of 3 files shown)


sergey-safarov:
Hello Victor @linuxmaniac 
Could you look this PR.
Changes is spited by commit.
After you approve i will squash commit into one and create new commit message like
```
pkg/kamailio/deb: Updated systemd service and tmpfile files
```

pkg/kamailio/deb/buster/kamailio.service
```diff
@@ -4,18 +4,16 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-Type=forking
+Type=simple
+User=kamailio
```
mckaygerhard: Hi!, thanks for this pull request.. finally after many fights and ignored mails, but now a small request (if it can be done): there's a way to pass the user and group from environment or invoke command? i mean for packaging.. in debian exits the /etc/default/kamailio file that set user and group...

sergey-safarov: To change username at runtime need place drop-in into `/run/systemd/system/kamailio.service.d` directory and apply changes by `systemctl daemon-reload`
For dynamic unit updates you cannot use `/etc/systemd/system/kamailio.service.d` directory.

Also you can use [service templates](https://fedoramagazine.org/systemd-template-unit-files/)

sergey-safarov: `there's a way to pass the user and group from environment or invoke command?`
Please look d3cc4e4c10820e538e7cc73b839140e38df09756
kamailio default have complete description how to change user/group of kamailio process.

pkg/kamailio/deb/buster/kamailio.service
```diff
@@ -4,18 +4,16 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
-Type=forking
+Type=simple
```
verticelo: How will this interact with the default setting of fork=yes in the config?

pkg/kamailio/deb/buster/kamailio.service
```diff
@@ -4,18 +4,16 @@ Wants=network-online.target
...
 Environment='PKG_MEMORY=8'
-Environment='USER=kamailio'
-Environment='GROUP=kamailio'
 EnvironmentFile=-/etc/default/kamailio
 EnvironmentFile=-/etc/default/kamailio.d/*
-# PIDFile requires a full absolute path
-PIDFile=/var/run/kamailio/kamailio.pid
 # ExecStart requires a full absolute path
-ExecStart=/usr/sbin/kamailio -P /var/run/kamailio/kamailio.pid -f $CFGFILE -m $SHM_MEMORY -M $PKG_MEMORY -u $USER -g $GROUP
+ExecStart=/usr/sbin/kamailio -DD -P /var/run/kamailio/kamailio.pid -f $CFGFILE -m $SHM_MEMORY -M $PKG_MEMORY
```
verticelo: Changing to this mode, will it have any impact on performance of Kamailio? If so, in what direction?

sergey-safarov: I did not observe the influence

pkg/kamailio/deb/buster/kamailio.service
sergey-safarov: This systemd setting not kamailio process.
Interactions is not exist.

mckaygerhard:
seems systemd made more problems thatn solves https://github.com/kamailio/kamailio/issues/1898

sergey-safarov:
@mckaygerhard your reference not relevant to systemd.

miconda:
@linuxmaniac - these changes are on systemd unit for debian packaging. Any conclusion on what to do with this PR?

If decided to merge, quash in a single commit and set a proper formatted commit message, because now the commits do not follow the rules.

sergey-safarov:
I ready to squash and update commit after review.

pkg/kamailio/deb/buster/kamailio.service
verticelo: But, doesn't this determine how systemd monitors the process after the process has started, ie. if the kamailio process forks or not?

sergey-safarov: yes, that is determines how to monitor kamailio process.
I suggest to start kamailio with `-DD`  option.
In this case kamailio main process control own child's, and systemd monitors only kamailio main process.

miconda:
@linuxmaniac - any conclusion this one?

miconda:
Does this one still needs to be addressed, or are we fine with the version included in the deb/rpm packages right now?

kamailio-sync:

> Does this one still needs to be addressed, or are we fine with the
> version included in the deb/rpm packages right now?

It works right now (with the syslog message), but according systemd
documentation, the dir is /run, not /var/run:

https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=

Perhaps the best would be to use $RUNTIME_DIRECTORY environment var.

-- Juha

kamailio-sync:
sergey-safarov writes:

> Now unit contains
> ```
> RuntimeDirectory=kamailio
> ```
...
> Look as current definition correct

Yes, RuntimeDirectory is correct, but not this:

PIDFile=/var/run/kamailio/kamailio.pid

since it hardwires /var/run rather than using $RUNTIME_DIRECTORY.

-- Juha

PR #1814, opened 2019-01-18
pv: add $def pseudo variable

lazedo:
#### Pre-Submission Checklist

- [X] Commit message has the format required by CONTRIBUTING guide
- [X] Commits are split per component (core, individual modules, libs, utils, ...)
- [X] Each component has a single commit (if not, squash them into one commit)
- [X] No commits to README files for modules (changes must be done to docbook files
in `doc/` subfolder, the README file is autogenerated)

#### Type Of Change
- [X] Small bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would change existing functionality)

#### Checklist:

- [ ] PR should be backported to stable branches
- [X] Tested changes locally
- [ ] Related to issue #XXXX (replace XXXX with an open issue number)

#### Description

allows the use definitions as pseudo-variables
```
#!trydef XIP "$HN(i)"
#!substdef "!MY_IP_ADDRESS!$def(XIP)!g"

#!substdef "!MY_HOSTNAME!$HN(f)!g"
#!trydef REGISTRAR_MIN_EXPIRES 300

loadmodule "registrar.so"
modparam("registrar", "min_expires", REGISTRAR_MIN_EXPIRES)

route[SOMETHING]
{
xlog("L_INFO", "configured value of registrar min-expires is $def(REGISTRAR_MIN_EXPIRES)\n");
xlog("L_INFO", "configured value of host is MY_HOSTNAME\n");
xlog("L_INFO", "configured value of ip is MY_IP_ADDRESS\n");
}

```

PR #1893, opened 2019-03-16
[5.2] pv: add $def pseudo variable

sergey-safarov:
That is back port of https://github.com/kamailio/kamailio/pull/1814
Kazoo still use 5.0/5.1 branch

I expect if created backport into 5.1 branch, than backport to 5.2 also need.


miconda:
We do not backport new features to stable branches. This set of new additions will be part of 5.3.

PR #1894, opened 2019-03-16
[5.1] pv: add $def pseudo variable

sergey-safarov:
That is backport of #1814
Kazoo still use 5.0/5.1 branch


miconda:
We do not backport new features to stable branches. This set of additions will be part of 5.3.

PR #1903, opened 2019-03-22
etc/kamailio.cfg: detect sipvicious as scanner

kevin-olbrich:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

### DIFF
--- a/etc/kamailio.cfg
+++ b/etc/kamailio.cfg
@@ -581,7 +581,7 @@ route[REQINIT] {
 		}
 	}
 #!endif
-	if($ua =~ "friendly-scanner|sipcli|VaxSIPUserAgent") {
+	if($ua =~ "friendly-scanner|sipcli|sipvicious|VaxSIPUserAgent") {
 		# silent drop for scanners - uncomment next line if want to reply
 		# sl_send_reply("200", "OK");
 		exit;


sergey-safarov:
Probable you want look other [scanners UA](https://github.com/2600hz/kazoo-configs-kamailio/blob/master/kamailio/default.cfg#L402-L407).

miconda:
Thanks!

kevin-olbrich:
> Probable you want look other [scanners UA](https://github.com/2600hz/kazoo-configs-kamailio/blob/master/kamailio/default.cfg#L402-L407).

IMHO adding these makes sense. What do you think?

henningw:
To add the missing user agents from the kazoo cfg snippet sounds good to me.

miconda:
I don't think it is good to add a lot of string there. The default configuration files is more like a reference. sipvicious is a well known one and made sense (although most of its versions use `friendly-scanner` in user-agent).

But, afaik, portsip is an sdk and pbx. Then, others like NiceGuy likely to be sipvicious with a different value set for user agent. If the attacker was changing the default value to something else, banning that one by regex won't get it fixed for too long, the attacker will use a different value soon.

In other words, if it is something very common or a tool used for scanning attacks, makes sense to add. If it is just different user agent name used by an attacker against a platform, it should not be added in the default config file. There is pike+htable that can detect DoS attacks.

PR #1910, opened 2019-03-27
AUTH_USERNAME_EXPIRED auth api return code

juha-h:
This pull request adds AUTH_USERNAME_EXPIRED auth api return code and uses it in ephemeral authentication when username has expired.  Script can then tell the user about the expiration with a suitable response.


miconda:
Thanks!

PR #1912, opened 2019-03-28
extend siptrace module tracing methods

ionutionita92:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Siptrace module has two different tracing concepts which do mostly the same thing. It has sip_trace function and it has trace_flag parameter which traces messages in a transaction. This pull request aims to have a single point of trace activation, that is when the user calls sip_trace() function from the script. Moreover, it adds 3 levels of tracing: message, transaction and dialog level tracing. This option is added as a parameter to sip_trace(), therefore the user can now summon sip_trace(duri, correlation_id, "m/t/d") where "m" stands for message, "t" for transaction and "d" for dialog.

The patch has been tested both locally and in our company's test system. Locally I've tested it using a few scenarios generated using sipp: successful call, negative reply(491 answer - hop-by-hop ACK), UAC CANCEL and dual bye scenario generated by Kamailio(dialog timeout). The only issue with current patch is it doesn't trace internally generated BYEs that is because they are only a buffer in Kamailio, not a parsed SIP message, therefore, after a few days of searching for solutions I didn't find any valid one. Parsing a SIP message for tracing didn't seem worth the processing time to me, but of course it can be added in the future.

Looking forward to your opinion or suggestions regarding the current patch.


lazedo:
@ionutionita92 hello, currently we can enable/disable the trace flag, this can be done in script processing in generic and/or specific routes. By removing this you are removing flexibility in script processing as there is no `sip_untrace` function.

ionutionita92:
Hi @lazedo. When writing the patch I thought one wouldn't have a reason to disable tracing if it's already enabled. One could simply not enable the feature. But I see in your situation you can't avoid these types of scenarios. I can add this feature back, of course.

adubovikov:
for me it looks good :-) :+1:  Thank you!

src/modules/siptrace/doc/siptrace_admin.xml
```diff
@@ -656,6 +654,11 @@ hlog("$hdr(P-MyID)", "Another one with a custom correlation ID");
 	Stateless forwarded messages (forward()) are not logged if you set the
 	flag, use sip_trace() inside <emphasis>onsend_route</emphasis> block.
 	</para>
+	<para>
+	If dialog level tracing is used internally generated BYE transaction(in
+	case of internal timeouts) won't be traced. At the moment kamailio offers
+	no posibility to trace those messages.
```
grumvalski: Actually it should be possible to capture these BYEs, as we do for ACK, in onsend_route or in tm's module local_route.

miconda:
So, to summarize what I could see being discussed here, if the sip flag that will allow disabling tracing is back, then it can be merged.

src/modules/siptrace/doc/siptrace_admin.xml
ionutionita92: Thanks for mentioning. I've found callbacks for these BYEs but inside them the message is not parsed, I can only access the char buffer. There's no sip_msg structure created, therefore I needed to parse it in order to trace it. I didn't know whether it would be a good idea to parse the message just for it to be traced since SIP message parsing is an expensive operation. What are your thoughts about this?

miconda:
@ionutionita92 - sip parsing is no longer something very expensive, if can be avoided, it is good of course. Eventually, it can be made a modparam to control if one wants to do it at that stage or not. Sometimes is better to have the option of enabling a feature at the expense of some processing time, given these days horizontal scalability can be done quite easily with cloud or containers.

I am not sure what are the headers you are looking for, but tm has some lightweight parsing option for few headers, iirc, they are used to match quickly the replies, or maybe something else, need to look into the code for exact answer here ... anyhow, we can move them in the core if they are useful somewhere else.

ionutionita92:
@miconda i need the first line, to, from and callid. I'll take a look in tm for the lightweight parsing.
I also got one question: The case when you want to use the duplicate_uri(modparam) and not the destination uri(siptrace parameter) is not covered. I don't exactly know what's the best practice in kamailio to pass a null parameter to a function. I tried with "$null", I get a NULL pvar in code but I didn't see this implemented in other modules(or maybe I didn't look well enough).

miconda:
To get the value of the variable, being $null or something else, then you have to use some internal fixup function -- maybe ds_select_dst() from dispatcher has something like that, not 100% sure though.

Anyhow, the best in my option would be an empty string in this case. There are couple of other functions with the same behaviour (e.g., t_on_branch("") ...).

ionutionita92:
Everything is ready, I added trace flag back. Moreover I fixed a bug in https://github.com/kamailio/kamailio/pull/1912/commits/1cd0aaade45872a38f26b60e5af985006c26f9b1, https://github.com/kamailio/kamailio/pull/1912/commits/688b258ce7a3561725af7857aa973bf11fd6fba0 and https://github.com/kamailio/kamailio/pull/1912/commits/16f24e482c2584dac0b227c2ebbaddb4915ba777. What happened: if modparam duplicate uri was not defined but siptrace uri param was defined messages were not traced because of a bogus condition in siptrace_send.c. Also, in this scenari, the code(which is useless for that scenario) removed in https://github.com/kamailio/kamailio/pull/1912/commits/200acfa2a94ab77c5b44692cc2399fc71a001969 would have crashed if the condition was fixed.

src/modules/siptrace/doc/siptrace_admin.xml
```diff
@@ -18,24 +18,17 @@
 	<para>
 		The SIPtrace module offer a possibility to store incoming and outgoing SIP
 		messages in a database and/or duplicate to the capturing server (using <acronym>HEP</acronym>,
-		the Homer encapsulation protocol, or plain SIP mode)
+		the Homer encapsulation protocol, or plain SIP mode). Since version 5.4 new levels of tracing
+		are available. Transactions and dialogs can be traced. Trace flag is now useless.
```
henningw: "flag is useless" - probably should be changed after the re-introducing as well.

src/modules/siptrace/doc/siptrace_admin.xml
henningw: Daniel commented in the main comment thread about the sip parsing, that it is actually not that expensive anymore

src/modules/siptrace/siptrace.c
```diff
@@ -304,10 +338,30 @@ static int mod_init(void)
...
 	if(load_tm_api(&tmb) != 0) {
 		LM_WARN("can't load tm api. Will not install tm callbacks.\n");
-	} else if(tmb.register_tmcb(0, 0, TMCB_REQUEST_IN, trace_onreq_in, 0, 0)
-			  <= 0) {
-		LM_ERR("can't register trace_onreq_in\n");
-		return -1;
+	}
+
+	if (load_dlg_api(&dlgb) < 0) {
+		LM_WARN("can't load dlg api. Will not install dialog callbacks.\n");
```
henningw: Below you log on ERROR level if the dialog tracing is not available, makes probably sense to log here with ERROR as well.

src/modules/siptrace/siptrace.c
```diff
@@ -636,85 +690,93 @@ static int sip_trace_store_db(siptrace_data_t *sto)
...
 	}
 
-	*param = (void *)dst;
 	return 0;
 }
 
+
 /**
- * Send sip trace with destination and correlation id
+ * TODO TODO TODO:
```
henningw: Is this still a TODO or was already added?

src/modules/siptrace/siptrace.c
```diff
@@ -1332,16 +1533,226 @@ static void trace_onreply_out(struct cell *t, int type, struct tmcb_params *ps)
 	sto.stat = siptrace_rpl;
 #endif
 
-	sip_trace_store(&sto, NULL, NULL);
-	return;
+	if (info->uriState == STRACE_RAW_URI) {
+		LM_BUG("uriState must be either UNUSED or PARSED here! must be a bug! Message won't be traced!\n");
+		abort();
```
henningw: Calling abort() will kill Kamailio, it is not possible to recover here?

src/modules/siptrace/siptrace.c
```diff
@@ -636,85 +690,93 @@ static int sip_trace_store_db(siptrace_data_t *sto)
...
-	if(p) {
-		free_proxy(p); /* frees only p content, not p itself */
-		pkg_free(p);
+		trace_type = parse_siptrace_flag(&sflags);
+		if (trace_type == SIPTRACE_NONE) {
+			LM_ERR("Failed to parse trace type!\n");
+			return -1;
+		}
+
+		*param = pkg_malloc(sizeof(trace_type));
```
henningw: pkg_malloc can fail, check for success.

src/modules/siptrace/siptrace.c
```diff
@@ -636,85 +690,93 @@ static int sip_trace_store_db(siptrace_data_t *sto)
...
-		// Empty URI, use the URI set at module level (dup_uri)
-		if(dup_uri) {
-			uri = *dup_uri;
-		} else {
-			LM_ERR("Missing duplicate URI\n");
-			return -1;
-		}
-	} else {
-		duri = (char *)*param;
+	for (idx = 0; idx < sflags->len; idx++) {
```
henningw: This looks that there a a lot of whitespaces, can you check? Maybe github shows it different.

src/modules/siptrace/doc/siptrace_admin.xml
ionutionita92: thanks, fixed

src/modules/siptrace/siptrace.c
ionutionita92: I'm not totally sure I understand correctly. My intention was the module not to depend on either tm or dialog, that's why I put LM_WARN there. That's because it can and should be used in stateless mode without any issues. Is that what you're referring to?

src/modules/siptrace/siptrace.c
ionutionita92: it's github diff that's behaving odd.

src/modules/siptrace/siptrace.c
ionutionita92: Thanks for this one.

src/modules/siptrace/siptrace.c
ionutionita92: No, thanks.

src/modules/siptrace/siptrace.c
ionutionita92: This was for debugging purposes, forgot it there. Thanks alot for this one.

src/modules/siptrace/siptrace.c
henningw: It was only a minor comment. But having though about one more time its makes sense now with your comment.

henningw:
Thank you for the fixes. One last thing, in 2c5e63f you've added a README file, can you please remove it again. Then it can be merged from my side, eventual remaining issues can be fixed directly in git master.

ionutionita92:
@henningw sure. I wasn't aware kamailio builds the readme automatically.

henningw:
Thank you, merged.

PR #1914, opened 2019-03-28
Small fixes to dialplan module when an avp is used as subst_exp

marcocapetta:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

If an AVP is used as subst_exp then each contained value is evaluated against the string.
In case a value doesn't match, a noisy error line was printed in the log. The patch fix this behavior in the following way:
 - printing just a debug line per each not matching value
 - printing an error line only if all the values fail the match

Additionally fixed the improper usage of match_exp instead of subst_exp when both values contain an AVP.


miconda:
Thanks!

PR #1921, opened 2019-04-03
Add entities for from_tag and to_tag and bump the fields to 128

oyvindkolbu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

We have locally needed to update the `from_tag` and `to_tag` columns in `acc`, `dialog` and `siptrace` as we get requests which are larger than the defaults can handle.

While bumping the limit, I added two entities to `entities.xml` to get consistent sizes for `from_tag` and `to_tag` in the tables.


henningw:
Looks good to me. There is no maximum to/from tag length defined in RFC 3261, so longer tags can certainly happens.

henningw:
This is change that affects several modules. But It should have no database size implications as MySQL only stores in a VARCHAR the actual used characters (plus lengths) [link](https://dev.mysql.com/doc/refman/8.0/en/storage-requirements.html#data-types-storage-reqs-strings).
It also should not affect backward compatibility, Kamailio 5.1.x/5.2.x should run just fine with the new table structure and vice-versa.

henningw:
Thank you, merged

Issue #1942, opened 2019-05-02
git master siptrace regression in sip_trace function


henningw:
Tracking ticket for the discussion on sr-dev list "current git master branch issue with the siptrace module".

It seems that the changes from #1912 introduced a regression. The sip_trace() should work without duplicate uri because it can write directly to the database. It should be fixed to work without those parameters for module or function.

Error message from logs:
/usr/bin/sip-proxy[1659]: ERROR: siptrace [siptrace.c:908]: w_sip_trace3(): no duplicate_uri modparam nor duplicate uri sip_trace() argument provided!

ionutionita92:
sorry for this, will fix it.

miconda:
Closing, expected to be fixed after the last commits. If not, a new issue should be opened with the symptoms of the master branch.

PR #1951, opened 2019-05-13
siptrace: fix tracing messages to database when destination uri not set

ionutionita92:
New additions to siptrace module broke previous functionality.
If sip_trace was used with no param and destination uri was not set
then the current message would not have been traced to database. The
current fix traces always the current message if siptrace function
was called.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #1942

### DIFF
--- a/src/modules/siptrace/doc/siptrace_admin.xml
+++ b/src/modules/siptrace/doc/siptrace_admin.xml
@@ -556,6 +556,9 @@ sip_trace("sip:10.1.1.2:5085");
 ...
 sip_trace("sip:10.1.1.2:5085", "$ci-abc");
 ...
+/* trace current dialog; needs to be done on initial INVITE and dialog has to be loaded */
+sip_trace("sip:10.1.1.2:5085", "$ci-abc", "d");
+...
 </programlisting>
 		</example>
 	</section>
--- a/src/modules/siptrace/siptrace.c
+++ b/src/modules/siptrace/siptrace.c
@@ -1011,7 +1011,7 @@ static int w_sip_trace3(sip_msg_t *msg, char *dest, char *correlation_id, char *
 		}
 	}
 
-	if(trace_is_off(msg)) {
+	if(trace_type != SIPTRACE_MESSAGE && trace_is_off(msg)) {
 		LM_DBG("trace off...\n");
 		return 1;
 	}


ionutionita92:
I did the fix but I think the implementation as it was it's a little broken. That is because you can trace only to database using sip_trace with no arguments, but you can't trace only to database if using correlation_id(this was the case in previous versions). With the new additions you can't trace a dialog only to database. I respected previous implementation, that's why duplicate URI was mandatory, but this check is skipped if sip_trace is used with 0 or 1 argument, which I didn't notice.

kamailio-sync:
Ionut Ionita writes:

> I did the fix but I think the implementation as it was it's a little
> broken. That is because you can trace only to database using sip_trace
> with no arguments, but you can't trace only to database if using
> correlation_id(this was the case in previous versions). With the new
> additions you can't trace a dialog only to database. I respected
> previous implementation, that's why duplicate URI was mandatory, but
> this check is skipped if sip_trace is used with 0 or 1 argument, which
> I didn't notice.

I don't understand from the above, if after this fix the very basic
existing usage of siptrace module (if flag is set sip_trace(); adds the
trace to db) works or not.

I would feel comfortable, if you would leave the 5.2 code of the simple
usage untouched and write your extensions in new sections of the code
rather than making the simple usage a special case in a more complex
implementation.

-- Juha

ionutionita92:
Yes it works but the code in 5.2 is broken. Here's an example, maybe it's more clear:

sip_trace(); /* traces to database only no db url */
sip_trace("", "$var(correlation_id)"); /* does not work; requires valid duplicate uri */
// with my additions
sip_trace("", "$var(correlation_id)", "d"); /* does not work; requires valid duplicate uri */

So there are basically 2 functions:
- no arguments - trace to dabase only;
- with any kind of arguments - trace to duplicate uri and(if set) to database;

kamailio-sync:
Ionut Ionita writes:

> Yes it works but the code in 5.2 is broken. Here's an example, maybe
> it's more clear:
> 
> sip_trace(); /* traces to database only no db url */
> sip_trace("", "$var(correlation_id)"); /* does not work; requires valid duplicate uri */
> // with my additions
> sip_trace("", "$var(correlation_id)", "d"); /* does not work; requires valid duplicate uri */
> 
> So there are basically 2 functions:
> - no arguments - trace to dabase only;
> - with any kind of arguments - trace to duplicate uri and(if set) to
> database;

Then leave implementation of the working part (sip_trace();) as is and
add code to fix the broken part.

-- Juha

ionutionita92:
That's what this commit does, it fixes sip_trace(). Regarding the other part, it's not that obvious to me because if no one reported this maybe it's not an issue and that's the desired usage, elseway I would have fixed it already.

miconda:
I am merging, because I want to do some further adjustments.

PR #1957, opened 2019-05-18
Fixed compile warnings on Fedora 30

sergey-safarov:
Warning example
*** WARNING: ./usr/src/debug/kamailio-5.3.0-dev1.0.fc30.x86_64/src/modules/app_python3/apy_kemi.c is executable but has empty or no shebang, removing executable bit
*** WARNING: ./usr/src/debug/kamailio-5.3.0-dev1.0.fc30.x86_64/src/modules/ims_auth/pvt_message.c is executable but has empty or no shebang, removing executable bit
*** WARNING: ./usr/src/debug/kamailio-5.3.0-dev1.0.fc30.x86_64/src/modules/auth/api.c is executable but has empty or no shebang, removing executable bit


miconda:
Pushed the changes manually to format the commit message and reflect better the cause, not the effect.

PR #1962, opened 2019-05-22
core: Added methods for ip conversation.

alexyosifov:
- The new methods are used in usrloc pscscf module, but they are universal and can be used anywhere.

core: Added debug messages in grep_sock_info.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/core/resolve.c
+++ b/src/core/resolve.c
@@ -31,6 +31,7 @@
 #include <arpa/nameser.h>
 #include <resolv.h>
 #include <string.h>
+#include <arpa/inet.h>  /* for inet_pton() */
 
 #include "resolve.h"
 #include "compiler_opt.h"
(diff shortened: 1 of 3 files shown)


miconda:
There are already functions to parse a string to ip struct for ipv4 and ipv4, see str2ip() and str2ip6() in `src/core/resolve.h `. In that struct you have the `af` field. In most of the code, the two functions were used one after the other to parse ip addresses, if str2ip() fails, then try str2ip6().

We can add a new function str2ipx() as a wrapper to this logic, but otherwise, I see no reason to add the new functions proposed by your pull request.

Also, the one for trimming `[ ]` for ipv6 with copy over is not that useful in a global scope from my point of view. Moreover, there are no safety checks over the size of the target pointer, nor comments requesting a minimum value for that. This can be in the module that needs it.

I am closing this PR without merge. If people want to discuss more, then keep adding comments and if it is decided, it can be reopened.

PR #1977, opened 2019-06-10
kamctl: enable kamctl test from shell variable.

alishir:
- we can enable kamctl test by setting `KAMCTL_TEST=true kamctl`
  instead of setting TEST variable in kamctl script.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/utils/kamctl/kamctl
+++ b/utils/kamctl/kamctl
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # control tool for maintaining Kamailio
 #
--- a/utils/kamctl/kamctl
+++ b/utils/kamctl/kamctl
@@ -11,7 +11,12 @@ PATH=$PATH:/usr/local/sbin/
 
 # for testing only, please don't enable this in production environments
 # as this introduce security risks
-TEST="false"
+if [ -z $KAMCTL_TEST ]; then
+	TEST="false"
+else
+	TEST="true"
+fi
+
 
 ### include config files
 
--- a/utils/kamctl/kamdbctl
+++ b/utils/kamctl/kamdbctl
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # control tool for maintaining Kamailio databases
 #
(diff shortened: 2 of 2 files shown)


miconda:
@alishir - what were the errors related to commit 2c8fe55? Stretch is out for long time, nobody reported the errors so far.

alishir:
@miconda It causes `Substitution Error`. I search for that error and I found that the cause is that `sh` links to `dash` in Debian Stretch. I change the `/bin/sh` to `/bin/bash` and error resolved.

miconda:
Can you paste the the command you run and the entire error message printed in the terminal to try to track it properly?

alishir:
Unfortunately I have no access to that system now, I will revert the change and test again.

alishir:
@miconda 
Here is the output of `kamdbctl create`:
```bash
MySQL password for root:
-e \E[37;33mINFO: creating database kamailio_simple_db ...
-e \E[37;33mINFO: granting privileges to database kamailio_simple_db ...
-e \E[37;33mINFO: creating standard tables into kamailio_simple_db ...
-e \E[37;33mINFO: Core Kamailio tables succesfully created.
Install presence related tables? (y/n): n
/usr/sbin/kamdbctl: 216: /usr/sbin/kamdbctl: Bad substitution

```

henningw:
I fear that applying 2c8fe55 will break compatibility for embedded systems again (issue #1815).
Can you execute the kamdbctl create with 'bash -x' prefixed and see where exactly it stops? Maybe it is possible to just make this substitution to confirm to POSIX standard as well.

miconda:
It seems that the issue is expanding the variable when getting the answer for y/n question:
  * https://github.com/kamailio/kamailio/blob/master/utils/kamctl/kamdbctl.base#L216-L218

Such expression seems to be specific for bash:
  * https://mywiki.wooledge.org/Bashism

We should decide what to do to have an acceptable solution. Or maybe focus to make kamcli a (full) replacement for kamctl/kamdbctl and get rid of those old-style shell/bash scripts.

miconda:
I applied a slightly different patch for enabling test mode from command line, enclosing the variable in double quotes - 4968e8994157f1152bf7461b16049d67e5c9cc59 .

For the bash related problems, I will open a dedicated issue on tracker to decide on a proper solution.

Issue #1911, opened 2019-03-27
rtpengine: hashing algorithm not offering a good enough distribution


ionutionita92:
We are having problems in our environment with the way rtpengine distributes RTP sessions. It doesn't provide a good enough distribution among RTP nodes and this is mostly because the current algorithm is quite simple, doing the sum of the characters in the callid and after that applying a 0xFF mask over this(which contributes even more to the distribution of this algorithm).  This may affect our systems under heavy load and we've been looking for an alternative.

To have something to compare I have designed a tesing application in which I generate CallIds, apply a hashing algorithm and assign the result to a set/node. For example I have 10 nodes each of weight 10. Let's say hash value is 125 I do 125 mod (10 * 10) = 25 therefore the hash will go to node 3.
I've been running some tests over various hashing algorithms such as: SHA 256, jenkins(simple hashing algorithm found on wikipedia), md5, sha1, ripemd, crc32 and the one used in the rtpengine. For each algorithm I test from 1000 randomly generated callids to 10 million with a multiplyer of 10(5 tests for each algorithm). The callIds have 16 randomly generated bytes and 16 fixed bytes which are always the same. This way I'm trying to reproduce the way callIds are generated most of them containing a randomly generated part and a fixed part. 

From the results I can tell for sure that multiple rounded hashing algorithms from libssl(md5, sha1, ripemd, sha 2) offer a much better distribution than a trivial hashing function so we're looking to change the current algorithm with SHA 1 because SHA 2 takes more time without significant improvement. 

It would be nice to push this upstream, but there are some issues. First, using these functions, rtpengine module will depend on libssl. What is your opinion about this? I've looked if we can copy any of the algos from libssl into the code but this is not as easy as it seems. Secondly, if you agree to replacing the algorithm, how would you approach the issue: only replace the current one or add the posibility for the user to select between multiple algorithms?

Any suggestion of testing scenarios or hashing algorithms that we've not considered is much welcome. 

The results of the tests are attached to his post.  
[h_funcs_compare.txt](https://github.com/kamailio/kamailio/files/3014285/h_funcs_compare.txt)

miconda:
If you want to use crypto hashing from libcrypto, the best way to do it is to add what you need to crypto module and then have an option via modparam in rtpengine module to bind to crypto module and use that function. Otherwise, it should be using the internal hashing function (can be changed from what is doing now with another hash function from those available in the core).

[...]

PR #1986, opened 2019-06-19
Extend rtpengine hashing algorithms

ionutionita92:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #1911 

#### Description

The goal was to have a different, more complex hashing algorithm that would offer a better distribution among nodes. We've seen in production that the current algorithm does not perform well under heavy load therefore we decided to add SHA1 hashing. After running performance tests we've decided that we need a cryptographic function in order to have the best results mostly when the callids have small entropy. 
In order to add SHA1, a new API was added in crypto module exporting a function applying SHA1 to a string. The API was used in rtpengine module over the callid. After this operation the hash was calculated as before except for applying 0xFF mask over the hash as it was done before, which restrains the power of this algorithm.

### DIFF
--- a/src/modules/crypto/api.c
+++ b/src/modules/crypto/api.c
@@ -1,0 +1,30 @@
+/*
+ * Copyright (C) 2019 1&1 Internet AG
+ *
+ * This file is part of Kamailio, a free SIP server.
+ *
+ * Kamailio is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version
+ *
+ * Kamailio is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


miconda:
From crypto module, I am ok to merge.

For rtpengine, maybe @rfuchs can review and decide the next steps.

Somehow related, just making a note here for the future: the src/lib/srutils/shautils.{c,h} have dome of the shaXYZ functions, but not sha1 -- maybe we should add it and eventually move the files in the core, as they become more commonly used.

rfuchs:
@miconda looks good to me, no objections

smititelu:
@miconda is it ok for me to backport this to 4.4 branch? Or shall be considered as new feature (starting from 5.3 branch)?

henningw:
You mentioned above in the patch description that it is a new feature. ;-) So it should be not backported, 4.4. is also EOL since 2018.

smititelu:
ok, thanks

PR #2008, opened 2019-07-15
Microsleep

urtho:
This is a ringless implementation of millisecond precision timer and millisecond async_sleep function.
I've added the functionality to the current async mod but this can me moved to a dedicated module.

For now only the async_ms_sleep is implemented but async_ms_route is just a few lines. 

Anyway how should I proceed with cleanup so this can be accepted ?

With millisecond precision sleep one can easily implement a leaky bucket request shaper that delays incoming requests to achieve perfect rate limits per customer/sip-trunk/whatever without dropping requests. No more spikes and no more 503s.

I've implemented external HTTP microservice that returns millisecond delays per dialog initiating request and tested it with SIPP with 3 flows/streams each with 250 CAPS.  No leaks nor other issues in 24hrs of testing.

### DIFF
--- a/src/modules/async/README
+++ b/src/modules/async/README
@@ -10,7 +10,7 @@ Daniel-Constantin Mierla
 
    <miconda@gmail.com>
 
-   Copyright © 2011-2016 asipto.com
+   Copyright (c) 2011-2016 asipto.com
      __________________________________________________________________
 
    Table of Contents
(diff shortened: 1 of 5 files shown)


miconda:
Hello,
thanks for contributing! To merge the code you should combine the patches in a single commit and then format the commit message as per contributing guidelines.

To merge the patches you can get the diff of the pull request:
  * https://patch-diff.githubusercontent.com/raw/kamailio/kamailio/pull/2008.diff

And apply it to a fresh clone of kamailio master. Then write the commit message based on:

Add also the docs for the new function in the admin guide located in doc/ subfolder of the module.

Given that async sleep is a bit sensitive in usage, add also the async_ms_route() that should be easy as you already said.

PR #2013, opened 2019-07-19
evapi:return code and log fixed for no evapi client while sending

surendratiwari3:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

Right now, if there is no evapi client to send message then no logs are getting printed. This make us to not set the alert if send failed. After this changes if there is no evapi client we are going to get that in line like send_failed.

### DIFF
--- a/src/modules/evapi/evapi_dispatch.c
+++ b/src/modules/evapi/evapi_dispatch.c
@@ -728,6 +728,7 @@ int _evapi_relay(str *evdata, str *ctag, int unicast)
 
 	int len;
 	int sbsize;
+	int evapi_send_count;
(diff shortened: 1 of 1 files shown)


miconda:
I will merge it, but then I will do some further updates:

  * print the log message as a warning, not error -- I think it is more appropriate, not being an error of processing. If you want, you can follow up with a new PR to add a modparam to set the log level for this message
  * edit a bit the message to reflect better the situation (of course, as I see it)

Thanks!

surendratiwari3:
@miconda agree.

PR #2022, opened 2019-08-01
core: Fix the cfg framework variable input type default

mancasg:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

What the affected line of code really wants to achieve is: if an
accepted inputs type mask is not specified then default to accepting
only the actual variable type. So we must mask the var type first,
then shift it by `CFG_INPUT_SHIFT`, before or-ing it with the rest.

What happened before was that the entire type was shifted, and that
included var type, input type and flags. What we end up with is some
additional higher bits set (for flags). I actually discovered this while
adding an additional flag that was meant to mark variables as private,
only accessible through an internal API (not available to modules such
as cfg_rpc).

### DIFF
--- a/src/core/cfg/cfg.c
+++ b/src/core/cfg/cfg.c
@@ -94,7 +94,7 @@ int cfg_declare(char *group_name, cfg_def_t *def, void *values, int def_size,
 
 		/* verify the type of the input */
 		if (CFG_INPUT_MASK(def[i].type)==0) {
-			def[i].type |= def[i].type << CFG_INPUT_SHIFT;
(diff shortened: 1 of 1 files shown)


henningw:
Hello, thank you for the pull request. Can you give a quick example what does not work in the cfg without the fix? Just to be able to test it as well.

mancasg:
Because of a coincidence(having 3 type flags) this doesn't affect kamailio at all. It will, however, if you add additional flags, as you can see from the description above. It's a bug that didn't have the opportunity to manifest itself but nonetheless a bug.

miconda:
Thanks! I will merge it.

PR #2040, opened 2019-08-23
modules/rtpengine: set_rtpengine_set fix

juha-h:
WIthout this fix, set_rtpengine_set(set1, set2) sets set1 twice and set2 not at all.


miconda:
Thanks! Maybe @rfuchs wants to do a review before merging.

juha-h:

> Thanks! Maybe @rfuchs wants to do a review before merging.

Sure, lets wait his review.  Looks like no-one has ever used that
feature before.

juha-h:
Thanks for the review.

PR #2045, opened 2019-08-26
nathelper: test sdp c line and sip source ip address is matched.

ycaner06:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

test sdp c line and sip source ip address is matched.

### DIFF
--- a/src/modules/nathelper/doc/nathelper_admin.xml
+++ b/src/modules/nathelper/doc/nathelper_admin.xml
@@ -49,7 +49,7 @@
 	<itemizedlist>
 		<listitem>
 			<para>
-			<emphasis>UDP packet</emphasis> - 4 bytes (zero filled) UDP 
+			<emphasis>UDP packet</emphasis> - 4 bytes (zero filled) UDP
 			packets are sent to the contact address.
 			</para>
 			<itemizedlist>
@@ -59,8 +59,8 @@
 				</para>
 				</listitem>
 				<listitem>
-				<para><emphasis>Disadvantages:</emphasis> unidirectional 
-				traffic through NAT (inbound - from outside to inside); As 
+				<para><emphasis>Disadvantages:</emphasis> unidirectional
+				traffic through NAT (inbound - from outside to inside); As
 				many NATs do update the bind timeout only on outbound traffic,
 				the bind may expire and closed.
 				</para>
@@ -69,19 +69,19 @@
 		</listitem>
 		<listitem>
 			<para>
-			<emphasis>SIP request</emphasis> - a stateless SIP request is 
+			<emphasis>SIP request</emphasis> - a stateless SIP request is
 			sent to the UDP contact address.
 			</para>
 			<itemizedlist>
 				<listitem>
 				<para><emphasis>Advantages:</emphasis> bidirectional traffic
-				through NAT, since each PING request from &kamailio; (inbound 
-				traffic) will force the SIP client to generate a SIP reply 
+				through NAT, since each PING request from &kamailio; (inbound
+				traffic) will force the SIP client to generate a SIP reply
 				(outbound traffic) - the NAT bind will be surely kept open.
 				</para>
 				</listitem>
 				<listitem>
-				<para><emphasis>Disadvantages:</emphasis> higher bandwidth 
+				<para><emphasis>Disadvantages:</emphasis> higher bandwidth
 				traffic, more expensive (as time) to generate by &kamailio;;
 				</para>
 				</listitem>
@@ -99,7 +99,7 @@
 			<itemizedlist>
 			<listitem>
 			<para>
-				<emphasis>usrloc</emphasis> module - only if the NATed 
+				<emphasis>usrloc</emphasis> module - only if the NATed
 				contacts are to be pinged.
 			</para>
 			</listitem>
@@ -109,7 +109,7 @@
 	<section>
 		<title>External Libraries or Applications</title>
 		<para>
-		The following libraries or applications must be installed before 
+		The following libraries or applications must be installed before
 		running &kamailio; with this module loaded:
 			<itemizedlist>
 			<listitem>
@@ -147,8 +147,8 @@ modparam("nathelper", "force_socket", "127.0.0.1:5060")
 	<section id="nathelper.p.natping_interval">
 		<title><varname>natping_interval</varname> (integer)</title>
 		<para>
-		Period of time in seconds between sending the NAT pings to all 
-		currently registered &ua;s to keep their &nat; bindings alive. 
+		Period of time in seconds between sending the NAT pings to all
+		currently registered &ua;s to keep their &nat; bindings alive.
 		Value of 0 disables this functionality.
 		</para>
 		<note><para>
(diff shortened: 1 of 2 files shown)


henningw:
Thanks Yasin, looks good to me.

miconda:
I see that the comparison of the IPs is done with string operations. That's ok for ipv4, but can be a mismatch with ipv6 because there can be different forms of the same address.

I suggest to convert the the ip in sdp to binary struct and compare with the src ip which is in binary struct -- see str2ipxbuf() or str2ipx() and ip_addr_cmp() functions/macros.

ycaner06:
@miconda 

I tried to convert string to ip_addr struct as you mentioned but  , functions cannot convert it. In addition , when i try to log "sdp ip address " it gives empty. I didnt understand reason.


converting with str2ip gives NULL.
```
struct ip_addr* cline_addr;

if(session->pf==AF_INET6){
	cline_addr=str2ip(&session->ip_addr);
}else if(session->pf==AF_INET){
	cline_addr=str2ip6(&session->ip_addr);
}else{
	LM_ERR("Couldn t get sdp address type \n");
	break;
}

if(!cline_addr){
	LM_ERR("Sdp ip address is empty  \n");
	break;
}

```

miconda:
Have you tested that session->ip_addr.len>0? Can you print the str value for session->ip_addr and see if it is a valid IP?

Then it would be better to use str2ipx(), because it tries internally both ipv4/6.

ycaner06:
@miconda 
I dont understand reason but it gives empty line.
loggin:
```
LM_ERR("SDP CLINE2 [%*.s][LEN: %d] \n",session->ip_addr.len,&session->ip_addr.s[0],session->ip_addr.len);
LM_ERR("SDP CLINE2 [%s] \n",&session->ip_addr.s[0]);
```

here is result

```
9(4455) ERROR: {1 1 INVITE 1-4478@192.168.1.39} nathelper [nathelper.c:1370]: test_sdp_cline(): SDP CLINE2 [            ][LEN: 12]
 9(4455) ERROR: {1 1 INVITE 1-4478@192.168.1.39} nathelper [nathelper.c:1371]: test_sdp_cline(): SDP CLINE2 [192.168.1.39
t=0 0
m=audio 6000 RTP/AVP 8
a=rtpmap:8 PCMA/8000
]

```

miconda:
It has to be `%.*s`, you used `*` before the `.` -- also you do not need that expression with address of the first char, just give `session->ip_addr.s` instead of `&session->ip_addr.s[0]`, it should avoid array access and get address of it -- probably there is not much difference in performance, but I find it easier to follow up.

miconda:
First a note for who is going to merge: the last commit doesn't have proper formatted commit message. So this PR should be squashed and merged or merged manually.

Then, I need a clarification: when this test is true: when all IPs in SDP match the source IP? If my logic is not missing something, the ip_addr_cmp() returns true if the two ips are the same, but you return -1 in that case.

Also, the result variable is set to 1 and used only for the return at the end. If you need it only for that, then remove it and just do return 1 at the end.

ycaner06:
Hello,

I used str2ipxbuf() instead of stripx(). 

Tested

- IPv6 short and long versions
- IPv4 versions.

```
ip_addr_t* str2ip6(str* st)
{
	static ip_addr_t ip;

	if(str2ip6buf(st, &ip)<0) {
		return NULL;
	}

	return &ip;
}
```

ycaner06:
> First a note for who is going to merge: the last commit doesn't have proper formatted commit message. So this PR should be squashed and merged or merged manually.
> 
> Then, I need a clarification: when this test is true: when all IPs in SDP match the source IP? If my logic is not missing something, the ip_addr_cmp() returns true if the two ips are the same, but you return -1 in that case.

yes ofc. If Connection Line IP address in SDP is matched  with SIP source address, there isnt any problem about NAT. 
For example;

UE has 192.168.1.18
A Router that  has ability Source NAT . so translate SIP Source IP address to 192.168.16.1

UE SIP request address is 192.168.16.1, Connection Lİne 192.168.1.18  so RTP flows cannot reach to UE. For solution , we add to rtpengine_manage function "SIP Source Address" flag.

256 flag helps to understand this scenario.

I will add more information to document.

miconda:
OK, so the title of PR is misleading, the test is true if there is mismatch, not when IPs are matched.

miconda:
As I looked at the entire diff -- do not change the formatting of existing code/docs files when adding a new features. Now the diff has a lot of changes in the xml doc files making hard to spot what was changed. Patches to fix existing indentation and whitespaces are more than welcome, but must be dedicated commits/pull requests. It can stay here, but have in mind for the future.

I think further clarifications are needed, because there can be many c= lines in SDP and you return -1 with the first one when IPs matches, but maybe before or after that c= line it is a mismatch (e.g., audio session matches, video session mismatches). How should this be considered?

You can leave it like now, just explain in the docs.

Or, an idea, you can add one (or two) variable(s) to count the c= lines that match (and those that mismatch). Then at the end of the function you test if matches is same value as the number of sessions and return different value if all matched, or there was some partial matches, or no matches.

Also, I think there should be like return -2 (or other negative value) if SDP fails to be parsed, to be able to differentiate between error and match.

ycaner06:
> As I looked at the entire diff -- do not change the formatting of existing code/docs files when adding a new features. Now the diff has a lot of changes in the xml doc files making hard to spot what was changed. Patches to fix existing indentation and whitespaces are more than welcome, but must be dedicated commits/pull requests. It can stay here, but have in mind for the future.

i dont change any indentation , vim can do it automatically. Sorry for it.
> I think further clarifications are needed, because there can be many c= lines in SDP and you return -1 with the first one when IPs matches, but maybe before or after that c= line it is a mismatch (e.g., audio session matches, video session mismatches). How should this be considered?

Well , i never see more c line so i didnt think about it. I will change it  as you mentioned.

> You can leave it like now, just explain in the docs.
> 
> Or, an idea, you can add one (or two) variable(s) to count the c= lines that match (and those that mismatch). Then at the end of the function you test if matches is same value as the number of sessions and return different value if all matched, or there was some partial matches, or no matches.
> 
> Also, I think there should be like return -2 (or other negative value) if SDP fails to be parsed, to be able to differentiate between error and match.

I will add more option but I hope it dont change other test results. For example , if there is a problem parsing sdp return 2 from nat_uac_test() function.

henningw:
Hi Yasin, any update about the patch?
if you don't change the code regarding the multiple c lines - some extensions in the docs would be necessary. A second point would be to add a return -2 if the SDP can't be parsed.

Let me know, then I can do a manual merge e.g. on Tuesday evening in time for the freeze.

ycaner06:
Hello,
I couldnt test multiple connection address but i am sure it works . I think parse_sdp cannot parse it so it always give me 1 address.
function on error returns -2.
if all address are matched , returns -1 instead of 0 (zero). Because if returns 0 , script exits.(I dont know why)

added more information to documentation.

henningw:
I merged your pull request, with some adaptions. I removed the white-space changes from the docs.

The nat_uac_test function return 1 when a NAT check matches for all existing checks and -1 otherwise. So i also adapted the check to work like this. A second reason for the change was that the way it was implemented  would stop working if another new flag would be added later on. I also adapted the c function return codes to the other existing code. 

The additionally discussed functionality to also return the number of matches lines could be easily exported as a dedicated function. This could be done with a small wrapper like the existing function is_rfc1918.

PR #2058, opened 2019-09-08
dialog: check xdata length when loading profile from db

grumvalski:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

When using a backend not supporting "NULL", like Redis, which store it as an empty string, xdata.s can be not null (it contains '\0'). This leads to an error trying to read it as a json document.
Not really a bug, but it leads to the error
"invalid json doc [[]]"
being printed each time a dialog is loaded from db (Redis).


miconda:
Thanks! It can be merged and backported if needed.

Issue #2060, opened 2019-09-08
db can't parse ipv6 address from config


vasilevalex:
### Description

If in kamailio.cfg I write IPv6 address for Database URL like this:
```
#!ifndef DBURL
#!define DBURL "mysql://kamailio:kamailiorw@[2001:5c8:4e00:11:a213:ba2d:579a:3e3d]/kamailio"
#!endif
```
Then kamailio can't connect to database.

#### Reproduction

Just try to use database server with IPv6-address in kamailio.cfg 

#### Log Messages

```
Sep  8 14:35:40 c7-01 /sbin/kamailio[10680]: ERROR: db_mysql [km_my_con.c:139]: db_mysql_new_connection(): driver error: Unknown MySQL server host '[2001' (2)
Sep  8 14:35:40 c7-01 /sbin/kamailio[10680]: ERROR: <core> [db.c:319]: db_do_init2(): could not add connection to the pool
Sep  8 14:35:40 c7-01 /sbin/kamailio[10680]: ERROR: usrloc [dlist.c:646]: register_udomain(): failed to open database connection
```

### Possible Solutions

Correct parsing of config file.

### Additional Information

```
version: kamailio 5.3.0-dev7 (x86_64/linux) 
flags: USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLACKLIST, HAVE_RESOLV_RES
ADAPTIVE_WAIT_LOOPS 1024, MAX_RECV_BUFFER_SIZE 262144, MAX_URI_SIZE 1024, BUF_SIZE 65535, DEFAULT PKG_SIZE 8MB
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
```

* **Operating System**:

CentOS 7

```
Linux sip.aaa.de 3.10.0-957.27.2.el7.x86_64 #1 SMP Mon Jul 29 17:46:05 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

henningw:
[...]

PR #2061, opened 2019-09-08
Db ipv6-address parsing in config file

vasilevalex:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #2060 

#### Description
Added parsing of IPv6 address in config file for DB URL.

### DIFF
--- a/src/lib/srdb1/db_id.c
+++ b/src/lib/srdb1/db_id.c
@@ -102,12 +102,13 @@ static int parse_db_url(struct db_id* id, const str* url)
 		ST_USER_HOST,  /* Username or hostname */
 		ST_PASS_PORT,  /* Password or port part */
 		ST_HOST,       /* Hostname part */
+		ST_HOST6,      /* Hostname part IPv6 */
 		ST_PORT,       /* Port part */
 		ST_DB          /* Database part */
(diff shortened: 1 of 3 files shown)


henningw:
Hello, thank you for the pull request. This is indeed a good addition. 
I had a look to the code - I think it is good to combine this with a small refactoring in the modules as well.

As git master branch is currently frozen, I will put this on hold until 5.3 is released in a few weeks.

miconda:
Not supporting IPv6 is a bug, Kamailio is supposed to work on IPv6 for its common use cases. Not sure why support in db_url was not discovered so far missing IPv6, we did a review couple of years ago, but we focused on SIP routing, this was somehow overlooked.

If the review of the code is ok, then this PR should be merged, unless many devs have something agains.

vasilevalex:
There is some code duplication, and I don't like it. But I don't know, may be it make sense, so I left it as it was.

henningw:
Yes, we should support IPv6. But there are probably still a few places left like this. In the end most people will use DNS for internal names.
I will merge it and do the refactoring then later.

PR #2066, opened 2019-09-14
dialog: run DLGCB_LOADED callbacks when loading dialog from db

grumvalski:
When a specific dialog is loaded from db via dlg_db_load_callid, the load callbacks are not executed so, for example, a CDR is not generated.
No sure if this should considered a new feature or a bug (and so eventually be backported).

### DIFF
--- a/src/modules/dialog/dialog.c
+++ b/src/modules/dialog/dialog.c
@@ -714,7 +714,6 @@ static int mod_init(void)
 		}
 	}
 
-	destroy_dlg_callbacks( DLGCB_LOADED );
 
 	/* timer process to send keep alive requests */
 	if(dlg_ka_timer>0 && dlg_ka_interval>0)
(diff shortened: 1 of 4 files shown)


henningw:
Thanks, it looks good to me. This should fix also the same problem for the dlg_db_load_extra() cfg function. One small thing - would be great if you fix the different code indention in the new function in file dlg_cb.c.

We integrated a similar fix for the uac module some time ago (aa7dee1ab874) also the stable branches. So if no other developers object in the next few days it can be merged and backported in my opinion.

grumvalski:
Thans for looking Henning, I fixed the indentation (automatic editor settings) and push a new commit. I can rebase/squash before merging.
Yes, the fix should also work for dlg_db_load_extra (mode 2 for the function).

miconda:
From my point of view it is ok to merge and backport if needed.

grumvalski:
Thanks. I'm going to merge and backport to 5.2 (where the new functions have been introduced).

Issue #2067, opened 2019-09-18
xavp_serialize_fields : adding an equal sign on empty strings


jchavanton:
I noticed the problem when trying ot use :
xavp_params_implode followed by xavp_params_explode no luck with that because
xavp_params_explode will fail to parse emtpy params.

Example:
`$xavp(x[0]=>y) = "";`

Will fail to parse because `y=;` is consider invalid by the param parser.

Glad to see that the core was exposing a reusable serialize function !

PR #2068, opened 2019-09-18
core: fix xavp_serialize_fields

jchavanton:
stop adding plus signs when serializing empty strings

github issue #2067


miconda:
Thanks!

PR #2099, opened 2019-10-10
dispatcher: ds_list_exist support pv as param (only logging level adjustment was kept)

jchavanton:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Change parameter one to support pseudo variables
```
if (ds_list_exist("12001") {
...
}
```
or
``` 
$var(id) = 10;
if (ds_list_exist($var(id)) {
...
}
```

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -1488,10 +1488,10 @@ int ds_list_exist(int set)
 	si = ds_avl_find(_ds_list, set);
 
 	if(si == NULL) {
-		LM_INFO("destination set [%d] not found\n", set);
+		LM_DBG("destination set [%d] not found\n", set);
 		return -1; /* False */
 	}
-	LM_INFO("destination set [%d] found\n", set);
+	LM_DBG("destination set [%d] found\n", set);
 	return 1; /* True */
 }
 
--- a/src/modules/dispatcher/dispatcher.c
+++ b/src/modules/dispatcher/dispatcher.c
@@ -1039,7 +1039,6 @@ static int ki_ds_list_exists(struct sip_msg *msg, int set)
 static int fixup_ds_list_exist(void **param, int param_no)
 {
 	return fixup_igp_null(param, param_no);
-	return 0;
 }
 
 static int ds_parse_reply_codes()


lazedo:
@jchavanton this is ok but you could already do 
```
if (ds_list_exist("$var(id)") {
```

miconda:
The current approach is Kamailio specific and I think it is fine to keep it, because it already allows giving the variable as parameter. The fixup proposed to be used in this patch is inherited from SER side.

jchavanton:
Thanks for the feedback,

Weird, I decided to change it because it was not accepting the was (complaining at startup).
I will dig it further.

We could add deprecated comments here and there to point to the favored alternatives.

jchavanton:
```
$var(list) = 10;
if (ds_list_exist($var(list))) {
...
```

```
 0(8) CRITICAL: <core> [core/cfg.y:3507]: yyerror_at(): parse error in config file /etc/kamailio/kamailio.cfg, line 200, column 21-44: function ds_list_exist: parameter 1 is not constant
```

jchavanton:
I see this is working :
```
$var(list) = 10;
if (ds_list_exist("$var(list)")) {
...
```

I guess I have what I need ...

jchavanton:
I looked again at the macros, I will always search in `core/mod_fix.h`
```
FIXUP_F2FP(igp_null      , 1   , 1   ,   1, FPARAM_INT|FPARAM_PVS, 0)
#define FIXUP_F2FP(suffix, minp, maxp, no1, type1, type2) \
```

jchavanton:
I converted the PR to a small logging level adjustments.

lazedo:
the `INFO` level for this message is actually misleading ops that get into panic saying `something's not working` when is perfectly normal not to have entries in some group. just merge it :)

miconda:
I consider it should be merged as well: there is no relevant need to have a log message other than debug to confirm one or the other goal of the function -- in this case the function is to detect if a destinations group exists or not, by returning appropriate code in each case, so custom log messages can be printed in config routing blocks.

Anyhow, to differentiate between error and not found, I will change a bit the return codes. So, I am going to close this one and push a slightly different commit soon.

PR #2106, opened 2019-10-21
lib: big integers should not be treated as negative ones

ivanmurashko:
0xffffffff00000000ULL mask does not take into consideration signed 32 bit
integers and as result 2147483648 will be stored as -2147483648.
Another mask should be used to correct the issue: 0xffffffff80000000ULL.
It processes correctly the case of integers that are greater than 2147483647.

There are several gdb tests below:
(gdb) p ((unsigned long long)2147483648 & 0xffffffff80000000ULL)
$1= 2147483648
(gdb) p ((unsigned long long)2147483647 & 0xffffffff80000000ULL)
$2 = 0
(gdb) p ((unsigned long long)2147483646 & 0xffffffff80000000ULL)
$3 = 0
(gdb) p ((unsigned long long)4147483646 & 0xffffffff80000000ULL)
$4 = 2147483648

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

### DIFF
--- a/src/lib/srdb1/db_ut.c
+++ b/src/lib/srdb1/db_ut.c
@@ -598,7 +598,7 @@ int db_val2pv_spec(struct sip_msg* msg, db_val_t *dbval, pv_spec_t *pvs)
 				db_longlong2str(dbval->val.ll_val, ll_buf, &pv.rs.len);
 				pv.rs.s = ll_buf;
 				/* if it fits, also store as 32 bit integer*/
-				if (! ((unsigned long long)dbval->val.ll_val & 0xffffffff00000000ULL)) {
+				if (! ((unsigned long long)dbval->val.ll_val & 0xffffffff80000000ULL)) {
 					pv.flags |= PV_VAL_INT | PV_TYPE_INT;
 					pv.ri = (int)dbval->val.ll_val;
 				}


miconda:
I am not sure about the right way to deal in this case. Because also the very high unsigned int values will be negative in a pv value (which is signed int). Maybe adding a global parameter to control this is better than just changing the existing behaviour. The value of a pv can be retrieved in an embedded script (e.g., Lua, JS) and casted to unsigned int, if one knows that should be that kind of value. By the new mask you introduce, practically the unsigned int values are truncated to half of the range.

ivanmurashko:
You are right in the observation. The fix is not ideal and the string representation will be used for integers that are greater than 2147483647. From other side the string representation will be definitely  used for any ints that are greater than 4294967295.  Thus the fix can lead to a performance degradation for values between  2147483647 and 4294967296 if the integer is **explicitly** interpreted as unsigned. That is a rare case. Please correct me if I am wrong there. 

From another side the current behaviour is completely wrong if the result is not **explicitly** interpreted as unsigned. For instance 2 different integers 2626734008 and -1668233288 will be stored as -1668233288 and the negative value (-1668233288) will be used as pv. I.e. any value in the range (2147483647, 4294967296) will be stored and interpreted as a negative integer. This leads to a strange service behaviour in my case.

henningw:
@miconda There have been several changes to core and also PVs (like AVP, XAVPs etc..) to use now internally long instead of int. Do you think it would make sense now to address this DB library topic?

PR #2110, opened 2019-10-24
Switch from deprecated /var/run to /run

guillemj:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [ ] Tested changes locally

#### Description

The FHS 3.0 ([/run](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html) and [/var/run](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s13.html)) has deprecated /var/run in favor of /run. Current distributions have done so for a long time (for example in Debian since 6.0 “squeeze”), even though they provide a compatibility symlink from /var/run to /run. But software like systemd have started emitting warnings when using /var/run, for example for its PIDFile directive, which pollute the logs.

I've included a change to a man page, because even though it looks like a similar case to the READMEs, I didn't find any part of the build system generating 8 section man pages. And I've not split this into components, as it is a global change, and seems preferable to have an atomic and bisectable commit.


miconda:
All now in master branch.

PR #2114, opened 2019-10-30
ims_ipsec_pcscf: closing mnl_socket in some exceptional situations

dolk13:
- fix leak socket when error handling errors in creation or destruction ipsec tunnel

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Fixed socket leak on error during ipsec tunnel creation or destruction.

### DIFF
--- a/src/modules/ims_ipsec_pcscf/cmd.c
+++ b/src/modules/ims_ipsec_pcscf/cmd.c
@@ -338,15 +338,18 @@ static int create_ipsec_tunnel(const struct ip_addr *remote_addr, ipsec_t* s)
     if(remote_addr->af == AF_INET){
         if(str2ipbuf(&ipsec_listen_addr, &ipsec_addr) < 0){
             LM_ERR("Unable to convert ipsec addr4 [%.*s]\n", ipsec_listen_addr.len, ipsec_listen_addr.s);
+            close_mnl_socket(sock);
             return 0;
         }
     } else if(remote_addr->af == AF_INET6){
         if(str2ip6buf(&ipsec_listen_addr6, &ipsec_addr) < 0){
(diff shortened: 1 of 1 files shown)


henningw:
I looked into the existing code of the function create_ipsec_tunnel(..)

Are you sure that the "return 0" in the error cases is correct? Should it not be -1, as in the one case in the bottom of the function? If yes - then it makes probably sense to combine them all into one "goto error" case.

alexyosifov:
> Thank you for the pull request. I looked into the existing code of the function create_ipsec_tunnel(..)
> 
> Are you sure that the "return 0" in the error cases is correct? Should it not be -1, as in the one case in the bottom of the function? If yes - then it makes probably sense to combine them all into one "goto error" case.

Hi,
I am pretty sure that all return codes in error cases should be changed to -1. I am working on this module actively and return code 0 in these error cases will cause problems for sure.

miconda:
@henningw, @alexyosifov - looking at the patch, the pull request here is not adding any `return 0`, they existed. The patch is adding a couple of `close_mnl_socket(sock)`. If the existing `return 0` in error cases are wrong, then they need to be fixed of course. But it is not a strict relation with this PR.

alexyosifov:
> @henningw, @alexyosifov - looking at the patch, the pull request here is not adding any `return 0`, they existed. The patch is adding a couple of `close_mnl_socket(sock)`.
> 
> If the existing `return 0` in error cases are wrong, then they need to be fixed of course. But it is not a strict relation with this PR.

Hi,
I'll fix the problem with the error codes in the next ipsec pull request from my side.

henningw:
@miconda - I know. :-) But as @alexyosifov was working on the code right now, I wanted to ask him about his opinion.

henningw:
Merged, will also backport to stable branches.

PR #2124, opened 2019-11-08
nathelper : added new function set_alias_to_pv

ycaner06:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

new function that read alias from Contact header then write to given avp as sip uri

### DIFF
--- a/src/core/parser/parse_uri.c
+++ b/src/core/parser/parse_uri.c
@@ -1439,26 +1439,3 @@ void proto_type_to_str(unsigned short type, str *s) {
 		*s = s_null;
 	}
 }
-
-void proto_type_int_to_str(int type, str *s) {
-       switch (type) {
-       case PROTO_UDP:
-               *s = s_udp;
-               break;
-       case PROTO_TCP:
-               *s = s_tcp;
-               break;
-       case PROTO_TLS:
-               *s = s_tls;
-               break;
-       case PROTO_SCTP:
-               *s = s_sctp;
-               break;
-       case PROTO_WS:
-       case PROTO_WSS:
-               *s = s_ws;
-               break;
-       default:
-               *s = s_null;
-       }
-}
--- a/src/core/parser/parse_uri.c
+++ b/src/core/parser/parse_uri.c
@@ -1439,3 +1439,26 @@ void proto_type_to_str(unsigned short type, str *s) {
 		*s = s_null;
 	}
 }
+
+void proto_type_int_to_str(int type, str *s) {
+       switch (type) {
+       case PROTO_UDP:
+               *s = s_udp;
+               break;
+       case PROTO_TCP:
+               *s = s_tcp;
+               break;
+       case PROTO_TLS:
+               *s = s_tls;
+               break;
+       case PROTO_SCTP:
+               *s = s_sctp;
+               break;
+       case PROTO_WS:
+       case PROTO_WSS:
+               *s = s_ws;
+               break;
+       default:
+               *s = s_null;
+       }
+}
--- a/src/modules/nathelper/nathelper.c
+++ b/src/modules/nathelper/nathelper.c
@@ -125,8 +125,8 @@ static int add_rcv_param_f(struct sip_msg *, char *, char *);
 static int nh_sip_reply_received(sip_msg_t *msg);
 static int test_sdp_cline(struct sip_msg *msg);
 
-static int ki_set_alias_to_avp(struct sip_msg *msg, char *uri_avp);
-static int set_alias_to_avp_f(struct sip_msg *msg, str *uri_avp);
+static int w_set_alias_to_avp(struct sip_msg *msg, char *uri_avp, char *hollow);
+static int ki_set_alias_to_avp(struct sip_msg *msg, str *uri_avp);
 static int alias_to_uri(str *contact_header, str *alias_uri);
 static int write_to_avp(struct sip_msg *msg, str *data, str *uri_avp);
 
@@ -228,7 +228,7 @@ static cmd_export_t cmds[] = {
 	{"is_rfc1918",         (cmd_function)is_rfc1918_f,           1,
 		fixup_spve_null, 0,
 		ANY_ROUTE },
-		{"set_alias_to_avp",   (cmd_function)ki_set_alias_to_avp,     1,
+		{"set_alias_to_avp",   (cmd_function)w_set_alias_to_avp,     1,
 		0, 0, ANY_ROUTE },
 	{0, 0, 0, 0, 0, 0}
 };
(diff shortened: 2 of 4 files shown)


ycaner06:
This function , later , will be used for keepalive modules. I have new updates for keepalive module.

src/core/parser/parse_uri.c
```diff
@@ -1439,3 +1439,26 @@ void proto_type_to_str(unsigned short type, str *s) {
 		*s = s_null;
 	}
 }
+
+void proto_type_int_to_str(int type, str *s) {
```
henningw: What about just using the existing function proto_type_to_str? You can easily cast the input parameter from int to unsigned short, or I am wrong?

src/modules/nathelper/nathelper.c
```diff
@@ -2445,6 +2452,175 @@ static int sel_rewrite_contact(str *res, select_t *s, struct sip_msg *msg)
...
+
+	return 1;
+
+}
+/**
+* write_to_avp function writes data to given avp
+* @param data  source data
+* @param uri_avp destination avp name
+*/
+void write_to_avp(struct sip_msg *msg, str *data, str *uri_avp){
```
henningw: It would be probably a good idea to also return an error code if this function fails, to later evaluate it in the set_alias_to_avp function.

src/modules/nathelper/nathelper.c
```diff
@@ -2445,6 +2452,175 @@ static int sel_rewrite_contact(str *res, select_t *s, struct sip_msg *msg)
...
+	valx.flags      =       PV_VAL_STR;
+	valx.rs.s       =       data->s;
+	valx.rs.len     =       data->len;
+
+	LM_DBG("result: %.*s\n", valx.rs.len, valx.rs.s);
+	pvresult->setf(msg, &pvresult->pvp, (int)EQ_T, &valx);
+
+}
+/**
+* function alias_to_uri  select alias paramter from @param contact_header
```
henningw: spelling - parameter

src/modules/nathelper/nathelper.c
```diff
@@ -2445,6 +2452,175 @@ static int sel_rewrite_contact(str *res, select_t *s, struct sip_msg *msg)
...
+* then writes to @param alias_uri
+* @param contact_header  Source contact header
+* @param alias_uri destination string
+*/
+void alias_to_uri(str *contact_header, str *alias_uri){
```
henningw: Same as above, consider to return an error in case this function fails to properly evaluate it in the calling function.

src/modules/nathelper/nathelper.c
```diff
@@ -2445,6 +2452,175 @@ static int sel_rewrite_contact(str *res, select_t *s, struct sip_msg *msg)
...
+* @param msg sip message
+* @param uri_avp given avp name
+*/
+static int set_alias_to_avp_f(struct sip_msg *msg, str *uri_avp){
+	str contact;
+	str alias_uri={0,0};
+
+
+	if(parse_headers(msg,HDR_CONTACT_F,0) < 0 ) {
+	       LM_ERR("Finding Contact Header is failed\n");
```
henningw: Spelling - "Could not find Contact Header"

src/modules/nathelper/nathelper.c
```diff
@@ -2445,6 +2452,175 @@ static int sel_rewrite_contact(str *res, select_t *s, struct sip_msg *msg)
...
+	       return -1;
+	}
+
+	if(!msg->contact)
+	       return -1;
+
+	if(parse_contact(msg->contact)<0 || !msg->contact->parsed ||
+	((contact_body_t *)msg->contact->parsed)->contacts==NULL ||
+	((contact_body_t *)msg->contact->parsed)->contacts->next!=NULL){
+	       LM_ERR("Parsing Contact Header is failed\n");
```
henningw: same

src/modules/nathelper/nathelper.c
```diff
@@ -2445,6 +2452,175 @@ static int sel_rewrite_contact(str *res, select_t *s, struct sip_msg *msg)
...
+	       return -1;
+	}
+
+
+
+	contact.s = ((contact_body_t *)msg->contact->parsed)->contacts->name.s;
+	contact.len = ((contact_body_t *)msg->contact->parsed)->contacts->len;
+
+	alias_to_uri(&contact,&alias_uri);
```
henningw: as referenced below, return value checks for both functions

src/modules/nathelper/nathelper.c
```diff
@@ -2445,6 +2452,175 @@ static int sel_rewrite_contact(str *res, select_t *s, struct sip_msg *msg)
...
+	}
+
+	//last char is proto 0,1,2,3,4..7
+	proto.s= &port.s[port.len+1];
+	proto_type_int_to_str(atoi(proto.s), &proto);
+
+	LM_DBG("Host [%.*s][port: %.*s][proto: %.*s] \r\n",host.len,host.s,port.len,port.s,proto.len,proto.s);
+
+	//sip:host:port;transport=udp
+	alias_uri->s =(char *) pkg_malloc(port.len+host.len+proto.len+16);
```
henningw: When is alias_uri->s freed?

ycaner06:
I will solve problems then add a docs. Thanks for support Henning

src/core/parser/parse_uri.c
ycaner06: i got a error for conversion .Now fixed it in next commit

src/modules/nathelper/nathelper.c
ycaner06: i looked some examples in codes for pv->setf and i dont think that it should be freed.

ycaner06:
Hello, 
if it is ok for function name , i will add doc.
thanks

miconda:
Looking at the resulting diff of the PR (https://patch-diff.githubusercontent.com/raw/kamailio/kamailio/pull/2124.diff), I see that the src/core/parser/parse_uri.h has a single empty line removed. I guess that is not really needed.

The the name of the C function exported to kamailio.cfg file is named `ki_set_alias_to_avp`, but the `ki_` prefix is used for functions exported to KEMI interface. Those exported to kamailio.cfg have usually the `w_` prefix. Then the prototype for kamailio.cfg functions is with three parameters (one sip_msg* and two char*), it is safer to set it like this even if the 2nd parameter is not used, because the core interpreter gives NULL as 2nd char* param when executing the function.

Then, you should export the new function to kemi interface as well, which is more or less the `set_alias_to_avp_f()`, eventually you name this one `ki_set_alias_to_avp`.

miconda:
Looking once again at the code, the function can set the value in any (pseduo-)variable, not only in AVP, right? Because it is using pv->setf() function, not the AVP specific functions.

Do you want to restrict to writing to AVP or it is fine writing to any variable?

ycaner06:
> Looking once again at the code, the function can set the value in any (pseduo-)variable, not only in AVP, right? Because it is using pv->setf() function, not the AVP specific functions.
> 
> Do you want to restrict to writing to AVP or it is fine writing to any variable?

I dont want to restrict because it will be use to add keepalive module as a destination or it can be use to setting/checking/comparing other variables like $du , $ru etc.

henningw:
If you don't want to restrict it to avps, probably it would be good to rename the function (code and configuration file export) to "set_alias_to_pv" or something like this. I guess this was intention of Daniels question.

henningw:
Thank you, merged. I will add 1-2 small formatting changes directly in git master. If there are more changes, they can be done directly in git master as well.

Issue #11, opened 2014-12-20
issue notifying subscribers with transport tcp/tls


lazedo:
Hi,

i noticed a weird behavior in 4.2 branch with this kamailio.cfg (snippet)
# !substdef "!UDP_SIP!udp:MY_IP_ADDRESS:5060!g"
# !substdef "!TCP_SIP!tcp:MY_IP_ADDRESS:5060!g"

listen=UDP_SIP
listen=TCP_SIP
###### # Routing Logic

route
{
    # log the basic info regarding this call
    xlog("L_INFO", "$ci|start|recieved $oP request $rm $ou");
 .....
}

when clients using TCP send a SUBSCRIBE we get log info that it is a UDP operation.
when we send the first notify we get tcp errors.
for reference, i'm using Bria 4 in testing.

```
Dec 20 16:49:29 kamailio-05 kamailio[5085]: INFO: <script>: YzYwYWVlOTFiMGU0YmNjM2IyMDA1NmQ4NjdiMGEwZDY|start|recieved UDP request SUBSCRIBE sip:2135@sip.com
Dec 20 16:49:29 kamailio-05 kamailio[5072]: INFO: <script>: received dialog update for sip:2135@sip.com
Dec 20 16:49:29 kamailio-05 kamailio[5072]: INFO: presence [notify.c:1604]: send_notify_request(): NOTIFY sip:user_7z7b6s@sip.com via sip:user_7z7b6s@192.168.16.12:62285;transpop on behalf of sip:2134@sip.com for event dialog
Dec 20 16:49:29 kamailio-05 kamailio[5072]: CRITICAL: <core> [pass_fd.c:157]: send_all(): send on 0 failed: Socket operation on non-socket
Dec 20 16:49:29 kamailio-05 kamailio[5072]: ERROR: <core> [tcp_main.c:2303]: tcpconn_send_put(): failed to get fd(write):Socket operation on non-socket (88)
Dec 20 16:49:29 kamailio-05 kamailio[5072]: ERROR: tm [../../forward.h:227]: msg_send(): tcp_send failed
Dec 20 16:49:29 kamailio-05 kamailio[5072]: ERROR: tm [uac.c:598]: send_prepared_request_impl(): t_uac: Attempt to send to precreated request failed

```

PR #2127, opened 2019-11-13
core: Improve handling of async-signal-unsafe code in signal handler

guillemj:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description

During a daemon shutdown via a SIGTERM, we got a crash in isatty(3) due to stack exhaustion from infinite nested calls within stdio coming from the syslog(3) in src/main.c:sig_usr():
    
      #0  0x00007f691903fe6d in __isatty (fd=2) at ../sysdeps/posix/isatty.c:24
            term = <error reading variable term (Cannot access memory at address 0x7fffa44f4fb0)>
      #1  0x00007f6918fc37bb in local_isatty (fd=<optimized out>)
      #2  __GI__IO_file_doallocate (fp=0x7f6919110680 <_IO_2_1_stderr_>)
      #3  0x00007f6918fd1700 in __GI__IO_doallocbuf (
      #4  0x00007f6918fd0988 in _IO_new_file_overflow (
      #5  0x00007f6918fcfa8f in _IO_new_file_xsputn (n=8, data=<optimized out>,
      #6  _IO_new_file_xsputn (f=0x7f6919110680 <_IO_2_1_stderr_>,
      7  0x00007f6918fa41fb in _IO_vfprintf_internal (
      #8  0x00007f6918fc32e8 in locked_vfxprintf (
      #9  0x00007f6918fc3469 in __fxprintf (fp=0x7f6919110680 <_IO_2_1_stderr_>,
      #10 0x00007f6918fd3a50 in __malloc_assert (
      #11 0x00007f6918fd5e6f in sysmalloc (nb=nb@entry=4112,
      #12 0x00007f6918fd72c9 in _int_malloc (
      #13 0x00007f6918fd83e3 in __GI___libc_malloc (bytes=bytes@entry=4096)
    
      [ next calls nested until stack exhaustion ]

      #41030 0x00007f6918fc371c in __GI__IO_file_doallocate (
      #41031 0x00007f6918fd1700 in __GI__IO_doallocbuf (
      #41032 0x00007f6918fd0988 in _IO_new_file_overflow (
      #41033 0x00007f6918fcfa8f in _IO_new_file_xsputn (n=8, data=<optimized out>,
      #41034 _IO_new_file_xsputn (f=0x7f6919110680 <_IO_2_1_stderr_>,
      #41035 0x00007f6918fa41fb in _IO_vfprintf_internal (
      #41036 0x00007f6918fc32e8 in locked_vfxprintf (
      #41037 0x00007f6918fc3469 in __fxprintf (fp=0x7f6919110680 <_IO_2_1_stderr_>,
      #41038 0x00007f6918fd3a50 in __malloc_assert (
      #41039 0x00007f6918fd5e6f in sysmalloc (nb=nb@entry=4112,
      #41040 0x00007f6918fd72c9 in _int_malloc (
      #41041 0x00007f6918fd83e3 in __GI___libc_malloc (bytes=bytes@entry=4096)
      #41042 0x00007f6918fc371c in __GI__IO_file_doallocate (
      #41043 0x00007f6918fd1700 in __GI__IO_doallocbuf (
      #41044 0x00007f6918fd0988 in _IO_new_file_overflow (
      #41045 0x00007f6918fcfa8f in _IO_new_file_xsputn (n=8, data=<optimized out>,
      #41046 _IO_new_file_xsputn (f=0x7f6919110680 <_IO_2_1_stderr_>,
      #41047 0x00007f6918fa41fb in _IO_vfprintf_internal (
      #41048 0x00007f6918fc32e8 in locked_vfxprintf (
      #41049 0x00007f6918fc3469 in __fxprintf (fp=0x7f6919110680 <_IO_2_1_stderr_>,
      #41050 0x00007f6918fd3a50 in __malloc_assert (
      #41051 0x00007f6918fd5e6f in sysmalloc (nb=nb@entry=8208,
      #41052 0x00007f6918fd72c9 in _int_malloc (
      #41053 0x00007f6918fd91a2 in __libc_calloc (n=n@entry=1,
      #41054 0x00007f6918fcc0b3 in __GI___open_memstream (
      #41055 0x00007f69190477e0 in __GI___vsyslog_chk (pri=189, flag=-1,
      #41056 0x00007f6919047d4c in __syslog (pri=<optimized out>,
      #41057 0x0000563a58b01c08 in sig_usr (signo=<optimized out>) at main.c:857
      #41058 sig_usr (signo=<optimized out>) at main.c:816
      #41059 <signal handler called>

While this is certainly not easy to trigger, it's still not correct to leave enabled in production builds. It might be an acceptable trade-off as a debugging aid, though. So we should disable this unsafe usage within the signal handler by default.

In addition the sig_usr() signal handler contains debug code to print memory usage and statistics that is not async-signal-safe, which can cause crashes or misbehavior. This code which only affects explicit SIGUSR1 signals on children processes, is not guarded with the SIG_DEBUG macro. We will protect that.


henningw:
Thank you for the pull request. You are right that its not nice if this error happens during a shut-down. So far we did not got many reports for this issue.

If I understand the patch correctly it would be disable the possibility to do the memory logging during shutdown without a recompilation. This memory debugging is a great help during complicated issues reported from our users.

For this reasons we enable some other settings as well in our source code by default, e.g. compile in debugging mode and without advanced optimization. So I don't want to merge it right now without proper consideration the pros and cons.

Lets wait for the opinion from other developers on this one.

guillemj:
Hey @henningw, actually it should not really remove the reporting during shutdown, as that is usually (when not using -D) handled by the handle_sigs() on the main_loop(). This pretty much affects only the children that get signals.

This still leaves the -D usage on the main process as unsafe, but I guess it's a similar compromise to the one of rebuilding with unsafe signal debugging enabled.

henningw:
Thank you for the clarification, I will have another look to it in the next days.

linuxmaniac:
@henningw ping about this one

miconda:
@linuxmaniac - thanks for the reminder. I also looked at this PR a while ago, but forgot to comment.

The version of kamailio is not provided to match the source code line from backtrace, but it seems to be related to printing the memory usage logs at shutdown.

Based on my understanding so far, I do not see the usefulness of this PR at all. The memory status report/summary is printed only if some global parameters are set, respectively `memlog` less or equal than `debug` and `mem_summary` different than `0`.

in my opinion, there are enough options to disable printing the memory logs at the shutdown, if one faces problems with them. Actually they are not printed using the default config, one needs to change those params to get them printed.

Anyhow, if I misunderstood, definitely I do not want a compile flag (macro) to disable/enable shutdown reports, it has to be a global param. It is not convenient to recompile the sources if it is need of more info for troubleshooting.

guillemj:
@miconda I could try to find the specific version for that backtrace, but this is the Sipwise downstream version with all our patches so this might not give you the expected line numbers when compared to a pristine upstream codebase anyway.

Re the usefulness, the first patch makes the code consistent so that when you disable SIG_DEBUG and enable PKG_MALLOC you still get safe behavior, matching the existing code dealing with PKG_MALLOC in the signal handler. For the default change I could take a look at making it run-time configurable, but IMO it should still default to off, but then see the next item.

Re the shutdown reports, please see my reply to @henningw  https://github.com/kamailio/kamailio/pull/2127#issuecomment-554650695.

miconda:
The first patch it is ok, SIG_DEBUG is defined by default.

Lately some of the devs put efforts in eliminating compile time macro switches that were not changed for very long time, to make the code easier to follow. Anyhow, I am fine with the first commit, if it is useful.

But turning this off by default at compile time is not ok, it is needed for tracking memory leaks in PKG and if someone reports it, likely it has it installed with default compile options. Asking to recompile with new flags is not convenient at all, production systems are typically not configured for building packages. And again, the PKG report should not be printed if the global parameters listed above were not changed.

The signal handler of the main process is printing only PKG reports for itself, the main process does not print PKG reports for child processes.

miconda:
I incorporated the first commit in this PR in one (31f0612) I did to add debug messages on other signals for coherence.

The second one that disables debug messages on shutdown by default won't be merged as it is, because getting pkg memory reports is very useful and should be possible without recompiling kamailio, even at the risk of a crash during shut down. Even more, getting pks report can be controlled via memlog global parameter, which can also be changed at runtime without restart.

If one wants to make a variant based on a global parameter instead on disabling at compile time, then make another pull request.

PR #2132, opened 2019-11-15
core: add support for parsing http/2 replies parsing

grumvalski:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

http_async_client module needs to parse http replies. Currently HTTP/2 is not supported by the core parser.

### DIFF
--- a/src/core/parser/parse_fline.h
+++ b/src/core/parser/parse_fline.h
@@ -49,6 +49,9 @@
 #define HTTP_VERSION "HTTP/1."
 #define HTTP_VERSION_LEN (sizeof(HTTP_VERSION)-1)
 
+#define HTTP2_VERSION "HTTP/2"
+#define HTTP2_VERSION_LEN (sizeof(HTTP2_VERSION)-1)
+
 #define CANCEL "CANCEL"
 #define ACK    "ACK"
 #define INVITE "INVITE"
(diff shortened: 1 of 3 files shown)


src/core/parser/parse_fline.h
```diff
@@ -47,7 +47,10 @@
 #define SIP_VERSION_LEN 7
 
 #define HTTP_VERSION "HTTP/1."
-#define HTTP_VERSION_LEN (sizeof(HTTP_VERSION)-1)
+#define HTTP_VERSION_LEN (sizeof(HTTP2_VERSION)-1)
```
miconda: Was the above change somehow done unintentionally? Because I see no reason to have `HTTP_VERSION_LEN` be based on size of `HTTP*2*_VERSION` value.

grumvalski: Yes, it was not done intentionally. I'll fix and force push if you agree.

miconda:
It can be merged. Probably can be considered for backporting at least for 5.3.x, otherwise over the time I expect won't work to do http queries in more and more many cases.

grumvalski:
I agree that the usage of http/2 will become more and more common. If you agree, I'll also backport it to 5.2.x since this is still a supported version.

henningw:
If the patch applies cleanly to 5.2 branch, a backport should be ok there as well.
If there are other (functional) changes necessary before it could be applied, then it would be better to ask people to upgrade in this case to 5.3.x

grumvalski:
The patches applied cleanly on both branches, I needed a couple of days to test. I've cherry-picked the commit.

PR #2133, opened 2019-11-16
keepalive : added new function del_destination and added cfg functions

ycaner06:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
added new function del_destination that deletes sip address from list
added add_destination module functions for kamailio.cfg
added counter parameter for attempt count. after count pass, module dont try until it adds again.
added kemi interfaces both of them
added lock to stack
added find_destination function

### DIFF
--- a/src/modules/keepalive/keepalive_api.c
+++ b/src/modules/keepalive/keepalive_api.c
@@ -54,6 +54,8 @@ int bind_keepalive(keepalive_api_t *api)
 	api->add_destination = ka_add_dest;
 	api->destination_state = ka_destination_state;
 	api->del_destination = ka_del_destination;
+	api->lock_destination_list = ka_lock_destination_list;
+	api->unlock_destination_list = ka_unlock_destination_list;
 	return 0;
 }
 
@@ -64,13 +66,14 @@ int ka_add_dest(str *uri, str *owner, int flags, ka_statechanged_f callback,
 		void *user_attr)
 {
 	struct sip_uri _uri;
-	ka_dest_t *dest=0;
+	ka_dest_t *dest=0,*hollow=0;
 
 	LM_DBG("adding destination: %.*s\n", uri->len, uri->s);
-
-	if(ka_find_destination(uri , owner , &dest , &dest)){
+	ka_lock_destination_list();
+	if(ka_find_destination(uri , owner , &dest , &hollow)){
 		LM_INFO("uri [%.*s] already in stack --ignoring \r\n",uri->len, uri->s);
 		dest->counter=0;
+		ka_unlock_destination_list();
 		return -2;
 	}
 
@@ -107,7 +110,9 @@ int ka_add_dest(str *uri, str *owner, int flags, ka_statechanged_f callback,
 	dest->next = ka_destinations_list->first;
 	ka_destinations_list->first = dest;
 
-	return 0;
+	ka_unlock_destination_list();
+
+	return 1;
 
 err:
 	if(dest) {
@@ -116,6 +121,8 @@ err:
 
 		shm_free(dest);
 	}
+	ka_unlock_destination_list();
+
 	return -1;
 }
 /*
@@ -124,7 +131,7 @@ err:
 ka_state ka_destination_state(str *destination)
 {
 	ka_dest_t *ka_dest = NULL;
-
+	ka_lock_destination_list();
 	for(ka_dest = ka_destinations_list->first; ka_dest != NULL;
 			ka_dest = ka_dest->next) {
 		if((destination->len == ka_dest->uri.len - 4)
@@ -133,7 +140,7 @@ ka_state ka_destination_state(str *destination)
 			break;
 		}
 	}
-
+	ka_unlock_destination_list();
 	if(ka_dest == NULL) {
 		return (-1);
 	}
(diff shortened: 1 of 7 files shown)


ycaner06:
There is two issue,

- I couldn't decide how to clean old address that passed counter, cleaning with a timer thread or when counter passed limit .

-  When address is a tcp connection , callback cant catch state. how to solve this problem?

After reviews , i will add doc for functions.

henningw:
@gbour - some comments from your side about this extensions from Yasin?

henningw:
It seems that Guilliaume recently changed jobs, see if some others developers can comment here.

miconda:
A quick look and there are some issues, I will post specific details soon.

src/modules/keepalive/keepalive_api.c
```diff
@@ -63,9 +64,15 @@ int ka_add_dest(str *uri, str *owner, int flags, ka_statechanged_f callback,
...
 	struct sip_uri _uri;
-	ka_dest_t *dest;
+	ka_dest_t *dest=0;
 
-	LM_INFO("adding destination: %.*s\n", uri->len, uri->s);
+	LM_DBG("adding destination: %.*s\n", uri->len, uri->s);
+
+	if(ka_find_destination(uri , owner , &dest , &dest)){
+		LM_INFO("uri [%.*s] already in stack --ignoring \r\n",uri->len, uri->s);
+		dest->counter=0;
```
miconda: The change of the `dest->counter` is done out of the mutex zone and this operation can be done on an invalid structure if the `dest` was removed meanwhile.

src/modules/keepalive/keepalive_api.c
```diff
@@ -142,3 +140,100 @@ ka_state ka_destination_state(str *destination)
...
+* @result 1 successful  , -1 fail
+*/
+int ka_del_destination(str *uri, str *owner){
+
+	ka_dest_t *target=0,*head=0;
+
+	if(!ka_find_destination(uri,owner,&target,&head)){
+		LM_ERR("Couldnt find destination \r\n");
+		return -1;
+	}
```
miconda: Similar issues, `target` and `head` are retrieved from `ka_find_destination()`, followed by code out of the mutex zone  and later used directly, but at that time any of them can be already deleted. So removal has to be done in the same mutex zone as it is searched.

Actually I do not see any good use for  `ka_find_destination()` returning `target` and `head`. The function can be useful to know if the address exists, but using `target` and `head` after this function expose to segfault. Either you keep the lock set in the function at the return time and unlock later, out of the function, or just make it to return true/false on finding the destination, without returning `target` and `head`.

src/modules/keepalive/keepalive_mod.c
```diff
@@ -52,19 +52,29 @@ static int ka_mod_add_destination(modparam_t type, void *val);
...
+static int w_add_destination(sip_msg_t *msg, char *uri, char *owner);
+static int w_del_destination(sip_msg_t *msg, char *uri, char *owner);
+
 
 extern struct tm_binds tmb;
 
 int ka_ping_interval = 30;
 ka_destinations_list_t *ka_destinations_list = NULL;
+str ka_ping_from = str_init("sip:keepalive@kamailio.org");
+int counter_del = 5;
```
miconda: I suggest adding `ka_` prefix to global variables that are not static and exposed in other files via `extern`.

src/modules/keepalive/keepalive_mod.c
```diff
@@ -52,19 +52,29 @@ static int ka_mod_add_destination(modparam_t type, void *val);
...
 
 
 static cmd_export_t cmds[] = {
 	{"is_alive", (cmd_function)w_cmd_is_alive, 1,
 			fixup_spve_null, 0, ANY_ROUTE},
 	// internal API
+	{"add_destination", (cmd_function)w_add_destination, 2,
+		fixup_add_destination, 0, REQUEST_ROUTE|BRANCH_ROUTE|ONREPLY_ROUTE},
+	{"del_destination", (cmd_function)w_del_destination, 2,
+		fixup_add_destination, 0, ANY_ROUTE},
```
miconda: As more functions are exported from this module, I think is better to also prefix the exported name with `ka_`, like most modules use a common prefix for their functions. Probably we should also add an alias named  `ka_is_alive()` for `is_alive()`.

src/modules/keepalive/keepalive_mod.c
```diff
@@ -126,6 +138,8 @@ static int mod_init(void)
  */
 static void mod_destroy(void)
 {
+	lock_release(ka_destinations_list->lock);
+	lock_dealloc(ka_destinations_list->lock);
 }
```
miconda: This two have to be enclosed in `if(ka_destinations_list) { ... }`, the mod_destroy() is executed even when mod_init() is not finished, in that case ka_destinations_list is NULL, crashing kamailio at shutdown.

ycaner06:
Thanks for review. I will work on these at weekend.

ycaner06:
if it is ok , will add doc.

miconda:
You can add the docs.

miconda:
Thanks!

Issue #2139, opened 2019-11-19
support sd_notify() to improve systemd integration


linuxmaniac:
https://www.freedesktop.org/software/systemd/man/sd_notify.html
> sd_notify() may be called by a service to notify the service manager about state changes. It can be used to send arbitrary information, encoded in an environment-block-like string. Most importantly, it can be used for start-up completion notification.

https://manpages.debian.org/buster/libsystemd-dev/sd_notify.3.en.html

miconda:
I don't think someone will (re-)compile with systemd support in the core.

We should add a core api that executes callbacks on the events you need and then have a module setting them.

In this case I see the patches are minimal, so I can take care converting this PR to a use of a module with new core api callbacks. Then you can add more to it if needed in the future.

linuxmaniac:
Great, thanks!

miconda:
See the newly added systemdops module -- beware it was not tested! Feel free to improve/change it as you need, I did more or less the framework to avoid the need to re-compile the core with libsystemd dependency.

linuxmaniac:
>  Feel free to improve/change it as you need

I will, thank you!

amessina:
Trying to test @miconda's commits on top of the 5.3 branch--can't get Kamailio to notify systemd on startup.  strace doesn't show anything attempting to send notification to systemd.  I'll keep trying...

amessina:
> Trying to test @miconda's commits on top of the 5.3 branch--can't get Kamailio to notify systemd on startup. strace doesn't show anything attempting to send notification to systemd. I'll keep trying...

Looking through `main.c`, I think this is because `sr_corecb_void_exec(app_ready);` is not included in `if (dont_fork)`.  The assumption with `Type=notify` is that we'd be using `kamailio -DD` where the main process isn't forked, correct?

amessina:
I tested a bit more in Fedora 31, but couldn't get this to work as a module.  Similar kamailio/systemd issues like #799, which I'll update there, subjectively.

Thanks again for working toward systemd integration.

amessina:
> I tested a bit more in Fedora 31, but couldn't get this to work as a module. Similar kamailio/systemd issues like #799, which I'll update there, subjectively.
> 
> Thanks again for working toward systemd integration.

Confirming that @linuxmaniac's initial patch does work with `kamailio -DD`.

sergey-safarov:
I can confirm kamailio process reach `sr_corecb_void_exec(app_ready);`
[...]

PR #1762, opened 2018-12-11
pkg/kamailio/obs: Added ruby package #1761

sergey-safarov:
#### Pre-Submission Checklist
- [x] Commit message has the format required by CONTRIBUTING guide
- [ ] Commits are split per component (core, individual modules, libs, utils, ...)
- [ ] Each component has a single commit (if not, squash them into one commit)
- [x] No commits to README files for modules (changes must be done to docbook files
in `doc/` subfolder, the README file is autogenerated)

#### Type Of Change
- [ ] Small bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would change existing functionality)

#### Checklist:
- [x] PR should be backported to stable branches
- [x] Tested changes locally
- [x] Related to issue #1761

#### Description
Ruby module is disabled on CentOS6 and RHEL6 dist

PR #2144, opened 2019-11-22
etc/kamailio.cfg: loaded systemdops.so by default

sergey-safarov:
#1762 ## Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #2139

#### Description
Enabled systemd notification by default.

Victor@linuxmaniac, is change acceptable for dist not have `systemd`?
For similar RPM dist i goes to delete this sting during packaging rules execution.

### DIFF
--- a/etc/kamailio.cfg
+++ b/etc/kamailio.cfg
@@ -306,6 +306,8 @@ loadmodule "xmlrpc.so"
 loadmodule "debugger.so"
 #!endif
 
+loadmodule "systemdops.so"
+
 # ----------------- setting module-specific parameters ---------------
 
 


linuxmaniac:
> @linuxmaniac, is change acceptable for dist not have systemd?

Yes, it is. I would not add this as default for now

miconda:
At this moment it is not going to be added to default configuration file, there are many distributions without systemd, some of them having it now even considering to replace systemd or provide alternatives at the same time.

henningw:
Small addition - to not forget about the non Linux user (various BSDs) and also embedded systems.

PR #2147, opened 2019-11-25
tm: add support for dropping messages in local-request event route

grumvalski:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Currently it is not possible to drop a local generated message in tm:local-request event route because there is no context associated when running the event route. With this PR I tried to solve this introducing a new returned error (E_DROP) and a context into t_run_local_req.

### DIFF
--- a/src/core/error.h
+++ b/src/core/error.h
@@ -48,6 +48,7 @@
 #define E_Q_EMPTY       -16 /* Empty q */
 #define E_Q_TOO_BIG     -17 /* q too big (> 1) */
 #define E_Q_DEC_MISSING -18 /* Decimal part missing */
+#define E_DROP          -19 /* Dropped in script */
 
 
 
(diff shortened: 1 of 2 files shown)


miconda:
Thanks!

I couldn't spot quickly the initialization of the context, something like:

```
init_run_actions_ctx(&ra_ctx);
```

grumvalski:
right, added now.

miconda:
OK, fine for me.

henningw:
One remark - the E_DROP error is planned to be used for all kinds of drop() or only for the local request event route type? If the latter - maybe making the name a bit more specific to highlight that is only used for this particular case.

grumvalski:
I haven't spotted any other usage for the moment. I could rename it like E_LOCAL_DROP.

henningw:
Ok, this might be better understandable in the years to come. :-) Good addition, thanks.

miconda:
If you keep it in core, then E_DROP is fine, it can be used in the future for similar purposes. If you rename it and make it specific to tm usage for local request route, then place the define in the tm module. We should not populate core with module specific needs.

grumvalski:
I did put it in the core to avoid conflicts with other returned codes and for beeing used in the future (in case). I'll leave it in the core.

PR #2156, opened 2019-11-28
pv_headers: Initial commit

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
The main goal of the module is to offload the intermediate header
processing into the XAVP dynamic container as well as provide with high
level methods and pseudovariables to simplify SIP message header
modifications.

### DIFF
--- a/src/Makefile.groups
+++ b/src/Makefile.groups
@@ -15,7 +15,7 @@ mod_list_basic=async auth benchmark blst cfg_rpc cfgutils corex counters \
 				   nat_traversal nathelper path pike pv ratelimit rr rtimer \
 				   rtpproxy sanity sdpops siputils sl statistics textops \
 				   textopsx tm tmx topoh xlog rtpengine stun sipt tcpops \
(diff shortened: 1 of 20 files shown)


henningw:
Interesting module addition, thank you for that. I just had a quick look today, as it is quite some code to read. :-) I noticed two things:

- some missing documentation (3.2 - 3.4) in the README, probably still to be added
- it makes probably sense to move the new str and del str functions to core/str.[h,c] and the str copy function to core/ut.h - what do you think?

miconda:
I looked quickly and I could not spot anything obvious, apart of what @henningw reported regarding the missing parts in the docs -- there are some `??????`. Being a new module, at this moment I don't plan to review more, so I am fine to be merged whenever everyone is ok with it.

linuxmaniac:
* about the missing documentation, It was in the xml I forgot to refresh the README file :-/
* about the str functions... I will create another PR to move them there.

PR #2158, opened 2019-11-28
avpops: avp_subst() support pvar as subst parameter

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
Allow pvar as second parameter of avp_subst() function

### DIFF
--- a/src/modules/avpops/avpops.c
+++ b/src/modules/avpops/avpops.c
@@ -67,6 +67,8 @@ static int fixup_pushto_avp(void** param, int param_no);
 static int fixup_check_avp(void** param, int param_no);
 static int fixup_op_avp(void** param, int param_no);
 static int fixup_subst(void** param, int param_no);
+static int fixup_subst_pv(void** param, int param_no);
+static int fixup_free_subst_pv(void** param, int param_no);
 static int fixup_is_avp_set(void** param, int param_no);
 
 static int w_print_avps(struct sip_msg* msg, char* foo, char *bar);
@@ -81,6 +83,7 @@ static int w_pushto_avps(struct sip_msg* msg, char* destination, char *param);
 static int w_check_avps(struct sip_msg* msg, char* param, char *check);
 static int w_op_avps(struct sip_msg* msg, char* param, char *op);
 static int w_subst(struct sip_msg* msg, char* src, char *subst);
+static int w_subst_pv(struct sip_msg* msg, char* src, char *param);
 static int w_is_avp_set(struct sip_msg* msg, char* param, char *foo);
 
 /*! \brief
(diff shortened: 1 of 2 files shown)


src/modules/avpops/avpops.c
```diff
@@ -1049,9 +1051,36 @@ static int w_op_avps(struct sip_msg* msg, char* param, char *op)
...
+	LM_DBG("%s preparing %s\n", exports.name, subst.s);
+
+	se=subst_parser(&subst);
+	if (se==0)
+	{
+		LM_ERR("%s: bad subst re %s\n", exports.name, subst.s);
+		return E_BAD_RE;
+	}
+
+	return ops_subst(msg, (struct fis_param**)src, se);
```
miconda: I think that se needs to be freed, because subst_parser() allocates the result.

linuxmaniac: Yes, indeed. Thanks!

miconda:
Looking further at this patch, it requires to have the second parameter as a single variable to be considered with "pvar" support, because I see you search for `/` in the second parameter. It makes it support like:

```
$var(se) = "/abc/xyz/";
avp_subst("$avp(x)", "$var(se)");
```

Right now it support PV in the replacement part of the subst expression, not allowing the match/regexp part.

Maybe is better to just add a new function that will always evaluate the subst parameter with variables, allowing more flexible format for subst parameter. At the end, you just need to have your added code in a dedicated function that evaluates first the parameter and then compiles the subst expression.

So we can have:

```
avp_susbtx("$avp(x)", "/$fU/$rU/g");
```
Allowing also:

```
$var(se) = "/" + $fU + "/" + $rU + "/";
avp_substx("$avp(x)", "$var(se)");
```
The new function can have a different name than avp_substx(), I used to exemplify here.

linuxmaniac:
Added ``avp_subst_pv()`` and now this is supported:
```
    $(avp(src)[*]) = "testME";
    $var(x) = "/e/j/gi";
    avp_subst_pv("$avp(src)", "$var(x)");

    $var(z) = "j";
    $var(y) = "e";
    $var(x) = "/" + $var(y) + "/" + $var(z) + "/gi";
    avp_subst_pv("$avp(src)", "$var(x)");
    avp_subst_pv("$avp(src)", "/$var(y)/$var(z)/gi");
    avp_subst_pv("$avp(src)", "/" + $var(y) + "/" + $var(z) + "/gi");
    avp_subst_pv("$avp(src)", "/e/j/gi");
```

henningw:
Is this is still work in progress or could be merged?

linuxmaniac:
I'm waiting for feedback. No more changes from my side.

miconda:
It is a new function now, it can be merged.

PR #2162, opened 2019-12-02
registrar: add path value to xavp_rcd

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
Add path value to xavp_rcd. This is needed when checking if a subscriber is still registered
with a certain parameter in the path.

### DIFF
--- a/src/modules/registrar/doc/registrar_admin.xml
+++ b/src/modules/registrar/doc/registrar_admin.xml
@@ -771,6 +771,11 @@ request_route {
 				<emphasis>received</emphasis> - the record's received value.
 			</para>
 		</listitem>
+		<listitem>
+			<para>
+				<emphasis>path</emphasis> - the record's path value.
+			</para>
+		</listitem>
 		</itemizedlist>
 		<para>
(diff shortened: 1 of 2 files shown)


henningw:
This is a small change and a good extension, I think it is fine to merge it.

miconda:
I am fine to merge, but probably we should add in short term a new parameter to control what attributes are cloned in xavp, some of them can be large in size like, path with couple of hops, and maybe not needed in specific cases.

linuxmaniac:
@miconda something like `modparam("registrar", "xavp_rcd_mask", 0)` so we can define what to load?

miconda:
@linuxmaniac - yes, but can be also the opposite condition: what to skip. What people find better.

henningw:
I like "what to skip" more, in the end most people would probably use it on smaller installations.

linuxmaniac:
Ok, I'm going to merge and I will work on the module parameter to define what values to skip

PR #2166, opened 2019-12-05
tls: add verify_client support

armenb:
Note: See below for more info about motivation of this feature.

#### Pre-Submission Checklist
- [ * ] Commit message has the format required by CONTRIBUTING guide
- [ * ] Commits are split per component (core, individual modules, libs, utils, ...)
- [ * ] Each component has a single commit (if not, squash them into one commit)
- [ * ] No commits to README files for modules (changes must be done to docbook files

#### Type Of Change
- [ * ] New feature (non-breaking change which adds new functionality)

#### Checklist:
- [ * ] Tested changes locally

#### Description

This feature aims to replace require_certificate and verify_certificate params with a single option, verify_client: 
* Provides flexibility: require_certificate and verify_certificate are both booleans, so there are only 4 max combinations of params, and only 3 of them make sense (require_certificate=1 and verify_certificate=0 does not).  In contrast, verify_client is a list of enumerated values, which can be more gracefully expanded by adding additional behaviors to the enum.
* Motivation for this feature is to enable optional_no_ca behavior, described in the docbook; Without this feature, that behavior cannot be represented by any combination of require_certificate and verify_certificate.  I figured if I need to add another variable to support desired behavior, it may as well be one that can hold more than just boolean values.
* This feature was inspired from a similar one in Nginx; that software has a similar "ssl_verify_client" option that takes the same "on", "off", "optional", and "optional_no_ca" values, which effectively implement the same feature.  Note that there is no shared code between implementations, and that these behaviors are implemented (in both cases) via a very thin layer of glue code on top of the OpenSSL library.
* Note that the only function definition in tls_verify.c, verify_callback(int pre_verify_ok, X509_STORE_CTX *ctx), has been compiled into the kamailio binary, but apparently not used.  Rather than modify the existing function, I added a simple 1-line function (2 if you count the log message too) to enable this feature.

Please let me know if I can answer any questions.  Thanks!

### DIFF
--- a/src/modules/tls/doc/params.xml
+++ b/src/modules/tls/doc/params.xml
@@ -1043,6 +1043,7 @@ modparam("tls", "renegotiation", 1)
 			<listitem><para>tls_method - (str) - TLS methods</para></listitem>
 			<listitem><para>verify_certificate - (bool) - see modparam</para></listitem>
 			<listitem><para>require_certificate - (bool) - see modparam</para></listitem>
+			<listitem><para>verify_client - (str) - see modparam</para></listitem>
 			<listitem><para>private_key - (str) - see modparam</para></listitem>
 			<listitem><para>certificate - (str) - see modparam</para></listitem>
 			<listitem><para>verify_depth - (int) - see modparam</para></listitem>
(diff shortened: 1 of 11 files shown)


miconda:
Thanks! As I can see, this is an alternative, not replacing the old variants.

Seems ok after quick look. I will squash&merge soon once I get a bit more time to check it.

armenb:
Hi Daniel,

Ah indeed, this is an alternative way of controlling certificate require/verify behavior, not a strict replacement.

I updated the docbook to include a conversion table between verify_certificate+require_certificate parameters and verify_client parameters (and fixed a docbook bug in the process).

Armen

PR #2175, opened 2019-12-09
dmq_usrloc: replicate_socket_info using dmq

nmreis:
- replicate_socket_info using dmq through a new modparam
 - useful for anycast scenarios


henningw:
Maybe Charles is not available right now. This looks to me like a small change, should also not affect existing replication functionality. Lets wait e.g. until the end of this week if there are no objections from other developers, and then merge it.

miconda:
Thanks! Merged, no comments for a while.

PR #2177, opened 2019-12-16
keepalive: added add/del/get/flush rpc commands

ycaner06:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

added add/del/get/flush rpc commands

### DIFF
--- a/src/modules/keepalive/keepalive_rpc.c
+++ b/src/modules/keepalive/keepalive_rpc.c
@@ -44,18 +44,21 @@
 static const char *keepalive_rpc_list_doc[2];
 static const char *keepalive_rpc_add_doc[2];
 static const char *keepalive_rpc_del_doc[2];
-
+static const char *keepalive_rpc_get_doc[2];
+static const char *keepalive_rpc_flush_doc[2];
 
 static void keepalive_rpc_list(rpc_t *rpc, void *ctx);
 static void keepalive_rpc_add(rpc_t *rpc, void *ctx);
 static void keepalive_rpc_del(rpc_t *rpc, void *ctx);
-
+static void keepalive_rpc_get(rpc_t *rpc, void *ctx);
+static void keepalive_rpc_flush(rpc_t *rpc, void *ctx);
 
 rpc_export_t keepalive_rpc_cmds[] = {
 	{"keepalive.list", keepalive_rpc_list, keepalive_rpc_list_doc, 0},
 	{"keepalive.add", keepalive_rpc_add, keepalive_rpc_add_doc, 0},
 	{"keepalive.del", keepalive_rpc_del, keepalive_rpc_del_doc, 0},
-
+	{"keepalive.get", keepalive_rpc_get, keepalive_rpc_get_doc, 0},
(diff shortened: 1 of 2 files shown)


linuxmaniac:
documentation is missing

miconda:
It can be merged if no one else finds something to be changed.

src/modules/keepalive/keepalive_rpc.c
```diff
@@ -87,3 +100,137 @@ static void keepalive_rpc_list(rpc_t *rpc, void *ctx)
...
+static void keepalive_rpc_add(rpc_t *rpc, void *ctx)
+{
+	str sip_adress = {0,0};
+	str table_name ={0,0};
+	int ret = 0;
+
+	ret = rpc->scan(ctx, "SS",&sip_adress,&table_name);
+
+	if (ret < 2) {
+		LM_ERR("not enough parameters - read so far: %d\n", ret);
```
oej: When reading the log file, you probably want to add the command name or some context 
LM_ERR("keepalived.add : not enough parameters - read so far: %d\n", ret);

src/modules/keepalive/keepalive_rpc.c
```diff
@@ -87,3 +100,137 @@ static void keepalive_rpc_list(rpc_t *rpc, void *ctx)
...
+		rpc->fault(ctx, 500, "Not enough parameters or wrong format");
+		return;
+	}
+
+	LM_DBG(" keepalive deletes [%.*s]\n", sip_adress.len , sip_adress.s);
```
oej: s /deletes/delete/

src/modules/keepalive/keepalive_rpc.c
```diff
@@ -87,3 +100,137 @@ static void keepalive_rpc_list(rpc_t *rpc, void *ctx)
...
+static void keepalive_rpc_flush(rpc_t *rpc, void *ctx)
+{
+	ka_dest_t *dest;
+	LM_DBG("Keepalive flushes  \n");
+	ka_lock_destination_list();
+
+	for(dest = ka_destinations_list->first; dest != NULL; dest = dest->next) {
+		free_destination(dest);
+	}
+	ka_destinations_list->first = 0;
```
oej: Are you sure you want to set it to "0" here? first seems to be of type ka_dest_t * so maybe NULL is better?

miconda: There is no strict policy on using `NULL`  or `0` for pointers, probably the former is more intuitive for pointers and eventually raise compile warnings if the type is not pointer, but there are many places in the code where `0` is used for this purpose. If someone wants to make it coherent everywhere by using `NULL` instead of `0` for pointers, I am fine with it.

Given that the PR is rather old here, I am going to merge it and fix what was pointed in the other comments.

src/modules/keepalive/keepalive_rpc.c
miconda: I haven't changed this one. The C function name is printed in each log message, being suggestive enough, but if you want to make an update, you can push it to the repo.

Issue #2167, opened 2019-12-06
Weight-based contacts call termination distribution


marcocapetta:
### Description
Implement a new TM logic to distribute calls to contacts belonging to the same subscriber similar to the weight mechanism that is used in lcr module. This mechanism should support primary/backup, load-sharing or a combination of both.

### Expected behavior
Currently, with the existing serial forking using q-value:
 - if the q-values are different, the devices will ring serially but always in the same order
 - if the q-values are equals, the devices will ring in parallel

With the new mechanism, instead, I have to be able, for example, to register 2 devices and send:
 - the 50% of the call to device 1 and then to device 2 as a backup
 - the 50% of the call to device 2 and then to device 1 as a backup

### Possible Solutions
A possible approach is to extend the existing load_contact function and re-use the q-value as a weight.
For example:
 - contact 1 has Q value 1
 - contact 2 has Q value 0.5
 - contact 3 has Q value 0.5
 - contact 4 has Q value 0
Then kamailio distributes the calls using the following proportion:
 Q : total_Q = call_prob : 100
So:
 - contact 1 has probability to ring of 50% (1 * 100 / 2)
 - contact 2 has probability to ring of 25% (0.5 * 100 / 2)
 - contact 3 has probability to ring of 25% (0.5 * 100 / 2)
 - contact 4 has probability to ring of 0 (0 * 100 / 2) but it is used as backup option if all other contacts fail

This should solves the limitations of serial forking and allows to achieve equal or predictable weight-based call termination distribution.

miconda:
The related PR was merged.

PR #2178, opened 2019-12-16
tm: new weight-based call-termination distribution

marcocapetta:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #2167

#### Description
Extended t_load_contacts function with the parameter 'mode'.
It can be used to define the algorithm to use for ordering the contacts.
If 0, or missing, the standard algorithm with be used.
If 1, the new weight-based algorithm will be used.

The new algorithm uses the q value as the weight of the branch.
Higher is the q value, higher is the probability to be tried as first. Equals q values means equals probability to be tried.
Branches with q values equals to 0 or lower are not considered by the ordering algorithm, but just added at the end of the list as backup option if all other branches fail.

### DIFF
--- a/src/modules/tm/t_serial.c
+++ b/src/modules/tm/t_serial.c
@@ -33,6 +33,8 @@
 #include "../../core/parser/msg_parser.h"
 #include "../../core/ut.h"
 #include "../../core/xavp.h"
+#include "../../core/sr_module.h"
+#include "../../core/rand/kam_rand.h"
 #include "config.h"
 #include "t_funcs.h"
 #include "t_reply.h"
@@ -40,6 +42,9 @@
 
 /* usr_avp flag for sequential forking */
 #define Q_FLAG      (1<<2)
+/* t_load_contacts modes/algorithms */
+#define T_LOAD_STANDARD     0
+#define T_LOAD_PROPORTIONAL 1
 
 extern str ulattrs_xavp_name;
 
@@ -57,6 +62,7 @@ struct contact {
 	unsigned short q_flag;
 	struct contact *next;
 	sr_xavp_t *ulattrs;
+	unsigned short q_index;
 };
 
 struct instance_list {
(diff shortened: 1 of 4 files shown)


miconda:
I looked quickly trying to see how the existing code was changed, which seems to be mainly moved in a dedicated function. I didn't dig in the new algorithm (mode), being specific for the feature wanted with this PR.

Maybe @juha-h wants to check a bit and see if the old mode (now STANDARD) is not affected somehow.

Otherwise, if nobody else has comments, I am fine to merge.

src/modules/tm/t_serial.c
```diff
@@ -159,26 +164,171 @@ void add_contacts_avp(str *uri, str *dst_uri, str *path, str *sock_str,
...
+ * only one contact at a time has to ring.
+ */
+int t_load_contacts_proportional(struct contact *contacts, char *sock_buf, int n, unsigned short q_total)
+{
+	int q_remove, n_rand, idx;
+	struct contact *curr;
+	sr_xavp_t *lxavp = NULL;
+
+	/* Initialize the random generator */
+	kam_srand(time(NULL));
```
henningw: This is not necessary (and will also overwrite the state for the global one). The random generator is initialized during startup.

src/modules/tm/t_serial.h
```diff
@@ -28,14 +28,19 @@
 #ifndef _T_SERIAL_H_
 #define _T_SERIAL_H_
 
+#define STANDARD 0
```
henningw: Makes probably sense to qualify the defines, e.g. with T_LOAD_ prefix or similar. Also consider to move them into the respective .c file, if they are not necessary in the .h

henningw:
Thank you for the change. @juha-h - comments from your side as well?

henningw:
Merged - eventual further changes could be done directly in git master.

Issue #2057, opened 2019-09-06
cfg parse error behaviour on a mismatch of #!ifdef/#!endif statements


henningw:
The #!ifdef/#endif statement parsing of the kamailio.cfg does not work correctly on a mismatch of #!ifdef/#!define statements. In some cases will just ignore the #!defines and parses the cfg file without the statements.

version: kamailio 5.3.0-pre0 (x86_64/linux) ed10d7

> henning@linux:~/repositories/kamailio/src> ./kamailio -c -f ../etc/test-define.cfg 
>  0(9501) CRITICAL: <core> [core/cfg.y:3539]: yyerror_at(): parse error in config file /home/henning/repositories/kamailio/src/../etc/test-define.cfg, line 12, column 8: syntax error
>  0(9501) CRITICAL: <core> [core/cfg.y:3539]: yyerror_at(): parse error in config file /home/henning/repositories/kamailio/src/../etc/test-define.cfg, line 12, column 8: unknown config variable
>  0(9501) CRITICAL: <core> [core/cfg.y:3536]: yyerror_at(): parse error in config file /home/henning/repositories/kamailio/src/../etc/test-define.cfg, line 12, column 9-11: 
> ERROR: bad config file (3 errors)

Test cfg to show the problem:

```
henning@linux:~/repositories/kamailio/etc> cat test-define.cfg
#!KAMAILIO

#!define foo

#!ifdef foobar1
#!ifdef foobar2
#!endif
#!endif
#!endif

#!ifdef bar
invalid=123
#else 
children=8
#!endif

request_route {
        ;
}

```

The "bar" condition is parsed, even if the #!define was not set. The WARN message is also not shown in this error case.

Probably the easist fix would be to stop kamailio startup on a mismatch of #!ifdef/#endif statements and output ERROR instead of the current WARN.

miconda:
While looking at the code I discovered that the index of preprocessor conditions could become negative and even be used for a wrongly placed #!else - I pushed a fix for it: 8f6e826576e52270f21c22a7d94b3026cc22c172 -- this probably needs to be backported.

Otherwise, it is fine for me to stop starting kamailio on a mismatch of ifdef/ifndef-endif pairs.

henningw:
Thank you Daniel, the example from above works now. Unfortunately the opposite condition still not work:

> $./kamailio -c -f ../etc/test-define.cfg -Y /root
[...]

PR #2183, opened 2019-12-23
core: stop start if ifdef check fails

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #2057

#### Description

If the config doesn't have the same amount of if[n]def endif directives, don't start it


miconda:
OK to merge from my point of view.

PR #2187, opened 2020-01-02
mqtt: module to support mqtt pubsub messaging

towe75:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

New Module to participate in a MQTT pub/sub messaging system. 

Initial Features:

- automatic reconnect with connected/disconnected event routes
- support mqtt last will feature
- subscribe and unsubscribe topics via script code
- receive messages via event route
- send messages from script code
- full KEMI support

### DIFF
--- a/src/modules/mqtt/README
+++ b/src/modules/mqtt/README
@@ -1,0 +1,349 @@
+MQTT Module
+
+Thomas Weber
+
+   <thomas.weber@pascom.net>
+
+Edited by
+
+Thomas weber
+
+   <thomas.weber@pascom.net>
+
+   Copyright © 2019 Thomas Weber
+     __________________________________________________________________
+
+   Table of Contents
+
+   1. Admin Guide
+
+        1. Overview
+        2. Dependencies
+
+              2.1. External Libraries or Applications
+
+        3. Parameters
+
+              3.1. host (str)
+              3.2. port (int)
+              3.3. keepalive (int)
+              3.4. id (str)
+              3.5. username (str)
+              3.6. password (str)
+              3.7. will_topic (str)
+              3.8. will (str)
+              3.9. event_callback (str)
+
+        4. Functions
+
+              4.1. mqtt_subscribe(topic, qos)
+              4.2. mqtt_unsubscribe(topic)
+              4.3. mqtt_publish(topic, message, qos)
+
+        5. Event routes
+
+              5.1. mqtt:connected
+              5.2. mqtt:disconnected
+              5.3. mqtt:message
+
+        6. Exported pseudo-variables
+
+   List of Examples
+
+   1.1. Set host parameter
+   1.2. Set port parameter
+   1.3. Set keepalive parameter
+   1.4. Set id parameter
+   1.5. Set username parameter
+   1.6. Set password parameter
+   1.7. Set will_topic parameter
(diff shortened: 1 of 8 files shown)


oej:
I would love to see TLS support clearly implemented and documented. After all, it's 2020 :-)

src/modules/mqtt/README
```diff
@@ -0,0 +1,349 @@
...
+event_route[mqtt:message] {
+        xlog("mqtt message received: $mqtt(topic) -> $mqtt(msg) !\n");
+}
+...
+
+6. Exported pseudo-variables
+
+     * $mqtt(topic) - Received topic (only in mqtt:message)
+     * $mqtt(msg) - Received message (only in mqtt:message)
+
```
oej: How do I reach the QoS value for the message?

oej:
Thank you for this pull request. Generally, I miss these improvements to your code:
- TLS
- Failover (multiple MQTT servers)

Maybe one could want to set event_route per subscription too.

src/modules/mqtt/README
towe75: @oej  it's not exposed ATM but it's simple to add it. I will enhance the module accordingly.

towe75:
@oej thank you for the **really** quick review (only a couple of minutes after submission). 

I will check the effort for TLS support. It should not be very high given the fact that libmosquitto exposes a highlevel API for it. 

Regarding event_route per subscription: it's a nice idea and saves the script author a bit of boilerplate code. On the other hand: it would have to support pattern matching for wildcard topics to be really useful. I will not add it for now because of expected complexity.

oej:
Regarding event_route per subscription: I was toying with the idea to add a new parameter to the subscribe function. I think you consider something more automatic, but that would be more complex. Why not just add a third parameter?

towe75:
@oej: it would IMHO only be simple for "static" subscriptions like "kamailio/foo" -> on_mqtt_foo(). 

But what about wildcard subscriptions like "kamailio/+/unregister" or "kamailio/#" ? I can, of course, forbit these mappings by some validation rule but a cleaner solution would be to pattern-match each incoming message against the wildcard subscription to find the correct route. 

It's IMHO a somewhat sophisticated feature for a later version, if the need arises. For now the mapping for incoming messages can be done in script code. We can discuss about it in more detail on the mailing list or some other channel, if you want.

oej:
Ok, let's add new features later and polish this module. It's a very good start compared with my core-dumping first attempt :-)

Will try to test it soon.

src/modules/mqtt/README
towe75: @oej, see commit ea69941

towe75:
I added TLS support in commit 81852e5. The documentation is enhanced as well. 

It's now possible to have a plain connection as well as TLS with configurable method, ciphers, client certificate and - if desired - disabled server cert checks. 

Parameters are more or less directly forwarded to libmosquitto and subsequently into openssl. So the default optionset depends on the installed openssl version and the available libmosquitto version.

miconda:
Going to merge it, is not a lot of code and further reviews and adjustments can be done directly in the repo afterwards. I will grant @towe75 git commit access so he can maintain the module. Thanks for the contribution!

PR #2197, opened 2020-01-20
pkg: fixed SystemD service install path

guss77:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
- `systemd/system/service` is not a real directory. Services should be in `systemd/system` to be loaded by SystemD.
- Also, fixed typo in `%files`.

### DIFF
--- a/pkg/kamailio/oracle/el7/kamailio.spec
+++ b/pkg/kamailio/oracle/el7/kamailio.spec
@@ -309,9 +309,9 @@ make install-modules-all group_include="k%PYTHON_MODULES"
 make install-modules-all group_include="k%GEOIP_MODULES"
 
 
-mkdir -p $RPM_BUILD_ROOT/usr/lib/systemd/system/service
+mkdir -p $RPM_BUILD_ROOT/usr/lib/systemd/system
 install -m644 $RPM_SOURCE_DIR/kamailio.service \
-              $RPM_BUILD_ROOT/usr/lib/systemd/system/service/kamailio.service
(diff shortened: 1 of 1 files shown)


miconda:
The commit message first line is not formatted according to the contributing guidelines:

You can reformat the commit message if you want, then the PR can be merged directly.

Otherwise the patch can be applied manually this time and you keep in mind for future commits.

Note also that the specs of oracle linux were not properly maintained lately, if you want to update their content for the last version, see the spec file from obs folder, that should be pretty actual.

sergey-safarov:
Oracle dist based in RHEL dist and must be packaged using this [packaging files](https://github.com/kamailio/kamailio/tree/master/pkg/kamailio/obs)

This already documented at Oracle [README.md](https://github.com/kamailio/kamailio/blob/master/pkg/kamailio/oracle/README.md)

I prefer to delete not maintained oracle packaging files and leave only `README.md` file.

guss77:
> I prefer to delete not maintained oracle packaging files and leave only `README.md` file.

Sounds like a valid solution - if the position of the project is that RPM builds should be based on `pkg/kamailio/obs` instead of any of the `pkg/kamailio/oracle` files, I can work with that.

I'm currently building RPMs for my use based on the oracle file, and as it has a few problems, I keep patching it locally for each release - which is getting tedious. I'll update my packaging to use the `obs` file and will submit patches as relevant (and I'll make sure to write comments as required, next time).

sergey-safarov:
You are welcome to update OBS packaging file.

PR #2206, opened 2020-01-27
core:tcp allow to handle TCP connections with tcp_accept_haproxy=yes flag even message does not contain PROXY protocol header

ovoshlook:
Covers setups when kamailio behind the service with HAPROXY protocol support but in other hand it also serves requests from other services that available per direct connection (such as B2BUA in the same local network).

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
- function tcpconn_read_haproxy() returns 2 if PROXY header NOT found. 
- Message about no header shown at the debug level.
- No errors
- TCP connection successfully establishes

### DIFF
--- a/src/core/tcp_main.c
+++ b/src/core/tcp_main.c
@@ -1109,8 +1109,8 @@ int tcpconn_read_haproxy(struct tcp_connection *c) {
 		return 1; /* EOF? Return "no IP change" in any case */
 	}
 	else {
-		/* Wrong protocol */
-		return -1;
+		/* Wrong protocol*/
+		return 2;
 	}
 
 done:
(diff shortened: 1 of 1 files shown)


miconda:
I applied the patches manually, with some updates to log messages. Test and if any issues, report on bug tracker. Thanks!

ovoshlook:
tested - works fine. IS it possible to push it into 5.3.x branch as well? I know there more changes in the tcp_main a 5.4.0
I can prepare PR for 5.3 branch.

vaizki:
Documentation still states in the core cookbook:

> Please note that enabling this option will reject any inbound TCP connection that does not conform to the PROXY-protocol spec.

So I started looking at making this a per-listen socket option.. And then ran into this commit. Apparently all connections would be accepted now in 5.6, including my Prometheus scraping through xhttp? So the only side effect of having non-proxyprotocol connections is a debug-level message about it?

ovoshlook:
Hi 
I think,  I just forgot to update this part of the doc.
It should be working fine for any connection type and yes. It should be only a debug-level message

But feel free to test it and let us know if something is wrong.

PR #2211, opened 2020-02-01
sanity: use debug log level when content length header is missing

lassebm:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
This is to align behavior with the remaining checks. The example configuration enable the content length sanity check and prior to this change would warn when SIP messages without body were received.

### DIFF
--- a/src/modules/sanity/sanity.c
+++ b/src/modules/sanity/sanity.c
@@ -582,7 +582,7 @@ int check_cl(sip_msg_t* msg) {
 		}
 		LM_DBG("check_cl passed\n");
 	} else {
-		LM_WARN("content length header missing in request\n");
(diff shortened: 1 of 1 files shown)


ovoshlook:
But WARNING looks correct behaviour because of [RFC](https://tools.ietf.org/html/rfc3261#page-169) specifications of content-length header, why it should be changed to DBG?

henningw:
Yes, it is indeed a SHOULD. In the end one could simply deactivate it in the cfg (with the sanity flag) or just increase the log level to not include WARN.

henningw:
Close this one

lassebm:
I agree with closing after reading the RFC section once again 👍

PR #2212, opened 2020-02-04
tm: t_continue() - do not print error and return error code

eschmidbauer:
- t_continue() - do not print error and return error code.
- change func to log INFO so errors are not printed twice (once by tm and once by func call)

### DIFF
--- a/src/modules/tm/t_suspend.c
+++ b/src/modules/tm/t_suspend.c
@@ -189,7 +189,7 @@ int t_continue_helper(unsigned int hash_index, unsigned int label,
 
 	if (t_lookup_ident_filter(&t, hash_index, label, 1) < 0) {
 		set_t(backup_T, backup_T_branch);
-		LM_ERR("active transaction not found\n");
+		LM_WARN("active transaction not found\n");
 		return -1;
 	}
 


henningw:
Hi, thanks for the pull request. Just briefly looked to the code, is t_continue not just a small wrapper around the t_continue_helper function? Where is the ERROR reported the second time?

eschmidbauer:
`tm_continue()` is an API endpoint inside `tm` so other modules such as `tmx` call it.

if you look at `t_continue()` in `tmx` module, it checks return code and prints it's own log message:
`	if(_tmx_tmb.t_continue(tindex, tlabel, act)<0)`

i think the API func should only print an error on critical failures. `t_lookup_ident_filter()` check is quiet graceful and should not alarm the system with an error message since it's very possible & often likely a transaction has completed when it's called.

henningw:
Ah, I did misunderstood you here. But IMHO to report an INFO when the API call returns an error, is probably not the expectation of a caller. In the end this could be also used from other modules. I would suggest a WARN, similar to the condition when the transaction is not suspended below.

eschmidbauer:
agreed. updated PR to use WARN level

henningw:
Thanks, merged

PR #2226, opened 2020-02-21
core: log parser error messages with debug level 

nferreyra:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description
* Make parser error logs as debug avoiding excessive logging for negative HTTP replies.
The issue was discussed here: https://lists.kamailio.org/pipermail/sr-users/2020-February/108232.html


henningw:
Thanks for the pull request.  Well, these are the core parsing functions, that are used for all kind of messages. I don't like to just change it to DBG, because it suppress all kind of parsing errors then. 
You pull request contains also three unrelated merge commits.
I think there should be another solution possible.

miconda:
Those message can be aready skipped by setting `core_log` global parameter.

  * https://www.kamailio.org/wiki/cookbooks/5.3.x/core#corelog

No need to force them to a specific log level.

As for topos, I pushed a patch to catch http replies. Try with:

  *  https://github.com/kamailio/kamailio/commit/fad795648f8cbf12cc099dea18092ff7beee9c75

nferreyra:
@miconda thank you for the corelog reference, it was there all the time and I didn't rtfm.
You topos patch is working fine. For the sake of consistency, shouldn't a similar patch be applied to topoh?
Thank you.

miconda:
topoh got a similar patch, if you can test that module as well and report the results, it would be great.

PR #2243, opened 2020-03-10
dialog: update event route statistics

smititelu:
If dialog was deleted while running the event route, update statistics
for early and active dialogs.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Please let me know if I can backport it to 5.3

### DIFF
--- a/src/modules/dialog/dlg_handlers.c
+++ b/src/modules/dialog/dlg_handlers.c
@@ -1810,6 +1810,16 @@ int dlg_run_event_route(dlg_cell_t *dlg, sip_msg_t *msg, int ostate, int nstate)
 		if (dlg0==0) {
 			LM_ALERT("after event route - dialog not found [%u:%u] (%d/%d) (%p) (%.*s)\n",
 					h_entry, h_id, ostate, nstate, dlg, evname.len, evname.s);
+			if (nstate == DLG_STATE_DELETED) {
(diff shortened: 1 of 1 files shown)


henningw:
Hi Stefan, thanks for the pull request. Can you elaborate a bit on the failure condition that this patch solves? Shouldn't be the dialog statistics updated by the "normal" code e.g. in dlg_handlers.c? Or are you refering to conditions when a dialog is deleted _by_ the event route?

smititelu:
Indeed, this needs a bit more explanation.

Check out this function flow:
```
dlg_onreply() -> dlg_run_event_route() -> dlg_lookup() -> if (dlg0 == 0)

```
When dlg_run_event_route() fails (returns -1) then dlg_onreply() just returns.

Now, in the same dlg_onreply() at line 571, the statistics are updated in case of new_state==DLG_STATE_DELETED. However, because of the above, this code is not called if dlg_run_event_route() fails.

Most probably I should update this pull request to increment also the failed dialogs, as done in dlg_onreply() at line 585.

Let me know if this clarifies your questions.

smititelu:
I've updated the "if" conditions for updating statistics in case of dlg_lookup() failure.

henningw:
Ok, thanks for the clarification. You can merge and also backport, its a bugfix.

smititelu:
Cool. Thanks for review.

PR #2253, opened 2020-03-16
p_usrloc: add new update function

smititelu:
Add also new "uniq" column.
The new update function will use this new column.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

### DIFF
--- a/src/modules/p_usrloc/config.c
+++ b/src/modules/p_usrloc/config.c
@@ -47,7 +47,7 @@ cfg_def_t	p_usrloc_cfg_def[] = {
 		" Specifies the error value on which a database shall be turned off. "},
 	{"failover_level",		CFG_VAR_INT | CFG_ATOMIC,	0, 0, 0, 0,
 		"Specifies the manner a failover is done (1 = turn off, 2 = find a spare) "},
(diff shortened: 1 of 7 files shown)


henningw:
Thanks for the pull request.
I think it might be better to use a value of "2" to specify the new scheme (at least in the docs). This would be more similar to the way other modules mode parameter work.
Did I understand it correctly that you add a new DB column? Is this for the reg_db table or the (partitioned) location table?

smititelu:
I've updated the pull request to consider only 0, 1, 2 values (docs and code).

I've added the new function to consider additional columns when updating database; "uniq_col" is one of them. We have some user agents that send "uniq" in Contact header of SIP REGISTER message.

henningw:
Thank you. So I understand correctly that the "new way" is using the uniq column for updates.

The p_usrloc module is not using the generic DB scheme infrastructure. So it would be great if you could also update the _location.sql_ file in the p_usrloc module directory to include the new column definition, just to have it somewhere defined.

smititelu:
Updated p_usrloc's location.sql to include "uniq" column.

henningw:
Thanks again, you can merge.

smititelu:
Thanks for review.

PR #2257, opened 2020-03-19
ratelimit: increase default number of queues

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

In our setup, we need one more ratelimit queue (in addition to the default of 10).
In this pull request, I am increasing the default number of ratelimit queus to 16.

Please let me know if this pull request is ok, or if I should do it via a modparam and dynamic allocation of ```static rl_queue_t queues```.

### DIFF
--- a/src/modules/ratelimit/ratelimit.c
+++ b/src/modules/ratelimit/ratelimit.c
@@ -47,8 +47,8 @@
 
 MODULE_VERSION
 
-#define MAX_PIPES       16
-#define MAX_QUEUES      10
+#define MAX_PIPES       32
+#define MAX_QUEUES      32
 
 /*
  * timer interval length in seconds, tunable via modparam


miconda:
If you don't want to make it modparam, then better to be increased a bit higher, maybe 32. Same for the number of pipes.

On the other hand, have you looked at pipelimit module? It doesn't have the concept of queues anymore, but they can be replaced with IF conditions in config script, iirc.

smititelu:
Updated the default number of pipes and queues to 32.

Thanks, will definitely have a look on the pipelimit module also.

miconda:
From my point of view, this can be backported if needed, is just increasing the capacity of existing features, as the limits were hit in specific deployments, no new feature added. Like we had to do in the past to increase internal buffer sizes to cope with larger values.

PR #2260, opened 2020-03-23
Fix returning multiple IP's for a cached DNS entry

nbruning:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
The method `dns_entry2he` was returning 10 `hostent` structures, containing the same record, because the rr counter was never updated (10 being the value of `DNS_HE_MAX_ADDR`).

With this fix, the cache correctly returns all records found by the resolver.

In the documentation of `dns_int_match_ip`, it says:
>> At this moment, the function might not check all the IP addresses as returned by dns_sys_match_ip(), because the internal resolver targets to discover the first address to be used for relaying SIP traffic.

I believe that with this patch, `dns_int_match_ip` now also works for DNS entries with multiple values. I'm not sure about mixed ipv4/ipv6 though.

### DIFF
--- a/src/core/dns_cache.c
+++ b/src/core/dns_cache.c
@@ -2304,6 +2304,7 @@ inline static struct hostent* dns_entry2he(struct dns_hash_entry* e)
 						return 0;
 				}
 				memcpy(p_addr[i], ip, len);
+				rr_no++;
 	}
 	if (i==0){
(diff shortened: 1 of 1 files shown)


miconda:
Thanks! Patch applied manually to update commit message to follow the format from contributing guideline:

PR #2264, opened 2020-03-27
core: strutils - add urlencode and urldecode functions

armenb:
- moved from modules/pv/pv_trans.c and exported, so they can be used in other modules
- genericized fn names


miconda:
Thanks!

PR #2268, opened 2020-03-31
tls: add support for escaped cert PVs and select

armenb:
- new PVs: $tls_peer_escaped_cert, $tls_my_escaped_cert
- new selects: @tls.peer.escaped_cert, @tls.my.escaped_cert

### DIFF
--- a/src/modules/tls/tls_select.c
+++ b/src/modules/tls/tls_select.c
@@ -41,6 +41,7 @@
 #include "../../core/ut.h"
 #include "../../core/cfg/cfg.h"
 #include "../../core/dprint.h"
+#include "../../core/strutils.h"
 #include "tls_server.h"
 #include "tls_select.h"
 #include "tls_mod.h"
@@ -58,6 +59,8 @@ enum {
 	CERT_SELFSIGNED,  /* self-signed certificate test */
 	CERT_NOTBEFORE,   /* Select validity end from certificate */
 	CERT_NOTAFTER,    /* Select validity start from certificate */
+	CERT_RAW,         /* Select raw PEM-encoded certificate */
+	CERT_URLENCODED,  /* Select urlencoded PEM-encoded certificate */
 	COMP_CN,          /* Common name */
 	COMP_O,           /* Organization name */
 	COMP_OU,          /* Organization unit */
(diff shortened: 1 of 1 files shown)


henningw:
Hi, thanks for the pull request. 
You are probably aware of the available transformation, does it not suits your requirement?
https://www.kamailio.org/wiki/cookbooks/devel/transformations#surlencodeparam

miconda:
@henningw - these seem to be completely new variables, returning the full certificate in escaped format. The transformation would help if there would be variables returning the full certificate in unescaped format.

armenb:
@henningw @miconda indeed, the full raw PEM-encoded certificate is not currently available via an existing PV.

Thinking through this a little bit more, would you prefer a full certificate access mechanism via something like $tls_peer_raw_cert{s.urlencode.param} instead?

Just to explain where I was coming from, nginx has deprecated access to the raw PEM-encoded certificate, encouraging users to use a new urlescaped cert mechanism instead:

http://nginx.org/en/docs/http/ngx_http_ssl_module.html

It is that avenue of thinking that I created this PR for accessing and urlencoding the URL cert with one PV.  In retrospect, nginx's urlencoding of the certificate is probably rooted in use cases that involve wrapping the cert in a header and passing it off to a different system for processing.  I did not consider use of the raw PEM-encoded certificate in a kemi-enabled scenario, where $tls_peer_escaped_cert would need to be urldecoded before being asn1parsed (or even re-encoded with some other encoding scheme in pv_trans.c...).

I'll create $tls_peer_raw_cert as a separate PR, since that seems simpler.  Does this plan sound good to you?

henningw:
Thanks, this sounds good. It is probably better to use a generic approach with an existing transformation than to add new PVs for special cases. In case a PV with a certain transformation is needed frequently, it can be still added, thought. I was just not sure if this is the case here.

miconda:
The format of the value can be toggled no matter the pv -- if encoded format, use decode transformation for counterpart value; if raw format use encode transformation. Adding both pvs is fine.

armenb:
I just pushed an amended commit for these 4 variables and their accompanying selects: $tls_{peer,my}_{raw,urlencoded}_cert.  The commit was amended to correct the variable description.

Note that I changed "escaped" to "urlencoded" since it is more specific in what type of escaping mechanism is being used.

@miconda @henningw please let me know what you think. Thanks!

miconda:
Thanks, I am merging it.

You have to add docs for the new vars at:

  * https://www.kamailio.org/wiki/cookbooks/devel/pseudovariables

armenb:
@miconda Thank you! I updated that wiki page.

miconda:
@armenb - there were some warnings that I got in macos compilation:

```
CC (gcc) [M tls.so]		tls_select.o
tls_select.c:765:7: warning: variable 'local' is used uninitialized whenever switch case is taken [-Wsometimes-uninitialized]
        case CERT_URLENCODED: urlencoded = 1; break;
             ^~~~~~~~~~~~~~~
tls_select.c:771:27: note: uninitialized use occurs here
        return get_ssl_cert(res, local, urlencoded, msg);
                                 ^~~~~
tls_select.c:764:7: warning: variable 'local' is used uninitialized whenever switch case is taken [-Wsometimes-uninitialized]
        case CERT_RAW: urlencoded = 0; break;
             ^~~~~~~~
tls_select.c:771:27: note: uninitialized use occurs here
        return get_ssl_cert(res, local, urlencoded, msg);
                                 ^~~~~
tls_select.c:759:11: note: initialize the variable 'local' to silence this warning
        int local, urlencoded;
                 ^
                  = 0
tls_select.c:763:7: warning: variable 'urlencoded' is used uninitialized whenever switch case is taken [-Wsometimes-uninitialized]
        case CERT_LOCAL: local = 1; break;
             ^~~~~~~~~~
tls_select.c:771:34: note: uninitialized use occurs here
        return get_ssl_cert(res, local, urlencoded, msg);
                                        ^~~~~~~~~~
tls_select.c:762:7: warning: variable 'urlencoded' is used uninitialized whenever switch case is taken [-Wsometimes-uninitialized]
        case CERT_PEER: local = 0; break;
             ^~~~~~~~~
tls_select.c:771:34: note: uninitialized use occurs here
        return get_ssl_cert(res, local, urlencoded, msg);
                                        ^~~~~~~~~~
tls_select.c:759:23: note: initialize the variable 'urlencoded' to silence this warning
        int local, urlencoded;
                             ^
                              = 0
```

I pushed a patch to initialize the variables:

  * https://github.com/kamailio/kamailio/commit/e2be8fe723ab98318816c18b59ffd086f7305040

But my guess that local and urlencoded have to be used as a combination, with both set, not only one set as it seems to be done in that switch. You know better the related code, check the commit and see if anything else needs to be done. Thanks!

Issue #2269, opened 2020-03-31
dialog: extract/dump a large amount


jchavanton:
### Description

The commands available to extract dialog information `dlg.profile_list` and `dlg.list`
Are not fast enough to output a large amount of dialogs.

After doing some benchmark the time is spent in in `rpc->struct_add`, `rpc->add etc`, etc.
it appears to be slowing down linearly.

### Possible Solutions

I made a quick POC to verify what I can get if I output to file instead and I got blazing fast performance, this is good news since it means that iterating over the hash table etc. is super fast.

https://github.com/jchavanton/kamailio/commit/f28c5a6b6364d79ee35acb5be439e57612ad67ed

```
 dialog: outputing large amount of dialogs to file

Test:
1) injecting 100K dialogs with voip_perf

2) confirmation that we have 100K+ dialogs
kamcmd dlg.profile_get_size outbound_call
100879

3) execute the quickly hacked modifications to output to /tmp/dialogs.txt
kamctl rpc dlg.profile_list outbound_call

3.1) verify how long it took
2020-03-31T22:28:18.398916476Z 12(31) : dialog [dialog.c:2174]: internal_rpc_profile_print_dlgs(): profile->lock
2020-03-31T22:28:18.554379411Z 12(31) : dialog [dialog.c:2190]: internal_rpc_profile_print_dlgs(): profile->lock release
// command took less than 20ms

3.2) verify the content of the output
wc   /tmp/dialogs.txt
  124355   117324 20547037 /tmp/dialogs.txt

Note about the content
// not everything is outputed in this quick hack but we can imagine that the impact will be very low.
call-id[uV.YzlhIZd1inE8SYDAlzmOLG12tYqVS]from[sip:voip_perf@147.75.65.147]to[sip:12060412757@52.42.12.21]h_entry[6207]h_id[44579]ref[2]contact[sip:voip_perf@147.75.65.147:5072;transport=udp]cseq[19528]
```

### Additional Information

I am thinking about adding commands to output to file in `json` and/or `csv`

PR #2272, opened 2020-04-02
dialog: adding dlg.dump

jchavanton:
alternative command to output dialogs data to file, far much faster than
dlg.list

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #2269 

#### Description
This command can output 100K dialogs in less than a second.

Almost the exact same format as dlg.list, except no using arrays for `variables` and `profiles`
since they can have a mixture of values and key / value pairs.

```
{
  "h_entry": 16383,
  "h_id": 4428,
  "ref": 2,
  "call_id": "ehm839mk6HwtIXfCpluxkNlbDq8H9im2",
  "from_uri": "sip:voip_perf@147.75.65.147",
  "to_uri": "sip:12063722016@52.42.112.121",
  "state": 4,
  "start_ts": 1585863821,
  "init_ts": 1585863821,
  "end_ts": 0,
  "timeout": 1585885421,
  "lifetime": 21600,
  "dflags": 1539,
  "sflags": 0,
  "iflags": 0,
  "caller": {
    "tag": "KK4KtrhQ--OjGMBNCvpXfKX-tinxQ3EK",
    "contact": "sip:voip_perf@147.75.65.147:5072;transport=udp",
    "cseq": "25945",
    "route_set": "",
    "socket": "udp:0.0.0.0:5060"
  },
  "callee": {
    "tag": "jlf1PLBMkC3utNwg-qwdkolYbaMx0no5",
    "contact": "sip:voip_perf@147.75.65.147:5060;transport=udp",
    "cseq": "0",
    "route_set": "",
    "socket": "udp:0.0.0.0:5060"
  },
  "profiles": {
    "outbound_call": ""
  },
  "variables": {
    "foo": "bar"
  }
}
```

### DIFF
--- a/src/modules/dialog/dialog.c
+++ b/src/modules/dialog/dialog.c
@@ -2197,6 +2197,118 @@ int mod_register(char *path, int *dlflags, void *p1, void *p2)
 
 /**************************** RPC functions ******************************/
 /*!
+ * \brief Helper method that outputs a dialog in a file
+ * \see rpc_dump_file_dlg
+ * \param dlg printed dialog
+ * \param output file descriptor
+ * \return 0 on success, -1 on failure
+ */
+static inline void internal_rpc_dump_file_dlg(dlg_cell_t *dlg, FILE* dialogf)
+{
+	dlg_profile_link_t *pl;
+	dlg_var_t *var;
+	srjson_doc_t jdoc;
+	srjson_t * jdoc_caller = NULL;
+	srjson_t * jdoc_callee = NULL;
+	srjson_t * jdoc_profiles = NULL;
+	srjson_t * jdoc_variables = NULL;
+
+	srjson_InitDoc(&jdoc, NULL);
+	jdoc.root = srjson_CreateObject(&jdoc);
+	if (!jdoc.root) {
+		LM_ERR("cannot create json\n");
+		goto clear;
+	}
+	srjson_AddNumberToObject(&jdoc, jdoc.root, "h_entry", dlg->h_entry);
+	srjson_AddNumberToObject(&jdoc, jdoc.root, "h_id", dlg->h_id);
+	srjson_AddNumberToObject(&jdoc, jdoc.root, "ref", dlg->ref);
+	srjson_AddStrToObject(&jdoc, jdoc.root, "call_id", dlg->callid.s, dlg->callid.len);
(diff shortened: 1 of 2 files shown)


jchavanton:
Hopefully we can also optimize JSON-RPC responses to support something that fast and use pipe-lining to make sure memory will never be a concern.

The dialog state can be quite valuable for platform developers, we can imagine many use cases.
I understand this ca already be use however it is not fast enough in some cases, seems like it could also be traffic impacting since it is holding a lock.

I think it would be a great improvement to make sure we get very fast and "reliable" JSON-RPC access to its state.

I understand 100K may seem to much, but in some scenarios "broken"  traffic can create leaking dialogs and extracting the dialog state is one good way to find out what is happening.

Also it is already slowing down a little bit at 10K and "may" impact traffic slightly.
I am not sure I would feel safe about building too much logic around doing many JSON-RPC requests to dialog.list on live servers. 

However with such a much faster version I do feel much better doing that.

miconda:
Is anything particular that jansson offers for this case compared with the internal srjson (which is based on cjson lib), see src/core/utils/srjson.{c,h} ?

Dialog module already uses srjson for profile storage and you can use srjson either with system malloc or internal memory.

jchavanton:
Missed this one, I guess I should use it.

miconda:
Yes, it is preferred to use the included srjson.{c,h} functions, because it doesn't introduce any external dependency.

jchavanton:
done

miconda:
Thanks, I am fine to merge it with one more remark.

Thinking a bit more about the rpc command name, I would suggest to change it to `dumpfile` or something else to suggest is writing to a file. Actually I would like to have something different that `dump` because other modules use this name to write back to rpc response. We have now a mixture between `list` and `dump` that write back to rpc response, adding a new `dump` with a different type of behaviour will make it incosistent. For example, if we add a similar command to usrloc for writing to a file, the `dump` command is already there. Let's add a name that can be (re-)used to other modules in the future, so we have some consistency from this perspective.

jchavanton:
I also hope we can find time to improve dlg.list to be much faster, this was a bigger project ...

jchavanton:
last modifications and docs tested once more before merging

jchavanton:
`dlg.dump` >> `dlg.dump_file` was not updated in the commit message but it was in the docs ...

Not a real concern.

PR #2273, opened 2020-04-05
siptrace flag fixes

grumvalski:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally


miconda:
You can go ahead and merge it.

I am thinking that it could make sense to add some dedicated functions to the TM module and TM API to get only the needed fields from a transaction, not to have to backup active transaction, lookup the one we need for checking and then restoring to the old one, as long as we do not need to do any other operation with the searched transaction. But this only as a future enhancement (the idea popped up a while ago, for some other parts of the code, but could be used here as well).

grumvalski:
Agree that such functions could be very useful in cases like this one.
Should I backport to 5.3?

PR #2278, opened 2020-04-08
siptrace: safety check for transaction when checking ACK

grumvalski:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
Testing with moderate load I had a crash in siptrace, in the newly added check for ACK for negative replies. Analyzing the core, the crash was caused by the invite transaction being 0x0 when t_gett is called after t_lookup_request. I was not able to reproduce it (how can t_lookup_request return 1 and then the transaction being 0x0?), but I've added a check to avoid crashing.


miconda:
It can be merged.

PR #2279, opened 2020-04-08
dialog: add dialog init, start end end timestamps to the output of dl…

grumvalski:
…g.briefing

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
IN some scenarios I found useful to have the timestamps (init, start, end) of the dialog as output of the dlg.briefing command.

### DIFF
--- a/src/modules/dialog/dialog.c
+++ b/src/modules/dialog/dialog.c
@@ -3075,7 +3075,7 @@ static void rpc_dlg_briefing(rpc_t *rpc, void *c)
 				rpc->fault(c, 500, "Failed to create the structure");
 				return;
 			}
-			if(rpc->struct_add(h, "ddSSSSSd",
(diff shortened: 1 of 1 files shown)


miconda:
Thanks! It can be merged.

However, for better flexibility, I was thinking to add an option parameter to this command to specify the fields to be returned in a string-of-char-flags manner. Have hash label, hash id and call-id always int attributes of the output, the the rest be specified in the parameter, like:

  * `f` - from uri
  * `F` - from tag
  * `t` - to uri
  * `T` - to tag
  * ...

Then use `kamcli rpc dialog.briefing "fFtT..."` -- the order of char-flags won't matter. When the parameter is not provided, a default value can be used for it (like what is printed now, before or after this PR).

grumvalski:
Nice idea, if I have time the next days I can start looking at it.

PR #2282, opened 2020-04-12
app_python3: update Python3 detection mechanism

micmac1:
This commit

  - removes the python calls whose output is never actually used.

  - changes the include discovery to use python3(.x)-config. This is
    preferable because it also works for cross-compiling. Calling
    python3(.x) directly will always provide host flags, which for
    cross-compiling is not feasible.

  - updates LDFLAGS discovery to also work with >= Python 3.8. To
    achieve this python3(.x)-config is first run with the argument
    "--embed". If this does not succeed (exit status 1)
    python3(.x)-config is run again without "--embed". This is the
    method suggested by Python upstream to provide backwards
    compatibility. See [1] for more details.

[1] https://docs.python.org/3/whatsnew/3.8.html

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Hi all,

While trying to get app_python3 going on OpenWrt I stumbled over this. We're cross-compiling and I got the wrong includes (host includes instead of target includes).

And then there was also the "-lpython3.x" missing because we're using python 3.8 which doesn't provide this linker flags by default anymore. So once I tried to load the module on the target device it threw an error:

```
Fri Apr 10 18:52:13 2020 daemon.err kamailio[15622]:  0(15622) ERROR: <core> [core/sr_module.c:512]: load_module(): could not open module </usr/lib/kamailio/modules/app_python3.so>: Error relocating /usr/lib/kamailio/modules/app_python3.so: PyTuple_New: symbol not found
Fri Apr 10 18:52:13 2020 daemon.err kamailio[15622]:  0(15622) CRITICAL: <core> [core/cfg.y:3543]: yyerror_at(): parse error in config file /etc/kamailio/kamailio.cfg, line 124, column 12-27: failed to load module
```

This pull requested fixes this.

Kind regards,
Seb

### DIFF
--- a/src/modules/app_python3/Makefile
+++ b/src/modules/app_python3/Makefile
@@ -11,18 +11,16 @@ NAME=app_python3.so
 # but no testing has been done with that.
 PYTHON3?=python3
 
-PYTHON3_VERSION=${shell ${PYTHON3} -c "import distutils.sysconfig;print(distutils.sysconfig.get_config_var('VERSION'))"}
-PYTHON3_LIBDIR=${shell ${PYTHON3} -c "import distutils.sysconfig;print(distutils.sysconfig.get_config_var('LIBDIR'))"}
-PYTHON3_LDFLAGS=${shell ${PYTHON3} -c "import distutils.sysconfig;print(distutils.sysconfig.get_config_var('LINKFORSHARED'))"}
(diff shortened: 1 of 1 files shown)


miconda:
Thanks!

PR #2289, opened 2020-04-17
tls: add sel for tls verified cert chain (requires OpenSSL 1.1+)

armenb:
* New Feature
* Changes Tested Locally

This sel variable allows a kam script to get access to not just the peer certificate (at index 0), but when a certificate is successfully verified, the entire chain of certificates that were used to verify the peer certificate (at index 1, 2, 3, etc).

This functionality is provided by OpenSSL's SSL_get0_verified_chain() function, which is only available in OpenSSL 1.1.0+ (which is why there is an #if for this feature)

This is important when a server trusts many CAs - without this addition, I don't think that it is definitively possible to tell which CA signed the verified certificate, leading to security issues if one of the trusted CAs was compromised and was used to sign certificates that look like they were signed by another issuing CA.

### DIFF
--- a/src/modules/tls/tls_select.c
+++ b/src/modules/tls/tls_select.c
@@ -840,7 +840,6 @@ static int get_verified_cert_chain(STACK_OF(X509)** chain, struct tcp_connection
 		goto err;
 	}
 
-	tcpconn_put(*c);
 	return 0;
 err:
 	tcpconn_put(*c);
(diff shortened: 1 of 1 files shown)


miconda:
Thanks, it will be merged!

Have you seen my comment on your previous PR:

  * https://github.com/kamailio/kamailio/pull/2268#issuecomment-609935161

PR #2295, opened 2020-04-22
siptrace: siptrace: fix memory leak in fake replies tracing

grumvalski:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

When tracing a faked reply (like 200 for CANCEL), the sip msg used to retrieve the tracing parameters is the transaction's uas one (uas.request) which might have the from header parsed or not, depending on the routing script actions (by default, if I'm not wrong, it is not parsed). If the header is not parsed, calling get_from in trace_onreply_out function will lead to a memory leak, beacause the header is parsed in pkg memory not freed when the transaction is deleted.

### DIFF
--- a/src/modules/siptrace/siptrace.c
+++ b/src/modules/siptrace/siptrace.c
@@ -1383,7 +1383,7 @@ static void trace_onreq_out(struct cell *t, int type, struct tmcb_params *ps)
 	}
 
 	if(sip_trace_msg_attrs(msg, &sto) < 0) {
-		return;
+		return; 
 	}
 
 	if(ps->send_buf.len > 0) {
@@ -1616,7 +1616,7 @@ static void trace_onreply_out(struct cell *t, int type, struct tmcb_params *ps)
 	}
 
 	if(sip_trace_msg_attrs(msg, &sto) < 0) {
-		return;
+		goto end;
 	}
 
 	if(faked == 0) {
@@ -1662,7 +1662,7 @@ static void trace_onreply_out(struct cell *t, int type, struct tmcb_params *ps)
 	sto.status.s = int2strbuf(ps->code, statusbuf, INT2STR_MAX_LEN, &sto.status.len);
 	if(sto.status.s == 0) {
 		LM_ERR("failure to get the status string\n");
-		return;
+		goto end;
 	}
 
 	memset(&to_ip, 0, sizeof(struct ip_addr));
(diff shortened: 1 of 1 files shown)


src/modules/siptrace/siptrace.c
```diff
@@ -1607,6 +1608,11 @@ static void trace_onreply_out(struct cell *t, int type, struct tmcb_params *ps)
 	if(msg == NULL || msg == FAKED_REPLY) {
 		msg = t->uas.request;
 		faked = 1;
+		/* check if from header has been already parsed.
+		 * If not we have to parse it in pkg memory and free iit at the end.
+		 */
+		if (msg->from && msg->from->parsed != NULL)
```
miconda, line 1614: Should the above condition have:

```
		msg->from->parsed == NULL
```

Otherwise it should be already parsed.

src/modules/siptrace/siptrace.c
```diff
@@ -1686,10 +1692,14 @@ static void trace_onreply_out(struct cell *t, int type, struct tmcb_params *ps)
...
 		LM_BUG("uriState must be either UNUSED or PARSED here! must be a bug! Message won't be traced!\n");
-		return;
+		goto end;
 	}
 
 	sip_trace_store(&sto, info->uriState == STRACE_PARSED_URI ? &info->u.dest_info : NULL, NULL);
+
+end:
+	if (faked && parsed_f) 
+		free_from(msg->from->parsed);
```
miconda, line 1702: At least for safety,  probably `msg->from->parsed` should be set back to NULL, not to be freed again in other places.

src/modules/siptrace/siptrace.c
grumvalski: you're right, I wrongly copied from local. I'm going to fix and force push.

miconda:
Looking at the function, I think two other `return` have to be replaced with `goto end`, lines 1619, 1665.

grumvalski:
Thank you Daniel and, again, you are right, I was working on 5.3 branch and didn't realize the new sip_trace_msg_attrs function. I'm pushing a fix. This also means that the patch cannot be backported to 5.3 as it is.

src/modules/siptrace/siptrace.c
```diff
@@ -1383,7 +1383,7 @@ static void trace_onreq_out(struct cell *t, int type, struct tmcb_params *ps)
 	}
 
 	if(sip_trace_msg_attrs(msg, &sto) < 0) {
-		return;
+		goto end; 
 	}
```
miconda, line 1387: This one is in another function -- mistake?

grumvalski: sorry, definitely not my day today...

miconda:
No worries, I had worse days. It is ok to merge. The patch doesn't look that big to be converted and applied for what 5.3 needs. Thanks.

PR #2304, opened 2020-04-28
usrloc: change "-" for "_" in stats name to be prometheus compliant

grumvalski:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

usrloc generated statistics (contacts, expires, users) contain "-" in their names (usrloc-contacs and so on), which is not valid according to prometheus naming convention.

### DIFF
--- a/src/core/cfg.lex
+++ b/src/core/cfg.lex
@@ -346,6 +346,7 @@ IP_FREE_BIND		ip_free_bind|ipfreebind|ip_nonlocal_bind
 
 PORT	port
 STAT	statistics
+STATS_NAMESEP	stats_name_separator
 MAXBUFFER maxbuffer
 SQL_BUFFER_SIZE sql_buffer_size
 CHILDREN children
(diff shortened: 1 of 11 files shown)


miconda:
Have anyone noticed use of `-` instead of `_` in other places/modules? If we do it, better doing it for all.

Then wondering if we should do this via some config option (maybe a global param) to set the separator char for these stats, so we have a way to let people use older name in case this change is going to break their exisiting monitoring apps.

grumvalski:
I've spot the same usage in p_usrloc and ims_usrloc_pcscf modules (they both derived from usrloc).
I didn't notice any other module building "derivate" stats (as those), so I'm not sure that implementing a global param is worth in this case.

miconda:
With the global parameter I was thinking of the cases when someone has already a monitoring/admin tool using existing stats names, practically this change will break them, requiring to update them as well, which is not always easy.

What you can do is to add a define in `core/config.h`, like:

```
#define KSR_STATS_NAMESEP '_'
```

Use it across those modules/patches and later I can add a global config param to replace it, if considered useful.

miconda:
It can be merged. Then I will add the global parm to set this char. The question is what to make the default, the old dash/minus `-` (for backwards compatibility); or underscore `_`, the new proposal in this PR.

sergey-safarov:
my vote is use `Prometheus` format "_".

grumvalski:
I can find some time in the next days to implement the global param. I'd also go for prometheus format by default.

miconda:
If you want to merge it now, I can add quickly what is needed to have the global parameter in the interpreter, as I am familiar with lex/yacc and then you can do further adjustments, if you need. Waiting that you do everything is more than fine for me, as well.

miconda:
@grumvalski - I think it makes no sense to added to the `cfg_group_core / default_core_cfg` structure, because that is for parameters that can be changed at runtime. But, imo, this is not set once and actually use only during startup, when the stats names are exported.

It can simply be a global variable like `int ksr_tcp_accept_hep3=0;`. It can be added to src/core/counters.c and exposed with extern in src/core/globals.h.

The cfg_group_* structures work as a block and making them larger without a reason should be avoided.

In the future probably we should create some structures for global attributes grouped per purpose, so instead of many individual global variables, there will be a few global structure.

As a side node, I would recommend using the `ksr_` prefix for global variables in the core, recently I encounter some issues with global variables being cross visible between object files, so db_mode of permissions module was inherited by usrloc module. Apparently some compilers (or operating systems) build libraries with visibility for all symbols, without adding any prefix.

grumvalski:
Right, I realized after doing it. I've pushed the new suggest implementation.

miconda:
OK to merge it. Do not forget to add docs in the core cookbook wiki.

grumvalski:
Thanks Daniel, documentation added to the wiki.

PR #2305, opened 2020-04-28
core: do not use tcp id for lookup if not needed

dsciarra:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [X] Tested changes locally

#### Description
In case the first lookup (with tcp id) is not successful, the second
attempt should have been performed without a tcp id.
Issue was introduced with dc43750644 (new global parameter
tcp_connection_match) in 5.3.


miconda:
Thanks!

Issue #2300, opened 2020-04-24
ndb_redis: redis_cmd() doesn't return error with a faulty command


linuxmaniac:
### Description

Detected that a command like this was not returning <0 
```
if(redis_cmd("test", "HSET dd field1", "r")) {
  [...]
}
```
Same command on redis-cli:
```
# redis-cli
127.0.0.1:6379> HSET dd field1
(error) ERR wrong number of arguments for 'hset' command
```

#### Reproduction

I've added some debug to the code and I find out that the command is returning:
```
Apr 23 10:55:02 spce proxy[12880]: DEBUG: ndb_redis [redis_client.c:956]: redisc_exec(): rpl->rplRedis->type:3
```
That's an INTEGER. From hiredis/read.h:
```
#define REDIS_REPLY_INTEGER 3
#define REDIS_REPLY_NIL 4
#define REDIS_REPLY_STATUS 5
#define REDIS_REPLY_ERROR 6
```
I've created a simple test code to check the same command:
```
#include <stdio.h>
#include <stdarg.h>
#include "hiredis.h"

void check(redisContext *c, const char *cmd, ...) {
        va_list ap;
        va_start(ap,cmd);
        redisReply *r = redisvCommand(c, cmd, ap);
        va_end(ap);

        if(r == NULL) {
                printf("reply NULL\n");
        }
        printf("reply->type:%d\n", r->type);
        freeReplyObject(r);
}

int main(void) {
        redisContext *c = redisConnect("127.0.0.1", 6379);

        if (c == NULL || c->err) {
            if (c) {
                printf("Error: %s\n", c->errstr);
            } else {
                printf("Can't allocate redis context\n");
            }
            return 1;
        }
        check(c, "HSET dd field");
        redisFree(c);
        return 0;
}
```
```
root@spce:/usr/local/devel# cc t.c -o t -lhiredis -I/usr/include/hiredis
root@spce:/usr/local/devel# ./t
reply->type:6
```
No problems there. The reply is ERROR.

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
version: kamailio 5.3.3 (x86_64/linux)
[...]

PR #2314, opened 2020-05-05
ndb_redis: redis_cmd() check reply type to detect command errors

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #2300

#### Description

from https://github.com/redis/hiredis#using-replies

> The standard replies that redisCommand are of the type redisReply. The type field in the redisReply should be used to test what kind of reply was received:
> 
> - REDIS_REPLY_STATUS:
>   The command replied with a status reply. The status string can be accessed using reply->str. The length of this string can be accessed using reply->len.
> 
>  - REDIS_REPLY_ERROR:
>     The command replied with an error. The error string can be accessed identical to REDIS_REPLY_STATUS.

adding checks for the reply type to detect command errors

### DIFF
--- a/src/modules/ndb_redis/redis_client.c
+++ b/src/modules/ndb_redis/redis_client.c
@@ -950,6 +950,14 @@ int redisc_exec(str *srv, str *res, str *cmd, ...)
 			goto error_exec;
 		}
 	}
+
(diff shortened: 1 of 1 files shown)


miconda:
It can be merged and backported if needed.

PR #2336, opened 2020-05-26
dispatcher: add attrs param to rpc 'add' call

grumvalski:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Expands the rpc 'add' command allowing to specify the "attrs" body.

### DIFF
--- a/src/modules/dispatcher/dispatcher.c
+++ b/src/modules/dispatcher/dispatcher.c
@@ -1800,15 +1800,16 @@ static void dispatcher_rpc_add(rpc_t *rpc, void *ctx)
 {
 	int group, flags;
 	str dest;
+  str attrs;
 
 	flags = 0;
 
-	if(rpc->scan(ctx, "dS*d", &group, &dest, &flags) < 2) {
+	if(rpc->scan(ctx, "dS*d", &group, &dest, &flags, &attrs) < 3) {
 		rpc->fault(ctx, 500, "Invalid Parameters");
 		return;
 	}
 
-	if(ds_add_dst(group, &dest, flags) != 0) {
+	if(ds_add_dst(group, &dest, flags, &attrs) != 0) {
 		rpc->fault(ctx, 500, "Adding dispatcher dst failed");
 		return;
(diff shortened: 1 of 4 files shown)


henningw:
Thanks, it looks good to me. I think it can be merged.

grumvalski:
Thanks Henning.

src/modules/dispatcher/dispatcher.c
```diff
@@ -1800,15 +1800,16 @@ static void dispatcher_rpc_add(rpc_t *rpc, void *ctx)
 {
 	int group, flags;
 	str dest;
+  str attrs;
 
 	flags = 0;
 
-	if(rpc->scan(ctx, "dS*d", &group, &dest, &flags) < 2) {
+	if(rpc->scan(ctx, "dS*d", &group, &dest, &flags, &attrs) < 3) {
```
miconda, line 1807: I think here is missing and extra `S` in scan's 2nd parameter.

src/modules/dispatcher/dispatcher.c
```diff
@@ -1800,15 +1800,16 @@ static void dispatcher_rpc_add(rpc_t *rpc, void *ctx)
 {
 	int group, flags;
 	str dest;
+  str attrs;
```
miconda, line 1803: Fix also the indentation here, since the `S` needs to be added below.

src/modules/dispatcher/dispatcher.c
miconda: The above condition needs some update. Because attrs is optional, then it is still fine to go on with return code 2 (which represents the number of items/params scanned).

Probably return code has to be stored in a variable, e.g., then if ret<2 is error, if ret==2 then attrs.s=NULL and attrs.len=0 (or attrs=STR_NULL when declared).

grumvalski: Right, I'm going to fix.

PR #2337, opened 2020-05-26
$xavi(...) 

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This change introduces a new type of variable $xavi - eXtended Attribute Value Insensitive case.
It's like a $xavp but keys are case insensitive:

```
$xavi(WhatEver=>fOo) = 1;

if($xavi(whatever[0]=>foo) == 1) {
....
}
```

### DIFF
--- a/src/core/pvar.h
+++ b/src/core/pvar.h
@@ -77,8 +77,8 @@ enum _pv_type {
 	PVT_DSTURI,           PVT_COLOR,             PVT_BRANCH,
 	PVT_FROM,             PVT_TO,                PVT_OURI,
 	PVT_SCRIPTVAR,        PVT_MSG_BODY,          PVT_CONTEXT,
-	PVT_XAVP,             PVT_XAVU,              PVT_HDRC,
-	PVT_OTHER,            PVT_EXTRA /* keep it last */
+	PVT_XAVP,             PVT_XAVU,              PVT_XAVI,
+	PVT_HDRC,             PVT_OTHER,             PVT_EXTRA /* keep it last */
 };
 
 typedef enum _pv_type pv_type_t;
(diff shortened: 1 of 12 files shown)


miconda:
I could not spot anything on a quick look over the patch. If nobody else wants to still keep it here for review, I am fine to merge it.

miconda:
I prefer to have separate functions, instead of going to some generic void* callbacks. Those are hard to track and prone to introduce bugs in existing functional code. The duplicated code should not be really big in size and having new dedicated functions for comparison, etc ... will result also in extra size of the overall code.

I actually looked at reusing the xavp code when I implementing xavu recently and the code got a lot of messy IF-ELSE conditions everywhere, functions with extra parameters to indicate what type of variable, etc ... so I reverted and just made dedicated functions, everything afterwards being easier to follow.

henningw:
I don't think the usage of function pointers is a big problem, was just an idea. If you need to pass to many "switch" type of variables, it gets hard to follow, sure.
The two functions xavi_add_value, xavi_add_value_after just operate on a global static list. This could be implemented type safe by just passing the list to it, without the need of void pointers.

miconda:
In the case of the two functions you mention, then `_xavi_list_crt` is hidden in the c file (being declared static, since it is controlled in very special cases related transaction states, ...), so if those functions are used outside of the file, then they still need  wrappers to call a common function (for xavp and xavi) with the adequate list.

And my remark was targeting exactly this case, because the functions have like 10 effective lines of code. A wrapper function is like 3 lines of code, with a jump on the stack. So overall it is not a significant gain.

It makes sense to have a function for a small number of lines of code, if the function is used in a lot of places, but if it going to be used in 2-3 places, then it has to be a significant number of codes of lines to really worth it.

Of course, it can be seen as a personal preference, but I prefer code easier to follow/search/maintain instead of modularising every snippet of code.

PR #2341, opened 2020-05-30
core: PV_IDX_NONE to point that pv had no index

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

if index is 0 there were no way to know if the index was there

$x_hdr(A) = "value" needs to append a value
$(x_hdr(A)[0]) = "value" needs to set the value

pv_parse_spec(str *s, pv_spec_p p) was always setting p->pvp.pvi.type
to 0 == PV_IDX_INT

### DIFF
--- a/src/core/pvapi.c
+++ b/src/core/pvapi.c
@@ -857,6 +857,7 @@ char* pv_parse_spec2(str *in, pv_spec_p e, int silent)
 	tr = 0;
 	pvstate = 0;
 	memset(e, 0, sizeof(pv_spec_t));
+	e->pvp.pvi.type = PV_IDX_NONE;
 	p = in->s;
 	p++;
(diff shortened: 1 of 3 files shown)


linuxmaniac:
If no objections, I will merge this PR tomorrow

henningw:
No objections, just one question - is my assumption correct that this will not change the PV behaviour for existing ***common*** use cases?

linuxmaniac:
> No objections, just one question - is my assumption correct that this will not change the PV behaviour for existing _**common**_ use cases?

I didn't spot any other problems a part of the ones fixed at 9ac856fc3794699c09717264694956f2f3f148f7

henningw:
Ok, thank you.

PR #2342, opened 2020-06-01
dispatcher: don't reset load table when adding/removing a destination…

grumvalski:
… or reloading the list

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

When a destination is added/removed or the whole list is reloaded, the calls that are being routed to another destination will fail.
This because ds_load_replace is returning an error not finding in the table the entry for the call being rerouted.
Another side effect of this is that when adding/removing destinations the actual load of the endpoints will be wrong since the current calls will never be deleted.
I've removed the calls to ds_clean_slots and modified ds_load_replace to not exit when the old destination is not found (the endpoint could have been removed after the failover list had been created) and to return a new error code (-2) when the new destination is not found (for the same reason).
Finally I've modified ds_update_dst to skip the current selected destination and select the next if ds_load_replace returns -2.

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -878,7 +878,7 @@ next_line:
 	/* Update list - should it be sync'ed? */
 	_ds_list_nr = setn;
 	*crt_idx = *next_idx;
-	ds_ht_clear_slots(_dsht_load);
+	
 	ds_log_sets();
 	return 0;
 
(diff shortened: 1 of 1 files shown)


miconda:
Thanks, it can be merged!

grumvalski:
Thank you Daniel, going to merge.
I also think it should be backported, adapted because there is no rpc add/remove command, to stable branches.

henningw:
It was cherry-picked into 5.3 and manually backported (with the one depending commit) to 5.2.

PR #2358, opened 2020-06-12
dmq_usrloc: add new parameter `usrloc_delete` to disable synchronizing delete actions

eschmidbauer:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
in high volume systems with ephemeral user location data (e.g. very low expiration times on user location records), the user location delete action can cause a high amount of DMQ traffic when UAs unregister. This new parameter allows the user location `delete` action to be disabled from being shared across DMQ nodes which helps reduce DMQ traffic on systems with high register/unregister rates and low expiration times.

### DIFF
--- a/src/modules/dmq_usrloc/dmq_usrloc.c
+++ b/src/modules/dmq_usrloc/dmq_usrloc.c
@@ -42,6 +42,7 @@ int _dmq_usrloc_batch_msg_contacts = 1;
 int _dmq_usrloc_batch_msg_size = 60000;
 int _dmq_usrloc_batch_usleep = 0;
 str _dmq_usrloc_domain = str_init("location");
+int _dmq_usrloc_delete = 1;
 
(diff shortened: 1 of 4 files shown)


abalashov:
Very nice feature, cheers for that! 

But just to be clear, how do the contact bindings get aged out, in the absence of delete operations pushed across DMQ? Are they subject to the same local `usrloc` expiration logic in that case?

eschmidbauer:
Yes, each kamailio server will remove contact records as they expire. In our use case, they expire within 60s on each kamailio server.

abalashov:
And this happens via the normal usrloc timer processes, right?

eschmidbauer:
Yes, exactly.

abalashov:
That really is an excellent feature, thank you for putting it together!

miconda:
Thanks!

Issue #2361, opened 2020-06-16
Dispatcher : Re-distribution of messages when hash used and destination not available


Kalki70:
# Description

Selected algorithm is a hash, over Call-ID, From, etc. If destination is not available, a new one must be selected, without breaking the hash assignation (all messages with the same hash must be forwarded to the same destination)

### Expected behavior

The load to the failed destination to be (hopefully) evenly distributed over the remaining destinations.

#### Actual observed behavior

All the messages that were supposed to be sent to the failed destination are sent to only one of the remaining, causing it to receive double load, comparing with the rest

#### Debugging Data

In this test, 6000 Calls were sent to a group of six destinations, using algorithm "0" (hash over Call-ID)

  | All destinations active
-- | --
Dest1 | 996
Dest2 | 997
Dest3 | 999
Dest4 | 1006
Dest5 | 1005
Dest6 | 997
Total | 6000

![all_active](https://user-images.githubusercontent.com/20492105/84825429-6719a580-afef-11ea-971c-d59152fc6550.png)

Repeating the test, but with destination Dest4 down:

  | OriginalImplementation
-- | --
Dest1 | 975
Dest2 | 993
Dest3 | 2022
Dest4 | 0
Dest5 | 1019
Dest6 | 991
Total |  6000

![one_down_original](https://user-images.githubusercontent.com/20492105/84826132-8107b800-aff0-11ea-9663-f6e8992ed5e2.png)

We can see that poor destination Dest3 receives double traffic

### Possible Solutions

I implemented a solution that does not break the expected behavior (messages with same hash are assigned the same destination, but with a better distribution over remaining destinations)

  | New Implementation
-- | --
Dest1 | 1175
Dest2 | 1206
Dest3 | 1191
Dest4 | 0
Dest5 | 1216
Dest6 | 1212
Total | 6000

![one_down_new](https://user-images.githubusercontent.com/20492105/84826051-62092600-aff0-11ea-8f89-f06102e732d5.png)

Basically, the solution is as simple as to execute a rehash of the original hash. 

I will create a pull request with the proposed change, which I already tested, as it can be seen on the previous chart.

### Additional Information

  * **Kamailio Version** 

```
 kamailio 5.4.0-dev4 (x86_64/linux) 0c29e8-dirty
```

* **Operating System**:

```
[...]

PR #2362, opened 2020-06-16
dispatcher : better distribution when using hash and destination is not available

Kalki70:
#### Pre-Submission Checklist

- [ X] Commits are split per component (core, individual modules, libs, utils, ...)
- [ X] Each component has a single commit (if not, squash them into one commit)
- [ X] No commits to README files for modules (changes must be done to docbook files

#### Type Of Change

#### Checklist:

- [ X] Tested changes locally
- [ X] Related to issue #2361

#### Description
When a destination is selected using one of the hash algorithms (over Call-ID, From, etc.)
and destination is not available, then the next is selected. This causes the following problem :
All messages to the unavailable destination will be sent to the same unique destination.
This will cause that one destination will receive double load and the rest will continue with the same.

Solution : To better distribute messages, while also guaranteeing the same distribution of the hash, a re-hash of the hash is made.

As the rehash could select the same or other failed destination, it can rehash several times. The number of rehashes made is controlled by :

modparam("dispatcher", "ds_rehash_max", -1)

Which can take the following values :

-1 : The total count of destinations is used 
0 : Disabled. Reverts to original implementation.
Positive number. Maximum number of rehashes.

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -79,22 +79,6 @@
 #define DS_ALG_RELWEIGHT 11
 #define DS_ALG_PARALLEL 12
 
-/* increment call load */
-#define DS_LOAD_INC(dgrp, didx) do { \
-		lock_get(&(dgrp)->lock); \
-		(dgrp)->dlist[didx].dload++; \
-		lock_release(&(dgrp)->lock); \
-	} while(0)
-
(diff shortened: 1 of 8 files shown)


Kalki70:
I dont' knwo how.... my pull request has many changes more than those that I did. Maybe my fork became out of sync when I created the pull request. Cancelling pull request.

PR #2365, opened 2020-06-19
registrar: expires_range should not make expires < min_expires

khoegh:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

expires_range should not make expires smaller than min_expires.
A valid value might be made to small, which triggers 423 when min_expires_mode is set to 1


khoegh:
When using min_expires=<min>, expires_range=<range> and min_expires_mode=1, the client might send REGISTER containing expires equal to min_expires.
expires_range make expires smaller than min_expires which triggers a 423.
As save("location") fails, the default config-file also calls sl_reply_error(), which sends 500 after 423

miconda:
Thanks!

PR #2368, opened 2020-06-23
Add some missing build requirements

guss77:
I'm building in Docker containers, and with the minimal CentOS/Fedora/whatever image you get from official repos, builds will fail without installing these.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/pkg/kamailio/obs/kamailio.spec
+++ b/pkg/kamailio/obs/kamailio.spec
@@ -272,6 +272,7 @@ Conflicts:  kamailio-xhttp-pi < %ver, kamailio-xmlops < %ver
 Conflicts:  kamailio-xmlrpc < %ver, kamailio-xmpp < %ver
 Conflicts:  kamailio-uuid < %ver
 BuildRequires:  bison, flex
+BuildRequires:  which, make, gcc, gcc-c++
 %if 0%{?rhel} != 6
 Requires:  systemd
 BuildRequires:  systemd-devel


miconda:
Not having the commit first line as per contributing guide, I added those build requirements as part of commit 71248a2f7394b69c85f030f103afb2a8f6a5a49e, along with other changes.

If any issue introduced (eg., I added also pkgconfig which makes it easier to detect path to libs and include files), open an issue, or make a PR,  but follow the contributing guidelines so the PR can be merged from here using the web ui:

sergey-safarov:
Hi Daniel
Think this change
```
%if 0%{?rhel} || 0%{?centos_ver}
%if 0%{?rhel} > 6 || 0%{?centos_ver} > 6
BuildRequires:  pkgconfig
%endif
%else
BuildRequires:  pkgconfig
%endif
```
Is equal to
```
%if 0%{?rhel}
%if 0%{?rhel} > 6
BuildRequires:  pkgconfig
%endif
%else
BuildRequires:  pkgconfig
%endif
```
And equal to
```
%if 0%{?rhel} > 6
BuildRequires:  pkgconfig
%else
BuildRequires:  pkgconfig
%endif
```
As `pkgconfig` is present in all RPM supported dists think we can use
```
BuildRequires:  pkgconfig
```

What you think?

miconda:
Not having CentOS/RHEL 6 around, I googled a bit and couldn't find it for sure that they ship with pkgconfig. If yes, go ahead and remove that condition and add pkgconfig to the BuildRequires listing make, gcc, etc... Thanks!

sergey-safarov:
yes, exists
```sh
[safarov@safarov-dell kamailio-upstream]$ docker run -it centos:6
[root@c8b81ef4309e /]# pkg-config --version
0.23
```
I make changes

Issue #2308, opened 2020-04-28
set_contact_alias() issue with clients adding it to their Contact


mtryfoss:
I'm using set_contact_alias() to fix nat registrations, in this case with a Bria client.
Somehow Bria learns that I'm using this - probably through the Contact in the 200 OK reply of save(). Bria then adds this to the Contact when replying to incoming INVITE.

After some time, something happens in the network at client side and the connection gets a new port through the router. Bria does not notice and still sends the old alias-parameter with the port not used anymore. When I'm calling handle_ruri_alias(), Kamailio reads this instead of the actual port and the ACK to 200 OK is sent to a port not used anymore.

This can probably be solved by either, in preferred order:
- rewrite set_contact_alias() and add_contact_alias() to replace existing alias parameters 
- regexp remove of alias parameter from Contact of all messages coming from clients
- regexp remove of alias parameter from Contact returned by save()

Br,
Morten

miconda:
set_contact_alias() should not be used for registrations, but for requests that are typically forwarded. If you do it on an edge proxy between device and registrar server, then you should use path module and leave the contact unchanged.

If you use it in the registrar server config, then you should replace it with fix_nated_register(), like in the default kamailio.cfg.

Otherwise, if you really want to use it, then in case you have to deal with such UA, then yes, you will have to remove (or replace) the alias parameter in Contact. Probably you can play with the replace/subst functions from textops and if you can't get it, then new C code has to be added. A pull request is of course appreciated.

mtryfoss:
Thank you for your reply.

Does the manual somewhere state that set_contact_alias() should not be used for REGISTER?
Actually, some examples suggest using this instead of fix_nated_contact()  - http://www.evaristesys.com/blog/server-side-nat-traversal-with-kamailio-the-definitive-guide/

The reason for using set_contact_alias() instead of fix_nated_contact() is that one client seemed to have an issue with the contact being changed.

I'm actually using in on an edge proxy and already using Path,  but I guess I missed setting this on the edge: modparam("path", "use_received", 1)

Fixed it already temp with a subst on the reply for REGISTERs, as you wrote.

pafels:
Would you please share how you fixed it with subst?

mtryfoss:
All done on edge proxy. I'm not a regex-pro.
It will not work if done locally on the registrar (the node doing the save()).

https://gist.github.com/mtryfoss/9a0811dbc4462b91afe87d7dd5a210e8

[...]

PR #2390, opened 2020-07-03
nathelper: add optional set_contact_alias([trim]) parameter

eschmidbauer:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #2308 

#### Description
add support for trimming an already existing alias parameter when calling `set_contact_alias()`

### DIFF
--- a/src/modules/nathelper/nathelper.c
+++ b/src/modules/nathelper/nathelper.c
@@ -110,6 +110,7 @@ static int fix_nated_contact_f(struct sip_msg *, char *, char *);
 static int add_contact_alias_0_f(struct sip_msg *, char *, char *);
 static int add_contact_alias_3_f(struct sip_msg *, char *, char *, char *);
 static int set_contact_alias_f(struct sip_msg *msg, char *str1, char *str2);
+static int w_set_contact_alias_f(struct sip_msg *msg, char *str1, char *str2);
 static int handle_ruri_alias_f(struct sip_msg *, char *, char *);
 static int pv_get_rr_count_f(struct sip_msg *, pv_param_t *, pv_value_t *);
 static int pv_get_rr_top_count_f(struct sip_msg *, pv_param_t *, pv_value_t *);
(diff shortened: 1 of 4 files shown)


src/modules/nathelper/nathelper.c
```diff
@@ -704,10 +707,12 @@ static int fix_nated_contact_f(struct sip_msg *msg, char *str1, char *str2)
  * Replaces ip:port pair in the Contact: field with the source address
  * of the packet.
  */
-static int set_contact_alias(struct sip_msg *msg)
+static int set_contact_alias(struct sip_msg *msg, int trim)
```
miconda, line 710: By changing the signature of `set_contact_alias()` you broke the kemi export of it.

eschmidbauer: > By changing the signature of `set_contact_alias()` you broke the kemi export of it.

sorry i haven't worked with the kemi export code yet. i pushed another commit which i _think_ fixes it

miconda:
Not really fixing it, because now set_contact_alias() from kemi require a parameter always, so existing kemi deployement will be broken on upgrade. A new function can be exported to kemi (e.g., set_contact_alias_trim()) and the old one should stay the same. So practically a new wrapper function has to be created to cope with the two cases.

eschmidbauer:
> Not really fixing it, because now set_contact_alias() from kemi require a parameter always, so existing kemi deployement will be broken on upgrade. A new function can be exported to kemi (e.g., set_contact_alias_trim()) and the old one should stay the same. So practically a new wrapper function has to be created to cope with the two cases.

thanks for the guidance on that.
 i think latest commit takes care of it correctly.

miconda:
Pretty much, I would only suggest to use `ki_` prefix for the kemi functions, like the other wraperss which have a specific prefix for exports to kemi (e.g., the last three in the kemi exports structure).

miconda:
If nobody else will have opinions against, then I will merge this PR before branching 5.4

PR #2398, opened 2020-07-13
rtp_media_server: update to mediastreamer 4.4.0

jchavanton:
#### Type Of Change
- [x] Updated build examples, tested with recent version of dependencies

Updated documentation to simplify usage using docker

### DIFF
--- a/src/modules/rtp_media_server/Makefile
+++ b/src/modules/rtp_media_server/Makefile
@@ -6,7 +6,7 @@ DEFS+=-I$(LOCALBASE)/lib
 
 ORTPLIBS=-lortp
 BCUNITLIBS=-lbcunit
-MS2LIBS=-lmediastreamer_voip -lmediastreamer_base
+MS2LIBS=-lmediastreamer
 
 LIBS=$(ORTPLIBS) $(BCUNITLIBS) $(MS2LIBS)
 DEFS+=-DKAMAILIO_MOD_INTERFACE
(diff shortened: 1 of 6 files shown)


miconda:
Nice that now depends on a release version, not development references. Maybe I find some time soon to give it a spin.

It can be merged from my point of view.

jchavanton:
Thanks for having a quick look, just tested working fine.

I think I will add the missing bits to make sure the image is easy to use, the following steps are still required :

```
#!/bin/sh
DIR_PREFIX=`pwd`
CONTAINER=rtp_media_server
IMAGE=jchavanton/rtp_media_server:latest
docker run -d --net=host --name=${CONTAINER} ${IMAGE} /bin/sh -c "tail -f /dev/null"
```

```
sudo docker exec -it rtp_media_server bash

cp git/kamailio/src/modules/rtp_media_server/config_example/kamailio.cfg /etc
// set the IP 
kamailio -m 64 -D -dd -f /etc/kamailio.cfg
```

Next, I am tempted to make a second attempt to use the dialog module, the new dlgs module does not expose an API yet, any toughts on trying to use/extend dlgs or dialog ?

miconda:
This should be pushed to 5.4, since it is about dependency on a stable release of external library and the code was not touched. I can do it before releasing 5.4.0.

Regarding the dialog vs dlgs: dlgs is in very early stage and for the moment targets offering statistics about active calls and a ability to group/tag them and count active dialogs on various criteria, without depending on tm module callbacks (or in other words, be as much as possible stateless). Therefore maybe dialog is better choice now, unless you want to work on dlgs module and add the new features you need.

jchavanton:
Thanks for the answer, I think dependencies on TM already needed in this case, I will favor the dialog module on my next attempt.
I only  merged into master

PR #2399, opened 2020-07-14
p_usrloc: reworking mdb_availability_control

lbalaceanu:
- simplify general implementation
- avoid deadlock due to a process trying to acquire same lock twice

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

On a timer, checking the availability of the master db used for writing and setting a status variable accordingly. In case the mdb_availability_control option is set, this status variable is consulted with no synchronization by each process when trying to do a write to the master db. As error mitigation, in case the per process handle associated with master db is NULL, p_usrlock recreates it. In case the per process handle is not NULL, Kamailio will try to recreate a broken connection by itself.

### DIFF
--- a/src/modules/p_usrloc/p_usrloc_mod.c
+++ b/src/modules/p_usrloc/p_usrloc_mod.c
@@ -185,7 +185,7 @@ str default_db_type   = str_init(DEFAULT_DB_TYPE);
 str domain_db         = str_init(DEFAULT_DOMAIN_DB);
 int default_dbt       = 0;
 int expire            = 0;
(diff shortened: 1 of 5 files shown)


lbalaceanu:
Hi,

If there are no comments, I will later on merge the pull requests.

@miconda, should I also backport this to 5.3, 5.4 since this also fixes a previously possible deadlock due to acquiring a lock twice?

Thank you,
Lucian

miconda:
Yes, should be backported if it fixes issues.

lbalaceanu:
Thank you.

PR #2422, opened 2020-08-01
acc: add to-tag for locally generated replies

jchavanton:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
When doing t_reply/send_reply stateful, the ACC module will not add the to-tag to the ACC events.
This patch will fix this and ensure that we get the to-tag that was locally generated used in the reply.

example :

```
modparam("acc", "acc_prepare_always", 1)
modparam("acc_json", "acc_missed_flag", 6)

failure_route[test] {
    setflag(6); // MISSED_ACC, the failure will be recorded by ACC
    t_flush_flags(); // This will set the flags in the newly created transaction
    t_reply("500", "Service Unavailable");
}
```
This does generate an ACC event, however the to-tag is not populated, my guess this is because the TM callback is done before the to-tag is generated.

This patch will populate the to-tag immediately.

### DIFF
--- a/src/modules/acc/acc_logic.c
+++ b/src/modules/acc/acc_logic.c
@@ -79,8 +79,7 @@ struct acc_enviroment acc_env;
 static void tmcb_func( struct cell* t, int type, struct tmcb_params *ps );
 
 
-static inline struct hdr_field* get_rpl_to( struct cell *t,
-														struct sip_msg *reply)
+static inline struct hdr_field* get_rpl_to( struct cell *t, struct sip_msg *reply)
 {
 	if (reply==FAKED_REPLY || !reply || !reply->to)
 		return t->uas.request->to;
@@ -88,12 +87,16 @@ static inline struct hdr_field* get_rpl_to( struct cell *t,
 		return reply->to;
 }
 
+void env_set_totag(struct cell *t, struct sip_msg *reply)
+{
+	if (reply==FAKED_REPLY || !reply || !reply->to)
+		tmb.t_get_reply_totag(t->uas.request, &acc_env.to_tag);
+}
 
 static inline void env_set_to(struct hdr_field *to)
 {
 	acc_env.to = to;
 }
-
 
 static inline void env_set_text(char *p, int len)
 {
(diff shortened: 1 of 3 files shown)


src/modules/acc/acc_logic.c
```diff
@@ -79,21 +79,29 @@ struct acc_enviroment acc_env;
...
 
 static inline void env_set_to(struct hdr_field *to)
 {
 	acc_env.to = to;
 }
 
+// static inline void env_set_totag(str *to_tag)
+// {
+// 	acc_env.to_tag = to_tag;
+// }
```
miconda, line 104: I guess this part can be removed if not intended to be used.

src/modules/acc/acc_logic.c
```diff
@@ -372,6 +380,9 @@ void acc_onreq( struct cell* t, int type, struct tmcb_params *ps )
 	int tmcb_types;
 	int is_invite;
 
+	LM_ERR("is_mc_on: %d\n", is_mc_on(ps->req));
+	LM_ERR("is_acc_on: %d\n", is_acc_on(ps->req));
+	LM_ERR("is_acc_prepare_on: %d\n", is_acc_prepare_on(ps->req));
```
miconda, line 385: There are couple of error messages which I guess are only for testing purposes. Before merging should be removed. I could spot in two other places, besides the three here.

jchavanton:
Sorry, about the delay, I left it unfinished before taking a long week-end off, I will push adjustments tomorrow

miconda:
Still working on it, or it is finished from your point of view?

jchavanton:
it is working, but as I integrate it in a complicated routing configuration, I may find more little modifications.

Just found one ...

We could wait a few more days.

jchavanton:
Would be good if I could also improve this to always have the real reason phrase sent.
Having a look at the options.

```
acc_env.code = code;
if (reply==FAKED_REPLY || reply==NULL) {
        /* code */
        acc_env.code_s.s =
                int2bstr((unsigned long)code, code_buf, &acc_env.code_s.len);
        /* reason */
        acc_env.reason.s = error_text(code);                                  
        acc_env.reason.len = strlen(acc_env.reason.s);
```

jchavanton:
not expecting more modifications based on testing.
ready to review/merge

miconda:
Thanks!

PR #2446, opened 2020-08-18
Fix missing "Route" header name in usrloc keepalive msgs

gedia:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

The keepalive functionality of the usrloc module seems to produce invalid messages when a location contact has Path information set. This should fix the issue.


miconda:
Your patch is breaking keepalives for contacts without path. I pushed a different patch to fix the issue, pushed to master and 5.4 branches. Test and see if works fine now. Thanks!

PR #2470, opened 2020-09-06
uac: contact_addr field added into db uacreg table  

ovoshlook:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
uac: contact_addr field added into db uacreg table

- Allows to pass unique address part into contact header for REGISTER requests for each uacreg entry
- If contact_addr wasn't passed into uacreg for a particular entry default reg_contact_addr will be used instead.

### DIFF
--- a/src/lib/srdb1/schema/entities.xml
+++ b/src/lib/srdb1/schema/entities.xml
@@ -1,5 +1,6 @@
 <!ENTITY user_len "64">
 <!ENTITY domain_len "64">
+<!ENTITY contact_addr_len "128">
 <!ENTITY socket_len "128">
 <!ENTITY from_tag_len "128">
 <!ENTITY to_tag_len "128">
--- a/src/lib/srdb1/schema/uacreg.xml
+++ b/src/lib/srdb1/schema/uacreg.xml
@@ -135,6 +135,14 @@
         <description>Initial registration delay</description>
     </column>
 
+    <column id="contact_addr">
+        <name>contact_addr</name>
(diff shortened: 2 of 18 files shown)


miconda:
The examples show that the contact addr field in database can be `udp:192.168.0.125:5060`, but from the code seems to be concatenated prefixed with `sip:UUID@`, meaning that it can result in `sip:UUID@udp:192.168.0.125:5060`, which is invalid SIP uri. Maybe I failed to notice where conversion is done.

Then, for database updates, you have to update the xml files from `src/lib/srdbb1/schema/` and then do `make dbschema` in the root folder of Kamailio to update the sql/db creation files.

miconda:
Thanks! I will do another review soon and if all ok, I will merge manually to have the proper commit message format and commits per components.

miconda:
I pushed your commits manually, then followed up with a few commits making the changes:

  * increased the size of contact_addr filed to 255 (it can have parameters and then can become long, e.g., cases for ims or mobile devices)
  * contact_addr is optional value on RPC add command, `.` can be provided in such case to use the modparam value (to be easier to have the existing mode in the future, when one does not want to add per record value
  * contact_addr is last parameter in RPC add command, being the last added (you made it the previous last, but existing deployments should be easier to upgrade their cli tools by adding at the end instead of in the middle of cli parameters)

No testing though, should there be any problem after my additions, just open a bug report.

PR #2472, opened 2020-09-06
msrp: fixed IPv6 address parsing

sergey-safarov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [*] Tested changes locally

#### Description
Fix relaying MSRP messages to IPv6 uri.

### DIFF
--- a/src/modules/msrp/msrp_parser.c
+++ b/src/modules/msrp/msrp_parser.c
@@ -491,6 +491,12 @@ int msrp_parse_uri(char *start, int len, msrp_uri_t *uri)
 	}
 	hook = &uri->host;
 	hook->s = s;
+	p = q_memchr(s, ']', e - s);
+	if(p!=NULL)
+	{
+		/* move s pointer to end of IPv6 address*/
+		s = p + 1;
+	}
 	p = q_memchr(s, ':', e - s);
 	if(p!=NULL)
 	{
@@ -521,6 +527,12 @@ int msrp_parse_uri(char *start, int len, msrp_uri_t *uri)
 	hook->len = e - hook->s;
 	trim(hook);
 
+	if(strncasecmp(uri->host.s, "[", 1)==0) {
+		/* required remove square brackets from IPv6 address */
+		uri->host.s++;
+		uri->host.len--;
+		uri->host.len--;
+	}
 	if(uri->host.len<=0)
 	{
 		LM_ERR("bad host part in [%.*s] at [%ld]\n",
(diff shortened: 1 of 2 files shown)


miconda:
Can you explain where the problem actually happens? Because the patch does not seem right, being in the URI parser, making host shorter (skipping [ ]), but that is not correct, and can have other side effects. If routing is not working, then the fix has to be done in the routing part, not in the SIP uri host parsing.

sergey-safarov:
The issue happens when Kamailio receive MSRP message like
```
MSRP e1df91d477cc632a SEND
To-Path: msrp://3.236.25.4:5060/s.57471.1.1260517458;tcp msrp://[2600:1f18:63e1:6937:c504:55a1:e141:6810]:2855/e3e5c58412a3479fb007238b3729403a;tcp
From-Path: msrp://192.168.122.2:2855/88fafa933e5144ffa59e;tcp
Message-ID: f7d1b51d7ec9ad2d
Byte-Range: 1-365/365
Success-Report: no
Failure-Report: no
Content-Type: message/cpim

From: +12134445566 <sip:+12134445566@wv.nga911.com>
To: <sip:+13044598081@wv.nga911.com>
DateTime: 2020-09-06T11:07:09.980093+03:00

MIME-Version: 1.0
Content-Type: application/im-iscomposing+xml

<?xml version='1.0' encoding='UTF-8'?>
<isComposing xmlns="urn:ietf:params:xml:ns:im-iscomposing"><state>active</state><contenttype>text</contenttype></isComposing>
```

Kamailio acts as MSRP relay between IPv4 and IPv6 clients.

sergey-safarov:
I removed brackets striping and make tests - also work.

miconda:
Thanks for the details. I will have to check the code and see if this patch covers it properly, somehow I am not confident in just jumping to `]` for the case of ipv6 in the uri.

miconda:
Can you test with commit 1cda029, I did a different approach to be sure there is `[` at beginning of host part, not to find `]` in uri parameters. The for the PV field, there is a function that already returns the ipv6 address with []. I haven't tested, thus feedback on how it works is appreciated in order to backport.

sergey-safarov:
I made test of 9645be commit (current master), now msrp messages properly parsed and relayed via IPv6.

miconda:
Thanks! The commits can be backported. If you need them quickly, you can go ahead and backport, if not I will do it during the next days.

Issue #2459, opened 2020-08-28
NATHELPER: fix_nated_sdp does not support a=rtcp headers


oej:
### Description

fix_nated_sdp() function with an IP address as argument supports modifying the media line and the connection line but not a=rtcp. It would be helpful if it did that line too.

### Expected behavior

a=rtcp:<port> IN ip4 <new IP>
#### Actual observed behavior
No modification

miconda:
This should be implemented, with the commits referenced above.

PR #2497, opened 2020-10-05
nathelper: fix_nated_sdp added support for a=rtcp param RFC3605

arsperger:
- GH #2459

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #2459 

#### Description
fix_nated_sdp() function with flags 0x02 and 0x08 and an IP address as argument (or without one) does modify the media line and the connection line. In addition it will modify IP address in "a=rtcp" parameter (rfc3605) in case if there is one. If a=rtcp param is not found the function doesn't return -1, but prints a DBG message and continues to execute (changing c= or o= lines)

### DIFF
--- a/src/modules/nathelper/doc/nathelper_admin.xml
+++ b/src/modules/nathelper/doc/nathelper_admin.xml
@@ -562,7 +562,7 @@ if (search("User-Agent: Cisco ATA.*") {fix_nated_contact();};
 				<listitem>
 					<para><emphasis>0x02</emphasis> - rewrite media
 					&ip; address (c=) with source address of the message
-					or the provided IP address (the provide IP address take
+					or the provided IP address. (a=rtpc) param will be rewritten if exists. (the provided IP address take
 					precedence over the source address).</para>
 				</listitem>
 				<listitem>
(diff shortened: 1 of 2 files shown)


oej:
Great! Thank you! Just checking - do you handle all variants of "a=rtcp"? From the RFC:

```
   m=audio 49170 RTP/AVP 0
    a=rtcp:53020

    m=audio 49170 RTP/AVP 0
    a=rtcp:53020 IN IP4 126.16.64.4

    m=audio 49170 RTP/AVP 0
    a=rtcp:53020 IN IP6 2001:2345:6789:ABCD:EF01:2345:6789:ABCD

```

https://tools.ietf.org/html/rfc3605

arsperger:
Hi Olle,

internally function acts the same way as before, so for the above cases it yields the following results:

If there is no IP address in the rtcp attribute - nothing will be changed.

If param has an IPv4  - it will be replaced with the msg source IP address - if only the first argument (flags 0x02, 0x08, 0x10) is passed. Or with the IP address passed as the second argument. 

If param contains IPv6 - nothing will be changed. 

a possible issue, I have found in my tests, is that it is possible to pass IPv6 as a second argument to the function and it will replace IPv4 address leading to address family mismatch. (e.g c=IN IP4 2001:2345:6789:ABCD:EF01:2345:6789:ABCD) 
Probably should be mentioned in the nathelper admin guide.

miconda:
@arsperger - you can make a PR to update the docs with a remark about being careful to not replace IP address with one from a different family.

miconda:
Thanks!

PR #2498, opened 2020-10-05
registrar: added new parameter sock_addr

tiglat:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

This feature was made half year ago. We would like to have inside location record address of the host where kamailio is installed. However when we installed it behind NAT we got private address in this field. 
This feature introduces possibility to set a socket string which will be written into location record. It could be public address instead address on which the register request was received. The default value is NULL and it works as usual. The feature was installed on our production servers in April and since that time it works well. I would like to propose this feature to others, because it could be useful.

### DIFF
--- a/src/modules/registrar/doc/registrar_admin.xml
+++ b/src/modules/registrar/doc/registrar_admin.xml
@@ -535,31 +535,6 @@ modparam("registrar", "sock_hdr_name", "Sock-Info")
 		</example>
 	</section>
 
-	<section id="registrar.p.sock_addr">
-		<title><varname>sock_addr</varname> (string)</title>
-		<para>
-		This parameters can be used to override value written into socket field when contact is saved.
-		The value should be given in the following format: proto:IP:port. 
-		</para>
-		<para>
-		It could be usefull when kamailio is installed behind NAT and it is necessary to store its public IP 
-		instead socket on which the register request was received.
-		</para>
-		<para>
-		<emphasis>
-			Default value is NULL.
-		</emphasis>
-		</para>
-		<example>
-		<title>Set <varname>sock_addr</varname> parameter</title>
-		<programlisting format="linespecific">
-...
-modparam("registrar", "sock_addr", "udp:104.215.148.63:5060")
-...
-		</programlisting>
-		</example>
-	</section>
-
 	<section id="registrar.p.method_filtering">
 		<title><varname>method_filtering</varname> (integer)</title>
 		<para>
--- a/src/modules/registrar/doc/registrar_admin.xml
+++ b/src/modules/registrar/doc/registrar_admin.xml
@@ -535,6 +535,31 @@ modparam("registrar", "sock_hdr_name", "Sock-Info")
 		</example>
 	</section>
 
+	<section id="registrar.p.use_advertised_address">
+		<title><varname>use_advertised_address</varname> (integer)</title>
+		<para>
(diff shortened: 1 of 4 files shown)


miconda:
This works when listening on a single socket, right?

While that is ok for some use cases, if you want to have the advertised address of a listen socket, then this can be coded in a different way, by saving `socket->useinfo.sock_str` instead of `socket->sock_str`.

If you still want to have a parameter like in your patch, it is fine for me, even it has rather limited usage, but in this case I think it should use a static variable in pack_ci() function, like all the other variables in the function that are used outside, to avoid the pkg_malloc() and the pkg_free(), so declare:

```
static struct socket_info si = {0};
```

Then have something like:

```
		if (sock_addr.len>0) {
			memset(&si, 0, sizeof(struct socket_info));
			si.sock_str = sock_addr;
			ci.sock = &si;
...
```

Then IF blocks to do `pkg_free(si);` and `pkg_free(ci->sock);` should dissapear, having the changes only inside the pack_ci() function.

tiglat:
Thank you for advise. I will fix tomorrow.

tiglat:
Hi Daniel!
I have question about useinfo structure.
When I store sock_addr in socket->useinfo.sock_str kamailio crashes.
Should I initialize other fields of the structure?
The comment against useinfo tells that it is for "details to be used in SIP msg". Actually we do not want to use this address in SIP messages.
Is that correct place to assign socket address?
May I keep it in socket->sock_str?

tiglat:
Daniel, i have implemented your proposal and found that useinfo doesn't override socket field in db.
So I think we need to keep first approach.

miconda:
The use of `useinfo` field is when you want to store `advertised` address set to `listen` parameter. It is not about using it for the value of the new parameter `sock_addr`.

As I mentioned above, setting the `sock_addr` value via modparm is ok only using kamailio with a single listen address. I find its purpose really limited from that perspective.

The socket field in database location table is important to match a local listen socket (by listen address or advertised address) when phones connect from behind nat.

So, what I meant to store the value from useinfo requires to update other parts of usrloc (or registrar) module to use it when is set (that happens when advertise is in listen value) and if not, store the filed done right now.

tiglat:
Daniel, please code below

Is that you meant before?

		if (_m->rcv.bind_address->useinfo.sock_str.len > 0) {
		    memset(&si, 0, sizeof(struct socket_info));
		    si.sock_str = _m->rcv.bind_address->useinfo.sock_str;
		    ci.sock = &si;

It works as we expected.

miconda:
Looking at the combined diff against existing code, I think the condition on sock_flag has to be processed first (higher priority), respectively:

```
if (_m->flags&sock_flag) {
```

Because that is a config admin instructing to take the value from a different place. With this PR, that will never work when advertised address is set.

Moreover, using advertised address (the `useinfo`) should be controlled via a modparam, because from sip routing point of view it is more useful to have local listen socket than the public ip. So it needs a new parameter for having this new feature enabled.

miconda:
Thanks! I will squash and merge, but I plan to do some changes to this code to make it a bit more open for extensions in the future (e.g., maybe one wants to store socket name at some point).

miconda:
Renamed the mod param to `sock_mode` and advertised address is stored when this param is set to 1. See docs for more details. If any problem with it, then open a bug in the tracker.

PR #2509, opened 2020-10-16
permissions: exported allow_trusted() to kemi

arsperger:
#### Type Of Change 

#### Checklist:
- [x] Tested changes locally

#### Description
permissions allow_trusted function is exported to kemi. function doesn't not accept any parameters.


arsperger:
just fixed previous PR which was broken. Thank you.

miconda:
The function does not accept parameters, but it is not the right way to export the one for the configuration file, because the signature differs and cannot be found for documentation, as well as can create problems in the future because of the need to cast.

I will export it and can be used as a reference for the future, if you want to contribute on exporting missing functions to kemi. Thanks!

miconda:
For reference - the commit 843b22b implements it.

arsperger:
Got it, thank you!

PR #2512, opened 2020-10-19
janssonrpcc: cleanups

gaaf:
#### Pre-Submission Checklist

#### Type Of Change
- [x] Bug fix (non-breaking change which fixes an issue)

#### Checklist:

- [x] Tested changes locally

#### Description

Small cleanups

### DIFF
--- a/src/modules/janssonrpcc/janssonrpc_funcs.c
+++ b/src/modules/janssonrpcc/janssonrpc_funcs.c
@@ -94,8 +94,7 @@ int jsonrpc_request(struct sip_msg* _m,
 		for (; pit;pit=pit->next)
 		{
 			if PIT_MATCHES("route") {
-				pkg_str_dup(&route, &pit->body);
-				CHECK_MALLOC_GOTO(route.s, end);
+				route = pit->body;
 
 			} else if PIT_MATCHES("timeout") {
 				timeout = atoi(pit->body.s);
@@ -135,7 +134,6 @@ skip_parse:
 
 end:
 	if(freeme) free_params(freeme);
-	if(route.s) pkg_free(route.s);
 	return retval;
 }
 
--- a/src/modules/janssonrpcc/janssonrpc_funcs.c
+++ b/src/modules/janssonrpcc/janssonrpc_funcs.c
@@ -44,7 +44,6 @@ int jsonrpc_request(struct sip_msg* _m,
 	str params;
 	str options;
 	str route;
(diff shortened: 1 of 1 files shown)


henningw:
Sorry about the regression, the problem is that "str" is not supposed to be null terminated.
Its probably better to use "STR_VTOZ" instead of adding another str_dup implementation.

miconda:
I would rather change pkg_str_dup() and shm_str_dup() to allocate len+1 and add the ending zero, right now there are not widely used and at the end is just adding an extra operation `newbuf->s[len] = '\0';`. IMO it will be better and safer in long term, as one can then use standard string operations with newbuf->s. For that reason, over the years the values for variables (like avp, xavp, ...) were made zero-terminated.

STR_VTOZ macro is useful when one knows that the string is stored inside a larger buffer, so temporarily setting an ending 0 doesn't go over allocated memory, otherwise can end up in segmentation fault.

gaaf:
> I would rather change pkg_str_dup() and shm_str_dup() to allocate len+1 and add the ending zero

That was my first approach. I just thought it would provoke more discussion (see below) and I wanted to fix this bug fast, so I just kinda reverted the commits that introduced it.

The problem with adding a zero on every copy is that a ```str``` is supposed to be useful without a terminating zero. Lots of ```str``` uses are within the parser where ```str``` just points into the message buffer and has no terminating 0. Having them mixed will inevitable lead to bugs (just like the one I'm trying to fix). All developers will have to remember that only copies are 0-terminated.

Introducing a separate type for it (```strz```) will need duplicates for all existing str functions as C doesn't allow type inheritance.

Another option is to create ```pkg_str_dupz()``` and ```shm_str_dupz()``` which will 0-terminate the copies.

miconda:
The `len` field will not include the 0, so copying such a cloned value will not be affected by ending 0. Besides the str fields that point in larger buffers (like with the sip parser fields that point inside msg->buf), most of the other str->s values are 0 terminating (e.g., most of pseudo-variables values, functions parameters evaluated by fixup_get_svalue(), ...). I am sure there are way more 0-terminated str->s fields than the non-0-terminated.

Again, here we discuss about cloning values, their result is not impacting anything else, they are supposed to be independent of other resources.

The rule is more like: do not rely on 0-terminated str->s value unless you are sure it is.

In this case it is made sure to be 0-terminated and used accordingly.

miconda:
Thanks, I will squash and merge.

PR #2519, opened 2020-10-20
core: Zero terminate str copies made with (pkg|shm)_str_dup()

gaaf:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [x] Related to PR #2512

#### Description

As discussed in #2512, zero-terminating all copies is preferred as most str
usage is already with zero-terminated str.s.

### DIFF
--- a/src/core/ut.h
+++ b/src/core/ut.h
@@ -748,6 +748,7 @@ static inline int strz2sint(char* _s, int* _r)
 
 /**
  * \brief Make a copy of a str structure to a str using shm_malloc
+ *        The copy will be zero-terminated
  * \param dst destination
  * \param src source
  * \return 0 on success, -1 on failure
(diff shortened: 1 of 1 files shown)


miconda:
Thanks, it can be merged!

miconda:
Merged myself. Somehow I thought @gaaf has commit access already, but I was mistaken.

gaaf:
> Somehow I thought @gaaf has commit access already, but I was mistaken.

That was before the migration to github.

henningw:
The docs in str.h should be probably changed as well (here is documented that str is not null-terminated). Also probably a good idea to review other invocation of this utilty functions in the modules, now there might be double null-termination done somewhere (might be harmless in most cases, though).

gaaf:
Something like this: #2520 ?

I think that chances of double zero-termination are low when using these functions as it required 2 allocations in succession. Most uses of zero-terminated str will be open-coded to prevent that.

Issue #2496, opened 2020-10-05
freeradius-client support for auth_radius on debian jessie


twmobius:
### Description

I've been trying to setup kamailio with the latest freeradius-client. I've been using the docker image `kamailio/kamailio:5.3.1-jessie` adding the freeradius-client (latest github pull).

However it seems that kamailio cannot create a proper radius request and fails with error 
`3(9) ERROR: auth_radius [sterman.c:264]: radius_authorize_sterman(): unable to add User-Name attribute`

#### Reproduction

Install freeradius-client

```
    git clone https://github.com/FreeRADIUS/freeradius-client.git \
    && cd freeradius-client/ \
    && ./configure \
    && make \
    && make install
```

Modify the `/usr/local/etc/radiusclient/dictionary` and append the following lines 

```
$INCLUDE /etc/radiusclient/dictionary.sip
$INCLUDE /etc/radiusclient/dictionary.kamailio # https://github.com/kamailio/kamailio/blob/master/src/modules/auth_radius/cfg/dictionary.kamailio
```

Configure auth_radius

```
loadmodule "auth_radius.so"
loadmodule "misc_radius.so"

modparam("auth_radius|misc_radius", "radius_config", "/usr/local/etc/radiusclient/radiusclient.conf")

route[AUTH] {
	if (is_method("REGISTER") || from_uri==myself) {
		# authenticate requests
		if (!radius_www_authorize("$fd")) {
			www_challenge("$fd", "0");
			exit;
		}
         }
}
```

#### Debugging Data

I've went on the tried to debug the issue from the freeradius-client's point of view since it was easier to compile and iterate, and I've added various logs here and there to try and figure out what was wrong. From what I've seen, kamailio invokes `INIT_AV` to setup the `attrs` dictionary containing the dictionary attributes and matching values. 

https://github.com/kamailio/kamailio/blob/360e012d9d93c80b62b6e75bfac6e3340407ee98/src/modules/misc_radius/radius.h#L106

I've traced the `rc_dict_findattr` and it seems that the library properly returns `User-Name` attribute with value `1`. 

However in the `radius_authorize_sterman` function 

https://github.com/kamailio/kamailio/blob/360e012d9d93c80b62b6e75bfac6e3340407ee98/src/modules/auth_radius/sterman.c#L221

[...]

PR #2530, opened 2020-10-23
auth_radius, misc_radius: Adds VENDOR() definition if the freeradius-client.h does not contain one

twmobius:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [X] Related to issue #2496 

#### Description

freeradius-client removed the `VENDOR()` definition after commit https://github.com/FreeRADIUS/freeradius-client/commit/50d78bb53f4f341aa708e196b8955cacbae59669, since a new way was added in the library to separate vendor attributes. Currently `(((x) >> 8) & 0xffffff)` to determine vendor id works but is considered old-style. This PR adds the `VENDOR()` definition back if the freeradius-client library we are linking against is after 50d78b.

### DIFF
--- a/src/modules/misc_radius/radius.h
+++ b/src/modules/misc_radius/radius.h
@@ -42,6 +42,9 @@
 #else
 #include <freeradius-client.h>
 #define DEFAULT_RADIUSCLIENT_CONF ""
+#ifndef VENDOR
+#define VENDOR(x) (((x) >> 16) & 0xffff)
+#endif
 #endif
 
 


henningw:
Thanks for the pull request. Why do you defined it differently in your implementation then in the library?

- your request: #define VENDOR(x) (((x) >> 16) & 0xffff)
- library: #define VENDOR(x)		(((x) >> 8) & 0xffffff)

twmobius:
The initial version of freeradius-client, which kamailio was cross-compiled against, had the vendor id shifted by 16 bits. It was changed to 8 bits after https://github.com/FreeRADIUS/freeradius-client/issues/72 which as stated was a minor improvement since "private enterprise numbers aren't yet at 16 bits" and collisions were not that possible. Later on freeradius-client completely removed `VENDOR()` as part of a different way of handling vendor ids. 

Also radcli uses 16 bits for shifting vendor id as seen here https://github.com/radcli/radcli/blob/f0d1a1704595d62182d6a03c980ed07e906374a4/include/radcli/radcli.h#L56

so it made more sense to me, to keep 16 bits shifting to both libraries. 

I should note, that adding `VENDOR()` is more of a quick workaround at this point to allow kamailio to compile against both libraries with minimum effort. Otherwise a different implementation should be used by kamailio if freeradius-client is used

henningw:
Thanks for the clarification. Yes, its indeed more a workaround.

Issue #2032, opened 2019-08-13
Kemi sqlops memory leak


SipSeb:
### Description

We are running Kamailio 5.1.x in a mixed native/Kemi environment. That means, most of the config is still native Kamailio language, but some newer functions were written in JS and included via `jsdt_run("...");`. 

In one of those functions, we place an sqlops query. And since introducing this code, Kamailio leaks memory. Slowly, but eventually, the private memory will fill up.

#### Reproduction

In Kamailio, the code looks like this:
```
if (!jsdt_run("ksrIsCallAccountingRequired", "$(var(accentries){s.select,10,|})", "$(var(accentries){s.select,12,|})")) {
                xlog("L_ERR", "JS Call of ksrIsCallAccountingRequired route failed.\n");
}
```

And our JS code containing sqlops looks like this:
```
extension = '1234567e0';
mastersipid = '1234567';
KSR.sqlops.sql_query(
    "dbconn",
    `SELECT username, grp FROM grp WHERE username IN ('${mastersipid}', '${extension}') AND grp = 'ourGroupName'`,
    user
  );
```

For each call, this query is executed once. Depending on the traffic, you can see the private package memory starting to fill up quicker or slower.

### Possible Solutions

We had a similar problem about two years ago with app_lua and getting entries from a hash table, which we solved by introducing a temporary variable for using in the hash table index instead of strings. We tried doing the same thing in our current setup:

```
KSR.pv.sets("$var(extensionForQuery)", extension);
KSR.pv.sets("$var(masterSipidForQuery)", masterSipid);
KSR.sqlops.sql_query(
    "dbconn",
    "SELECT username, grp FROM grp WHERE username IN ('$var(masterSipidForQuery)', '$var(extensionForQuery)') AND grp = 'ourGroupName'",
    user
  );
```

However, Kamailio doesn't replace the variables in the query string. Although the documentation of the sqlops module says, it supports pseudovars in the query string. We even tried putting the whole query into a variable, but ended up with a sqlops error.

### Additional Information

```
ersion: kamailio 5.1.4 (x86_64/linux)
[...]

PR #2536, opened 2020-10-29
ndb_redis: Fix leaks from PV parsing

weili-jiang:
- Add missing pkg_free for the dynamic PV name created during PV parsing
- Add missing pkg_free for any GPARAM_TYPE_PVS created during PV parsing

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description

We are using NDB_REDIS in anger, fetching data from changing hash structures. We very quickly found that we were running out of package memory from using NDB_REDIS. Enabling Kamailio's own memory debug logging quickly revealed that every query of a `$redis(...)` PV caused a package memory leak.

In order to handle structured data returned from Redis commands, the NDB_REDIS module creates dynamic PVs that vary depending on the nature of the data being fetched from the reply. In order to return such dynamic PVs, it has a custom parser that allocates a dynamic PV structure as well as potentially multiple nested parameter structures. These are never freed.

It appears that the allocated structures are only used within the subsequent handler to fetch the data. Therefore, to resolve the leak, the allocated structures are freed after the data for the PV in question has been fetched.

### DIFF
--- a/src/modules/ndb_redis/ndb_redis_mod.c
+++ b/src/modules/ndb_redis/ndb_redis_mod.c
@@ -795,10 +795,7 @@ error_key:
 	return -1;
 }
 
-/**
- *
- */
-static int pv_get_redisc(struct sip_msg *msg,  pv_param_t *param,
+static int pv_get_redisc_result(struct sip_msg *msg,  pv_param_t *param,
 		pv_value_t *res)
 {
 	redisc_pv_t *rpv;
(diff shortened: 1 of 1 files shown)


weili-jiang:
It seems like the same sort of problematic pattern may exist in other modules, such as NDB_MONGO. I have not looked into those in detail, but this form of fix may be applicable there also.

weili-jiang:
Should I create an issue for this leak bug as well?

weili-jiang:
There still seems to be a leak somewhere. Will re-open once found.

miconda:
Are you using it in native kamailio.cfg script or in KEMI scripting (Lua, Python, ...)?

weili-jiang:
> 
> 
> Are you using it in native kamailio.cfg script or in KEMI scripting (Lua, Python, ...)?

KEMI Python3. The code in question doesn't seem to relate to KEMI at all, however.

miconda:
Then just reuse the redis result (reply) ids (e.g., static strings), likely you do not need to work with more than 2-3 results at the same time.

The design of PVs, which was done for the native scripting language, has impact here -- in the native scripting the PV names are resolved at startup, but with Kem, PVs can be created dynamically and can result in filling the memory. It is not a leak per se, because they are still referenced.

A similar case was also reported with sqlops module in #2032, in case you want to read about.

weili-jiang:
> Then just reuse the redis result (reply) ids (e.g., static strings), likely you do not need to work with more than 2-3 results at the same time.
I was finding an issue with array results, actually. But I can see how your reasoning still works, in that as long as arrays are bounded in length, there should be a limit at some point.

One final question then - is the ID space separate between workers? i.e. can worker 1 use ID 1 and worker 2 use ID 1 without worrying about colliding with each other?

miconda:
The results are kept in private memory, there is no collision between workers. If you want to discuss more, use sr-users@lists.kamailio.org mailing list.

Issue #2539, opened 2020-10-29
NAPTR resolution on IP addresses in kamailio 5.4.x


jchavanton:
### Description

The improvement to "use all NAPTR records" commit `28049aa` also introduced a regression where Kamailio is now doing resolution on IP addresses, adding unnecessary delays.

The modification is disabling the check
```
/* check if it's an ip address */
                if ( ((tmp_ip=str2ip(name))!=0)
                          || ((tmp_ip=str2ip6(name))!=0)
```
and always do an
```
if ((e=dns_get_entry(name, T_NAPTR))==0)
```

-----
### Initial Description
I noticed a test failure when validating Kamailio 5.4.x, and after looking at the SIP traces, Kamailio does not forward the first ACK.
However, it does it after receiving a 200 OK re-transmission .

```
                                         192.0.2.24:5060             
                                     ──────────┬─────────                
                                               │       200 coco (SDP
                                               │ <────────────────           
          200 coco (SDP)                       │                    
<───────────────────────────────────────────── │             
 ACK                                           │                    
─────────────────────────────────────────────> │    (not forwarded !)                               
                                               │       200 coco (SDP
                                               │ <────────────────
          200 coco (SDP)                       │                    
<───────────────────────────────────────────── │                            
 ACK                                           │                    
─────────────────────────────────────────────> │                                    
                                               │             ACK    
                                               │ ──────────────────>
```

### Troubleshooting
I did proceed by isolation and ended up concluding that this commit was triggering this behavior.
Simply, reverting this commit is fixing the problem.
```
commit 28049aafc8dd06c160ce5e7b8d5e4fc728441b0c
Author: Semen Darienko <semen.darienko@wildix.com>
Date:   Sun May 3 12:26:45 2020 +0200

    core: dns - use all NAPTR records
    
    - enable using of all NAPTR records instead of the first one ordered by
    priority
    - GH #2290
```

I did not look at the code modification and/or anything config specific yet, I will try to find some time to debug.

henningw:
Thanks for the report, are you testing 5.4.2? Do you have NAPTR records configured in your setup? Did not noticed something in this regards on my setups.

jchavanton:
Hi Henning, 

I was testing 5.4.2 but I ended up going back as far a May 3.

While trying to reproduce I just noticed that now I see 300-400ms to relay the ACK. (not enough to trigger a re-transmission anymore)

- First INVITE and BYE are relayed in 1ms.
[...]

PR #2541, opened 2020-11-04
core: DNS stop resolving NAPTR for IP addresses

jchavanton:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [x] Related to issue #2539
 
#### Description

@sem32 , maybe you can help to review this fix ?

### DIFF
--- a/src/core/dns_cache.c
+++ b/src/core/dns_cache.c
@@ -3292,6 +3292,9 @@ inline static int dns_naptr_sip_resolve(struct dns_srv_handle* h,  str* name,
 		}
 		try_lookup_naptr = 1;
 	}
+	/* check if it's an ip address, dns_srv_sip_resolve will return the right failure */
+	if (str2ip(name) || str2ip6(name))
+		goto naptr_not_found;
 	/* do naptr lookup */
 	if ((e=dns_get_entry(name, T_NAPTR))==0)
 		goto naptr_not_found;


src/core/dns_cache.c
```diff
@@ -3292,6 +3292,9 @@ inline static int dns_naptr_sip_resolve(struct dns_srv_handle* h,  str* name,
 		}
 		try_lookup_naptr = 1;
 	}
+	/* check if it's an ip address, dns_srv_sip_resolve will return the right failure */
```
sem32, line 3295: @jchavanton, can you give an example for your case?
I suppose that your check will never work, because we have the same check in line 3279 for first call of dns_naptr_sip_resolve

jchavanton: Hi, thanks for having a look.

Since you moved `dns_get_entry` outside of the following check, it can now be executed without checking for IP.


```
if (((h->srv==0) && (h->a==0)) && /* first call */
			 proto && port && (*proto==0) && (*port==0)){
```

jchavanton (commented):
In the issue I am sharing a simple ACK with an IP in the RURI, I am guessing this is taking place everytime, since there is no NAPTR record for the IP address, maybe you need more than one DNS server or something, I wanted to avoid getting dragged in spending to much time and  just by looking at the code I can see we are now accepting to do `dns_get_entry` on IP addresses.

However if this is raising more concerns, I can help to troubleshoot further.

sem32 (approved):

PR #2555, opened 2020-11-18
jansson: Fix another case of inverse boolean logic

gaaf:
Copy-pasto from before the other fix.


henningw:
related to  0125c1c8d8c1179

henningw:
Thanks, merged

PR #2556, opened 2020-11-19
smsops: fix len calc for a concatenated sms

alexyosifov:
- fixed the calculation for concatenated SMSs
based on TS 23.040, Sec. 9.2.3.16

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/modules/smsops/smsops_impl.c
+++ b/src/modules/smsops/smsops_impl.c
@@ -195,7 +195,7 @@ static int ascii_to_gsm(str sms, char * output_buffer, int buffer_size) {
 		}
 	}
 
-	if (i <= sms.len)
+	if (i < sms.len)
 		output_buffer[output_buffer_length++] =	(sms.s[i] & BITMASK_7BITS) >> (carry_on_bits - 1);
 
 	return output_buffer_length;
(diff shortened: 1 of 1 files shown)


src/modules/smsops/smsops_impl.c
```diff
@@ -618,6 +627,7 @@ int decode_3gpp_sms(struct sip_msg *msg) {
 					// Coding: 7 Bit
 					if (rp_data->pdu.coding == 0x00) {
 						// We don't care about the extra used bytes here.
+						rp_data->pdu.payload.sm.len = len;
 						rp_data->pdu.payload.sm.len = gsm_to_ascii(&body.s[p], len, rp_data->pdu.payload.sm, fill_bits);
```
miconda, line 631: Above are now two assignments one after the other to `rp_data->pdu.payload.sm.len`, is it `rp_data->pdu.payload.sm.len` used inside `gsm_to_ascii()`? Asking because len is also provided as parameter to `gsm_to_ascii()`.

alexyosifov: Yes "rp_data->pdu.payload.sm" is used in the gsm_to_ascii(...). You can see that sms.len is used in a few "if" statements.

miconda: OK, I will merge it. Somehow confusing is that `len` is also explicitly passed as a parameter to that function.

PR #2558, opened 2020-11-24
rtpengine: add CRC32 hash algo

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Add CRC32 selection algo.

### DIFF
--- a/src/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/src/modules/rtpengine/doc/rtpengine_admin.xml
@@ -2022,13 +2022,14 @@ modparam("rtpengine", "control_cmd_tos", 144)
 		<title><varname>hash_algo</varname> (integer)</title>
 		<para>
 			Hashing algorithm to be used in node selection algorithm. Now there are 2 possibilities: legacy
-			algorithm - 0(very basic hash over callid) or SHA1 - 1(apply sha1 over the callid and calculate hash).
+			algorithm - 0(very basic hash over callid), SHA1 - 1(apply sha1 over the callid and calculate hash) or
+			CRC32 - 2(calculate crc32 sum over the callid).
 		</para>
 		<para>
 			Default value is 0, legacy algorithm.
 		</para>
 		<para>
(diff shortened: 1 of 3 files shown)


miconda:
Thanks! Looks not intrusive, maybe @rfuchs wants to check it, otherwise can be merged from my point of view.

src/modules/rtpengine/doc/rtpengine_admin.xml
```diff
@@ -2028,14 +2028,17 @@ modparam("rtpengine", "control_cmd_tos", 144)
 			Default value is 0, legacy algorithm.
 		</para>
 		<para>
-			The values not falling into the range <quote>0-1</quote> are ignored.
+			The values not falling into the range <quote>0-2</quote> are ignored.
```
rfuchs, line 2031: The meaning of `2` should probably be mentioned in the paragraph just above (line ~2025). Otherwise LGTM.

smititelu:
Thanks for the merge. I've backported it to 5.3 and 5.4 branches.

henningw:
@smititelu its a new functionality, and should therefore not backported by default to stable branches, unless it qualifies as a bugfix.
Can you provide please more information if this is indeed a bugfix? Otherwise I will revert the commit there.

henningw:
Like, have the existing algorithm certain problems that are fixed with the new hash algorithm.

smititelu:
Hi @henningw,

We are obtaining smoother rtpengine session distribution results using crc32. Also crc32 is an alternative for those who don't want to load crypto.so module.

So, IMHO, this PR fixes the above two.

Thank you,
Stefan

henningw:
Hello @smititelu, sorry this does not sound like a bug fix to me. With the same argumentation we could e.g. merge the new dispatcher algorithm 13 to stable, as it provides a better call distribution in certain scenarios. :-)
I am reverting this commits therefore in stable branches.
Please ask next time another developer before backporting some patches.

PR #2561, opened 2020-11-25
tls: log version

jchavanton:
log the version of openssl used


henningw:
Thanks, I've changed it to LM_INFO - as for the other stuff we log on startup and committed it.

Issue #2547, opened 2020-11-10
dialog DMQ replication worker stops processing msgs and consume all shmmem


pwakano:
### Description

Kamailio is configured to use the DMQ replication for dialog.
When running a load test Kamailio DMQ workers eventually becomes unresponsive and stops processing DMQ messages. As a side effect the shm memory is fully consumed then SIP processing stops happening properly and everything else stops working due to that.

### Troubleshooting

Investigation showed that the DMQ workers are locked in the dmq_send_all_dlgs.
Debugging and studying the function it was identified that it is prone to data inconsistency because it uses a copy of the (d_table->entries)[index] and not a pointer to it. This gives a small window where a change in the d_table for that entry can occur and changing the lock information before the copy can actually get the lock making the process to get stuck in the lock get.
There is a need for some expert to validate this conclusion.

#### Reproduction

Run a simple load test making calls at a rate of ~5 cps and keep around ~2000 calls connected all the time. A higher cps seems to make it easier to reproduce the problem.
Adding network degradation to the environment makes the problem happens, but when running a tool such as SIPp for the load test, the retransmission can be forced by simply killing the SIPp instance receiving calls which will then force Kamailio to retransmit.

#### Debugging Data

```
7236 # DMQ WORKER
#0  0x00007fa7bd760bf9 in syscall () from /lib64/libc.so.6
#1  0x00007fa793395484 in futex_get (lock=0x7ffd072acaa4) at ../../core/futexlock.h:121
#2  0x00007fa7933a123b in dmq_send_all_dlgs (dmq_node=0x0) at dlg_dmq.c:657
        (gdb) p entry
        $1 = {first = 0x7fa79f8540d0, last = 0x7fa79f8540d0, next_id = 86622, lock = {val = 2}, locker_pid = {val = 7211}, rec_lock_level = 0}

        (gdb) p &entry->lock
        $2 = (gen_lock_t *) 0x7ffd072acaa4

        (gdb) p index
        $3 = 187

       (gdb) p (d_table->entries)[index]
       $1 = {first = 0x0, last = 0x0, next_id = 86652, lock = {val = 0}, locker_pid = {val = 0}, rec_lock_level = 0}

       (gdb) p entry
       $2 = {first = 0x7fa79f8540d0, last = 0x7fa79f8540d0, next_id = 86622, lock = {val = 2}, locker_pid = {val = 7211}, rec_lock_level = 0}

[...]

PR #2570, opened 2020-11-30
dialog: access dialog table by reference for dmq_send_all_dlgs

pwakano:
- Access to the d_table entries should be done by reference so there is
no risk of race conditions to get and release the lock

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #2547

#### Description

It was seen the access to the d_table was not being made by reference on the function dmq_send_all_dlgs. It was then leading to race conditions where the same entry could have its lock changed elsewhere and then a dead lock would occur. By checking the code in other places where the d_table is used, everywhere it is accessed by reference, this function was the only one not following this logic.
Changes were done to have "entry" as a reference to d_table and not a copy.

### DIFF
--- a/src/modules/dialog/dlg_dmq.c
+++ b/src/modules/dialog/dlg_dmq.c
@@ -646,22 +646,22 @@ error:
 
 int dmq_send_all_dlgs(dmq_node_t* dmq_node) {
 	int index;
-	dlg_entry_t entry;
+	dlg_entry_t *entry;
 	dlg_cell_t *dlg;
 
 	LM_DBG("sending all dialogs \n");
 
 	for(index = 0; index< d_table->size; index++){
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the pull request. Could you please amend the commit log, that it confirms to the contribution guidelines? Like:
`dialog: access dialog table by reference for dmq_send_all_dlgs`

miconda:
Thanks, this time I applied the patch manually, removing the unnecessary parenthesis when getting entry reference. In the future try to follow the contributing guidelines, as recommended to be easy to merge from the web interface here.

pwakano:
Thank you very much @miconda!

PR #2584, opened 2020-12-13
lost: removed not used namespace

sergey-safarov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [ ] Tested changes locally

#### Description
Caller LIS server may use namepace prefix `gml` or any other.
When created LOST request, then `location-info` payload created from `pidf-lo` and `p2` prefix namespace not used.

I suggest make some code cleanup and remove not used strings.

### DIFF
--- a/src/modules/lost/utilities.c
+++ b/src/modules/lost/utilities.c
@@ -955,12 +955,15 @@ https://tools.ietf.org/html/rfc5222
 	/* set properties */
 	xmlNewProp(ptrFindService, BAD_CAST "xmlns",
 			BAD_CAST "urn:ietf:params:xml:ns:lost1");
-	xmlNewProp(ptrFindService, BAD_CAST "xmlns:p2",
-			BAD_CAST "http://www.opengis.net/gml");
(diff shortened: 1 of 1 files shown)


wkampich:
The module includes a fallback in case `xpath` fails  due to a malformed `pidf-lo` (observed at ETSI Plugtests events). In such a case it just looks for attributes like `lat/lon` (point) or `lat/lon/radius` (circle) and creates a new location object, which I guess requires a `gml` namespace, therefore I would suggest the following change:
 
`
if(loc->xpath == NULL) {xmlNewProp(ptrFindService, BAD_CAST "xmlns:p2", BAD_CAST "http://www.opengis.net/gml");}
`

sergey-safarov:
@wkampich i have updated PR.
Could you look

wkampich:
Works for me! Thanks

miconda:
@wkampich - you are the developer/maintainer of the module, if you do not want to wait for other comments, you can merge the PR at any time.

PR #2588, opened 2020-12-16
modules/siptrace: fix regression preventing variables to be used

tsearle:
Commit fa7eb2a switched the 2 parameter version of sip_trace from
using the builtin fixup_spve_spve to the custom fixup method to
using the custom fixup_siptrace. As it is a custom fixup method,
the corresponding free method can not be auto-detected causing the
config parser to require the parameters to be a constant.  This
patch adds a free method, allowing variables to be once again passed
as the 2nd parameter of this method (as well as fixing a memory leak
for the 3rd parameter).

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

### DIFF
--- a/src/modules/siptrace/siptrace.c
+++ b/src/modules/siptrace/siptrace.c
@@ -793,7 +793,9 @@ static int fixup_free_siptrace(void **param, int param_no)
 		return fixup_free_spve_all(param, param_no);
 	} if (param_no == 3) {
 		/* tracing type; string only */
-		return fixup_free_str_all(param, param_no);
+		if (*param) {
+			pkg_free(*param);
+		}
 	}
 
 	return 0;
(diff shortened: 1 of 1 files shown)


miconda:
Thanks, I am going to merge it, but the leak is about using the function from kemi with the `KSR.x.modf()`? Or, if you used in native kamailio.cfg, what kind of parameters did you provide? Fixup is done usually only once at startup, and free fixup at shutdown.

tsearle:
kamailio.cfg was used as follows
sip_trace("sip:ip:port", $(ci{s.substr,0,26}));
it would result in the following error on startup
function sip_trace: parameter 2 is not constant

miconda:
That style was inherited from SER with the merge in 2008, the traditional Kamailio style is to enclose in quotes the parameters fixed up with `fixup_spve`, like:

```
sip_trace("sip:ip:port", "$(ci{s.substr,0,26})")
```

The parameter is evaluated and variables inside are replaced.

PR #2590, opened 2020-12-17
tm: 200 OK not processed correctly by Proxy after final non-2xx

annema-Kontron:
#### Pre-Submission Checklist

- [x ] Each component has a single commit (if not, squash them into one commit)
- [x ] No commits to README files for modules (changes must be done to docbook files

#### Type Of Change
- [x ] Breaking change (fix or feature that would change existing functionality)

#### Checklist:

- [x ] Tested changes locally -- tested at 5.1.6

#### Description

   Problem:
      Forwarding of 200 OK while sent 488 waits for ACK, destroys the UAS send buffer
   Scenario:
      During suspension of 200 OK by ims_qos module function Rx_AAR() at terminating PCSCF,
      The PCRF sends an AA Answer with result code DIAMETER_TOO_BUSY (3004), which triggers
      the PCSCF to send a 488 ‘Sorry no QoS available’ to the originating side (ims_dialog
      module function dlg_terminate()).
      Afterwards neither the 200 OK nor the ACK(488) is processed correctly by the PCSCF.
   Solution:
      The UAS send buffer should not be overwritten during processing of 200 OK,
      because non-2xx is needed to associate the ACK message in a correct way.
      200 OK must be forwarded statelessly.
   Side-Effect (potentially breaks existing function):
      Some callbacks cannot be called for the 200 OK, to avoid messing the stored 488.

### DIFF
--- a/src/modules/tm/t_reply.c
+++ b/src/modules/tm/t_reply.c
@@ -2042,40 +2042,42 @@ enum rps relay_reply( struct cell *t, struct sip_msg *p_msg, int branch,
 			goto error02;
 		}
 
-		/* attempt to copy the message to UAS's shmem:
-		 * - copy to-tag for ACK matching as well
-		 *   -  allocate little a bit more for provisional as
-		 *   larger messages are likely to follow and we will be
-		 *   able to reuse the memory frag
-		*/
-		if (res_len<=0) {
-			LM_ERR("invalid new buffer len\n");
-			goto error03;
-		}
-		uas_rb->buffer = (char*)shm_resize( uas_rb->buffer, res_len +
-			(msg_status<200 ?  REPLY_OVERBUFFER_LEN : 0));
-		if (!uas_rb->buffer) {
-			LM_ERR("cannot alloc reply shmem\n");
-			goto error03;
-		}
-		uas_rb->rbtype = relayed_code;
-		uas_rb->buffer_len = res_len;
-		memcpy( uas_rb->buffer, buf, res_len );
-		if (relayed_msg==FAKED_REPLY) { /* to-tags for local replies */
-			update_local_tags(t, &bm, uas_rb->buffer, buf);
-			t_stats_rpl_generated();
-		}
+		if (reply_status != RPS_PUSHED_AFTER_COMPLETION) {
+			/* attempt to copy the message to UAS's shmem:
+			 * - copy to-tag for ACK matching as well
+			 *   -  allocate little a bit more for provisional as
+			 *   larger messages are likely to follow and we will be
+			 *   able to reuse the memory frag
+			*/
+			if (res_len<=0) {
+				LM_ERR("invalid new buffer len\n");
+				goto error03;
+			}
+			uas_rb->buffer = (char*)shm_resize( uas_rb->buffer, res_len +
+				(msg_status<200 ?  REPLY_OVERBUFFER_LEN : 0));
+			if (!uas_rb->buffer) {
+				LM_ERR("cannot alloc reply shmem\n");
+				goto error03;
+			}
+			uas_rb->rbtype = relayed_code;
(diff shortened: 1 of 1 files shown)


henningw:
According to e-mail discussion with the author, there seems to be some adaptions necessary. Set to draft for now.

christoph-v-kontron:
Hi @henningw : we did our internal check now and no change is necessary. Please remove the WIP label

henningw:
Thanks for the confirmation.

christoph-v-kontron:
Maybe one question about or potential future contributions: how is the testing planned by the community?
I mean, we have an environment for patching stable release of kamailio (currently 5.1.6), where we can test our local (IMS related) use cases. We cannot test general use cases for the master branch. So any change must somehow pass some regression at the community, mustn't it?

henningw:
@christoph-v-kontron testing is usually done during the development and also in the freeze period before a major release is done. Of course it would be good if you can test your extensions also in master regarding the expected behavior, being the developer of them.

You mention some side effects of this patch due to the changes in transaction call-back behavior in the commit log. This should be probably reviewed and discussed before merging.

christoph-v-kontron:
[...]This should be probably reviewed and discussed before merging[...]
Of course. Should we start a discussion at the mailing list?

miconda:
@christoph-v-kontron - trying to understand the scenario from sip messages flow point of view ... is here about the case when 200ok is received, then suspended (via `t_suspend()`), and while it is suspended, the 488 arrives and it is forwarded, later the 200ok is resumed (`t_continue()`), being forwarded as well?

christoph-v-kontron:
@miconda almost correct - your scenario would lead to the same error, but it is a scenario with ims_qos module.
The **t_suspend()** is not explicitly called, but it happens implicitly within the function 
                $var(aarret) = **Rx_AAR**("**MT_aar_reply**","term","",-1);
                 switch ($var(aarret)) {
                        case 0:
                                #suspend was successful and we must break
                                exit;

The module sends a DIAMETER AAR message to some peer while the 200 OK is suspended, but the DIAMETER response (AAA) indicates a failure.

The 488 is created by **dlg_terminate()** from ims_dialog module.
route[**MT_aar_reply**]
{
        #this is async so to know status we have to check the reply avp
        switch ($avp(s:aar_return_code)) {
                case 1:
                        xlog("L_DBG", "Diameter: Orig AAR success on media authorization\n");
                        break;
                default:
                        #comment this if you want to allow even if Rx fails
                        if(dlg_get("$avp(CALLID_CUSTOM_AVP)","$avp(FTAG_CUSTOM_AVP)","$avp(TTAG_CUSTOM_AVP)")){
                                **dlg_terminate("all", "Sorry no QoS available")**;
                                usleep("100000");  #prm2272
                                exit;
                        }
        }
}

This leads to the situation that the 488 and the 200 OK are sent nearly at the same instance of time (only a few ms distance) and hence the ACK(488) is received AFTER the 200 OK is sent.

I can send you per e-mail a powerpoint from our analysis, OK?

henningw:
@christoph-v-kontron any updates on this patch, does the (apparently) private discussion clarified it?

christoph-v-kontron:
@henningw : no private discussion. @miconda asked a public question (see above) and I answered (see above). So I assume, it's his turn.

christoph-v-kontron:
@henningw : we are fine with the current version of the PR

miconda:
@christoph-v-kontron - I thought of it couple of time and I could not decide what is the right approach. It is a very particular case it happens in your scenario with a 200ok suspended.

Somehow make sense what the patch tries to do, but the side effects are not completely asserted. I would say that we should add a modparam to control this behaviour at the moment, with a default value making it to behave as the patch proposed. But still to have the option to turn it of and behave as before the patch.

So I am going to merge the PR, then I will add a new modparam to be able to control its behaviour.

Issue #2554, opened 2020-11-18
topos_redis: update record-route set on re-invite


balajeesv:
### Description
Kamailio not updating record-route value stored in redis (topos module) during re-invite

config:

modparam("ndb_redis", "server", "name=srv1;addr=localhost;port=6379;db=1")
modparam("topos", "storage", "redis")
modparam("topos_redis", "serverid", "srv1")
modparam("topos", "branch_expire", 300)
modparam("topos", "dialog_expire", 10800)
modparam("topos", "sanity_checks", 1)
modparam("topos", "clean_interval", 20)

#### Reproduction
1. Initial Invite from Kamailio to Server1 (Transaction complete)
2. Server1 is down
3. Re-invite from Kamailio to Server2
4. 200 OK contains record-route of Server2
5. ACK is forwarded to Server1 (record-route stored in dialog Key still has old record-route)
 

#### Debugging Data
b_rr value remains same even after 200 OK for re-invite has a different record-route 

```
**redis data after Initial Invite:**
127.0.0.1:6379[1]> hgetall d:z:atpsh-5fb50084-5b44-1
 1) "b_rr"
 2) "<sip:10.130.74.130;lr;ftag=11368SIPpTag091;did=07f.8fa;nat=yes>"
 3) "as_contact"
 4) "<sip:atpsh-5fb50084-5b44-1@3.225.0.56>"
 5) "s_rr"
 6) "<sip:3.225.0.56;lr;ftag=11368SIPpTag091;nat=yes>"
 7) "b_tag"
 8) "2"
 9) "bs_contact"
10) "<sip:btpsh-5fb50084-5b44-1@3.225.0.56>"
11) "rectime"
12) "1605698163"
13) "a_tag"
14) "11368SIPpTag091"
15) "s_cseq"
16) "1"
17) "s_method"
18) "INVITE"
19) "b_contact"
20) "sip:atpsh-5fb3efba-4a95-1@52.55.224.188;alias=10.130.72.128~5060~1"
21) "a_callid"
22) "1-11368@10.130.74.171"
23) "a_uuid"
24) "atpsh-5fb50084-5b44-1"
25) "b_uuid"
26) "btpsh-5fb50084-5b44-1"
27) "a_contact"
28) "sip:sipp@10.130.74.171:5060;alias=10.130.74.171~5060~1"
29) "iflags"
30) "2"

**redis data after re-Invite:**

[...]

PR #2594, opened 2021-01-03
topos: update record-route set on re-invite

balajeesv:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

topos: update record-route set on re-invite
Detailed information mentioned in https://github.com/kamailio/kamailio/issues/2554

### DIFF
--- a/src/modules/topos/doc/topos_admin.xml
+++ b/src/modules/topos/doc/topos_admin.xml
@@ -413,6 +413,26 @@ modparam("topos", "b_contact_avp", "$avp(tps-bct)")
 </programlisting>
 		</example>
 	</section>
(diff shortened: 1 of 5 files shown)


miconda:
After this patch, does the storage of RR headers happen when using DB backend, or only when using topos_redis?

balajeesv:
@miconda Changes only done for topos_redis

miconda:
I applied the patches manually, to break them by module, plus adding the support for db storage. Note that the module parameter was renamed to `rr_update`, being more generic in case we need in the future other modes for rr update, it can be reused with different value. Test and if any problem is with this version, open a bug report. Thanks!

Issue #2404, opened 2020-07-20
dual-stack bonding when used network interface name


sergey-safarov:
### Description

I prefer to use interface name for `listen`

*example*
```
listen=udp:eth0:5060
```

When `kamailo` started then bounded only to IPv4 address.
Could be added feature dual-stack bounding when used interface name.

tested latest master.

sergey-safarov:
for dual-stack Kamailio may be used `advertise` option like
```
listen=udp:eth0:5060 advertise "dual-stack.kamailio.org"
```

miconda:
I haven't worked with this one, but have you set:

https://www.kamailio.org/wiki/cookbooks/devel/core#auto_bind_ipv6

sergey-safarov:
Look as good approach but now has some error
I configured
```
listen=udp:eth0:5060                                         
auto_bind_ipv6=1                                                    
bind_ipv6_link_local=0
```
And when start kamailio get error
```
/ # kamailio -DD -E
Listening on 
             udp: 192.168.2.10:5060
             udp: 2600:70FF:B0CF:0:6C3E:2B62:BA0C:A6CC:5060
             udp: FE80:0:0:0:B951:EF1F:76C8:E5A2:5060
Aliases: 
             udp: fe80::b951:ef1f:76c8:e5a2:5060
             udp: 2600:70ff:b0cf:0:6c3e:2b62:ba0c:a6cc:5060
             udp: kazoo.aio:5060

 0(32) INFO: <core> [core/tcp_main.c:4982]: init_tcp(): using epoll_lt as the io watch method (auto detected)
 0(32) INFO: rr [../outbound/api.h:52]: ob_load_api(): unable to import bind_ob - maybe module is not loaded
 0(32) INFO: rr [rr_mod.c:185]: mod_init(): outbound module not available
 0(32) INFO: <core> [main.c:2834]: main(): processes (at least): 33 - shm size: 67108864 - pkg size: 8388608
 0(32) INFO: <core> [core/udp_server.c:154]: probe_max_receive_buffer(): SO_RCVBUF is initially 212992
 0(32) INFO: <core> [core/udp_server.c:206]: probe_max_receive_buffer(): SO_RCVBUF is finally 425984
 0(32) INFO: <core> [core/udp_server.c:154]: probe_max_receive_buffer(): SO_RCVBUF is initially 212992
[...]

PR #2595, opened 2021-01-04
Fix 2 issues that prevent auto_bind_ipv6 from happening on my setup

theraphim:
#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description

Without the first commit, on my setup - which has lots of interfaces with lots of addresses - auto_bind_ipv6 will make kamailio stuck on startup. It actually happens in get_flags, where data sent back from kernel can't fit into 8K anymore.

Second commit is fixing obvious typo while we're at it.

### DIFF
--- a/src/core/socket_info.c
+++ b/src/core/socket_info.c
@@ -1085,7 +1085,7 @@ static int get_flags(int family){
 	int nll = 0;
     int nl_sock = -1;
 
-	fill_nl_req(req, RTM_GETLINK, AF_INET);
+	fill_nl_req(req, RTM_GETLINK, family);
 
 	if((nl_sock = nl_bound_sock()) < 0) return -1;
 
--- a/src/core/socket_info.c
+++ b/src/core/socket_info.c
@@ -1070,6 +1070,7 @@ error:
 	req.g.rtgen_family = family;\
 	} while(0);
 
+#define NETLINK_BUFFER_SIZE 32768
 	
 static int get_flags(int family){
 	struct {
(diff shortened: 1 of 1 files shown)


sergey-safarov:
Is this fix this issue too?
https://github.com/kamailio/kamailio/issues/2404#issuecomment-661060298

when used
```
bind_ipv6_link_local=0
```

theraphim:
> Is this fix this issue too?
> [#2404 (comment)](https://github.com/kamailio/kamailio/issues/2404#issuecomment-661060298)
> 
> when used
> 
> ```
> bind_ipv6_link_local=0
> ```

I am not sure. The whole identification code is too convoluted for a casual analysis. If you can apply and test, we'll get the answer back.

As I said, this whole part will benefit from a rewrite. It'll be smaller, easier to analyze, more efficient and less error prone - for example, right now it's performing almost the same exchange with netlink twice and it can't handle dropped netlink messages.

src/core/socket_info.c
```diff
@@ -1184,6 +1189,16 @@ static int build_iface_list(void)
 		nll = 0;
 		p = buf;
 		while(1) {
+			// This (and the loop above in get_flags) is
+			// broken in more than one aspect. For example,
+			// if a receive is shorter than nlmsghdr,
+			// nlp->nlmsg_type will point at garbage.
+			// Proper fix for this (and static buffer sizing
+			// problem) would be to use libnl.
```
miconda, line 1197: This comment does not belong in the source code.

If there is still any issue left after this commit, then report it in the issue tracker. The developers do not scan the code for comments pointing at possible issues.

theraphim: Removed.

miconda:
Thanks! I am going to merge it.

Related to the comment, at a quick look, `libnl` seems to be linux specific, not usable on macos or *bsds. Anyhow, as I said, if something still needs to be fixed/tuned, report on issue tracker.

PR #2602, opened 2021-01-13
dispatcher: updated failover example

sergey-safarov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [ ] Tested changes locally

#### Description

we send "480 All servers busy" when no servers available for failover


miconda:
I am not really against, but also not seeing a real value on this approach, instead it can make troubleshooting harder for upstream, because it hides the downstream reply codes. In this case is about 500 or 408, but that can show a difference between a busy server and an unresponsive server.

Of course, there are cases when one wants to hide the downstream codes, but it is more likely a case per case specific.

Typically one has to edit that part of config to tune the list of reply codes for which to do re-routing.

On the other hand, the patch has syntax error, the opening `{` for `else` is missing.

sergey-safarov:
Thanks, Daniel for the clarification.

PR #2603, opened 2021-01-18
core: added xavu_serialize_fields function

nchaigne:
Also added a function called by the three "*_serialize_fields" to reduce code duplication.

The interface is not modified.

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
Small pull request that adds a function "xavu_serialize_fields" (with same interface as "xavp_serialize_fields" and "xavi_serialize_fields"), and an internal helper function to reduce code duplication.

### DIFF
--- a/src/core/xavp.c
+++ b/src/core/xavp.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2009 Daniel-Constantin Mierla (asipto.com) 
+ * Copyright (C) 2009 Daniel-Constantin Mierla (asipto.com)
  *
  * Permission to use, copy, modify, and distribute this software for any
  * purpose with or without fee is hereby granted, provided that the above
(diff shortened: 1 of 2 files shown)


miconda:
Actually the duplicate of code in this case was intentional, because the code for xavp, xavi and xavu should stay independent of each other, because they can diverge more in the future and a change in one can break the others.

There are a lot of functions that are somehow duplicate for the three types of xavps, when it becomes clear that they are going to stay very tight, then many functions may be combined, but till then I prefer to keep everything specific for one type. Also, when an eventual merging is decided, we have to balance the complexity of adding many IF conditions (to detect the type, then to figure out comparison of the names to be case sensitive or not, a.s.o.) vs some code duplication but clearer api to maintain.

You can make the pull request that just adds the independent xavu_serialize_fields() function and it will be merged.

nchaigne:
Thanks for your feedback.
I'll amend this pull request with only the xavu function then.

miconda:
Thanks!

Issue #2604, opened 2021-01-21
[kamailio 5.4.3] xavp_rm_by_index removes (N+1) first indexes instead of index N


nchaigne:
As previously discussed on the mailing list, this is a bug.
I have a fix that works, so I'll provide a pull request.

### Description

I’m trying to use function xavp_rm_by_index to remove a specific index from a given xavp.
I’m observing an unexpected behaviour, I think it might be a bug.
Instead of just removing the index, the function removes the index and all others before it.

E.g : if called with « idx = 1 », it removes indexes 0 and 1.

Reading the code of xavp_rm_internal (which is called by xavp_rm_by_index), I don’t think it behaves at it should :

```
/* Remove xavps
* idx: <0 remove all xavps with the same name
*      >=0 remove only the specified index xavp
* Returns number of xavps that were deleted
*/
```

For example when I do this :

```
LM_ERR("COUNT BEFORE: %d\n", xavp_count(rname, NULL));
int rm_count = xavp_rm_by_index(rname, 1, NULL); // TEST
LM_ERR("REMOVED: %d\n", rm_count);
LM_ERR("COUNT AFTER: %d\n", xavp_count(rname, NULL));
```

« rm_count » returned is 1 (which is expected). However, « count after » is decremented by 2.
And when I look at the contents of my xavp, I notice that the first two elements have been removed.

Likewise if xavp_rm_by_index is invoked with idx = 2 => the first 3 elements are removed.

#### Reproduction

Here is some code that shows the issue :

1) a test cmd_function in a module defined as follows :

```
static int was_test(struct sip_msg *msg, char *param1, char *param2);

static cmd_export_t cmds[] = {
       {"was_test", (cmd_function)was_test, 2, 0, 0, ANY_ROUTE},
       {0, 0, 0, 0, 0, 0}
};

/**
* For testing purposes.
*/
static int was_test(struct sip_msg *msg, char *param1, char *param2)
{
       LM_WARN("Called with param1: [%s], param2: [%s]\n", param1, param2);

       str my_str;
       my_str.s = param1;
       my_str.len = strlen(param1);

       int idx = atoi(param2);
       str *rname = &my_str;

       LM_ERR("COUNT BEFORE: %d\n", xavp_count(rname, NULL));

       int rm_count = xavp_rm_by_index(rname, idx, NULL);

       LM_ERR("REMOVED: %d\n", rm_count);
[...]

PR #2605, opened 2021-01-22
core: fix to xavp_rm_internal (#2604)

nchaigne:
This fixes the following issue:
https://github.com/kamailio/kamailio/issues/2604

When called to remove a specific index from a given xavp, function xavp_rm_by_index removes the index (as expected) but also all others before it.

E.g :
If called with idx = 1, it removes indexes 0 and 1.
Likewise if invoked with idx = 2 => the first 3 elements are removed.

This bug is located in function xavp_rm_internal. An assignment was missing when looping over the xavp list.
Same for xavi_rm_internal.

### DIFF
--- a/src/core/xavp.c
+++ b/src/core/xavp.c
@@ -454,6 +454,8 @@ static int xavp_rm_internal(str *name, sr_xavp_t **head, int idx)
 				if(idx>=0)
 					return 1;
 				count++;
+			} else {
+				prv = foo;
 			}
 			n++;
 		} else {
@@ -1914,6 +1916,8 @@ static int xavi_rm_internal(str *name, sr_xavp_t **head, int idx)
 				if(idx>=0)
 					return 1;
 				count++;
+			} else {
+				prv = foo;
 			}
 			n++;
 		} else {


miconda:
Thanks!

Can you amend the commit message to describe what it fixes? If not, I can apply the patch manually.

The reference to the issue number is ok to be included, but the fix still needs to be described, because in the future we may have to relocate the repository (e.g., if github goes premium only -- over the 20 years we have been on berlios, sourceforge, self hosting and now github) and then this issue number is no longer relevant.

Also, when looking at the git log, having just a issue reference number when generating release change log file, or decide on backporting, ...

nchaigne:
OK, I'll amend the comment.

miconda:
It does not have to be very long, just like: `fixed removal of previous xavps having the same name when providing the index`, so one can figure out from the log what was it about.

nchaigne:
Ah sorry didn't see your last comment in time. Well, I hope the message I pushed is acceptable. :)

miconda:
Thanks, this time should be ok.

PR #2612, opened 2021-01-27
http_client: http_client_request to include default clientcert, clien…

wkampich:
- the lost module uses http_client API functions and in the course of NG112
  client certificates are used for authentication when querying LIS or ECRF,
  the fix allows these to be read out via http_client module parameters.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

To use client certificates for authentication when sending HTTP requests via the http_client
API function `http_client_request`, the auth. specific module parameters (default values) are not
passed to curl. The fix allows these to be read out via http_client module parameters to
set proper curl config parameters. Maybe it's a new feature rather than a fix.

### DIFF
--- a/src/modules/http_client/functions.c
+++ b/src/modules/http_client/functions.c
@@ -673,6 +673,18 @@ int http_client_request(
 			query_params.http_proxy_port = default_http_proxy_port;
 		}
 	}
+	if(default_tls_clientcert.s != NULL && default_tls_clientcert.len > 0) {
+		query_params.clientcert = default_tls_clientcert.s;
+	}
+	if(default_tls_clientkey.s != NULL && default_tls_clientkey.len > 0) {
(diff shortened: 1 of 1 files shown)


oej:
Thank you for your contribution! I like that the API interface comes up to par with the functions in the module.

miconda:
The patch seems to be ok, from my point of view.

@oej - iirc, you authored the module, so you can merge if you do not need to wait for more reviews.

oej:
I will, was just keeping it open if anyone else had opinions...

miconda:
@oej - I guess this can be merged.

oej:
I wanted to test before merging. Today was a Kamailio day, but other tests came first. I will check tomorrow - test and merge :-)

Issue #2615, opened 2021-02-01
Add display name support to "pua_dialoginfo"


false-vacuum:
### Description

It doesn't look like the "pua_dialoginfo" module has support for the "display" attribute in the identity tag. This tag is used by some phones to update the phone display with caller/callee information for the respective key. I'm wondering how difficult it would be to add this feature. Thanks in advance for anyone who reads this!  

### Expected behavior

Kamailio includes display attribute in identity tags in generated publish messages. 

<pre>
  &lt;dialog-info xmlns="urn:ietf:params:xml:ns:dialog-info"  version="1" state="full"&gt;  
    &lt;dialog id="as7d900as8"  call-id="a84b4c76e66710"  local-tag="1928301774" remote-tag="456887766" direction="initiator"&gt;     
    &lt;state&gt;early&lt;/state&gt;
        &lt;local&gt;
            &lt;identity <b><i>display="Alice"</i></b>&gt;sip:alice@example.com&lt;/identity&gt;
            &lt;target uri="sip:alice@phone11.example.com"/&gt;
        &lt;/local&gt;
        &lt;remote&gt;
            &lt;identity <b><i>display="Bob"</i></b>&gt;sip:bob@example.org&lt;/identity&gt;
            &lt;target uri="sip:bobster@phone21.example.org"/&gt;
        &lt;/remote&gt;
    &lt;/dialog&gt;  
  &lt;/dialog-info&gt;
</pre>

#### Actual observed behavior

Kamailio generates a publish request for dialog events similar to the one below. The display attribute is not included in the "identity" tags. This means that when a notify message is sent to a watcher for this publish, the name is not rendered on the phone display. 

<pre>
  &lt;dialog-info xmlns="urn:ietf:params:xml:ns:dialog-info"  version="1" state="full"&gt;  
    &lt;dialog id="as7d900as8"  call-id="a84b4c76e66710"  local-tag="1928301774" remote-tag="456887766" direction="initiator"&gt;     
    &lt;state&gt;early&lt;/state&gt;
        &lt;local&gt;
            <b>&lt;identity&gt;sip:alice@example.com&lt;/identity&gt;</b>
            &lt;target uri="sip:alice@phone11.example.com"/&gt;
        &lt;/local&gt;
        &lt;remote&gt;
            <b>&lt;identity&gt;sip:bob@example.org&lt;/identity&gt;</b>
            &lt;target uri="sip:bobster@phone21.example.org"/&gt;
[...]

PR #2617, opened 2021-02-01
pua_dialoginfo: Add display attributes to identity tags.

false-vacuum:
pua_dialoginfo: Add display attributes to identity tags.

- This is an example of how "display" attributes could be added to identity tags in the PUBLISH message XML body.
- This is not a working solution, the strings("RemoteCallerName" & "LocalCallerName") are "hard coded". I could not figure out how to get the caller/callee information from the SIP message.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [ ] Related to issue #2615

#### Description
This is not a working solution. Opening the PR as requested by @miconda in issue #2615. I just added a couple lines to test how adding a "display" attribute to the identity tags would look like. This allows the phone to update it's display with relevant caller/callee information. I "hard coded" the values to verify that this works. However, I could not figure out how to get the display name from the SIP message(I'm not experienced with C). 

It seems that the "local" identity should have it's display name set to the name in the From header. And, the remote identity should have its display name set to the user part of the URI in the "To" header.

### DIFF
--- a/src/modules/pua_dialoginfo/dialog_publish.c
+++ b/src/modules/pua_dialoginfo/dialog_publish.c
@@ -188,6 +188,8 @@ str* build_dialoginfo(char *state, str *entity, str *peer, str *callid,
 			LM_ERR("while adding child\n");
 			goto error;
 		}
+		/* Testing(POC) - Add a display attribute to the remote/identity tag */
+		xmlNewProp(tag_node, BAD_CAST "display", BAD_CAST "RemoteCallerName");	
 		tag_node = xmlNewChild(remote_node, NULL, BAD_CAST "target", NULL);
 		if( tag_node ==NULL)
 		{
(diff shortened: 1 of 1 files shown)


false-vacuum:
I apologize for the commit message. I used the Github GUI to add the commit. I didn't realize it had a title field that would normally have been the first line.

src/modules/pua_dialoginfo/dialog_publish.c
```diff
@@ -188,6 +188,8 @@ str* build_dialoginfo(char *state, str *entity, str *peer, str *callid,
 			LM_ERR("while adding child\n");
 			goto error;
 		}
+		/* Testing(POC) - Add a display attribute to the remote/identity tag */
+		xmlNewProp(tag_node, BAD_CAST "display", BAD_CAST "RemoteCallerName");
```
miconda, line 192: I understand that you want the user part from SIP URI set in identity tag, then practically you have to parse it with parse_uri() function, which fills a structure with a field point to user part. Like:

```
parse_uri(remotetarget->s, remotetarget->len, &puri);
```
Then use `puri.user.s` and `puri.user.len`, which point to the start of user part and give the length. Similar will be for localtarget part.

miconda:
In master, see #2615.

PR #2621, opened 2021-02-03
rtpengine : add node fallback if node ran out of ports

dvillaume:
When rtpengine answers with error-reason 'Ran out of ports' adopt the same behavor that when node answers 'Parallel session limit reached'

#### Pre-Submission Checklist

#### Type Of Change
- [x ] New feature (non-breaking change which adds new functionality)

#### Checklist:

- [] PR should be backported to stable branches
- [ ] Tested changes locally

#### Description
Backport of the 'sessions limit' behavor for the 'Ran out of ports' error-reason.

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -294,6 +294,9 @@ static pv_spec_t *media_duration_pvar = NULL;
 #define RTPENGINE_SESS_LIMIT_MSG "Parallel session limit reached"
 #define RTPENGINE_SESS_LIMIT_MSG_LEN (sizeof(RTPENGINE_SESS_LIMIT_MSG)-1)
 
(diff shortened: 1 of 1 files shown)


miconda:
@rfuchs - any opinion on this PR?

rfuchs:
> @rfuchs - any opinion on this PR?

Personally I wouldn't want to rely on a textual error message to control behaviour, but it's already been done before. So fine with me.

miconda:
I haven't looked at the source code or the protocol, is there a response/error code that can be used instead of the text?

rfuchs:
> I haven't looked at the source code or the protocol, is there a response/error code that can be used instead of the text?

Well, it would have to be added :) We don't currently return error codes, but ideally something like that should be used instead of the text. But as we don't have that right now, can go with the error string for now.

miconda:
OK, thanks! Going to merge it.

PR #2622, opened 2021-02-04
korayvt/db postgres con params

korayvt:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

New configuration parameter is added to set postgres connection parameters.


henningw:
Thanks for the pull request. It seems that you are removing the timeout parameter. This is fine, as it can be replaced with the newly introduced connection parameters. But should be also reflected in the module XML docs (removal of the timeout parameter docs, and addition of a remark that it can be replaced with the new parameter).

korayvt:
> ut should be also reflected in the module XML docs (removal of the timeout parameter docs, and addition of a remark that it can be replaced with the new parameter).

Hi Henning,

Timeout parameter is used in two different places:
First one is like connect timeout which is replaced with my commit.
The other one is like query timeout which we need to keep in the code.
Therefore I decided not to remove the timeout parameter.
Maybe it's better to change the name and definition of the timeout parameter in xml doc.
What do you think?

miconda:
@korayvt - a variant that keeps also the backward compatibility is to go over the list of connections parameters and if `connect_timeout` is missing, then add it like it is done now, when `pg_timeout>0`.

On another aspect, I would suggest to go for the style of 'param_name=param_value", it seems that the postgres param names are compatible with sip param names, as I can see at the link you provided:

  * https://www.postgresql.org/docs/12/libpq-connect.html#LIBPQ-PARAMKEYWORDS

So instead of:

```
modparam("db_postgres", "con_param", "name=connect_timeout;value=15")
modparam("db_postgres", "con_param", "name=tcp_user_timeout;value=5000")
```

It can be:

```
modparam("db_postgres", "con_param", "connect_timeout=15")
modparam("db_postgres", "con_param", "tcp_user_timeout=5000")
```

And the above can also be compacted to:

```
modparam("db_postgres", "con_param", "connect_timeout=15;tcp_user_timeout=5000")
```

It would require some changes in function `postgres_con_param()`.

amessina:
Purely as an observer and a db_postgres user, wouldn't it be easier just to pass entire connection strings or connection URIs?  That way, Kamailio doesn't need to worry about parsing the params, just to rebuild them for input to libpq https://www.postgresql.org/docs/12/libpq-connect.html#LIBPQ-CONNSTRING

korayvt:
> @korayvt - a variant that keeps also the backward compatibility is to go over the list of connections parameters and if `connect_timeout` is missing, then add it like it is done now, when `pg_timeout>0`.
> 
> On another aspect, I would suggest to go for the style of 'param_name=param_value", it seems that the postgres param names are compatible with sip param names, as I can see at the link you provided:
> 
> * https://www.postgresql.org/docs/12/libpq-connect.html#LIBPQ-PARAMKEYWORDS
> 
> So instead of:
> 
> ```
> modparam("db_postgres", "con_param", "name=connect_timeout;value=15")
> modparam("db_postgres", "con_param", "name=tcp_user_timeout;value=5000")
> ```
> 
> It can be:
> 
> ```
> modparam("db_postgres", "con_param", "connect_timeout=15")
> modparam("db_postgres", "con_param", "tcp_user_timeout=5000")
> ```
> 
> And the above can also be compacted to:
> 
> ```
> modparam("db_postgres", "con_param", "connect_timeout=15;tcp_user_timeout=5000")
> ```
> 
> It would require some changes in function `postgres_con_param()`.

Hi Daniel,

I will consider your suggestions and make a new commit.

Thanks.

korayvt:
> Purely as an observer and a db_postgres user, wouldn't it be easier just to pass entire connection strings or connection URIs? That way, Kamailio doesn't need to worry about parsing the params, just to rebuild them for input to libpq https://www.postgresql.org/docs/12/libpq-connect.html#LIBPQ-CONNSTRING

Thanks for your suggestion. 
Currently PQconnectdbParams function is used to make connections.
This function takes two char arrays as keywords and values.
So we need to parse parameters.

amessina:
@korayvt, thank you for that detail.  I look forward to this pull request.

miconda:
Thanks!

PR #2643, opened 2021-02-21
Added ;sn param check to loose.c is_myself()

juha-h:
Added ;sn param check to loose.c is_myself() so that loose_route() would recognize such a Route URI denoting itself.

### DIFF
--- a/src/modules/rr/loose.c
+++ b/src/modules/rr/loose.c
@@ -120,7 +120,11 @@ static inline int find_first_route(struct sip_msg* _m)
  */
 static inline int is_myself(sip_uri_t *_puri)
 {
-	int ret;
+	int ret = 0;
+	param_hooks_t phooks;
+	param_t* plist = NULL;
+	param_t *pit = NULL;
+	str s;
 
 	if(_puri->host.len==0) {
 		/* catch uri without host (e.g., tel uri) */
@@ -144,6 +148,27 @@ static inline int is_myself(sip_uri_t *_puri)
 		/* match on host:port, but if gruu, then fail */
 		if(_puri->gr.s!=NULL)
 			return 0;
+	}
+
+	/* Check if URI has ;sn parameter naming existing socket */
+	if (ret == 0 && _puri->params.len > 0) {
+		s = _puri->params;
+		if (s.s[s.len-1] == ';')
+			s.len--;
+		if (parse_params(&s, CLASS_ANY, &phooks, &plist) < 0) {
+			LM_ERR("bad sip uri parameters: %.*s\n", s.len, s.s);
+			return 0;
+		}
+		for (pit = plist; pit; pit=pit->next)
+			if (pit->name.len == SOCKNAME_ATTR_LEN
+			    && strncasecmp(pit->name.s, SOCKNAME_ATTR,
+					   SOCKNAME_ATTR_LEN) == 0) {
+				if (pit->body.len > 0 &&
+				    ksr_get_socket_by_name(&pit->body) != NULL)
+					return 1;
+				else
+					break;
+			}
 	}
 
 	return ret;


miconda:
The commit message first line has to start with the name of the module ("rr: ...") like per contributing guide:

It is important for generating changelogs and releases. I made a similar remark to your latest commit as well.

Then, related to the commit itself, `myself` matching is related to routing address (ip/domain) matching of the server, this should be the default behaviour. Otherwise, on this case, it can happen that is matching headers added by another node using same socket names, which can have unexpected side effects in processing .

Personally I do not really see the need for what this patch brings, because you can specify the domains associated with the sever on several ways: config alias parameters, domain module (with register myself) for dynamic or large number of domains and corex module for subdomains.

However, I am fine to add the behaviour of the patch, but controlled by a new rr module parameter (e.g, myself_mode) , default being disabled (e.g., myself_mode=0) so the match is done like so far, on address elements only.

juha-h:

> The commit message first line has to start with the name of the module
> ("rr: ...") like per contributing guide: 
> 
>   * https://github.com/kamailio/kamailio/blob/master/.github/CONTRIBUTING.md#commit-message-format
> 
> It is important for generating changelogs and releases. I made a
> similar remark to your latest commit as well.

OK, will add if this leads to anywhere.

> Then, related to the commit itself, `myself` matching is related to
> routing address (ip/domain) matching of the server, this should be the
> default behaviour. Otherwise, on this case, it can happen that is
> matching headers added by another node using same socket names, which
> can have unexpected side effects in processing .

If such conflicts can happen or not, depends on the proxy deployment
environment.

> Personally I do not really see the need for what this patch brings,
> because you can specify the domains associated with the sever on
> several ways: config alias parameters, domain module (with register
> myself) for dynamic or large number of domains and corex module for
> subdomains.

In multi-tenant setup, MS Teams does not know which tenant INVITE from
Kamailio tries to reach if first R-R URI does not contain FQDN of the
tenant and, like everyone knows, MS does not care sh*t about standards
compatibility.  So the FQDN needs to be there and also loose_route()
needs to work.  I don't think that the other options you mention in the
above can help here.

> However, I am fine to add the behaviour of the patch, but controlled
> by a new rr module parameter (e.g, myself_mode) , default being
> disabled (e.g., myself_mode=0) so the match is done like so far, on
> address elements only.

OK, I'll try to add if the patch otherwise is OK.

-- Juha

miconda:
What I said that the list of FQDNs for tenants can be specified via alias global params, corex or domain modules to Kamailio, then this patch is not needed, the loose route should do its job. It is irrelevant for what loose route is supposed to do that you need it to interconnect with MS Teams.

And of course conflicts can happen depending on the deployment, but IP addresses and FQDN are global resources owned through a standard process. Socket names are random values that everyone can choose, and many go with suggestions from documentation. The it is likely to see interconnect providers using `s1`, `s2`, ... on their route headers, like it happens for `did` parameter added by dialog module. But to know it is the route header for the instance, the match is done on address, not custom parameters, because the did value from another provider can match another local dialog.

One of the benefits of open source is flexibility and trying to help solving easier various needs, but as a general remark, the default behaviour should be the one closest to the standards or the very common use cases. For specific needs, it has to be a parameter to turn it on. As I said, from my perspective it is ok to add it if controlled by a modparam.

juha-h:

> What I said that the list of FQDNs for tenants can be specified via
> alias global params, corex or domain modules to Kamailio, then this
> patch is not needed, the loose route should do its job. It is
> irrelevant for what loose route is supposed to do that you need it to
> interconnect with MS Teams.

I'll give corex alias_subdomains a try and close the pull request if it
works as expected.

--- Juha

juha-h:
Can be replaced by corex alias_subdomains.

PR #2647, opened 2021-02-23
uac_redirect: Make default q-value configurable

gaaf:
The old, quite arbitrary, default of 0.01 is preserved.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/modules/uac_redirect/rd_funcs.c
+++ b/src/modules/uac_redirect/rd_funcs.c
@@ -33,9 +33,9 @@
 
 
 extern sruid_t _redirect_sruid;
+extern int default_q_value;
 
 #define MAX_CONTACTS_PER_REPLY   16
-#define DEFAULT_Q_VALUE          10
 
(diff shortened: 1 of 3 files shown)


src/modules/uac_redirect/rd_funcs.c
```diff
@@ -33,9 +33,9 @@
 
 
 extern sruid_t _redirect_sruid;
+extern default_q_value;
```
miconda, line 36: This one is missing the type of the variable, expecting some compilers to at least throw a waring about, should be:

```
extern int default_q_value;
```

gaaf: Of course. Missed that. Fixed.

miconda:
Going to merge it, but then rename the new parameter to q_value, it is shorter and it comes to amend what is defined as default value for q, otherwise we have to specify what is default q value for the default q value parameter. Thanks!

PR #2656, opened 2021-03-03
etc: kamailio.cfg- use SIP-source-address

fredposner:
* default behavior of rtpengine is trust-address
* use SIP-source-address for nat_uac_test("8")

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/etc/kamailio.cfg
+++ b/etc/kamailio.cfg
@@ -851,9 +851,9 @@ route[NATMANAGE] {
 
 #!ifdef WITH_RTPENGINE
 	if(nat_uac_test("8")) {
-		rtpengine_manage("replace-origin replace-session-connection");
+		rtpengine_manage("SIP-source-address replace-origin replace-session-connection");
 	} else {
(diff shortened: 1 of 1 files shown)


fredposner:
> 
> Not sure if `SIP-source-address` is much useful, I would rather leave it with default behaviour for natted SDP. Also, in case of a chain of RTP relays in the cloud, the RTP relay app may run on a different system than the SIP server.
>

Valid points and I guess it goes to the old question of "what is the default kamailio used for?" I don't have an answer. I think if we do have the check, it would be best to switch to the behavior for rtpengine, but don't have a strong opinion either way.

In essence, wouldn't the daisy chain scenario be the same for rtpproxy as well?

miconda:
Let's merge this one for now to have backward compatibility with expected behaviour and we can review again the entire config file before releasing next major version, as we use to do it at that time. Thanks!

PR #2662, opened 2021-03-08
support subscribe dialog (topos + topos_redis)

fgaisnon:
#### Pre-Submission Checklist

- [x ] Each component has a single commit (if not, squash them into one commit)
- [x ] No commits to README files for modules (changes must be done to docbook files

#### Type Of Change
- [x ] New feature (non-breaking change which adds new functionality)

#### Checklist:

- [x ] Tested changes locally

#### Description
topos_redis: support SUBSCRIBE dialog
topos: SUBSCRIBE dialog
topos: documentation for SUBSCRIBE dialog
topos: add description for dialog_expire key

Add support for dialog subscribe when using topos.
When a Subscribe is received, a Subscribe dialog is created. Dialog TTL is set according to expire received in SIP request.
At each resubscribe the TTL is refresh according to expires set in SIP request.
Notify (in-dialog) with header event set are supposed to be inside a Subscribe dialog.

bugfix:
topos: when contact mode is set to 1, contact uri created is malformed if received contact has no user part
In this case we have a sip uri like : sip:@host

htable: ht_dmq_replicate_action was always called on ht_rm_items even if dmq was not activated

### DIFF
--- a/src/modules/htable/htable.c
+++ b/src/modules/htable/htable.c
@@ -484,7 +484,7 @@ static int ht_rm_items(sip_msg_t* msg, str* hname, str* op, str *val,
 		case 2:
 			if(strncmp(op->s, "re", 2)==0) {
 				isval.s = *val;
-				if (ht_dmq_replicate_action(HT_DMQ_RM_CELL_RE, &ht->name, NULL,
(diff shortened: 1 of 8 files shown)


miconda:
Thanks! All commits look ok, I will merge the PR.

PR #2665, opened 2021-03-10
 cplc: permanent contact not skipped by lookup anymore

easyrider14:
- a permanently registered contact has an "expires" value of 0 in usrloc
  when processing lookup, cplc module checks that contacts are not expired by
  checking that the "expires" value of the contact is not lower than current
  time
  as a consequence, permanently registered contact are alsways skipped and
  never targeted

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

when an address-of-record has a permanent contact set, and when a cpl is set, the lookup function of the cplc module skips the permanent contact and no branch to this contact is created. When no cpl is set, usrloc lookup will create the expected branch. So there is a different behavior depending on the presence of a cpl.
the proposed fix will correct this behavior : the cplc module will create a branch to the permanent contact during lookup

### DIFF
--- a/src/modules/cplc/cpl_run.c
+++ b/src/modules/cplc/cpl_run.c
@@ -283,7 +283,7 @@ static inline char *run_lookup( struct cpl_interpreter *intr )
 		} else {
 			contact = r->contacts;
 			/* skip expired contacts */
-			while ((contact) && (contact->expires <= tc))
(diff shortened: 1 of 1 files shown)


miconda:
Thanks!

PR #2685, opened 2021-03-21
rtpengine: remove square brackets from IPv6 address for getaddrinfo

Den4t:
rtpengine: IPv6 address in square brackets interpreted as FQDN.

- After loading from DB, rn_address is in form:` [x:x:x:x:x:x:x:x]:nnnn`, getaddrinfo function expect ipv6 in node argument without square brackets, otherwise, the argument is interpreted as FQDN.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -1829,6 +1829,7 @@ static int build_rtpp_socks(int lmode, int rtest) {
 		rtpe_reload_lock_get(rtpp_list->rset_lock);
 		for (pnode=rtpp_list->rn_first; pnode!=0; pnode = pnode->rn_next) {
 			char *hostname;
+			char *hp;
 
 			if (pnode->rn_umode == 0) {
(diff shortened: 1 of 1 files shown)


miconda:
@Den4t: have you checked that setting 0 in place of `]` is not having other side effects? I didn't notice any restore.

@rfuchs: any opinion on this patch?

Den4t:
@miconda 
>>have you checked that setting 0 in place of ] is not having other side effects? I didn't notice any restore.

Yes, hp - is a pointer to pkg_malloc-ed hostname variable, this var is freed right after getaddrinfo and not used any more.
I tested this patch localy, work as expected.

rfuchs:
> @Den4t: have you checked that setting 0 in place of `]` is not having other side effects? I didn't notice any restore.
> 
> @rfuchs: any opinion on this patch?

Looks good to me

miconda:
Thanks, merging it!

Issue #2716, opened 2021-04-27
TLS Client Verification OpenSSL 1.1.1(P-256) to OpenSSL 1.0.2 (P-521) is failing


space88man:
### Description
I am connecting two kamailio (5.4.5) using TLS with client cert verification. When one side is OpenSSL 1.1.1 (CentOS 8) + P-256 cert and the other side is OpenSSL 1.0.2 (CentOS 7) + P-521 cert *and* `cipher_list` is set(e.g., `DEFAULT`) on OpenSSL 1.1.1, the ClientHello fails because the only supported curve sent is P-256 (server has P-521).

If OpenSSL 1.1.1 is the TLS Server, the handshake also fails with `Alert Illegal Parameter 47`.

### Troubleshooting

1. If `cipher_list` is *not* set at all in `[client:default]`, then the ClientHello sends 4 supported EC groups and the handshake succeeds.
1. If `cipher_list` is set in `[client:defaut]` (e.g. `DEFAULT` or `ANY`) then the ClientHello sends only 1  supported EC group P-256 (I guess because the client is using a P-256 cert), then the handshake will fail.
1. In the reverse scenario if `cipher_list` is not set in `[server:default]` then the handshake will succeed, otherwise the handshake will fail.

This is so weird because if `cipher_list` is not set. then the value of `DEFAULT` (or maybe `DEFAULT:!KRB51`) is used. But if you explicitly enter that in the `tls.cfg` then the ClientHello is narrowed to only send one Supported Group(P-256). 

#### Reproduction

1. Connect two kamailos one on CentOS 8 sending SIP to one on CentOS 7, enable TLS, with client cert verification. Put a P-256 cert on CentOS 8 and a P-521 cert on CentOS 7.

#### Debugging Data

1. `cipher_list` is not set: TLS ClientHello Supported Groups is  4 groups: secp256r1 secp521r1 secp384r1 secp256k1
2. `cipher_list` is set to `DEFAULT`: TLS ClientHello Supported Groups is  1 group: secp256r1.
 

#### Log Messages

```
ERROR: tls [tls_util.h:42]: tls_err_ret(): TLS accept:error:1408A0C1:SSL routines:ssl3_get_client_hello:no shared cipher
```

#### SIP Traffic
NA - couldn't complete TLS handshake so no SIP was transferred.

### Possible Solutions

[...]

PR #2717, opened 2021-04-27
[tls] Don't use OpenSSL<1.0.2 fallback on 1.1+

space88man:
Addresses #2716. Also see https://bugs.python.org/issue29697.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X ] Tested changes locally
- [ ] Related to issue #2716 

#### Description
For OpenSSL 1.1.x initialization of EC SSL contexts has changed — we shouldn't be using the < 1.0.2 technique on OpenSSL 1.1+. This addresses a corner case where a TLS client(OpenSSL 1.1.1) with P-256 cert would not handshake with a TLS server with a P-521 cert. Adopted from the way that Python `_ssl.c` does initialization. Python, in turn, took this from Apache's `mod_ssl`.


miconda:
Thanks!

PR #2726, opened 2021-05-04
usrloc: cancel unnecessary db queries

korayvt:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

It looks like the db query is made before checking the configuration parameters for the following features:
- internal keep-alive mechanism
- contact-expired callback execution

For example, although keepalive mode is turned off, the location table is periodically queried.
This is also true when there is no usrloc:contact-expired route.

In this commit, checks are added before these two db queries.


miconda:
Thanks!

PR #2735, opened 2021-05-17
Update uac-create.sql to reflect UACREG_TABLE_VERSION in uac_reg.h

cbxbiker61:
Table version needs to be updated updated in uac-create.sql, to match uac_reg.h version 5. Version 5 reflects the addition of the contact_addr column.

### DIFF
--- a/utils/kamctl/db_sqlite/uac-create.sql
+++ b/utils/kamctl/db_sqlite/uac-create.sql
@@ -18,5 +18,5 @@ CREATE TABLE uacreg (
     CONSTRAINT uacreg_l_uuid_idx UNIQUE (l_uuid)
 );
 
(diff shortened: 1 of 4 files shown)


miconda:
Thanks for report and PR!

The database definition rules are kept in xml files inside `src/lib/srdb1/schema`, after changes there, `make dbschema` has to be run to generate the sql and the other db files for kamctl. It requires the xml toolkit to be installed. For uacreg I pushed the changes accordingly with commits: 8da9b7936c1be8b5c4df80d716344dfa4461b494 and 58d5f06aacce6ba7bb26f7be928b9bf71d39f9c3.

Also, for the future, be sure you format the commit message according to the guidelines in:

PR #2744, opened 2021-05-25
pike: fixed regression

drTr0jan:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
There's an regression in 5.5 on `pike.so`. After c9dc0336a33a0ecfe776975be7fbfab8f3c91b48 the module has stopped to work. `mark_node()` had returned a NULL on every IP-address.

### DIFF
--- a/src/modules/pike/ip_tree.c
+++ b/src/modules/pike/ip_tree.c
@@ -301,10 +301,6 @@ pike_ip_node_t* mark_node(unsigned char *ip,int ip_len,
 		}
 	}
 
-	if(node==NULL) {
-		return NULL;
-	}
-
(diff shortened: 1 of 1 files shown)


miconda:
Is this the fix for regression or just removal of an unnecessary return NULL? I understood from previous PR that the issues was with str case search.

drTr0jan:
@miconda, yep, it's the fix for regression. Pike (`pike_check_req()`) doesn't work without the fix at all. :(

It has nothing to do with previous PR (str case search). I've investigated the regression and found two issues (str case search and unnecessary NULL-return).

miconda:
Thanks!

PR #2752, opened 2021-05-29
siptrace: added siptrace pseudo var

sergey-safarov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description
I want to get the source (destination) IP address of the message during `event_route[siptrace:msg]` route.
I added the ability to get socket info in this route block.

Now added
```
$siptrace(src_uri)
$siptrace(dst_uri)
```
This pseudovars return string like
```
udp:8.8.8.8:5060
```

I think `src_uri` and `dst_uri` is not correct here because this socket info, not SIP uri.

I expecting suggestion on how to get these params and from current implementation:
```
src_ip
src_port
src_proto
dst_ip
dst_port
dst_proto
```

As reference, I take `sipdump` pseudovariable

### DIFF
--- a/src/modules/siptrace/siptrace.c
+++ b/src/modules/siptrace/siptrace.c
@@ -195,6 +195,10 @@ static str trace_local_ip = {NULL, 0};
 static db1_con_t *db_con = NULL; /*!< database connection */
 static db_func_t db_funcs;		  /*!< Database functions */
 
+int pv_parse_siptrace_name(pv_spec_t *sp, str *in);
+int pv_get_siptrace(sip_msg_t *msg, pv_param_t *param,
+		pv_value_t *res);
+
 /*! \brief
  * Exported functions
  */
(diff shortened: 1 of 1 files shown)


miconda:
I hope to get the time to check if this is the best way during the next days. I expected it works with the usual variables, but maybe I am wrong and this new variable class is needed.

miconda:
Thanks! Merging it.

miconda:
@sergey-safarov - you have the list the new variables in the pv cookbook:

  * https://www.kamailio.org/wiki/cookbooks/devel/pseudovariables

sergey-safarov:
Done
https://www.kamailio.org/wiki/cookbooks/devel/pseudovariables#siptrace_module

I think feature updates still required.
I think src_uri and dst_uri are not correct here because of this socket info, not SIP uri.

miconda:
I changed the code to be $siptrace(src_addr) and $siptrace(dst_addr), because the URI in the keyword would suggest is a SIP URI format. I also updated the wiki.

You can parse "proto:ip:port" with `parse_phostport()` and then get its attributes to have variables for ip, port, ...

Issue #2745, opened 2021-05-25
app_perl: Ignore __DIE__ signals that come from eval blocks


rodolfojcj:
### Description

If possible, I would like that the `kamailio/src/modules/app_perl/lib/perl/Kamailio.pm` file does not generate error messages to the system log files, which are not actual errors and are caused by `eval` blocks of third-party Perl modules that my code depends on.

To illustrate the current setup, let's consider the following:

- In file `/etc/kamailio/kamailio_52.cfg`
```perl
  modparam( "app_perl", "filename", "/opt/myprogram/perllib/MyKamailio.pm" )
```

- In file `/opt/myprogram/perllib/MyKamailio.pm`:
```perl
  use Crypt::JWT qw( decode_jwt encode_jwt );
```

- It happens that `Crypt::JWT` depends on `CryptX`, which in turn needs a `JSON::XS` implementation, which could be either `Cpanel::JSON::XS` or `JSON::XS`or `JSON::PP`, in that order of preference. That need is resolved at runtime, as we can see in the following code fragment taken from the file at https://github.com/DCIT/perl-CryptX/blob/master/lib/CryptX.pm:
```perl
BEGIN {
  if (eval { require Cpanel::JSON::XS }) {
    Cpanel::JSON::XS->import(qw(encode_json decode_json));
    $has_json = 1;
  }
  elsif (eval { require JSON::XS }) {
    JSON::XS->import(qw(encode_json decode_json));
    $has_json = 2;
  }
  elsif (eval { require JSON::PP }) {
    JSON::PP->import(qw(encode_json decode_json));
    $has_json = 3;
  }
  else {
    $has_json = 0;
  }
}
```

- In my system, `Cpanel::JSON::XS` is not installed because it is not needed. I could install it, but unfortunately that module intermittently causes a segmentation fault, as described in the issue report at https://github.com/rurban/Cpanel-JSON-XS/issues/179.

### Expected behavior

I would expect two things:

- The operating system log files, like `/var/log/syslog`, adds only entries like the following ones after restarting Kamailio:

      May 25 14:30:06 myserver systemd[1]: kamailio.service: Succeeded.
      May 25 14:30:06 myserver systemd[1]: Stopped Kamailio server.
      May 25 14:30:06 myserver systemd[1]: Starting Kamailio server...
      May 25 14:30:06 myserver systemd[1]: Started Kamailio server.

- The output of commands like `systemctl status kamailio.service` only shows informational messages of success.

[...]

PR #2754, opened 2021-06-01
app_perl change to ignore "die" signals from eval blocks

rodolfojcj:
Proposed fix for issue #2745.


miconda:
I am not using app_perl, also not that familiar with Perl, so if nobody else has comments, it can be merged soon.

PR #2756, opened 2021-06-02
Smsops 7bit support special chars

alexyosifov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally


miconda:
If no comments from other devs, it can be merged soon.

riccardv:
Hi @alexyosifov ,

a little comment regarding the fill bits, it's just for information.
There is a more easy way to perform decoding and encoding of 7bit concatenated sms.

Let me explain

3GPP has added the fill bits for retro compatibility with old devices that doesn't support the concatenated sms and then the UDH header. 
With this assumption the decoding of entire PDU may be done with the effect of first N ASCII characters (due to udh header) are not valid but the rest is ok. So it's possible to decode the entire user-data pdu as a normal non udh pdu discarding the first N decoded ASCII string.
N is equal to the septet of udh header + 1

In encoding the concept is the same, just add N preamble (3GPP recommends N '\n') characters to the ASCII string, encode as 7bit and overwrite the initial octets with the UDH header.

Hope my comment was useful
Regards

miconda:
@riccardv - if you want to do improvements to this part, just do a PR. This one will be merged.

Issue #2758, opened 2021-06-03
ERROR: <core> [core/resolve.c:941]: get_record(): unknown type 41


webstean:
#### Debugging Data

```
(paste your debugging data here)
```

#### Log Messages

```
(paste your log messages here)
```

#### SIP Traffic

```
(paste your sip traffic here)
```

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
version: kamailio 5.1.2 (x86_64/linux) 
flags: STATS: Off, USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MEM, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLACKLIST
, HAVE_RESOLV_RES
ADAPTIVE_WAIT_LOOPS=1024, MAX_RECV_BUFFER_SIZE 262144, MAX_LISTEN 16, MAX_URI_SIZE 1024, BUF_SIZE 65535, DEFAULT PKG_SIZE 8MB
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
id: unknown 
compiled with gcc 7.3.0

I check the source, type 41 isn't supported on the current source tree.
```

* **Operating System**:

```
(paste your output here)
```

henningw:
Thanks for the report. Did you tried current already a recent maintained version (e.g. 5.4,x, 5.5.x)?

miconda:
Closing, based on remarks in #2759.

PR #2759, opened 2021-06-03
Update resolve.c handle type 41

webstean:
Possible solution to https://github.com/kamailio/kamailio/issues/2758

#### Pre-Submission Checklist

#### Type Of Change
- [ X] Small bug fix (non-breaking change which fixes an issue)

#### Checklist:

- [ X] Tested changes locally
- [ ]X Related to issue #XXXX (replace XXXX with an open issue number)

### DIFF
--- a/src/core/resolve.c
+++ b/src/core/resolve.c
@@ -978,6 +978,7 @@ again:
 				last=&(rd->next);
 				break;
 			case T_OPT:
+			case 41:
 				/* skip DNS extensions, e.g. EDNS0 */
 				rd->rdata=0;
 				*last=rd;


henningw:
Thanks for the pull request. This should already taken care of it in the current versions:

/*
 * Older glibc < 2.25 does not include T_OPT in nameser_compat.h yet.
 * On alpine linux musl library it is also not defined. There is no
 * musl feature test macro, so we look for glibc instead.
 */
#if (defined __GLIBC__ &&  __GLIBC__ == 2 && __GLIBC_MINOR__ < 25) || !defined __GLIBC__
#ifndef T_OPT
#define T_OPT ns_t_opt
#endif
#endif

~/repositories/kamailio/src/core$ ack ns_t_opt /usr/include/arpa/nameser.h
    ns_t_opt = 41,

PR #2765, opened 2021-06-06
uac: registration refresh should in half time of expire

surendratiwari3:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
Currently re-registration is not happening in half of expire time. sometime its reaching near to expire time then sending re-registration. usually sip servers expect re-registration to happen in half of expire time.

### DIFF
--- a/src/modules/uac/uac_reg.c
+++ b/src/modules/uac/uac_reg.c
@@ -1205,7 +1205,7 @@ int uac_reg_update(reg_uac_t *reg, time_t tn)
 		reg->flags |= UAC_REG_INIT;
 	}
 
-	if(reg->timer_expires > tn + reg_timer_interval + 3)
+	if ((reg->expires/2) < reg->timer_expires - tn)
 		return 3;
 
 	return uac_reg_send(reg, tn);


henningw:
Hello, thanks for the pull request. If I do not misread the standard ([link](https://datatracker.ietf.org/doc/html/rfc3261#section-10.2.4)), there is no particular requirement for your change.
Have you tried to just set a lower reg_retry_interval? E.g.:
- reg_retry_interval: 180
- expires: 360

surendratiwari3:
@henningw reg_retry_interval is used for failed registration retry. i am looking for refreshing registration. Also just think about having multiple pbx that expect different expire then reg_timer_interval will not work. Also what changes i made that's apply to every where.

henningw:
I quoted the wrong parameter, i meant the reg_timer_interval: 
https://kamailio.org/docs/modules/devel/modules/uac.html#uac.p.reg_timer_interval

henningw:
The refresh threshold could be of course configured with a module parameter, keeping the current behavior as default.

surendratiwari3:
yeah i checked this too reg_timer_interval but this is not getting apply if kamailio is sending register to different pbx that are expecting different expire time.

miconda:
Doing registration refresh at the half of the expires interval is not a common practice, nor a requirement from SIP RFC, like @henningw pointed out. I haven't noticed such behaviour by sip phones, either. I second that such behaviour needs a module parameter to control it, default behaviour should stay like it is now.

surendratiwari3:
@henningw @miconda agreed its not hard core requirement from sip rfc , i will setup an module parameter for the same and will keep default behaviour as it is. :-)

miconda:
@surendratiwari3 - do you still plan to use this PR or will you open a new one once you get the time to work further on it?

PR #2774, opened 2021-06-09
statsd: adding support for histograms

jchavanton:
adding support for histograms

#### Pre-Submission Checklist

#### Type Of Change

#### Description
```
Histograms

The histograms are also a measure of time, but they are calculated at the server side. As the data exported by the client is the same, this is just an alias for the Timer type.

<metric name>:<value>|h
```


miconda:
I am fine to merge it.

PR #2825, opened 2021-08-13
mqueue: add two new module parameters

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Add 2 new mqueue module parameters to simplify config queue definition.
Add 1 new rtimer module parameter to set default timeout if not present.
Updated docs.

### DIFF
--- a/src/modules/mqueue/doc/mqueue_admin.xml
+++ b/src/modules/mqueue/doc/mqueue_admin.xml
@@ -150,6 +150,64 @@ modparam("mqueue", "mqueue", "name=qaz")
 	    </example>
 	</section>
 
+
+	<section id="mqueue.p.mqueue_name">
+	    <title><varname>mqueue_name</varname> (string)</title>
+	    <para>
+		Definition of a memory queue, just by name.
+	    </para>
+	    <para>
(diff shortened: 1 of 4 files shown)


miconda:
Thanks, however I can't see a relevant simplification of the config, but could be a matter of taste, not being a real technical matter -- instead of one modparam line with both queue name and size, now can be two modparams, one with name and one with size.

I will wait a little bit more to see if any other comments, then I can merge it.

smititelu:
Hi Daniel,

Thanks for the input. Indeed not a real technical matter.

The actual usefulness of this is that after the commits one can set and use config #!defines for the new modparams (config "simplification" was not the right word; config "organization" is better).

Thank you,
Stefan

DForester-FTC:
Hi Stefan,

I receive emails addressed to ***@***.******@***.***> (forwarded to me)

Dean is no longer with us and I cannot unsubscribe since I do not have his password.
You can find Dean at ***@***.******@***.***>

Please unsubscribe ***@***.******@***.***> from the email list

Multzumesc

AT

Issue #2797, opened 2021-07-10
JSON CEE structured logging


dpocock:
### Description

Kamailio recently added JSON logging support, described on these two pages:
https://www.kamailio.org/w/2020/04/structured-json-logging/
https://www.kamailio.org/wiki/cookbooks/devel/core#log_engine_cli_parameter

Structured logging is particularly useful when:

1. there are multiple applications involved in a call (e.g. Kamailio, Kurento, coTurn, Apache)
2. all the applications create JSON logs using identical field names

I recently made contributions to reSIProcate, GStreamer and Kurento to support the JSON CEE field names.  It would be very easy to replicate in Kamailio.
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/847
https://github.com/Kurento/kurento-media-server/pull/17
https://github.com/resiprocate/resiprocate/commit/d1df9aeb956be79253560fe3628b3f37a4ef94f2

Notice that ```rsyslog``` can receive legacy Syslog events from VoIP hardware, Cisco and Polycom phones and it can convert them to JSON CEE and push them into the document store where they will meet the records from Kamailio.  This is another case where it is really useful if all the JSON from different applications is using CEE.

Here are the JSON CEE field names:

https://cee.mitre.org/language/1.0-beta1/core-profile.html

### Expected behavior

Extend the config setting so that people can request CEE on the command line.

#### Debugging Data

```
(paste your debugging data here)
```

#### Log Messages

```
(paste your log messages here)
```

#### SIP Traffic

```
(paste your sip traffic here)
```

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
(paste your output here)
```

* **Operating System**:

```
(paste your output here)
```

dpocock:
Here are some of the commits and filenames relevant to the current implementation:

https://github.com/kamailio/kamailio/search?q=json+log&type=commits

henningw:
As the pull request(s) were merged, i think this issue can be closed. Please re-open if you want to continue to work on this area.

PR #2826, opened 2021-08-16
core: logging: JSON: add CEE schema support

dpocock:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #2797 (replace XXXX with an open issue number)

#### Description

See the feature request in #2797 

I've provided similar implementation in reSIProcate binaries, Kurento and GStreamer pull requests so they can all share the identical CEE schema.

Tested like this:

```
make

src/kamailio -f ./misc/examples/ims/pcscf/kamailio.cfg --log-engine=json:U -D -dddd -E
```

I copied one of the JSON log lines to a file and parsed it like this to check it is valid JSON:

```
python -m json.tool /tmp/kamailio.json 
{
    "appname": "kamailio",
    "file!line": 97,
    "file!name": "core/mem/pkg.c",
    "hostname": "ws.example.org",
    "msg": "destroying memory manager: q_malloc\n",
    "native!function": "pkg_destroy_manager",
    "pname": "kamailio",
    "pri": "DEBUG",
    "proc!id": "16264",
    "proc!tid": 140525261209856,
    "subsys": "core",
    "time": "2021-08-16T19:04:43.641230480Z"
}
```

### DIFF
--- a/src/core/dprint.c
+++ b/src/core/dprint.c
@@ -27,6 +27,7 @@
  */
 
 
+#include <pthread.h>
 #include <stdlib.h>
 #include <stdarg.h>
 #include <stdio.h>
@@ -451,6 +452,33 @@ static pv_elem_t *log_prefix_pvs = NULL;
 static char log_prefix_buf[LOG_PREFIX_SIZE];
 static str log_prefix_str = STR_NULL;
 
+void log_init(void)
+{
+	struct addrinfo hints, *info;
+	int gai_result;
+	char hostname[1024];
+
+	hostname[1023] = '\0';
+	gethostname (hostname, 1023);
+
+	memset (&hints, 0, sizeof (hints));
+	hints.ai_family = AF_UNSPEC;    /*either IPV4 or IPV6 */
+	hints.ai_socktype = SOCK_STREAM;
+	hints.ai_flags = AI_CANONNAME;
+
+	if ((gai_result = getaddrinfo (hostname, 0, &hints, &info)) != 0) {
+		log_fqdn = "?";
+	} else if (info == NULL) {
+		log_fqdn = "?";
+	} else {
+		log_fqdn = strdup (info->ai_canonname);
+	}
+
+	freeaddrinfo (info);
+
+	dprint_init_colors();
+}
+
 void log_prefix_init(void)
 {
 	str s;
@@ -611,6 +639,10 @@ static void ksr_slog_json_str_escape(str *s_in, str *s_out, int *emode)
 	" \"" NAME ".line\": %d, \"" NAME ".function\": \"%s\", \"" NAME ".callid\": \"%.*s\", \"" NAME ".logprefix\": \"%.*s\"," \
 	" \"%smessage\": \"%.*s\" }%s"
 
+#define KSR_SLOG_JSON_CEEFMT "{\"time\":\"%s.%09luZ\",\"proc!id\":\"%d\",\"proc!tid\":%ju,\"pri\":\"%s\",\"subsys\":\"%s\"," \
+        "\"file!name\":\"%s\",\"file!line\":%d,\"native!function\":\"%s\",\"msg\":\"%s\"," \
+        "\"pname\":\"%s\",\"appname\":\"%s\",\"hostname\":\"%s\"}%s"
+
 void ksr_slog_json(ksr_logdata_t *kld, const char *format, ...)
 {
 	va_list arglist;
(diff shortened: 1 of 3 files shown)


henningw:
Thanks for the pull request. Just a few notes, other might be able to comment more:
- instead of extending the core and adding another command line parameter for it, I think it should be implemented as a extension module, similar e.g. to log_systemd module
- the code includes pthread.h and also calls pthread_self(), this should be not necessary as there is no pthread_create() in the core right now (Kamailio uses mainly the multi-process model)

dpocock:
> Thanks for the pull request. Just a few notes, other might be able to comment more:
> 
>     * instead of extending the core and adding another command line parameter for it, I think it should be implemented as a extension module, similar e.g. to log_systemd module

This can make it more difficult to log unusual things during the startup phase before the modules are loaded.

There is already JSON support in core so I feel there is a good reason to keep the JSON code together in the same place.  The ```log_systemd``` code is much more different.

There is also a possibility that this code could replace the existing JSON code.  That would be easier to maintain but it would break compatibility for anybody who already depends on the old JSON schema.

> 
>     * the code includes pthread.h and also calls pthread_self(), this should be not necessary as there is no pthread_create() in the core right now (Kamailio uses mainly the multi-process model)

If any of the extension modules or libraries start threads then it is useful to capture the thread ID.  However, the field is optional, it can be completely removed and the rest of the pull request would still be valid without this.

henningw:
You are right, there is json support in the core. My impression was this PR was more for a specific dialect of it, that might not be used from a lot of people, therefore the remark with the module. But as I did not implemented the core json part, others might be able to comment more.
Regarding the thread_id - understood. Right now there is only the lwsc module that uses pthread_create (i was not aware that we are using threads at all yet). The remark with the libraries is a valid point.

dpocock:
> My impression was this PR was more for a specific dialect of it, that might not be used from a lot of people, therefore the remark with the module

I'm proposing the same CEE schema for Asterisk and other projects.

If all of us use the same schema then it will be more useful when troubleshooting problems that span multiple processes

https://issues.asterisk.org/jira/browse/ASTERISK-29517

There are many third-party documents that help explain CEE, for example, the blogs from the ```rsyslog``` developers.  Therefore, we can use those documents as a reference and save a lot of documentation effort.

miconda:
The `--log-engine` cli parameter exists in the core, this adds just a new attribute for it, `U` (which I can't relate to CEE schema, but it's fine for me, it has to be documented in the core cookbook in the wiki), so overall that is ok.

The pthread use probably needs some adjustments. If kamailio is compiled on a system with libssl 1.1+ (and pkg-config available), then the core is linked against libpthread in order to overwrite the mutex functions. But on old distros, with libssl 1.0, or distros without libssl installed, the core is not linked against libpthread and I expect to result in missing symbols. libssl1.1 is in most of latest stable Linux distros, but some may still use older ones (debian 9 or centos 7 probably have libssl 1.0 - I have no such system at hand to check right now).

Solutions I see now:

  * link core always against libpthread
  * make a module dependent of libpthread that sets a callback in the core to return the thread id -- who wants to have the thread id, loads this module. Such module can bring other features, like launching a thread from config to trigger a notification to external systems (e.g., do send_udp() from corex module)

dpocock:
The ```U``` can be for ```unified``` or ```universal``` logging as it will hopefully unify the logging schema from multiple projects in the VoIP and RTC ecosystem.  The letter ```c``` was already taken.

I can amend the patch to check for ```#ifdef HAVE_PTHREAD``` or something similar.  If ```pthread``` is not linked then it will simply send the JSON without the ```proc!tid``` value.

miconda:
You can push another commit to enclose getting thread id in some ifdef/endif -- that can be the easiest to merge it now, and it will be worked out in the future to improve it in a way or another and remove the ifdef/endif.

dpocock:
thanks for the feedback, I've added the logic for ```HAVE_PTHREAD``` and I also decided to stop using the ```proc!id``` notation and use nested objects instead, here is an example from my testing:

```
$ python3 -m json.tool /tmp/k1.json 
{
    "time": "2021-08-27T23:38:46.639132214Z",
    "proc": {
        "id": "56513"
    },
    "pri": "CRITICAL",
    "subsys": "core",
    "file": {
        "name": "core/cfg.y",
        "line": 3686
    },
    "native": {
        "function": "yyerror_at"
    },
    "msg": "parse error in config file foo/kamailio/kamailio/./misc/examples/ims/pcscf/kamailio.cfg, line 481, column 92: unknown command, missing loadmodule?\n\n",
    "pname": "kamailio",
    "appname": "kamailio",
    "hostname": "some-host.example.org"
}
```

miconda:
Thanks, merging!

PR #2832, opened 2021-08-23
lcr: source port check for from_any_gw() and from_gw().

zenichev:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
A simple change, which allows to check, if a coming request matches
not only by an IP address and a transport protocol,
but also using the source port of a request.

Improvements are done for the following functions of the lcr module:
- from_any_gw()
- from_gw()

Back compatibility provided, hence it does not affect already existing behavior.


juha-h:
I haven't tested the new feature, but from backwards compatibility point of view, the implementation looks good to me.  Thus I have nothing against the merge.

PR #2838, opened 2021-08-30
slack: new module, send message to slack channel

arsperger:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
this is a tiny module which provides integration with slack over webhooks. currently it exports only one function which accepts formatted string as an argument, evaluates pseudo-variables if any and sends the msg to slack channel (channel name, username and emoji are set as a module parameters). Kemi function is exported as well.


src/modules/slack/slack.c
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	// TODO: parse webhook to multiple channels? eg.: chan1=>https://AAA/BBB/CC, chan2=>...
+
+	slack_url = (char*)pkg_malloc(len + 1);
+	if (slack_url==NULL) {
+		PKG_MEM_ERROR;
+		return -1;
+	}
+	strcpy(slack_url, val);
```
henningw, line 177: would be good to use strncpy, as you already have the len.

src/modules/slack/slack.c
```diff
@@ -0,0 +1,298 @@
...
+	CURLcode res;
+	// LM_DBG("sending to[%s]\n", uri);
+
+	datasz = snprintf(NULL, 0, BODY_FMT, slack_channel, slack_username, post_data->s, slack_icon);
+	send_data = (char*)pkg_malloc((datasz+1)*sizeof(char));
+	if(send_data==NULL) {
+        LM_ERR("Error: can not allocate pkg memory [%d] bytes\n", datasz);
+        return -1;
+    }
+    snprintf(send_data, datasz+1, BODY_FMT, slack_channel, slack_username, post_data->s, slack_icon);
```
henningw, line 120: see above

src/modules/slack/slack.c
```diff
@@ -0,0 +1,298 @@
...
+
+
+/**
+ * Exported parameters
+ */
+static param_export_t mod_params[] = {
+	{ "slack_url", 			PARAM_STRING|USE_FUNC_PARAM, (void*)_slack_url_param},
+	{ "channel",			PARAM_STRING, &slack_channel }, // channel starts with #
+	{ "username",			PARAM_STRING, &slack_username },
+	{ "icon_emogi",			PARAM_STRING, &slack_icon },
```
henningw, line 55: I think the parameter _icon_emogi_ is mispelled, i think it should be named _icon_emoji_. Would be good to fix this in code and docs.

henningw (commented):
Thanks for the pull request. I just had a quick look and added a few comments, related to parameter naming and error checking.

src/modules/slack/slack.c
arsperger: thank you for your notes. I am wondering, if the above snprintf check is passed, what could be a reason for this one to fail? do we really need to error check this one?

henningw: This of course depends on the working of snprintf, the docs unfortunately do not give much more details. But generally it is good "defensive" programming practice to check for errors on outside library calls that you do not control/own.

oej:
We have too many modules who call the curl library, which can lead to issues with the TLS initialisation. This is why we created an API in the curl module. Please look into using that API instead of calling the curl library directly.

miconda:
The snprintf() for getting datasz should be checked, because can fail on unexpected characters - from the manual:

```
[EILSEQ]           An invalid wide character code was encountered.
```

That will make it safer when wanting to send data taken from SIP traffic (e.g., body of requests).

On negative return code, likely pkg_malloc() will fail, requested size being big.

The second snprintf() is unlikely to fail, but I would use pkg_mallocxz() or set send_data[0] = '\0' before it so in case it fails for what so ever unknown reason, then the CURL library doesn't get a buffer with garbage and does not go beyond allocated buffer to find the '\0'.

For reusing http_client, it would be good to review all modules. Can be done for this one also after merging, it is easier to work directly in repo than as PR.

miconda:
Thanks @arsperger! I am going to merge the PR and invite you to get commit access so you can develop and maintain the module in Kamailio repo. You can still do pull requests for this module, if you prefer it.

If you want to push code to other components or module, it is highly recommended to do pull requests, so other developers can review and be sure there is no conflict with ongoing work on those components.

Do consider using the API provided by http_client module as suggested by @oej -- we should do it for the other modules as well.

PR #2841, opened 2021-09-02
sipt: added one function and 8 pseudo-variables

MikhailKalashnikov:
- sipt_has_isup_body()
- $sipt(redirection_information.reason)
- $sipt(redirection_information.original_reason)
- $sipt(redirecting_number.nai)
- $sipt(redirecting_number)
- $sipt(original_called_number.nai)
- $sipt(original_called_number)
- $sipt(generic_number.nai)
- $sipt(generic_number)

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/modules/sipt/sipt.c
+++ b/src/modules/sipt/sipt.c
@@ -371,9 +371,9 @@ static int sipt_get_redirection_info(struct sip_msg *msg, pv_param_t *param, pv_
 		return -1;
 	}
 
-	if((body.s[0] != ISUP_ACM) && (body.s[0] != ISUP_CPG) && (body.s[0] != ISUP_IAM))
+	if((body.s[0] != ISUP_ACM) && (body.s[0] != ISUP_CPG))
 	{
-		LM_DBG("message not an ACM or CPG or IAM\n");
+		LM_DBG("message not an ACM or CPG\n");
 		return -1;
 	}
 	
--- a/src/modules/sipt/sipt.c
+++ b/src/modules/sipt/sipt.c
@@ -47,14 +47,27 @@ static int sipt_get_hop_counter(struct sip_msg *msg, pv_param_t *param, pv_value
 static int sipt_get_event_info(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
 static int sipt_get_cpc(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
 static int sipt_get_calling_party_nai(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
+static int sipt_get_calling_party(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
 static int sipt_get_presentation(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
 static int sipt_get_screening(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
 static int sipt_get_called_party_nai(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
+static int sipt_get_called_party(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
 static int sipt_get_charge_indicator(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
 
 static int sipt_get_redirection_info(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
 static int sipt_get_redirection_number_nai(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
 static int sipt_get_redirection_number(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
+
+static int sipt_get_redirection_reason(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
+static int sipt_get_original_redirection_reason(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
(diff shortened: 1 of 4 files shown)


miconda:
@tsearle - would you like to check this PR for sipt module?

henningw:
Thanks for the pull request. I did a review and generally things are looking good.
One question, you changed the implementation and (it seems also documentation) for the sipt_get_redirection_info(..) function. Was this a bug fix or a refactoring motivated from the added functionality from your PR?

MikhailKalashnikov:
@henningw I did not change sipt_get_redirection_info behavior. It returns the same result.
I only fixed documentation to describe actual behavior.

The old documentation says that redirection_info `Returns redirection info header...`
But in fact it does not get information from redirection info header ([ISUP_PARM_REDIRECTION_INFO](https://github.com/kamailio/kamailio/blob/master/src/modules/sipt/ss7.h#L121)), but gets from diversion info ([ISUP_PARM_DIVERSION_INFORMATION](https://github.com/kamailio/kamailio/blob/master/src/modules/sipt/ss7.h#L138))
You can check this link to see that another ISUP field was used
https://github.com/kamailio/kamailio/blob/master/src/modules/sipt/ss7_parser.c#L350

henningw:
I was referring to line 374/376 of the code:

```
	if((body.s[0] != ISUP_ACM) && (body.s[0] != ISUP_CPG) && (body.s[0] != ISUP_IAM))
	{
		LM_DBG("message not an ACM or CPG or IAM\n");
		return -1;
	}

```

MikhailKalashnikov:
@henningw Thank you for this point, It was my mistake, because the old documentation of sipt_get_redirection_info provided incorrect information, I first made it for IAM, and then realized that it returns different field, but did not rollback this change. 
I fixed it now

henningw:
@MikhailKalashnikov thanks. Two more things ;-)

- please check also line 904 in the patch, here is another change probably related to this
- please document the new added PVs in the wiki as well, there is already a sipt section: [link](https://www.kamailio.org/wiki/cookbooks/devel/pseudovariables)

Thank you.

MikhailKalashnikov:
@henningw 
- In this line 904. Behavior remains the same, because I also corrected mapping for the old [REDIRECTION_INFO](https://github.com/SIfoxDevTeam/kamailio/commit/4e6cdcfd54fc15dd9db10a3c08cb7fe9743a37ab#diff-9490a688e57f40dfedd7efae72fdc4ae90439a93e28b6fdbeaa3f3e9ae02c586R118) PV to make it clear that it returns ISUP_PARM_DIVERSION_INFORMATION.


- Could you give me (username: mish_gun) a write permission to edit wiki?

henningw:
@MikhailKalashnikov - i see you refactored it on two places so the behaviour stays the same.
Regarding the wiki, do you already tried to [register](https://www.kamailio.org/wiki/cookbooks/devel/pseudovariables?do=register)?

MikhailKalashnikov:
@henningw yes. I've registered with username `mish_gun`. 
and waiting for write permissions

henningw:
@MikhailKalashnikov thanks, your user should have now write permissions (before it was open for everybody, but we just got to much spam in the past on the wiki)

src/modules/sipt/sipt.c
```diff
@@ -102,10 +115,30 @@ static sipt_header_map_t sipt_header_mapping[] =
 		{{"CHARGE_INDICATOR", 1}, 
 			{NULL, 0}
 		}},
-        {"REDIRECTION_INFO", ISUP_PARM_REDIRECTION_INFO, 
-                {{NULL, 0}}  },
-        {"REDIRECTION_NUMBER", ISUP_PARM_REDIRECTION_NUMBER, 
-            		{{"NATURE_OF_ADDRESS", 1}, 
+	{"REDIRECTION_INFO", ISUP_PARM_DIVERSION_INFORMATION,
```
tsearle, line 118: wouldn't it be better to keep "REDIRECTION_INFO" as is
and add
"DIVERSION_INFO" as an extra parameter?

henningw: @tsearle It seems to be just a refactoring, have a look to the line 904 of the sipt.c diff/patch. It still returns the same function.

MikhailKalashnikov: This existing "REDIRECTION_INFO" actually gets information from diversion info (ISUP_PARM_DIVERSION_INFORMATION) and not from redirection info header (ISUP_PARM_REDIRECTION_INFO).
So this old name is incorrect. It would be better to rename the old "REDIRECTION_INFO" to "DIVERSION_INFO". 
But it will break existing behavior. If someone uses this old "REDIRECTION_INFO" he will get wrong results.
So I did not change existing behavior, only corrected documentation of "REDIRECTION_INFO"

tsearle:
seems ok for me

MikhailKalashnikov:
I updated wiki page with PVs

henningw:
Thank you, merged

MikhailKalashnikov:
@henningw Is it possible to backport this PR to a stable branch (to 5.5)?

henningw:
Its not possible, sorry, as this is a new functionality which should not be introduced into the stable branch. You could compile a 5.5.x version and apply the patch locally.

PR #2848, opened 2021-09-09
core: logging: JSON: CEE: add the @cee prefix in Syslog messages

dpocock:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

The JSON CEE output from ```stderr``` is OK

When the messages are sent through Syslog, they need this extra prefix ```@cee:``` before the JSON string

Tested with ```rsyslog-omelasticsearch``` and [OpenSearch Dashboard (replacement for Kibana)](https://opensearch.org/)

```
module(load="mmjsonparse")

# we handle LOG_LOCAL5 only, use the same setting in the application config
local5.* :mmjsonparse:

template(name="isJSON" type="list") {
  property(name="$!all-json")
}

module(load="omelasticsearch")
local5.* action(type="omelasticsearch"
           template="isJSON"
           server="some-server"
           serverport="9200"
           searchIndex="log"
           searchType="_doc/"
           uid="admin"
           pwd="admin")
```

### DIFF
--- a/src/core/dprint.c
+++ b/src/core/dprint.c
@@ -652,6 +652,10 @@ static void ksr_slog_json_str_escape(str *s_in, str *s_out, int *emode)
         "\"file\":{\"name\":\"%s\",\"line\":%d},\"native\":{\"function\":\"%s\"},\"msg\":\"%s\"," \
         "\"pname\":\"%s\",\"appname\":\"%s\",\"hostname\":\"%s\"}%s"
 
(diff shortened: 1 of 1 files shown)


dpocock:
The prefix is [in the specification here](https://cee.mitre.org/language/1.0-beta1/clt.html#cee-event-flag) and it is used in all other implementations and examples that I've seen.

henningw:
Thanks, looks good and its a small change, merge.

PR #2859, opened 2021-09-17
lcr: improve binary search to support a match including src port

zenichev:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [X] Tested changes locally

#### Description
Improve binary search in the lcr module and add a possibility
to do a matching not only based on an IP address of a GW, but also using a source port.

When a possibility to use 'src_port' parameter in from_gw() and from_any_gw()
was introduced here: 14e6fc80b3d2389567c73c4a2196bf8e6d92d8d2
the bsearch() remained untouched, and hence the matching (iteration through existing GWs)
is now done only based on an IP address.

This leads to the issue, when there are more than one GW with the same IP address in gws table,
and from_gw() and from_any_gw() functions are used with the 'src_port' parameter,
it can happen that a wrong GW is picked out by bsearch() from gws table (lcr_gw) and
a check by from_gw() and from_any_gw() returns False.

Hence the matching based on IP address and source port is required for bsearch(),
when from_gw() and from_any_gw() functions are used with the 'src_port' parameter.

This means backwards compatibility is still present (when one uses functions without 'src_port').

### DIFF
--- a/src/modules/lcr/lcr_mod.c
+++ b/src/modules/lcr/lcr_mod.c
@@ -3035,8 +3035,7 @@ static int do_from_gw(struct sip_msg *_m, unsigned int lcr_id,
 
 	/* Store tag and flags and return result */
 	if((res != NULL)
-					&& ((transport == PROTO_NONE) || (res->transport_code == transport))
-					&& ((src_port == 0) || (res->port == src_port))) {
+					&& ((transport == PROTO_NONE) || (res->transport_code == transport))) {
 		LM_DBG("request came from gw\n");
 		if(tag_avp_param) {
 			val.s.s = res->tag;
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the pull requests, looks ok on a first look to me. Lets wait some a few more days if there are more comments or remarks.

juha-h:
After the binary match change, is it so that there is no need to change this test?
```
/* Store tag and flags and return result */
	if((res != NULL)
					&& ((transport == PROTO_NONE) || (res->transport_code == transport))
					&& ((src_port == 0) || (res->port == src_port))) {
```

zenichev:
> After the binary match change, is it so that there is no need to change this test?
> 
> ```
> /* Store tag and flags and return result */
> 	if((res != NULL)
> 					&& ((transport == PROTO_NONE) || (res->transport_code == transport))
> 					&& ((src_port == 0) || (res->port == src_port))) {
> ```

Yes, you're right. Since the updated bsearch() already has ports comparison,
I think the following part looks excessive (doesn't break anything though):
`&& ((src_port == 0) || (res->port == src_port))) {`

I'll remove it.

juha-h:
Indentation does not look good  in the if statement.

henningw:
I'll merge the request and fix the white-space directly in the repository.

PR #2868, opened 2021-10-04
tm: Add callid_cseq_matching param

smititelu:
Enable transaction matching using callid and cseq values
instead of via md5 value.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Hello,

We had an issue with kamailio that the generated tm p_cell->md5 was not sent out in Via "branch=" parameter. Instead, other value was sent. Thus when comparing the replies Via "branch=" md5 value back to p_cell->md5, the transactions mismatched and dialogs were not setup. This happened for the most of the transactions/dialogs.

We could not really figure out why the above was happening. Because this was an isolated system, we decided to do this workaround using callid and cseq matching for transactions, and it helped. Now most of the dialogs are setup.

Let me know what you think, if other had this problem and if this PR helps.

Thank you,
Stefan

### DIFF
--- a/src/modules/tm/config.c
+++ b/src/modules/tm/config.c
@@ -47,6 +47,7 @@ struct cfg_group_tm	default_tm_cfg = {
 	1,	/* ruri_matching */
 	1,	/* via1_matching */
 	0,	/* callid_matching */
+	0,	/* callid_cseq_matching */
 	FR_TIME_OUT,	/* fr_timeout */
 	INV_FR_TIME_OUT,	/* fr_inv_timeout */
 	INV_FR_TIME_OUT_NEXT, /* fr_inv_timeout_next */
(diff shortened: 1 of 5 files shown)


miconda:
The indentation inside `src/modules/tm/t_lookup.c` does not seem right -- it should be consistent with the existing content in order to be easy to read the changes and figure out the impact. Please fix it (use tabs, not whitespaces).

smititelu:
Updated t_lookup.c with tabs indentation.

miconda:
Merging, the new modparm controls the behaviour.

henningw:
Is this matching actually compliant to the [standard](https://datatracker.ietf.org/doc/html/rfc3261#section-17.2.3)? I just wondered, it should be probably documented in the README then.

smititelu:
No, this is non-RFC compliant. I will update the tm doc for this new param and push it.

henningw:
thanks!

PR #2869, opened 2021-10-05
core: allow empty matches

btriller:
Anchors like ^ and $ match beginning resp. ending of strings but their
matches are zero length, which is ok.
The anchors can be used to pre-/append a prefix/suffix via e.g.
subst_hf("X-My-Header", "/^/prefix /", "a");
subst_hf("X-My-Header", "/$/ suffix/", "a");

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

### DIFF
--- a/src/core/data_lump.c
+++ b/src/core/data_lump.c
@@ -299,7 +299,7 @@ struct lump* del_lump(struct sip_msg* msg, int offset, int len,
 		return 0;
 	}
 	if (len==0){
-		LM_WARN("0 len (offset=%d)\n", offset);
+		LM_DBG("0 len (offset=%d)\n", offset);
 	}
 
 	tmp=pkg_malloc(sizeof(struct lump));
--- a/src/core/re.c
+++ b/src/core/re.c
@@ -440,10 +440,6 @@ struct replace_lst* subst_run(struct subst_expr* se, const char* input,
 				LM_ERR("Unknown offset?\n");
 				goto error;
 			}
-			if (pmatch[0].rm_so==pmatch[0].rm_eo) {
-				LM_ERR("Matched string is empty, invalid regexp?\n");
-				goto error;
-			}
 			*crt=pkg_malloc(sizeof(struct replace_lst));
 			if (*crt==0){
 				PKG_MEM_ERROR;


miconda:
Patch is small, but I wonder what side effects it can bring to existing configs, this behaviour likely is there for almost 2 decades. Maybe a global param can be added to control it. Any other opinions?

henningw:
Yes, i also think it should be probably configurable, and probably default deactivated. If there is another way of achieving this (for example in the textops module, maybe this is an even better way instead of adding another core parameter and code path.

miconda:
@btriller - looking again at this one, wouldn't it be the same to do:

```
subst_hf("X-My-Header", "/^(.+)$/prefix \1/", "a");
subst_hf("X-My-Header", "/^(.+)$/\1 suffix/", "a");
```

instead of:

```
subst_hf("X-My-Header", "/^/prefix /", "a");
subst_hf("X-My-Header", "/$/ suffix/", "a");
```

miconda:
And as another idea, maybe adding new functions like suffix_hf()/prefix_hf() would be even better, not to rely on regex matching for these basic operation.

btriller:
> @btriller - looking again at this one, wouldn't it be the same to do:
> 
> ```
> subst_hf("X-My-Header", "/^(.+)$/prefix \1/", "a");
> subst_hf("X-My-Header", "/^(.+)$/\1 suffix/", "a");
> ```
> 
> instead of:
> 
> ```
> subst_hf("X-My-Header", "/^/prefix /", "a");
> subst_hf("X-My-Header", "/$/ suffix/", "a");
> ```

Knowing that
```
% echo blub|sed 's/^/prefix /'
prefix blub
% echo blub|sed 's/$/ suffix/'
blub suffix
```
works, it felt weird to write
```
subst_hf("X-My-Header", "/^(.)/prefix \1/", "a");
subst_hf("X-My-Header", "/(.)$/\1 suffix/", "a");
```
As most of the time, there are other ways to do it (above suggested or `remove_hf()`, looping over all headers and `append_hf()`).
I understand the concerns, so I don't mind if the PR is closed.

miconda:
Closing it then, in this form it changes how it used to work so far.

You can propose another PR that makes this new behaviour configurable or adds new functions to offer the functionality.

PR #2873, opened 2021-10-06
Expose the domain in certificate validation errors

taf2:
We ran into an issue recently with certificate validation and this modification to the source code helped us quickly identify the root cause.  Previously we had no way of seeing which domain was invalid.  I didn't like using malloc here but hopefully freeing it in the same context ensures no memory leak as well avoid any memory corruption... with the memcpy and '\0'... hope this helps...

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x ] Tested changes locally

### DIFF
--- a/src/modules/tls/tls_util.h
+++ b/src/modules/tls/tls_util.h
@@ -39,7 +39,11 @@ static inline int tls_err_ret(char *s, tls_domains_cfg_t **tls_domains_cfg) {
 	{
 		while((err = ERR_get_error())) {
 			ret = 1;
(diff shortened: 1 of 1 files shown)


miconda:
I am going to close this one, after a second look at it, seems to be specific to your case, because it looks at the default server profile only, also it does not check if server_name is set or not, because it is not a mandatory field.

taf2:
Ok thank you for checking .  This was very helpful for us to quickly isolate the cert issue… perhaps there is an opportunity to improve or add some additional logging mode for the tls module in the future

miconda:
I pushed a commit trying to get sni from ssl structure.

PR #2888, opened 2021-10-19
core: tcp - add alias for cinfo dst IP

grumvalski:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

As I reported to developers' mailing list (https://lists.kamailio.org/pipermail/sr-dev/2021-October/064243.html) if support for proxy protocol is enabled, kamailio is not able to match a tcp DEST_IP/DEST_PORT to an alias of an already existing tcp connection. 
This beacause the (local) destination ip on which the connection was initiated is overwritten by the destination IP and port carried by the proxy protocol header. Since the original connection info is saved when parsing the protocol, we can use that information to build a new alias and to use it to match an outgoing tcp connection later.


henningw:
Thanks for the pull request. You probably considered that, but would it not make sense to only add the additional two aliases if the proxy protocol ("tcp_accept_haproxy") is actually enabled?

Issue #2909, opened 2021-10-29
permissions: crash during a RPC reload of trusted


linuxmaniac:
### Description

We are having several customers with crashes during calls to RPC `permissions.trustedReload` even after introducing the basic safety for concurrent RPC reload https://github.com/kamailio/kamailio/commit/c87f5202d778cf110ce0fbbc554310c148e997cb

### Possible solution

It seems that we need to implement locking in order to be sure the memory is not gone in the middle of a read.

#### Reproduction

It's quite difficult to reproduce since `match_hash_table` call has to happen during a RPC reload 

#### Debugging Data

```
(gdb) bt full
#0  match_hash_table (table=<optimized out>, msg=msg@entry=0x7fc3d800a808, src_ip_c_str=src_ip_c_str@entry=0x7fc3d7f217b8 "A.A.A.A", proto=proto@entry=1,
    from_uri=from_uri@entry=0x7fc3d7f1b750 "sip:fake@B.B.B.B") at hash.c:280
        __func__ = "match_hash_table"
        ruri = <optimized out>
        ruri_string = "sip:41584006001@B.B.B.B:5060\000\264{\224\377\177\000\000\000\240\200\331\302\177\000\000\300\000\000\000\000\000\000\000\332\003\000\000\000\000\000\000\300\067~\372PV\000\000\000\220
\324m,\233߁\260\341D\334\302\177\000\000\260go\372PV\000\000P\343!\332\302\177\000\000\004\340{\372PV\000\000`;~\372PV\000\000\300\067~\372PV\000\000@\024\000\000\000\000\000\000`\342!\332\302\177\000\000\001\00
0\000\000\000\000\000\000\310\066\204\331\302\177\000\000\000\240\200\331\302\177\000\000\300\000\000\000\000\000\000\000\300\067~\372PV\000\000`;~\372PV\000\000\332\003\000\000\000\000\000\000\000"...
        preg = {__buffer = 0x0, __allocated = 0, __used = 224, __syntax = 16843462, __fastmap = 0x0, __translate = 0x0, re_nsub = 0, __can_be_null = 0, __regs_allocated = 0, __fastmap_accurate = 1,
          __no_sub = 1, __not_bol = 0, __not_eol = 0, __newline_anchor = 0}
        np = 0x455449564e492035
        src_ip = {s = 0x7fc3d7f217b8 "A.A.A.A", len = <optimized out>}
        val = <optimized out>
        count = 0
[...]

PR #2910, opened 2021-10-29
permissions: fix crash during RPC trusted reload

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [ ] Related to issue #2909

### DIFF
--- a/src/modules/permissions/doc/permissions_admin.xml
+++ b/src/modules/permissions/doc/permissions_admin.xml
@@ -873,6 +873,27 @@ modparam("permissions", "reload_delta", 1)
 		</programlisting>
 		</example>
 	</section>
+	<section id ="permissions.p.trusted_cleanup_interval">
+		<title><varname>trusted_cleanup_interval</varname> (int)</title>
+		<para>
+		If module parameter <varname>db_mode</varname> is set to 1 a cleanup timer
+		is set to the trusted_cleanup_interval value in seconds.
+		This process will cleanup the previous values from memory after a RPC reload.
+		</para>
+		<para>
+		<emphasis>
+			Default value is <quote>60</quote>.
+		</emphasis>
+		</para>
+		<example>
+		<title>Set <varname>trusted_cleanup_interval</varname> parameter</title>
+		<programlisting format="linespecific">
+...
+modparam("permissions", "trusted_cleanup_interval", 120)
+...
+		</programlisting>
+		</example>
+	</section>
 	</section>
 
 	<section>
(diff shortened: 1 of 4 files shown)


miconda:
I am not using permissions with trusted table, but it looks like now it's going to be no parallel use of the in-memory table. It's safe against reloads done quickly one after the other, but all kamailio processes that do permissions check wait one after the other always.

If one ensures the reloads are not triggered quickly one after the other, then the patch is slowing down, specially if the number of records is high. An option would be to add a mod param to enable this lock or not at startup.

Anyhow, as I said, I am not using this feature, so I can't evaluate the impact. Maybe @juha-h can comment if he uses this feature, or some other devs.

linuxmaniac:
Yes, you're right. I didn't realize I was destroying parallel use of the hash. I was focusing on be sure the memory was not going to go away in the middle of a search. I need to find a better solution.

miconda:
Iirc, with other rpc commands related to db reload it was introduced a "safety delay" for them, so they cannot be done very often.

linuxmaniac:
Yes, I implemented the delta parameter [0] but that doesn't help if you're in the middle of a search

https://github.com/kamailio/kamailio/commit/01fa7503433f4e23dea156e1034a87bbb49a4a1d

miconda:
Some modules keep also the old data till next reload (dispatcher ?!?), I thought is the case also here.

Then the solution could be to have a list of old hash tables that are destroyed on timer, after some configurable delay. Practically, instead of destroying immediately, add it to the list with a timestamp, and destroy after N secs with a timer callback.

linuxmaniac:
Yes, that would be a nice approach. I'm going to implement it like that.

Thank you!

miconda:
I guess there is no possibility to perform another reload as long as the old hash table is not freed on timer. If yes, then fine for me, the new modparm has to be documented, if not done.

linuxmaniac:
> I guess there is no possibility to perform another reload as long as the old hash table is not freed on timer. If yes, then fine for me

[here](https://github.com/kamailio/kamailio/pull/2910/commits/7ce815c2612d6277b55aabf6519c70a4ecec6502#diff-ce4f68a3bf731efbaaa1d243de3f81a8c6a068fb2c8693195558dbda1b4bb389R228-R230)  I check that no reload was done before last tick.

if a reload is called before the timer interval, nothing changed. [reload_trusted_table](https://github.com/kamailio/kamailio/blob/7ce815c2612d6277b55aabf6519c70a4ecec6502/src/modules/permissions/trusted.c#L94-L100) It was already removing the old previous values before loading new values there before the switch.


> the new modparm has to be documented, if not done.

To be added if we agree in the implementation

miconda:
Docs can be added and then merged.

PR #2915, opened 2021-11-03
nats: Add publish function

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Added a publish function to nats module. When the function is called, the routing process tries to connect to the url (param) and tries to send the payload (param) to the subject queue (param). No libuv is used for this.

Probably a better approach is to make publisher workers (like subscriber workers) and pass payload from routing processes to publisher worker processes via shared memory. Right now I see this as an improvement and find the new function good enough, for now.

@eschmidbauer what do you think of this?

Thank you,
Stefan

P.S. Will update doc after review.

### DIFF
--- a/src/modules/nats/nats_mod.c
+++ b/src/modules/nats/nats_mod.c
@@ -32,6 +32,10 @@ nats_consumer_worker_t *nats_workers = NULL;
 int _nats_proc_count;
 char *eventData = NULL;
 
+static int nats_publish_1_f(struct sip_msg * msg, char *payload);
+static int nats_publish_2_f(struct sip_msg * msg, char *payload, char *url);
+static int nats_publish_3_f(struct sip_msg * msg, char *payload, char *url, char *subj);
+
 static pv_export_t nats_mod_pvs[] = {
 		{{"natsData", (sizeof("natsData") - 1)}, PVT_OTHER,
 				nats_pv_get_event_payload, 0, 0, 0, 0, 0},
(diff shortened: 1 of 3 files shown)


eschmidbauer (commented):
it looks like this code will connect to NATS server each publish request. I think it would be better to handle the connections outside of the `nats_publish_f` function to avoid having to connect each request.

smititelu:
Updated PR to connect to urls while parsing the given "nats_url" modparams. Also the new function now supports either 1, 2 or 3 params; url and subject are now optional; it will publish to all urls and all subject queues if none given.

Basically tested it with 3 local nats servers. And calling the new config function either with 1 param or 3 params.

Let me know what you think of this.

Thank you,
Stefan

smititelu:
force-pushed with doc update :D

Thanks,
Stefan

smititelu:
re-updated PR to add callbacks to publisher connections; right now the publisher callbacks are the same as the subscriber callbacks.

Thus, kamailio nats module can make use of nats lib reconnection(e.g. natsOptions_SetRetryOnFailedConnect() function) and not implement a reconnect mechanism by itself.

I will later make a new PR to make the parameters that control nats lib reconnect mechanism, configurable:
```
        natsOptions_SetMaxReconnect(opts, 10000); // how many times nats lib tries to reconnect before giving up; -1 is forever if i'm not wrong
        natsOptions_SetReconnectWait(opts, 2 * 1000); // the interval between reconnect retries
```

Thanks,
Stefan

eschmidbauer:
@smititelu this looks good, nice work and thank you! 
I would like to pull & test this branch if you don't mind waiting a few days (sorry I'm super swamped with work)

src/modules/nats/nats_mod.c
```diff
@@ -519,6 +614,7 @@ int init_nats_server_url_add(char *url)
 		n = n->next;
 	}
 	n = _init_nats_server_list_new(url);
+	_init_nats_server_conn(n);
```
eschmidbauer, line 617: this initiates connections to the servers but each subscriber still maintains a separate connection to each server. i think it would make sense to change all subscriber workers to use the same connections as publishers

eschmidbauer:
@smititelu I made some changes to better handle the NATS connections and share the connections between publishes and subscribers. This eliminates duplicate connection handling within the module and allows it to extend into independent connection handling (e.g. add/remove/label nats connections while kamailio is running) later.
I've submitted the PR to your fork [here](https://github.com/smititelu/kamailio/pull/2/files) so you can review and test.
It's not a common implementation to specify a nats url per publish request therefore the message publish will be sent to a single nats connection (which can container up to 10 nats urls). Also, I did not see the benefit of mixing queue group subjects with publish requests, therefore, I thought it best to also specify subject on publish request. I think there is more to be done here but mostly around testing and fixup of publish parameters. 
Please let me know your thoughts, and if you don't mind, let's move this conversation to your fork so we can discuss best path forward before submitting the code to kamailio repo.

smititelu:
Closing this, because #2978 has a better approach, using publisher workers.

Issue #2885, opened 2021-10-14
kemi python3: exception occur when import logging module


hnimminh:
### Description
kemi python3: exception occur when import logging module

#### Reproduction
very simple to reproduce is add this line to python kemi file
```
import logging
```

#### Debugging Data

```
(paste your debugging data here)
```

#### Log Messages

```
 0(49) INFO: <core> [core/tcp_main.c:4983]: init_tcp(): using epoll_lt as the io watch method (auto detected)
 0(49) WARNING: outbound [outbound_mod.c:121]: mod_init(): "stun" module is not loaded. STUN is required to use outbound with UDP.
 0(49) INFO: <core> [core/kemi.c:105]: sr_kemi_core_info(): KAMAILIO PYTHON3 APPLICATION INSTANTIATE CLASS OBJECT
 0(49) INFO: <core> [core/kemi.c:105]: sr_kemi_core_info(): KAMAILIO INITIATION
 0(49) INFO: <core> [main.c:3030]: main(): processes (at least): 25 - shm size: 268435456 - pkg size: 16777216
 0(49) INFO: <core> [core/udp_server.c:154]: probe_max_receive_buffer(): SO_RCVBUF is initially 212992
 0(49) INFO: <core> [core/udp_server.c:206]: probe_max_receive_buffer(): SO_RCVBUF is finally 425984
Exception ignored in: <function _releaseLock at 0x7fb7ed1d3f28>
Traceback (most recent call last):
  File "/usr/lib/python3.7/logging/__init__.py", line 226, in _releaseLock
    _lock.release()
RuntimeError: cannot release un-acquired lock
Ignoring exception from logging atfork <_StderrHandler <stderr> (WARNING)> release method: cannot release un-acquired lock
 1(50) INFO: <core> [core/kemi.c:105]: sr_kemi_core_info(): KAMAILIO CHILD PROCESS INITIALIZED 1
Exception ignored in: <function _releaseLock at 0x7fb7ed1d3f28>
Traceback (most recent call last):
  File "/usr/lib/python3.7/logging/__init__.py", line 226, in _releaseLock
Exception ignored in: <function _releaseLock at 0x7fb7ed1d3f28>
Traceback (most recent call last):
  File "/usr/lib/python3.7/logging/__init__.py", line 226, in _releaseLock
Exception ignored in: <function _releaseLock at 0x7fb7ed1d3f28>
Traceback (most recent call last):
[...]

PR #2919, opened 2021-11-05
core: add a module hook for post-fork cleanup

aalba6675:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #2885

#### Description
This PR provides a hook for modules when all `fork_process()`s are completed.
This is so that the module in the main process context can perform post-fork cleanup.

### DIFF
--- a/src/core/sr_module.h
+++ b/src/core/sr_module.h
@@ -144,6 +144,7 @@ typedef int (*param_func_t)( modparam_t type, void* val);
 #define PROC_FIFO      PROC_RPC  /**< FIFO attendant process */
 #define PROC_TCP_MAIN -4  /**< TCP main process */
 #define PROC_UNIXSOCK -5  /**< Unix socket server */
+#define PROC_POST_FORK -6  /**< Main process after all children are forked */
 #define PROC_ATTENDANT -10  /**< main "attendant process */
 #define PROC_INIT     -127 /**< special rank, the context is the main ser
 							  process, but this is guaranteed to be executed
--- a/src/main.c
+++ b/src/main.c
@@ -1489,6 +1489,14 @@ int main_loop(void)
 			goto error;
 		}
 
+
+		/* call it also w/ PROC_POST_FORK to make sure modules that cleanup
+		 * post-fork get a chance to run */
+		if (init_child(PROC_POST_FORK) < 0) {
+			LM_ERR("init_child(PROC_POST_FORK) -- exiting\n");
+			goto error;
+		}
+
 		/* We will call child_init even if we
 		 * do not fork - and it will be called with rank 1 because
 		 * in fact we behave like a child, not like main process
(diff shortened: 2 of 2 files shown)


aalba6675:
A user of this hook would be the app_python3 module: it needs to know when all the child processes are forked, then the main process would perform internal state cleanup. 

See the description of `PyOS_AfterFork_Parent`: https://docs.python.org/3/c-api/sys.html

aalba6675:
How this feature will be used in app_python3 - this gives a hook to call `PyOS_AfterFork_Parent`:
```
/**
 *
 */
static pid_t main_pid = 0; /* record pid_t of main process */

static int child_init(int rank)
{
	if(rank==PROC_INIT) {
		/*
		 * this is called before any process is forked
		 * so the Python internal state handler
		 * should be called now.
		 */
#if PY_VERSION_HEX >= 0x03070000
		PyOS_BeforeFork() ;
#endif
		main_pid = getpid();
		return 0;
	}
	if(rank==PROC_POST_FORK) {
		/*
		 * this is called after all processes are forked
		 * so the Python internal state handler
		 * should be called.
		 */
#if PY_VERSION_HEX >= 0x03070000
		PyOS_AfterFork_Parent() ;
#endif
		return 0;
	}
	_apy_process_rank = rank;

	/*
	 * If non-daemonize we are called in the main context
	 * with rank = PROC_SIPINIT
	 */
	if (main_pid != getpid()) {
		/* we really are a child process */
#if PY_VERSION_HEX >= 0x03070000
		PyOS_AfterFork_Child();
#else
		PyOS_AfterFork();
#endif
	}
	if (cfg_child_init()) {
		return -1;
	}
	return apy_init_script(rank);
}
```

miconda:
Introduction of another execution of init_child() with a different rank impacts all the modules, it may have side effects for modules that handle only the ranks expected so far. Therefore, even the patch is small, has to be analysed the impact.

An alternative would be to add another core event with arming in the mod_register().

I will think a bit about and see what would be the best.

aalba6675:
Unfortunately I found code snippets like this:

```
static int child_init(int rank)
{
        int i, j;
        int len;
        xhttp_rpc_mod_cmds_t *cmds;
        /* rpc_export_t *rpc_e; */

        if(rank==PROC_MAIN || rank==PROC_TCP_MAIN)
                return 0; /* do nothing for the main process */
```

So some modules are "aware" of which ranks refer to the main process.

aalba6675:
@miconda — have you had the chance to consider a better solution to this issue?
I am thinking of going through the various modules and see how they check for rank:

   /* ???? have an escape for a module if the module needs to do 
     * does something for PROC_INIT, PROC_MAIN and rank > 0 (and small negative integers PROC_TCP etc)
     * then do nothing and return here ???? */
  if (rank < PROC_MAIN_ONLY) return 0;  // PROC_MAIN_ONLY == -128 
  
Then to apply this PR I could set

    PROC_POST_FORK = -129
    // other new ranks can be created with values < -128 and all modules would not be affected

miconda:
@aalba6675 - had a busy week, I thought of introducing a mechanism that should allow modules to specify in mod_init() if they want to execute special types of ranks for child_init(), so it will not impact other modules at all (because I remember when PROC_INIT was introduced, we discovered some side effect long time after). I hope to find some time for it this week.

miconda:
I pushed the commit 7f89e4cc5c84b720928c2936c9e2e0630aebec9e for this purpose.

You have to execute `ksr_module_set_flag(KSRMOD_FLAG_POSTCHILDINIT)` inside `mod_init()` and then handle rank `PROC_POSTCHILDINIT` in `child_init()`.

Hoping it offers what is needed for this one.

aalba6675:
Thank! Closing with d2c22d2.

PR #2930, opened 2021-11-11
presence: add option to return 200 instead of 202 on subscriptions

lnicola:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

RFC 6665 [says](https://datatracker.ietf.org/doc/html/rfc6665#section-8.3.1):

> Implementations conformant with the current specification MUST treat an incoming 202 response as identical to a 200 response and MUST NOT generate 202 response codes to SUBSCRIBE or NOTIFY requests.

I'm not sure if Kamailio is supposed to conform to the older RFC, so this change might not be _acceptable_.

### DIFF
--- a/src/modules/presence/doc/presence_admin.xml
+++ b/src/modules/presence/doc/presence_admin.xml
@@ -792,7 +792,7 @@ modparam("presence", "subs_remove_match", 1)
 			so newer documents have higher priority.</para>
         </listitem>
         <listitem>
-			<para><emphasis>delete_subscription</emphasis> - integer value to 
+			<para><emphasis>delete_subscription</emphasis> - integer value to
 			give extra control of deleting the subscription after processing of
 			event_route[presence:notify-reply]. If value = 1, it deletes the subscription.
 			If xavp_cfg parameter is set but this attribute is not in the avp,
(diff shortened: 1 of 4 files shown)


lnicola:
CC @adita76 who asked me to file this. I don't know anything about SIP.

miconda:
I would suggest to make this a module parameter, so the current behaviour can be preserved and people can configure what response code to used.

henningw:
If there are no more comments, the documentation (XML) for the newly added parameter is missing and should be added before merging.

lnicola:
Added docs.

henningw:
Thanks for the additions. Just one small thing, could you please remove the README file from the PR. This is done from an automatic job in the repo to ensure consistency.

lnicola:
Done, sorry. I noticed some `Makefile` targets, but wasn't sure of the process.

henningw:
Thanks, merged. Being a small extensions, further work could be done in git master, if necessary.

lnicola:
@henningw if I backport this to the 5.5 branch, do you think it would be possible to cut a new release?

henningw:
Thanks for the changes. There will be a new release for 5.5.x next week. Regarding back-porting, technically its a new feature that should not be backported. On the other hand it fixes compliance with RFC 6665 for the module. @miconda what are your thoughts on this?

lnicola:
I filed a backport PR, we can discuss it there.

kamailio-sync:
Henning Westerholt writes:

> Thanks for the changes. There will be a new release for 5.5.x next
> week. Regarding back-porting, technically its a new feature that
> should not be backported. On the other hand it fixes compliance with
> RFC 6665 for the module. @miconda what are your thoughts on this?

RFC 6665 tells

   Implementations conformant with the current specification MUST treat
   an incoming 202 response as identical to a 200 response and MUST NOT
   generate 202 response codes to SUBSCRIBE or NOTIFY requests.

So should default behavior be 200 response.  At least new users of the
module would expect is it works according to standard out of the box.

lnicola:
I wasn't sure which RFC was implemented. Then we can change the default on `master` -- and _if_ we backport it to 5.5 we can leave the previous behavior as the default.

miconda:
I haven't checked the RFC 6665 in detail, but it is anyhow a new feature that updates the current behaviour,  which isbased on older specs, it is not a wrong implementation.

If decided to be backported to stable branches, then it has to preserve the current behaviour.

In master the default can be changed if people consider it better, and the docs updated to reflect that. I assume that many still run old phones for BLF, presence, ...

Regarding the backport, I am fine either way, what people prefer more. At the end we cannot backport implementations of new RFC always to stable branches. Here the patch is minimal, but as practice so far we always have to consider stability of released branches vs new/changed behaviour.

lnicola:
Filed #2949 to change the default.

Regarding the port to 5.5, is there anything we can do to push this forward?

PR #2931, opened 2021-11-12
TSILO / TM. Lookup and branch creating by contact (location record).

zenichev:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Commits are separated per component: TSILO and TM.
Important, this implementation has been tested internally and is in use by us already.

---

TSILO: Add support of a lookup and branch creating by contact

Improvement of the TSILO module, which allows to run a lookup using the provided RURI,
but only to create a new branch for the Contact, which is set in the currently processed REGISTER.
Or optionally a Contact URI value given as a parameter to the function.
If the Contact URI for a lookup, is given as a parameter,
it is possible to provide it as a pseudo-variable.

Hence it is now possible to append new branch(es) for only UAC(s)
getting REGISTERED at the moment of calling script function 'ts_append_by_contact()'.

Previously it was only possible to create new branches
for all previously present location records (for this specific URI), which was in some circumstances undesired.

New script functions introduced:
- ts_append_by_contact(table, ruri [, contact])

New module functions introduced:
- w_ts_append_by_contact2() / ki_ts_append_by_contact()
- w_ts_append_by_contact3 / ki_ts_append_by_contact_uri()

The documentation has been updated accordingly.

Tagging people for reviewing TSILO part: @grumvalski 

---

TM: API improvement, 't_append_branches()' improvement

This commit introduces a possibility of TM's API to append a branch
based on a specific location (Contact's URI).

The function 't_append_branches()' has been extended, and is now able
to obtain a parameter Contact of str type and takes care of appending a branch only for it.

The way how the function acts now depends on if the contact parameter of str type
is given or not (empty or not). If the contact parameter is given, then only a desired location
is meant for appending. If not found in the location table, an append will not happen for this AOR.
    
Otherwise create branches for all existing location records
of this particular AOR. Search for locations is done in the location table.

Tagging people for reviewing TM part: @miconda @juha-h

### DIFF
--- a/src/modules/tsilo/ts_append.h
+++ b/src/modules/tsilo/ts_append.h
@@ -24,5 +24,7 @@
 
 int ts_append(struct sip_msg* msg, str *ruri, char *table);
 int ts_append_to(struct sip_msg* msg, int tindex, int tlabel, char *table, str *uri);
+int ts_append_by_contact(struct sip_msg* msg, str *ruri, str *contact, char *table);
+int ts_append_by_contact_to(struct sip_msg* msg, int tindex, int tlabel, char *table, str *uri, str *contact);
 
 #endif
--- a/src/modules/tsilo/ts_append.h
+++ b/src/modules/tsilo/ts_append.h
@@ -22,9 +22,7 @@
 #ifndef _TS_APPEND_H
 #define _TS_APPEND_H
 
-int ts_append(struct sip_msg* msg, str *ruri, char *table);
-int ts_append_to(struct sip_msg* msg, int tindex, int tlabel, char *table, str *uri);
-int ts_append_by_contact(struct sip_msg* msg, str *ruri, str *contact, char *table);
-int ts_append_by_contact_to(struct sip_msg* msg, int tindex, int tlabel, char *table, str *uri, str *contact);
+int ts_append(struct sip_msg* msg, str *ruri, str *contact, char *table);
+int ts_append_to(struct sip_msg* msg, int tindex, int tlabel, char *table, str *uri, str *contact);
 
 #endif
--- a/src/modules/tm/t_append_branches.c
+++ b/src/modules/tm/t_append_branches.c
@@ -46,202 +46,18 @@
 #include "t_reply.h"
 #include "t_append_branches.h"
 
-int t_append_branches(void) {
-	struct cell *t = NULL;
-	struct sip_msg *orig_msg = NULL;
-	struct sip_msg *faked_req;
-	int faked_req_len = 0;
-
-	short outgoings;
-
-	int success_branch;
-
-	str current_uri;
-	str dst_uri, path, instance, ruid, location_ua;
-	struct socket_info* si;
-	int q, i, found;
-	flag_t backup_bflags = 0;
-	flag_t bflags = 0;
-	int new_branch, branch_ret, lowest_ret;
-	branch_bm_t	added_branches;
-	int replies_locked = 0;
-	int ret = 0;
-
-	t = get_t();
-	if(t == NULL)
-	{
-		LM_ERR("cannot get transaction\n");
-		return -1;
-	}
-
-	LM_DBG("transaction %u:%u in status %d\n", t->hash_index, t->label, t->uas.status);
-
-	/* test if transaction has already been canceled */
-	if (t->flags & T_CANCELED) {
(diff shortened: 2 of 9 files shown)


henningw:
Thank you for the pull request. Just one comment for the tm extension - this is really a lot of duplicated and non-trivial code for this particular function, which doubles maintenance tasks for bug fixes etc.. I assume the difference between the two functions (t_append_branch, t_append_branch_by_contact) is rather small? I would suggest move the common parts to one new function, and then call it from the old and the new one.

grumvalski:
Thank you for the PR, indeed an useful addition. I've read the code and it looks fine (I'll have another read), however I have the same remark as Henning about t_append_branch and t_append_branch_by_contact code sharing.

zenichev:
@henningw @grumvalski thanks for giving an advice. I will prepare an additional commit to improve this part.

zenichev:
@henningw @grumvalski what if we merge functions t_append_branches() and t_append_branch_by_contact() into one?
Hence leaving only the t_append_branches().

From what I can see, it seems for now it's only TSILO who directly uses t_append_branches(). Correct me please if I'm wrong.

Proposed approach assumes, that the function signature will be changed, and a function parameter 'contact' of str type will be present. Then depending on if it's empty or not, act accordingly:
- If the contact parameter is given, then only a desired location  is meant for appending. If not found in the location table, an append will not happen for this AOR.
- Otherwise create branches for all existing location records of this particular AOR.

I have prepared the implementation and even tested that locally,
it seems to be functioning just fine when either calling TSILO's script function ts_append() and ts_append_by_contact().

Please let me know if you see it reasonable, then I will upload a new commit.

henningw:
Extending the existing function (as you proposed) is probably a good idea, especially because that TSILO is the only one user of the function, as you also mentioned.

zenichev:
@henningw @grumvalski I have uploaded two additional commits, each one per module (tm and tsilo).
Please have a look when you have some time and let me know if there are other remarks.

As I mentioned this code has been tested locally and showed itself working.

grumvalski:
Thank you @zenichev. I've given another read at the changes on tsilo and, as for tm changes, I think that there is a lot of duplicated code between ts_append/ts_append_by_contact and ts_append_to/ts_append_by_contact. E.g. the only diffrerence between ts_append_to and ts_append_by_contact_to is just the way tm's t_append branches is called (with or without contact). 
I think that it could be cleaner to expand the current functions' signature to add an optional contact.

zenichev:
@grumvalski done, I've merged similar functionality in ts_append.c implementation.
tsilo.c implementation has been updated accordingly. Can you please check out and see if it can be merged like that?
I have tested usage of kamailio's script functions ts_append() and ts_append_by_contact() with a new code, and it worked fine.

P.S.: Previously I had an intention to keep as much back-compatibility as possible to let the new code be not overlapping with previously existing. 

---

@henningw TM's part had been updated previously as requested. Could you please have a look? Thank you.
(Please remember, it looks like t_append_branches() function is only used by TSILO now).

henningw:
The TM change looks fine to me now. As you mentioned its only used from tsilo module. Lets wait for some feedback from @grumvalski about the other part.

src/modules/tsilo/ts_append.h
```diff
@@ -22,7 +22,7 @@
 #ifndef _TS_APPEND_H
 #define _TS_APPEND_H
 
-int ts_append(struct sip_msg* msg, str *ruri, char *table);
-int ts_append_to(struct sip_msg* msg, int tindex, int tlabel, char *table, str *uri);
+int ts_append(struct sip_msg* msg, str *ruri, str *contact, char *table);
```
grumvalski, line 25: Any specific reason to have the contact arg inserted between ruri and table args, while in ts_append_to is put at the end?

grumvalski (approved):

grumvalski:
Looks good to me too, just a small comment on the code. Thank you @zenichev.

src/modules/tsilo/ts_append.h
zenichev: To be precise, the only reason was to keep it just after the ruri parameter.
So that in both cases with ts_append() and ts_append_to() I have put it unconsciously just after uri (ruri).

PR #2935, opened 2021-11-15
dialog: add local_only paramter to RPC dlg.stats_active

### LINKED ISSUE #2920: add dlg.stats_active parameter to filter dialogs that belongs to the server
### Description

If two kamailio servers share the same dialog db table, on restart, they both load all dialogues info stored in db into memory and
dlg.stats_alive returns the same amount of alive dialogues on both serves.

### Expected behavior

I would expect that dlg.stats_alive would return just the dialogues info that belongs to the server ( socket_info is not empty )

### Possible Solutions

Introduce an optional parameter to the RPC command to filter just the dialogues that belongs to the server

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #2920

#### Description
allows to filter out non local dialogs from stats using a new argument

fix #2920

### DIFF
--- a/src/modules/dialog/dialog.c
+++ b/src/modules/dialog/dialog.c
@@ -2858,16 +2858,21 @@ static void rpc_dlg_stats_active(rpc_t *rpc, void *c)
 {
 	dlg_cell_t *dlg;
 	unsigned int i;
+	int dlg_own = 0;
 	int dlg_starting = 0;
 	int dlg_connecting = 0;
 	int dlg_answering = 0;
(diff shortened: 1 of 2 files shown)


miconda:
OK for me.

henningw:
Only a small question regarding the docs:
"No bind_addr info in memory." - this maybe could a bit extended, as I do not understood it in this context.

miconda:
After merge, I added a few more details about bind_addr role.

PR #2936, opened 2021-11-17
pv: new variable $xavp_count

nchaigne:
Retrieve the number of rows stored in an xavp

As discussed on the mailing list, this is useful to at least two people.

#### Type Of Change
- [ x ] New feature (non-breaking change which adds new functionality)

#### Checklist:
- [ x ] PR should be backported to stable branches
- [ x ] Tested changes locally

### DIFF
--- a/src/modules/pv/pv.c
+++ b/src/modules/pv/pv.c
@@ -96,6 +96,9 @@ static pv_export_t mod_pvs[] = {
 	{ {"xavp", sizeof("xavp")-1}, /* xavp */
 		PVT_XAVP, pv_get_xavp, pv_set_xavp,
 		pv_parse_xavp_name, 0, 0, 0 },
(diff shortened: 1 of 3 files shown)


nchaigne:
> If I understand correctly, it works only for `$xavp(key)`, not for `$xavp(key1=>key2)`, right?
> 
> Also, I wonder if it is not better to enhance the `$cnt(...)` variable for this purpose, now it is for avps.


Yes… it works only with a single key. This is all I needed.

But you’re right, it would be better to also handle sub-keys.
That was a quick PR since I already had the code. Feel free to discard it. :)

nchaigne:
I'll do another PR using $cnt(...) as you suggested.

miconda:
OK, closing this one.

PR #2937, opened 2021-11-17
presence: add option to return 200 instead of 202 on subscriptions

lnicola:
Backport of #2930


henningw:
We do not pull-requests for backports, this is usually done with cherry-pick directly in the repo.

PR #2941, opened 2021-11-19
Fix a possible deadlock bug due to unreleased lock

### LINKED ISSUE #2939: Possible deadlock bug in the method acquire_spi and release_spi?
In the method `acquire_spi`, the condition of acquiring and releasing is not aligned. It seems the lock would be acquired but released only when `spi_data->spi_val == initial_val`, Would it lead to missing lock release?
https://github.com/kamailio/kamailio/blob/1ddc27f199061025a6a43da3e8a1388fcaddfe19/src/modules/ims_ipsec_pcscf/spi_gen.c#L83-L85
https://github.com/kamailio/kamailio/blob/1ddc27f199061025a6a43da3e8a1388fcaddfe19/src/modules/ims_ipsec_pcscf/spi_gen.c#L109-L111

In the method `release_spi`, the lock may be re-acquired so missing lock release in the method `acquire_spi` could lead to a deadlock.
https://github.com/kamailio/kamailio/blob/1ddc27f199061025a6a43da3e8a1388fcaddfe19/src/modules/ims_ipsec_pcscf/spi_gen.c#L126-L134

ryancaicse:
In acquire_spi, making condition of acquiring the releasing lock aligned.

Fixes  #2939

### DIFF
--- a/src/modules/ims_ipsec_pcscf/spi_gen.c
+++ b/src/modules/ims_ipsec_pcscf/spi_gen.c
@@ -106,9 +106,9 @@ uint32_t acquire_spi()
 			spi_data->spi_val = spi_data->min_spi;
         }
 
-		if(spi_data->spi_val == initial_val){ //there are no free SPIs
-			pthread_mutex_unlock(&spi_data->spis_mut);
-            return ret;
+
+		pthread_mutex_unlock(&spi_data->spis_mut);
+		return ret;
         }
 
     }


henningw:
Thanks for the pull request. I am not 100% sure if this approach is the right one.
* in line 83 the lock is allocated, sure
* the while loop basically has two exit conditions
  * there are no free SPIs - first unlock, line 109
  * a free SPI was found - second unlock, line 121

Your pull request changes this logic to unconditionally unlock, and then do a second unlock again, if I am not wrong?

ryancaicse:
My observation is that because the lock is acquired first (Line83) so the lock should be safely released before returning.
https://github.com/kamailio/kamailio/blob/1ddc27f199061025a6a43da3e8a1388fcaddfe19/src/modules/ims_ipsec_pcscf/spi_gen.c#L109-L111

henningw:
I've missed the missing braces around the if condition, right. Let me have another look at the logic later today.

henningw:
There is an build error reported from the auto-build, maybe you can have a look.

ryancaicse:
@henningw Thank you for your time.
The code can always ensure that the lock is released at line 109 before returning.

PR #2958, opened 2021-11-27
Fix TM Module Doc unmatched_cancel param from string to integer

ZionDials:
Currently, the documentation states the unmatched_cancel parameter accepts a string input, but it only accepts an integer input. Fixing documentation to match the code's expectations.

https://github.com/kamailio/kamailio/blob/b6cebaa76eaa4b98bbba6e271c50dbfa14a8f9f7/src/modules/tm/config.c#L159


henningw:
Thank you, will merge it. For future pull requests, please prefix the module name in the commit message (e.g. in this case "tm: fix unmatched.."). Further details can be found in our contribution guide document.

PR #2959, opened 2021-11-27
dispatcher: Not possible to deactivate probing of destination by reload

mtryfoss:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
If ds_probing_mode is set to 3, flag in the dispatcher entry needs to include 8
for probing to happen.

Changing a destination to not include this anymore while the destination do
not respond to OPTIONS will cause the probing flag to be re-added because of the
delayed response. A restart will fix it.

We need to check against the probing flag in the CB function to determine if
we should continue probing this destination. If probing is deativated while
waiting for a reply, we should abort.

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -3529,6 +3529,11 @@ static void ds_options_callback(
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


miconda:
Thanks!

PR #2960, opened 2021-11-28
db_redis: don't leave unconsumed redis replies in case of error

apogrebennyk:
- Execute db_redis_consume_replies on failed update same as in db_redis_insert:
if db_redis_perform_update encounters an error it was leaving unconsumed data
thus causing unexpected reply later on when reading from the same connection.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally


miconda:
@agranig - any comment on this one, in case you are still working with db_redis?

PR #2962, opened 2021-12-04
auth_ephemeral: fix sha256/384/512

justin-lavelle:
- sha256/384/512 broken due to use of sha1 password length, change to check and use proper lengths for each
- sha384 mistakenly using sha256 method
- update logging to include length


justin-lavelle:
Seems the LGTM analysis failed due to a server issue...
```
A fatal error occurred: Severe disk cache trouble (corruption or out of space) at /work/working/private/out/snapshot/db-cpp/default/cache/pages/93/30.pack: Failed to write item to disk
(eventual cause: IOException "No space left on device")
```

henningw:
Thanks for the pull request. I triggered the LGTM job again, fine now.
The patch looks good to me, if there are no other remarks i will merge in the next days.

miconda:
Thanks! Merging to backport for new releases.

PR #2964, opened 2021-12-08
dispatcher: Fix handling of inactive destination for alg 13

mtryfoss:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Alg 13 did try to distribute calls to inactive destinations.
If the highest priority destination is inactive, hash is not updated
and the xavp is not set. This is resulting in failover mechanism
not working at all for the given call.

When the hash variable is not updated, it makes alg 13 behave like
round robin if the scenario above occurs. If you got two destinations
and the highest priority is out of service, 50% of the calls will fail.

Now I tried a more simple approach updating hash with the first
entry of the sorted list.

Also fixed a typo in a variable name.

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -2201,7 +2201,7 @@ int ds_manage_route_algo13(ds_set_t *idx, ds_select_state_t *rstate) {
 	}
 
 	for(y=0; y<idx->nr ;y++) {
-		int latency_proirity_handicap = 0;
+		int latency_priority_handicap = 0;
 		ds_dest_t * ds_dest = &idx->dlist[z];
 		int gw_priority = ds_dest->priority;
 		int gw_latency = ds_dest->latency_stats.estimate;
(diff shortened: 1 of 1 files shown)


miconda:
@jchavanton - any comments on this PR?

mtryfoss:
Please notice, we use this with latency stats disabled. Just a way of having distribution with round robin within each priority for more than two destinations.

For example, in normal operation use round robin against two destinations and a third as failover if both the two first goes offline.

miconda:
@mtryfoss - have you checked ds_select_routes(), isn't it more suitable for what you need?

mtryfoss:
It seems like that's more of a thing to combine multiple dispatcher sets using potentially different algos?
That could be useful in some other scenarios, but in the current setup this is working perfectly with the applied patch.

My point here is that algo 13 seems to be broken (failover does not work) if the first selected (highest priority) destination is marked as inactive. The xavp var is not set. 

While it's being probed but not yet inactive, the call distribution will time out - but the xavp will be set and the on failure logic can trigger as expected.

miconda:
I am not using that algorithm, so I have no experience with expected behaviour. If @jchavanton has no comments soon, probably will be merged, I see it affects only code specific to the algorithm.

PR #2965, opened 2021-12-08
rtpengine: Fix set selection

smititelu:
When set_rtpengine_set() config function is called.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

After calling set_rtpengine_set(), the new set is not set for further rtpengine delete(). This is because active_rtpp_set is set to default_rtpp_set when msg id != current_msg_id (which is set once by set_rtpengine_set())

I've removed the msg id != current_msg_id condition, and initialized active_rtpp_set in mod_init().

Basically tested locally with 2 sets with 1 rtpengine each.

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -1810,6 +1810,7 @@ mod_init(void)
 
 	/* select the default set */
 	default_rtpp_set = select_rtpp_set(setid_default);
+	active_rtpp_set = default_rtpp_set;
 	if (!default_rtpp_set) {
 		LM_NOTICE("Default rtpp set %u NOT found\n", setid_default);
 	} else {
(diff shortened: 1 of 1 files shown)


miconda:
At the first sight, does not seem good.

That protection with msg_id is to ensure that a rtpengine group (set) is used only for the SIP message it was set for.

Now it seems you remove that protection and setting a rtpengine group stays until is set again.

Anyhow, the end of the call (BYE) can be handled by a different SIP worker process than the one which handled the corresponding INVITE, so you can't really have (expect) the rtpengine group to be set to same value from INVITE till BYE. You should store it inside a dlg variable (if you use dialog module) or in a hashtable. It is also an option to add it as a param to record-route header.

smititelu:
Thank you for the feedback! Closing this PR.

Issue #2967, opened 2021-12-08
Using undefined environmental variables with #!defenv 


whosgonna:
### Description

Currently, an undefined environmental variable with `#!defenv` results in Kamailio failing to start.  The proposal  here that if the environmental variable is not defined, then the macro value would be undefined as well.  For example:

```
!#KAMAILIO
loadmodule "xlog"
loadmodule "pv"

#!defenv FOO

request_route {
    #!ifdef FOO
        xlog("L_INFO","Environmental variable $$FOO is set\n");
    #!else
        xlog("L_WARN","Environmental variable $$FOO is NOT set\n");
    #!endif
    forward();
}
```

Currently this configuration will fail to load if environmental variable is not set.  My proposal is to either allow `#!envdef` to accept an undefined environmental variable and have the preprocessor macro undefined, or to add a new keyword (`#!envdefn` perhaps?) to allow this behavior.

whosgonna:
Proposed changes are in https://github.com/whosgonna/kamailio/tree/iss2967

whosgonna:
As indicated by @miconda, I've refactored the keywords from `defenvn` to `trydefenv` and from `defenvs` to `trydefenvns`.

whosgonna:
@miconda indicated in the Kamailio mailing list that they code for `#!defenvs` would be pushed.  Once that's done, I'll merge to my fork, then submit a PR.

miconda:
PR was merged.

PR #2969, opened 2021-12-09
core: cfg.lex support for undefned env vars

whosgonna:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #2967 

#### Description
Add new preprocessor keywords `trydefenv` and `trydefenvs` that work like `defenv` and `defenvs` respectively, with the difference that if the environmental variable is not defined, then the value in Kamailio will not be defined.

For example:
```
#!KAMAILIO

loadmodule "xlog"
loadmodule "pv"

#!trydefenv EXTRALOG

request_route {
    #!ifdef EXTRALOG
       xlog("L_I", "Some Extra Logging here\n");
    #!endif
    forward ();
}
```
If the environmental variable `$EXTRALOG` is defined, the log line will be written.  If the environmental variable is not set, then in Kamailio `EXTRALOG` will not be defined and the `#!ifdef` section will not be loaded.


miconda:
Thanks! I am merging it, then probably do a few cosmetic changes to make the new parameter more generic, to be able to reuse in the future if needed.

PR #2971, opened 2021-12-10
jwt: possibility to add headers when generating jwt

RodrigoAS28:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
Adding an optional forth parameter to jwt module to add headers.

### DIFF
--- a/src/modules/jwt/jwt_mod.c
+++ b/src/modules/jwt/jwt_mod.c
@@ -41,7 +41,8 @@ static int  mod_init(void);
 static int  child_init(int);
 static void mod_destroy(void);
 
-static int w_jwt_generate(sip_msg_t* msg, char* pkey, char* palg, char* pclaims);
+static int w_jwt_generate_4(sip_msg_t* msg, char* pkey, char* palg, char* pclaims, char* pheaders);
+static int w_jwt_generate_3(sip_msg_t* msg, char* pkey, char* palg, char* pclaims);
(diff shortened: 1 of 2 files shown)


miconda:
I see the PR is changing the prototype of ki_jwt_generate() function by adding a new parameter. However that is exported to KEMI and expects the same number of parameters as so far.

You can rename ki_jwt_generate() to ki_jwt_generate_hdrs() and add the new parameter, then re-add ki_jwt_generate() as a wrapper to ki_jwt_generate_hdrs() calling it with the new parameter being NULL.

src/modules/jwt/jwt_mod.c
```diff
@@ -172,18 +175,28 @@ static int jwt_fcache_add(str *key, str *kdata)
  *
  */
 static int ki_jwt_generate(sip_msg_t* msg, str *key, str *alg, str *claims)
+{
+	return ki_jwt_generate_hdrs(msg, key, alg, claims, NULL);
+}
+
```
miconda, line 181: This function has to be moved after `ki_jwt_generate_hdrs()` in order to get it compiled.

miconda:
Looking again over the patch, I didn't notice any free of the parsed params and duplicated string for new headers value. It should be done as it is for claims, or am I missing something?

RodrigoAS28:
> Looking again over the patch, I didn't notice any free of the parsed params and duplicated string for new headers value. It should be done as it is for claims, or am I missing something?

No, you are absolutely right.
Please have another look.

PR #2978, opened 2021-12-18
NATS: add new function `nats_publish()`

eschmidbauer:
#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
Adds Publish function to NATS module.

### DIFF
--- a/src/modules/nats/defs.h
+++ b/src/modules/nats/defs.h
@@ -1,0 +1,92 @@
+/*
+ * NATS module interface
+ *
+ * Copyright (C) 2021 Voxcom Inc
+ *
+ * This file is part of Kamailio, a free SIP server.
+ *
+ * Kamailio is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version
+ *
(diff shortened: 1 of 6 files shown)


miconda (approved):

miconda:
Fine to merge it from my point of view, only one comment as I noticed in the makefile it needs both libuv and libevent. I thought they are alternative for each other, not expecting to be needed at the same time, but maybe they are required in this manner because of dependencies.

smititelu:
I've basically tested this PR locally and works good. Nice work!

Thanks,
Stefan

eschmidbauer:
> Fine to merge it from my point of view, only one comment as I noticed in the makefile it needs both libuv and libevent. I thought they are alternative for each other, not expecting to be needed at the same time, but maybe they are required in this manner because of dependencies.

@miconda thank you for the feedback. I agree with you, we should probably not use two different eventing libraries. 
I will work on switching to libuv for the publish worker and submit a new PR.

eschmidbauer:
@miconda @smititelu 
I've updated this PR to use libuv _only_ and added parameter `num_publish_workers` which allows you to have N number of publish workers. Thanks!

miconda:
@eschmidbauer - fine to merge from my point of view.

PR #2980, opened 2021-12-20
tcpclose event_routes issue

vlitvinov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

We faced an issue that tcp:close (timeout and reset) event routes don't execute in rare cases. Debugging and testing show that in the case when a connection is marked as BAD or write timeout etc our event routes don't execute. 
I found that executing event routes is in the read function only. So in case when a connection is destroyed triggered by other reasons we don't have a notification about this. 
I propose to do this in main file every time when we destroy connections. But in this case, we don't have I reason to destroy (timeout reset or EOF).
In our lab with these changes issue is gone. But sure this is just a concept.

### DIFF
--- a/src/core/tcp_conn.h
+++ b/src/core/tcp_conn.h
@@ -193,12 +193,22 @@ typedef struct tcp_conn_alias {
 #endif
 
 
+enum tcp_closed_reason {
+	TCP_CLOSED_EOF = 0,
+	TCP_CLOSED_TIMEOUT,
+	TCP_CLOSED_RESET,
+
+	_TCP_CLOSED_REASON_MAX /* /!\ keep this one always at the end */
+};
+
+
 typedef struct tcp_connection {
 	int s; /*socket, used by "tcp main" */
 	int fd; /* used only by "children", don't modify it! private data! */
 	gen_lock_t write_lock;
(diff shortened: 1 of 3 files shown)


miconda:
Thanks for the PR!

I would suggest to add a new field to `tcp_connection` structure, like `event`, which is set to the event type of closing the tcp connection, so what was removed by current PR:

```
-				if (errno == ETIMEDOUT) {
-					tcp_emit_closed_event(c, TCP_CLOSED_TIMEOUT);
-				} else if (errno == ECONNRESET) {
-					tcp_emit_closed_event(c, TCP_CLOSED_RESET);
-				}
```

 becomes like:

```
				if (errno == ETIMEDOUT) {
					c->event = TCP_CLOSED_TIMEOUT;
				} else if (errno == ECONNRESET) {
					c->event = TCP_CLOSED_RESET;
				}
```

Same in other similar cases.

Then, when calling the emit function, if the `event` field is set, then it is used.

vlitvinov:
@miconda I added a new field into the tcp_connection structure and also added logic for work with it. Can you please review the changes?

miconda:
Thanks, merging!

Issue #2984, opened 2021-12-22
Adding listen flag to ignore matches in grep_sock_info when an IP is not active locally


rhys-hanrahan:
### Description

As per my thread on sr-users: **Active/active keepalived - Kamailio includes non-local virtual IP in "myself" which breaks DMQ call routing**, I am seeing issues due to the behavior of `check_self` and more specifically `grep_sock_info`. I have attached a patch with a proposed new feature that would allow changing this behaviour slightly, although I've avoided changing any default behaviour.

I have a setup with keepalived in active/active mode, with two virtual IPs. The goal is that each VIP is active only one one node at a time, but in the event of a failure, both VIPs would become active on one node.

The issue I am facing is that in order for this to work, I need to have both nodes listen on both virtual IPs at all times, but this means that both nodes always consider traffic to those IPs local to themselves, even when that is not the case. This breaks many things, basically anything with `uri == myself`. But even internally in Kamailio itself, this is causing problems.

For example: PATH support is broken  and this is not easily fixed in the config because `lookup()` is actually using `check_self()` internally. I am sure there are many other features which will not work right in the above scenario, and this is why I'm proposing a change in the code itself, instead of just trying to solve it in the config.

I've attached a patch file with my proposed change and would appreciate any feedback as to my overall approach. If this seems OK then I will put together a pull request for your review.

Thanks!

### Expected behavior

Kamailio should detect what IP addresses are currently active locally on it's system before considering the socket a match. If the IP is not currently there, it should ignore the match as if it is not currently listening on this IP (because it really is not!)

#### Actual observed behavior

Right now, Kamailio looks blindly at all "listen" sockets and looks for a match, regardless of if that IP is currently active locally or not, and this causes the false results of `check_self`

### Possible Solutions

As mentioned, I'm attaching a patch showing my overall approach. Basically:
* I am using pre-existing functions as much as possible for best compatibility e.g. IPv6 and IPv4.
* I use `dns_resolvehost()` which seems to have a caching function built-in and so I am hoping the performance impact of my change would be negligible.
* My code only makes a change _if_ the virtual flag is set, and if not, existing behaviour is kept so as to avoid any breaking changes.

[RT59671-kamailio-add-listen-virtual-check2.patch.txt](https://github.com/kamailio/kamailio/files/7762131/RT59671-kamailio-add-listen-virtual-check2.patch.txt)

rhys-hanrahan:
[...]

PR #2985, opened 2021-12-22
core: Add virtual flag to listen config to ignore inactive ip addresses

rhys-hanrahan:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #2984

#### Description

A new option to "listen" has been added called "virtual". This sets a flag on the listening socket to modify the behaviour of grep_sock_info. When this flag is set, grep_sock_info will only consider the listening IP a match if the IP is found in the system's current list of local IP addresses. If the IP is not currently local, then the matching IP is ignored.

If the virtual flag is not set on the socket then existing behaviour used instead.

This feature is intended for scenarios where you have two Kamailio nodes in active/active setup, with virtual IPs, where you need to be able to listen on a virtual IP but it may not always be active locally. With this flag applied to a `listen=` socket, Kamailio will only consider traffic destined to this IP local, when the IP is actually active on the system, and other times the match is ignored.

### DIFF
--- a/src/core/cfg.lex
+++ b/src/core/cfg.lex
@@ -305,6 +305,7 @@ XAVPVIAPARAMS	xavp_via_params
 XAVPVIAFIELDS	xavp_via_fields
 LISTEN		listen
 ADVERTISE	advertise|ADVERTISE
+VIRTUAL		virtual
 STRNAME		name|NAME
 ALIAS		alias
 SR_AUTO_ALIASES	auto_aliases
@@ -741,6 +742,7 @@ IMPORTFILE      "import_file"
 <INITIAL>{XAVPVIAFIELDS}	{ yylval.strval=yytext; return XAVPVIAFIELDS; }
 <INITIAL>{LISTEN}	{ count(); yylval.strval=yytext; return LISTEN; }
 <INITIAL>{ADVERTISE}	{ count(); yylval.strval=yytext; return ADVERTISE; }
+<INITIAL>{VIRTUAL}	{ count(); yylval.strval=yytext; return VIRTUAL; }
 <INITIAL>{STRNAME}	{ count(); yylval.strval=yytext; return STRNAME; }
 <INITIAL>{ALIAS}	{ count(); yylval.strval=yytext; return ALIAS; }
 <INITIAL>{SR_AUTO_ALIASES}	{ count(); yylval.strval=yytext;
--- a/src/core/cfg.y
+++ b/src/core/cfg.y
@@ -328,6 +328,7 @@ extern char *default_routename;
 %token XAVPVIAFIELDS
 %token LISTEN
 %token ADVERTISE
+%token VIRTUAL
 %token STRNAME
 %token ALIAS
 %token SR_AUTO_ALIASES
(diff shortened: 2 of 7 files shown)


rhys-hanrahan:
Some things I am concerned about/would like feedback on if possible:
* Right now I'm using DNS caching but would like to not use this if possible, because it would delay how long it takes for Kamailio to know when active IPs change. I am just not sure if using `_resolvehost()` to find local IPs on each call is too heavy or not?
* I am not sure if any other code would be caching results based on `grep_host_info` and expecting they never change? This is the only scenario I can see causing problems in future, but I can't see anywhere in the code where this is happening.

rhys-hanrahan:
Note, I have just re-based with a change to switch from using `dns_resolvehost` to `_resolvehost` to avoid DNS caching, as I think that's more sensible, unless suggested otherwise, for the intended use-case.

miconda:
I haven't had time to analyze the source code, but the change in `etc/kamailio.cfg` should not be done, the new attribute has to be added to the core cookbook in the wiki (after the merge). The listen has other attributes (e.g., name), not all have to be listed in the default config file and the new one is probably not going to be commonly used.

Also, is it sure it is needed to be added in `ims_ipsec_pcscf`?

rhys-hanrahan:
Hi Daniel, thanks for this! That's fine by me - both of those changes were just me trying to be thorough about making sure everything is being updated. Happy to remove the change to `kamailio.cfg`.

I don't mind if you don't want the flag added to `ims_ipsec_pcscf`, I just thought it would be good if this flag showed in the output, matching the other flags, for consistency? But if you think not, then I'm happy to remove this too.

rhys-hanrahan:
Just confirming I have rebased and dropped the commit for the cfg file change. Let me know if you want me to drop the commit for `ims_ipsec_pcscf` too. Thanks!

rhys-hanrahan:
Just wanted to add a note for one limitation I stumbled onto. As per this thread, if your hosts file contains an entry for your local machine name, then my method for listing local IPs does not work: https://stackoverflow.com/questions/1160963/how-to-enumerate-all-ip-addresses-attached-to-a-machine-in-posix-c

I just tested this out and confirmed it doesn't work. My machine had default entries for localhost only - so this code has worked fine in my testing. Right now I am just thinking of making a note of this requirement in the cookbook.

```
127.0.0.1       localhost.localdomain   localhost
::1             localhost6.localdomain6 localhost6
```

If there's any suggestions for a better way of getting a list of local IPs I'm open to that, but so far I haven't seen an alternative that doesn't have the same issue and is multi-platform.

mtryfoss:
Would be nice if this could disable (or at least, skip logging an error) for monitoring using OPTIONS in the same scenario where the server do listen to the required socket address. Currently we're relying on some external script enabling/disabling monitoring based on switchover, but that's not so elegant.

rhys-hanrahan:
@mtryfoss This sounds like something I'd be interested in implementing. Can you provide some more specifics of what you're doing? E.g. Using nathelper or keepalive module? What error is being logged? I am guessing if you have two DMQ nodes that both of them would be sending OPTIONS packets for each usrloc entry, but the inactive node would be timing out, and these are the errors you mean?

I am still not live with Kamailio yet and I'm only using nathelper in my lab, and it looks like timeouts are disabled by default, so I guess this explains why I haven't run into any errors being logged due to OPTIONS ping packets, but this is something I'll have to investigate as well.

Thanks for your input!

mtryfoss:
I guess multiple modules could benefit from this, but we're using the dispatcher module to distribute calls to static destinations, while monitoring availability of them. I see this PR might not be directly related.

Anyway, if socket is specified in the dispatcher entry and that's not available locally it will throw a lot of errors in the log.
Currently we use "kamcmd dispatcher.ping_active <1 or 0>", controlled by the software responsible for moving the VIP between hosts. But this needs to be re-executed after a Kamailio restart.

You'll have a similar issue if udp ping is enabled for registrations and you got those replicated from the live node, as you mentioned.

rhys-hanrahan:
Aha, makes sense. Yes I think you're right, multiple modules can probably benefit - at least dispatcher and nathelper that I'll be using, I'm sure. Using dispatcher instead of forwarding requests to a single PBX is literally the next item on my list, so I'll be running into this scenario very soon I'm sure. 

I will probably start looking at how I could handle this soon, not sure if I'll put it in this PR but probably put it in another PR at least, as it seems like a good fit for this flag.

miconda:
Thanks, merging! For further enhancements, new PR can be created!

PR #2997, opened 2022-01-07
add filtering based on start timestamp to dlg.list_match

Riccardo-78:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally


miconda:
Thanks for PR!

I would suggest to use str2int() (or suitable variant from core/ut.h) function to convert input string parameter to integer value, because the pkg str dup (and free) can be avoided in that case.

Riccardo-78:
now it uses str2int, thanks!

miconda:
Merging!

PR #3000, opened 2022-01-12
Fix PHONENUM, RTP_MEDIA_SERVER and WEBSOCKET for build on FreeBSD

drTr0jan:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
`websocket` and `phonenum` are not build on FreeBSD environment.
The proposed patch fixes it and makes modules are consistent with the rest of modules.

### DIFF
--- a/src/modules/phonenum/Makefile
+++ b/src/modules/phonenum/Makefile
@@ -5,13 +5,11 @@ include ../../Makefile.defs
 auto_gen=
 NAME=phonenum.so
 
-CXX=g++
-LD=g++
-LIB_DIR = /opt/local/lib
+CXX?=g++
+LD?=g++
 
-LIBS= -L${LIB_DIR} cphonenumber.o -lphonenumber -lgeocoding
-
-DEFS+= -I/opt/local/include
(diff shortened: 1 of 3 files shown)


src/modules/phonenum/Makefile
```diff
@@ -5,13 +5,8 @@ include ../../Makefile.defs
 auto_gen=
 NAME=phonenum.so
 
-CXX=g++
-LD=g++
```
miconda, line 9: Probably instead of removing, it's better to set CXX and LD if not set yet.

src/modules/phonenum/Makefile
```diff
@@ -5,13 +5,8 @@ include ../../Makefile.defs
...
-LIB_DIR = /opt/local/lib
-
-LIBS= -L${LIB_DIR} cphonenumber.o -lphonenumber -lgeocoding
-
```
miconda, line 13: The cphonenumber.o is needed for linking, not sure it would work without, expecting missing symbols. Have you started kamailio with this module, or just compiled it?

drTr0jan: Yep, you are right.

PR #3004, opened 2022-01-17
[htable] replacing wrong function return with correct one

saurabh-iitrke:
Replacing wrong return function with correct one in `w_ht_iterator_setex`


miconda:
Thanks for spotting it and PR. I pushed a slightly different fix because you PR didn't follow contributing guidelines related to commit message:

PR #3005, opened 2022-01-21
dialog: Add dlg_get_matches function

rhys-hanrahan:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
I've marked this as a "breaking change" only because it slightly alters the error handling of the RPC dlg.list_match command, to draw attention to it. But the actual functionality is unchanged.

This PR adds a new script function to the dialog module: `dlg_get_matches`. This has the same behaviour as the RPC `dlg.list_match` command, but allows use in scripts. The results are returned in an XAVP of a specified name, as an array of fields.

```
dlg_get_matches(mkey, mop, mval, xavp_name[, max_results]);

Returns number of matches found. Or -1 if there's an error.
```

An example of use would be:
```
        xlog("Extension is: $var(ext)\n");

        $avp(dlg_count) = dlg_get_matches("turi", "sw", "sip:nexusone$var(ext)@", "dlg_matches");

        xlog("Got $avp(dlg_count) matches.\n");
        $var(i) = 0;
        $var(matched) = 0;
        while ($var(i) < $avp(dlg_count))
        {
                //Skip dialogs that are not early state
                if ($xavp(dlg_matches[$var(i)]=>state) == 2)
                {
                        xlog("Found matching dlg[$var(i)]: $xavp(dlg_matches[$var(i)]=>from_uri)\n");
                }
        }
```

Some things to get feedback on
----
* Right now I am parsing pvar inputs outside of the "fixup" function. Is this required/suggested to be in fixup? I tried doing it this way but was getting random segfaults because the `pv_elem_t` ptr was getting corrupted somehow, so I removed this and do it all in the main function.
* Initially I just implemented `dlg_get_matches` on it's own and was torn on whether to refactor to avoid duplicating the matching logic in `rpc_dlg_list_match_ex`. As there's been a recent enhancement here, this pushed me to go ahead and refactor to share the common logic as much as possible in `dlg_list_matches`. Are there any issues with how I've approached this? Still slightly concerned it adds more complexity than it removes, but overall I think it's the right choice.
* In terms of my code in `dlg_list_matches`, I've tried to copy the dialogs as safely as possible but a second set of eyes would be great to make sure I am not missing anything that might cause corruption of the actual dialogs.

### DIFF
--- a/src/modules/dialog/dialog.c
+++ b/src/modules/dialog/dialog.c
@@ -460,8 +460,9 @@ static int fixup_dlg_get_matches(void **param, int param_no)
 		if(strncmp(s.s, "ruri", s.len) != 0 
 				&& strncmp(s.s, "turi", s.len) != 0
 				&& strncmp(s.s, "furi", s.len) != 0
-				&& strncmp(s.s, "callid", s.len) != 0) {
-			LM_ERR("param %d must be 'ruri', 'turi', 'furi' or 'callid'.\n",
+				&& strncmp(s.s, "callid", s.len) != 0
+				&& strncmp(s.s, "start_ts", s.len) != 0) {
+			LM_ERR("param %d must be 'ruri', 'turi', 'furi', 'callid' or 'start_ts'.\n",
 					param_no);
 			return E_CFG;
 		}
@@ -477,7 +478,7 @@ static int fixup_dlg_get_matches(void **param, int param_no)
 		}
 	//No extra validation for param 3
 
-	//VAXP result variable
+	//XAVP result variable
 	} else if(param_no == 4) {
 		/* We don't to evaluate the variable, just accept a string name to use as the XAVP class name.
 		 * no further validation needed.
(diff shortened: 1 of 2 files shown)


rhys-hanrahan:
For some reason I had totally ignored that dlg vars, profiles and callbacks are all referenced by the RPC command (and probably should be by my function). Given I don't know what the call backs might be doing to the original dialogs, it doesn't seem safe anymore to make copies of the original dialogs.

I just tried re-writing the code to return an array of pointers to the original dialogs, to avoid copying them. This would mean I need to perform locking around any results of `dlg_list_matches`. However I've realised that I won't know what "table entry" to lock as this could be different for each match.

I'm feeling a bit stuck right now - I suppose I could use a struct that tracks each dlg ptr match and what table index it belongs to, when returning results. I was trying to avoid creating extra things just for this small bit of code, though.

Any suggestions on this would be appreciated. Maybe it's just easier to not share the matching logic and keep it simple and just duplicate all the matching logic?

miconda:
I haven't checked the patches of the PR so far, but fyi, you should be able to execute an RPC command with jsonrpcs_exec(), get the output in a variable that you can parse with jansson module.

Having a dedicated function for config is probably better if you need to use it often.

rhys-hanrahan:
Thanks Daniel! As it happens, I just finished re-writing this code. I did end up creating a struct to return each matching dlg ptr, along with it's table entry index, to allow for locking. After some basic testing, this seems to be working for me, without copying the dialogs.

There's more cleanup I'd do before merging but I wanted to check if you have any issues with this overall approach before spending more time on it.

Thanks for the tip on jsonrpcs_exec - I was trying to find something like this originally but couldn't. As of right now, the solution in this PR is working well for me. But l will look at jsonrpcs_exec as a plan B. 

Thanks!

rhys-hanrahan:
@miconda Hi Daniel, just confirming this is cleaned up now and ready for whenever you have a chance to look at it - but let me know if you suggest any other changes. The only thing I haven't done is move the `pv_parse_format` into the fixup function, because as per the notes in my PR I had some weird segfaults there. But I can work on this if needed.

Otherwise, I've left the PR split into two commits in-case you're against sharing the matching logic for the dlg.match RPC command and this new dlg_add_matches, though I am happy with how things ended up with the shared function.

Thanks!
Rhys.

miconda:
Looking looking quickly at the final patch, there seems to be a race that can happen between building the list of matching dialogs and using the items in the list to push to xavps or to rpc output. The slots in the dialog hash table are released after it was walked and matching dialogs were linked in the matching list. It can take time till other slots are walked for matching other dialogs and during that time some dialogs in the previous slots can be destroyed (e.g., bye or timeout). This is one example, but dialogs might be destroyed also during the walk of matching list to convert it to xavp or rpc output.

The clone to xavp or printing to rpc output should happen when the slot of the matching dialog is acquired and before it is released again. If the lock is released after linking to matching list and then re-acquired for cloning to xavp or printing to rpc, then such races can happen and dlg field in matching list may point to invalid memory address.

rhys-hanrahan:
Thanks Daniel! Really appreciate the feedback, and thanks for catching that race condition. Unfortunately I'm still in a lab setting so haven't come across anything like that.

I will look at changing to use a callback style with variable arguments. For some reason at the time I didn't think this was an option but I can't see why this wouldn't work at the moment? And this would allow me to maintain the same lock all the way through. 

I will work on this as soon as I get some time and come back to you. Thanks!

henningw:
@rhys-hanrahan Any update on this one? Just wanted to ask because the freeze period for the next kamailio release starts soon.

rhys-hanrahan:
@henningw Sorry for the delayed reply. I'm still planning to re-work this with the feedback, but unfortunately my priorities shifted away from Kamailio for the moment. Hoping to come back to this in the next few weeks but I guess I'll be missing the freeze for next release, unfortunately.

henningw:
@rhys-hanrahan any update from your side?

miconda:
Closing this one being now 1 year old. When the code is updated, a new PR can be made if easier, or this one re-opened.

Issue #3018, opened 2022-02-07
dialog keep-alive only for own records


Riccardo-78:
### Description

When dialog module needs to send keep-alive to caller and/or called it doens't check if the ip address and port of the dialog is the current listening ip/port. If you have two distinct kamailio instances sharing a single dialog database, both instances are sending the OPTIONS and not just the one which handled the call at setup time.

#### Reproduction

Start two instance with a shared dialog database and db_mode set to 1 (realtime), set property ka-src for every dialog in kamailio cfg. Setup a long call and restart the instance not involved in the call. You should see it starting to send OPTIONS.

### Possible Solutions
In dlg_send_ka, if the ip/port stored in the dialog db is not present in the listening ip/port set, skip the check.

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
version: kamailio 5.5.2 (x86_64/linux) 55e232
flags: USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLOCKLIST, HAVE_RESOLV_RES
ADAPTIVE_WAIT_LOOPS 1024, MAX_RECV_BUFFER_SIZE 262144, MAX_URI_SIZE 1024, BUF_SIZE 65535, DEFAULT PKG_SIZE 8MB
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
id: 55e232
compiled on 14:20:31 Aug 25 2021 with gcc 4.8.5
```

* **Operating System**:
Centos 7

miconda:
I assume you mean the dialogs that are loaded upon kamailio restart, otherwise kamailio is using the records kept in memory and uses those ones.

Initially the dialog module was not designed to share the database table with another instance, but of course, a contribution can be made for this. Some similar work was done in the commit 506672e523a48985f89e2c35bdab3d024df0a34d, maybe you can look at it and come up with a PR for this feature request.

PR #3023, opened 2022-02-10
dialog keep-alive only for own records

Riccardo-78:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3018

#### Description

Skip keepalive check if dialog is not local, do the same in case of dialog timeout

### DIFF
--- a/src/modules/dialog/dialog.c
+++ b/src/modules/dialog/dialog.c
@@ -104,6 +104,7 @@ str dlg_extra_hdrs = {NULL,0};
 static int db_fetch_rows = 200;
 static int db_skip_load = 0;
 static int dlg_keep_proxy_rr = 0;
+int dlg_ignore_non_local_dlg = 0;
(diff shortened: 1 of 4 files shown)


miconda:
Thanks for the PR! I would suggest to add a module parameter to control this behaviour, to avoid looking up the socket every time. Also, there are cases when a restart is done to change the ip addresses and if the Routes were with FQDN, it is wanted to continue processing dialogs even if the old ip is not found.

Riccardo-78:
it makes sense, added new parameter (ignore_non_local_dlg) to control this behaviour

miconda:
Thanks!

I will merge it soon, but then I will rename the new parameter to something like `dlg_filter_mode` and behave on a flags-like value, because it could be useful to reuse it for other purposes, like having a filter mode where dialogs are not loaded from database if they don't have local socket.

Issue #2069, opened 2019-09-19
erlang: getting null value in erl_xbuff in response of erl_rpc request


vipulpatel310:
### Description
**I did:**
I have installed kamailio 5.2.4 in ubuntu system and configured kamailio script to connect with erlang node also it sends rpc request to erlang node and get response which I have printed in a logs.
...
 $erl_list(L) = "two";
 $erl_list(L) = "one";
$erl_list(args) = $erl_list(L);
 xlog("L_ALERT","Before RPC: $erl_list(args=>format)\n");
 erl_rpc("erlang", "list_to_tuple", "$erl_list(args)", "$erl_xbuff(repl)");
 xlogl("L_ALERT","type(reply): $erl_xbuff(repl=>type), format(repl): $erl_xbuff(repl=>format)\n");
...

**Expected to happen:**
Response should be get from erlang node against RPC request and it should be store in erl_xbuff. And it should be print logs like: ALERT: <script>: 501:type(reply): tuple, format(repl): {"one", "two"}

**Actually happened:**
Response is getting properly. but the problem is to store data into xbuff so it print logs like: ALERT: <script>: 501:type(reply): <null>, format(repl): <null>

### Troubleshooting
I have check with older version 4.3.0 and it works fine. getting this problem after 5.2.0 version
So, I have compare 5.2.0 and 4.3.0 version code. I found one change in pv_xbuff.c file which is creating a problem when I revert the change then it works fine.

#### Reproduction
It will be reproducible everytime. Just sends rpc request and get the value in xbuff. check xbuff value

#### Log Messages
Problem logs:
```
ALERT: <script>: Before RPC: [["one", "two"]]
WARNING: <core> [core/mem/q_malloc.c:480]: qm_free(): WARNING: free(0) called from erlang: pv_xbuff.c: xavp_decode(1209)
ALERT: <script>: 501:type(reply): <null>, format(repl): <null>
```
logs after change:
```
ALERT: <script>: Before RPC: [["one", "two"]]
WARNING: <core> [core/mem/q_malloc.c:480]: qm_free(): WARNING: free(0) called from erlang: pv_xbuff.c: xavp_decode(1209)
ALERT: <script>: 501:type(reply): tuple, format(repl): {"one", "two"}
```

### Possible Solutions
revert the 6542a367638ec5f8a2b62994f8c309e7a2533962 commit. It will work fine

### Additional Information

```
version: kamailio 5.2.4 (x86_64/linux) 759867
[...]

PR #3024, opened 2022-02-13
Erlang: Fixes to get the module functional again

jvanvleet:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #2069 

#### Description
The Erlang module is not working at all with newer Kamailio versions.   One of the issues was already found in #2069 but no code was submitted.   The second one I believe is related to https://github.com/kamailio/kamailio/commit/7bb2669528a819ec29a8193eebc9c641edd69993 in that the erlang module assumes that pvi.type is zero and uses that field to type the custom fields it manages and this commit starts populating pvi.type with a 4. 

I don't have a solid grasp on pvapi.c so if the experts have a better way to fix this up I am all ears.  These two commits do get the module working properly and I am able to use both the custom erlang PV terms and the RPC feature from my proxy now.

### DIFF
--- a/src/modules/erlang/pv_atom.c
+++ b/src/modules/erlang/pv_atom.c
@@ -51,6 +51,9 @@ int pv_atom_parse_name(pv_spec_t *sp, str *in)
 
 	name.s = p;
 
+	/*pvi.type now defaults to 4, breaking the erlang module's type management*/
+	sp->pvp.pvi.type = 0;
+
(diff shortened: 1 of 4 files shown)


miconda:
Thanks!

@seudin: any comment on this PR?

Otherwise, if nobody has comments in the near future, it will be merged.

PR #3047, opened 2022-03-04
uac: enable "uac:reply" event route for UAC SIP message timeouts

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

This PR adds 1 UAC module function and 1 TM module modparam. It enables "uac:reply" event route trigger on timeout or non-2xx replies. Basically it:
- adds TMCB_ON_FAILURE callback for UAC
- copies the TM uac content into uas; thus failure route will have the uas to work with. => this part of the code was already there, i just enabled it via (another) modparam. 

So far, is basically tested on our side, without any apparent issues. More testing will follow.

Please have a look on this and tell me what you think about it.

Thank you,
Stefan

### DIFF
--- a/src/modules/tm/config.c
+++ b/src/modules/tm/config.c
@@ -95,7 +95,8 @@ struct cfg_group_tm	default_tm_cfg = {
 		  CANCELs; on by default */
 	1,  /* e2e_cancel_reason -- copy the Reason headers from incoming CANCELs
 		  into the corresp. hop-by-hop CANCELs, on by default */
-	0   /* relay_100 -- by default, assume stateful proxy and do not relay SIP 100 */
+	0,   /* relay_100 -- by default, assume stateful proxy and do not relay SIP 100 */
+	0  /* enable_uac_fr */
 };
 
 void	*tm_cfg = &default_tm_cfg;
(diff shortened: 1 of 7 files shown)


smititelu:
Hi, I see PR build fails because i changed the header of UAC module api `uacb->req_send = uac_req_send;`. The solution(s) I can see right now are either:
1. Guard the uac_req_send(void/int) by some compilation #define
or
2. Get the tm module 'enable_uac_fr' config parameter value inside `uac_req_send(void)` function and decide adding TMCB_ON_FAILURE if it is enabled or not.
or
3. Adapt the test function with an parameter of "0"

miconda:
Wouldn't it be better to try to leverage somehow `$uac_req(evroute)`? Now it has to be set to `1` to get the event route executed. The field can be turned into a flags meaning, so if it is set to 2, then adds the execution as per this PR:

  * https://www.kamailio.org/wiki/cookbooks/5.5.x/pseudovariables#uac_req_key

By that, the prototype of the function does not change.

smititelu:
Good idea! Thank you! I will update the PR soon.

miconda:
Thanks!

PR #3050, opened 2022-03-07
ims_registrar_scscf: use xmlParseMemory instead of xmlParseDoc

kristiyan-peychev-flolive:
Fix issues where a non-zero terminated string would get passed to `parse_user_data`.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

### DIFF
--- a/src/modules/ims_registrar_scscf/userdata_parser.c
+++ b/src/modules/ims_registrar_scscf/userdata_parser.c
@@ -916,7 +916,7 @@ static ims_subscription* parse_ims_subscription(xmlDocPtr doc, xmlNodePtr root)
 
 /**
  * Parses the user data XML and copies data into a new ims_subscription structure.
- * @param xml - the input xml (NB must be null terminated)
+ * @param xml - the input xml
  * @returns the ims_subscription* on success or NULL on error
  */
 ims_subscription *parse_user_data(str xml)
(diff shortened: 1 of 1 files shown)


miconda:
@alexyosifov - any comment on this PR?

alexyosifov:
> @alexyosifov - any comment on this PR?

I don't know what is fixed/changed in the code with using of xmlParseMemory instead of xmlParseDoc, but hope that is tested by @kristiyan-peychev-flolive. Maybe he can elaborate a little bit more.

kristiyan-peychev-flolive:
The bug I encountered was quite the corner case, I'm not surprised nobody's noticed it. It's because the strings here are not zero-terminated by default, but `parse_user_data` requires a zero-terminated string by virtue of documentation. The bug I got was during diameter Cx registration termination requests with very specific XMLs being received, I managed to hit a place that was passing a string that was not zero-terminated. The end result was a failed check against the XML schema and the request being interpreted as invalid.
I believe I've solved the issue fundamentally by using `xmlParseMemory` instead of `xmlParseDoc` because the latter requires a zero-terminated string, while the former works with a buffer length, which is exactly what the strings here do.

alexyosifov:
> The bug I encountered was quite the corner case, I'm not surprised nobody's noticed it. It's because the strings here are not zero-terminated by default, but `parse_user_data` requires a zero-terminated string by virtue of documentation. The bug I got was during diameter Cx registration termination requests with very specific XMLs being received, I managed to hit a place that was passing a string that was not zero-terminated. The end result was a failed check against the XML schema and the request being interpreted as invalid. I believe I've solved the issue fundamentally by using `xmlParseMemory` instead of `xmlParseDoc` because the latter requires a zero-terminated string, while the former works with a buffer length, which is exactly what the strings here do.

Thanks!

miconda:
Thanks!

PR #3053, opened 2022-03-14
mqueue: add mqueue_mode modparam

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Add a mqueue parameter to control the mq_add() mode. 
0 - default, no changes in behavior
1 - add unique keys, keep oldest
2 - add unique keys, keep newest

### DIFF
--- a/src/modules/mqueue/doc/mqueue_admin.xml
+++ b/src/modules/mqueue/doc/mqueue_admin.xml
@@ -131,6 +131,31 @@ val character varying(4096) DEFAULT "" NOT NULL
 				Default value is 0 (no db table interaction).
 				</para>
 			</listitem>
+			<listitem>
+				<para>
+				<emphasis>addmode</emphasis>: how to add new (key,value) pairs.
+					<itemizedlist>
+					<listitem>
+						<para>
+						<emphasis>0</emphasis>:
+							Will push all new (key,value) pairs at the end of the queue. (default)
+						</para>
+					</listitem>
+					<listitem>
+						<para>
+						<emphasis>1</emphasis>:
+							Will keep oldest (key,value) pair in the queue, based on the key.
+						</para>
+					</listitem>
+					<listitem>
+						<para>
+						<emphasis>2</emphasis>:
+							Will keep newest (key,value) pair in the queue, based on the key.
+						</para>
+					</listitem>
+					</itemizedlist>
+				</para>
+			</listitem>
 			</itemizedlist>
 		</listitem>
 		</itemizedlist>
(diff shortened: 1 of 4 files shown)


miconda:
Walking the list of items without acquiring the lock has the risk of race with a consumer.

Then, the module was designed to define queues with attributes specified as part of `mqueue` modparm. I understood that having a variant with dedicated modparam can help sometime, but it should not be only as dedicated modparam, therefore this has to be supported also as `mqueue` modparm attribute.

smititelu:
Updated PR. Changed the name to "mqueue_addmode" to be more explicit. Can be set either as a modparam, or when defining the mqueue, and can be different for each defined mqueue. Updated doc.

Currently under testing

smititelu:
what I've found out while testing is that for big mqueue(s) (> 100 000) process(es) that mq_add() call become slower, because they have to search whole mqueue, under lock.

miconda:
It is expected to be slower for larger number of items in the queue. There can be some improvements by keeping a hash table with the names of the items in the key to speed up the search, but that will also add to complexity.

From my point of view, this would be a specific use case where one has to be aware of consequences to processing speed. But if it useful in some cases, then that's how it is.

In other words, I am still fine to merge it, but I think there has to be no unlock (lock release) after the if with the mode:

  * https://github.com/kamailio/kamailio/pull/3053/commits/8a3587004c27f1dc2aaa078c6508f5606124ccba#diff-32730fc08fc59be73bbefa3ec2551f08013719505f9f057d865e4c688a896a79R342

And the next lock_get() has to be removed. Practically, everything has to be done in a single lock_get() ... lock_release(), otherwise there is a short time frame open for races: the if block does not find the item with the name, does lock release, loses CPU for a while, and during that time another process adds an item with the same name.

The penalty is that the shm alloc and copy for the new item is done inside the lock. But probably a condition on `mh->addmode` can be leveraged to do the lock/unlock at the better place, because `mh->addmode` is set at startup and never changes, plus that `mh` does not get destroyed during runtime.

smititelu:
Hi, thank you for the feedback!

Yes, I was also thinking about that double lock_get()/release(). Actually in my tests I updated the code to be under a single lock_get()/release(), kept shm alloc outside the locks and shm released it when/if necessary inside lock_get()/release().

Since this new addmodes might have valid use cases in some scenarios, I will update the PR soon. In the short time frame, got no intention to update the code to using hash table, but maybe in the longer time frame will do it.

Thank you,
Stefan

smititelu:
Updated PR to take the lock sooner and check for addmode. Release it before returning. Shm alloc new "mi" item only if need.

miconda:
After merging I pushed another change to use a local variable to track locking and optimize a bit what operations are done under lock. Maybe you have some time to run your tests again and see if still ok.

smititelu:
In the tests so far it is ok. Thank you.

PR #3063, opened 2022-03-23
stirshaken: Add PVs to allow access to x509 subject and ppt grants

mrtrev:
- added vs_certsubject_pvname and vs_pptgrants_pvname config params
- adjusted log level of load/unload events

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

The stirshaken module currently does not allow any access to the information provided in the Identity header.  These changes allow key components of a signed call to be available in user-specified pseudo variables.

Specifically, the Subject of the certificate used to sign the call will be written to the PV specified in vs_certsubject_pvname and the JWT payload that has been verified will be written to the PV specified in vs_pptgrants_pvname.

There is no change to any behaviour whether these configuration options are set or not other than to write the appropriate values into the indicated PVs.

### DIFF
--- a/src/modules/stirshaken/doc/stirshaken_admin.xml
+++ b/src/modules/stirshaken/doc/stirshaken_admin.xml
@@ -256,6 +256,47 @@ modparam("stirshaken", "vs_cache_expire_s", 15)
 </programlisting>
 		</example>
 	</section>
+		<section>
+		<title><varname>vs_certsubject_pvname</varname> (str)</title>
+		<para>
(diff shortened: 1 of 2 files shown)


miconda:
Thanks for the PR!

Is no free needed after using `stir_shaken_cert_get_subject()`?

mrtrev:
> Is no free needed after using `stir_shaken_cert_get_subject()`?

I don't believe so.  The stir_shaken_cert_destroy(&cert_out); below seems to clean it up.

miconda:
OK.

PR #3065, opened 2022-03-25
uac: fix shm leak

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Setup: uac_req_send() with evroute=2 and tm module enable_uac_fr=1

Noticed following shm errors while intensively testing uac_req_send(), happening quite soon after a fresh restart (e.g. 10 min):
```
kamailio[5384]: ERROR: <core> [core/mem/q_malloc.c:297]: qm_find_free(): qm_find_free(0x7fe8943bc000, 40632); Free fragment not found!
kamailio[5384]: ERROR: <core> [core/mem/q_malloc.c:430]: qm_malloc(): qm_malloc(0x7fe8943bc000, 40632) called from uac: uac_send.c: uac_send_info_clone(105), module: uac; Free fragment not found!
kamailio[5384]: ERROR: uac [uac_send.c:108]: uac_send_info_clone(): no more shm memory
kamailio[5384]: ERROR: uac [uac_send.c:837]: uac_req_send(): cannot clone the uac structure
```
I tried monitoring system memory using `free` tool and noticed that the "used" memory didn't grow alot. However, I noticed the "buff/cached" memory had a steep growth when 200OK(s) started to come, for my setup.

Noticed that, for my setup, when 200OK come, event_route[uac:reply] is not called, so the tm callback set by UAC module is not called. => the "tp" callback parameter shm allocated by UAC module is not freed.

By analogy, when using uac_req_send() with evroute=1 and timeout happens (e.g. no reply comes at all), event_route[uac:reply] won't be called and same leak should happen => *i didn;t tested that*, was just an analogy.

I updated the code to call the callback on TMCB_DESTROY, and make sure tp parameter is freed, *one time*, by what TMCB event happened first. 

Let me know what you think of this... in my setup/tests I got no kamailio shm logs anymore. Also "buff/cached" memory didn't have that steep growth anymore.

Thank you,
Stefan

### DIFF
--- a/src/modules/uac/uac_send.c
+++ b/src/modules/uac/uac_send.c
@@ -717,14 +717,16 @@ void uac_resend_tm_callback(struct cell *t, int type, struct tmcb_params *ps)
 	}
 	tp = (uac_send_info_t*)(*ps->param);
 
-	if(tp->evroute!=0) {
+	if(tp->evroute!=0 && ps->code > 0) {
 		uac_req_run_event_route((ps->rpl==FAKED_REPLY)?NULL:ps->rpl,
 				tp, ps->code);
 	}
 
 done:
-	if(tp!=NULL)
+	if(tp!=NULL){
 		shm_free(tp);
+		*ps->param = NULL;
+	}
 	return;
 
 }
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the pull request. Just a small note regarding the debugging - standard system monitoring tools like free will in most cases not be that useful for detecting memory leaks, due to the Kamailio internal memory manager. But as you already got out of memory log messages in the logs, it strongly suggest a memory leak. You could use the kamailio memory debugging to get more detailled information, but in this case its probably not necessary.

smititelu:
Updated PR to NOT call event_route if code==0. This was happening for TMCB_DESTROY event.

PR #3079, opened 2022-04-11
pua_dialoginfo: Determine dialog event types in mod_init

btriller:
Determine dialog event types for dialog callback on module
initialization, since these will not change.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/modules/pua_dialoginfo/pua_dialoginfo.c
+++ b/src/modules/pua_dialoginfo/pua_dialoginfo.c
@@ -101,6 +101,7 @@ int disable_callee_publish_flag = DEF_DISABLE_CALLEE_PUBLISH_FLAG;
 char * pubruri_caller_avp  = DEF_PUBRURI_CALLER_AVP;
 char * pubruri_callee_avp  = DEF_PUBRURI_CALLEE_AVP;
 int publish_dialog_req_within = DEF_PUBLISH_DIALOG_REQ_WITHIN;
+int dialog_event_types = DLGCB_FAILED| DLGCB_CONFIRMED_NA | DLGCB_TERMINATED | DLGCB_EXPIRED | DLGCB_EARLY;
 
 int puadinfo_attribute_display = 0;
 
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the pull-request. I have not followed the complete code path yet, but I am not 100% sure this is the correct way to go.

In the existing code we are registering the DLGCB_REQ_WITHIN conditionally for the register_dlgcb(..) call. This will allocate new memory and copy the call back types to the new allocated structure. If you later change the callback types (as in your new code), it will be not synchronized with the structure inside the dialog module, or I am wrong?

henningw:
@btriller any comment regarding my previous question?

btriller:
In mod_init, pua_dialoginfo registers a dialog callback for newly created dialogs. When this callback runs, it registers another callback for this newly created dialog, w/ the interesting dialog event types. modparam `publish_dialog_req_within` is checked whether DLGCB_REQ_WITHIN should be included.
IMHO this check can be moved to mod_init, since event types aren't going to change.

Sorry for missing that out in the PR description.

src/modules/pua_dialoginfo/pua_dialoginfo.c
```diff
@@ -622,25 +623,12 @@ struct dlginfo_cell* get_dialog_data(struct dlg_cell *dlg, int type, int disable
...
-		if (dlg_api.register_dlgcb(dlg,
-				DLGCB_FAILED| DLGCB_CONFIRMED_NA | DLGCB_TERMINATED
-				| DLGCB_EXPIRED | DLGCB_EARLY,
-				__dialog_sendpublish, dlginfo, free_dlginfo_cell) != 0) {
-			LM_ERR("cannot register callback for interesting dialog types\n");
-			free_dlginfo_cell(dlginfo);
-			return NULL;
-		}
-        }
+	if (dlg_api.register_dlgcb(dlg, dialog_event_types,
```
henningw, line 626: This will allocate new memory and copy the call back types to the new allocated structure.

src/modules/pua_dialoginfo/pua_dialoginfo.c
```diff
@@ -856,6 +844,9 @@ static int mod_init(void)
 		LM_DBG("configured to use headers for uri values\n");
 	}
 
+	if(publish_dialog_req_within)
+		dialog_event_types |= DLGCB_REQ_WITHIN;
```
henningw, line 848: This will change the callback types, it will be not synchronized with the structure inside the dialog module.

henningw (commented):
I was probably not clear in my previous comment. I have added it to the appropriate places in your patch, maybe its more clear this way.

henningw:
You are right, there are multiple callbacks and the one that you've added is done in mod_init. So it should be fine. Lets wait a bit for other feedback, otherwise I will merge it.

PR #3095, opened 2022-04-28
pv_headers: fix possible mem. overflow issue and wrong size string

Tomohare:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

- remove a variable creation
Put two variable accessible to the entire function, renaming one to have some clarity. 
Remove the useless variable.  

- Remove from snprintf that fills the string
This fix possible a memory overflow while using `snprintf`. This also add a \0 at the end of the structure.

- change the destination of an allocated space being smaller than the source
Elements were being put into a smaller string, this commit makes the destination bigger then the source.

### DIFF
--- a/src/modules/pv_headers/pvh_func.c
+++ b/src/modules/pv_headers/pvh_func.c
@@ -383,13 +383,13 @@ int pvh_header_param_exists(struct sip_msg *msg, str *hname, str *hvalue)
 
 int pvh_remove_header_param(struct sip_msg *msg, int idx, str *hname, str *elements, str *toRemove)
 {
+	int notTarget, writtenChars;
 	int offset = 0;
 	int ret = -1;
 	char *next_token;
 	char *token;
 	char *result = (char*)pkg_malloc(elements->len - toRemove->len);
-	char *t = (char*)pkg_malloc(elements->len);
-	int maxSize = elements->len;
+	char *t = (char*)pkg_malloc(elements->len+1);
 
 	if (result == NULL || t == NULL)
 	{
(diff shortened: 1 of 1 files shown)


src/modules/pv_headers/pvh_func.c
```diff
@@ -383,51 +383,51 @@ int pvh_header_param_exists(struct sip_msg *msg, str *hname, str *hvalue)
...
 
 	if (result == NULL || t == NULL)
 	{
 		PKG_MEM_ERROR;
 		goto clean;
 	}
 
-	snprintf(result, elements->len - toRemove->len, "%*s", elements->len - toRemove->len, "");
-	snprintf(t, elements->len+1, "%s", elements->s);
+	memset(result, 32, elements->len - toRemove->len);
```
miconda, line 400: What is the purpose of filling `result` buffer with space (char code `32`)?

src/modules/pv_headers/pvh_func.c
```diff
@@ -383,51 +383,51 @@ int pvh_header_param_exists(struct sip_msg *msg, str *hname, str *hvalue)
...
-			offset += n;
+			offset += writtenChars;
 		}
 	}
 
 	if (elements->len-toRemove->len > 0)
 	{
-		snprintf(elements->s, elements->len, "%*s", elements->len-toRemove->len, "");
+		memset(elements->s, 32, elements->len - toRemove->len);
 		snprintf(elements->s, (strlen(result)%elements->len)+1, "%s", result);
```
miconda, line 431: Another `memset` with char code 32 (space), then `snprintf` in the same buffer.

Maybe it is ok, but `memset` with `space` is quite unusual, therefore asking for further details.

src/modules/pv_headers/pvh_func.c
Tomohare: Yes, here this is strange. No need of this line entirely.

src/modules/pv_headers/pvh_func.c
Tomohare: As above, this is not needed here.

henningw:
@Tomohare were all previous remarks addressed from your latest pushed version? @linuxmaniac any remarks from your side as well?

PR #3100, opened 2022-05-02
core: add an option to trigger also an RST on tcp connection close

dunst0:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

We have seen that in deployments where there are strict port requirements on both side, there are tcp connections in TIME_WAIT state causing no fresh TCP connections possible until the TIME_WAIT is gone. In a lot of these case the clients are sending from them self a RST after they send a FIN,ACK from there side but some aren't doing this. The basic idea of this PR is to give an extra option for the tcp configuration from the cfg to trigger always an RST on connection close from Kamailio side. This currently done by enabling SO_LINGER with a value of 0 just before close.


henningw:
Thanks for the pull request, which you just closed again? Just a note for myself, [this](https://notes.shichao.io/unp/ch7/#so_linger-socket-option) seems to be a good summary of this kernel option.

dunst0:
I think I had some outdated got clone on my side I will open it again after updating my clone.

PR #3102, opened 2022-05-05
userblocklist: fix restart errors

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Spotted this ERRORs on kamailio restart:
```
ERROR: <core> [db.c:481]: db_use_table(): invalid connection parameter
ERROR: userblocklist [db.c:123]: db_reload_source(): cannot use db table 'globalblocklist'
ERROR: userblocklist [userblocklist.c:425]: load_source(): cannot load source from 'globalblocklist
```

Debugging it further, seen that the above ERRORs appear on ```check_allowlist()``` function fixup. This was happening because db connection is closed in mod_init and re-done in child_init. But the fixups happen before child_init and they need db connection.

Decided to not close the db connection after mod_init because it is later, first closed and then re-opened, in each child. Figured out this fix after checked where each function of db_userblocklist.c is used.

ERRORs no longer appear on restart.

Thank you,
Stefan

### DIFF
--- a/src/modules/userblocklist/db_userblocklist.c
+++ b/src/modules/userblocklist/db_userblocklist.c
@@ -91,7 +91,6 @@ int userblocklist_db_init(void) {
 		userblocklist_db_close();
 		return -1;
 	}
-	userblocklist_db_close();
 	return 0;
 }
 


henningw:
Thank you. I will push another commit that fixes the function comment as well later.

PR #3104, opened 2022-05-10
ims_qos: rx_aar - unreference dialog in case of aar update

atanasdb:
- release the dialog reference even for aar updates as the dialog was
  already referenced when issuing the aar request

### DIFF
--- a/src/modules/ims_qos/rx_aar.c
+++ b/src/modules/ims_qos/rx_aar.c
@@ -180,6 +180,8 @@ void async_aar_callback(int is_timeout, void *param, AAAMessage *aaa, long elaps
             STR_SHM_DUP(*passed_rx_session_id, aaa->sessionId->data, "cb_passed_rx_session_id");
             LM_DBG("passed rx session id [%.*s]", passed_rx_session_id->len, passed_rx_session_id->s);
             dlgb.register_dlgcb_nodlg( data->dlg, DLGCB_TERMINATED | DLGCB_DESTROY | DLGCB_EXPIRED | DLGCB_RESPONSE_WITHIN | DLGCB_CONFIRMED | DLGCB_FAILED, callback_dialog, (void*) (passed_rx_session_id), free_dialog_data);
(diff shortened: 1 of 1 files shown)


atanas-wg2:
If we have multiple Rx_AAR in the script (e.g. for 183, then for UPDATEs) since the second AAR is considered an update, without that patch, the dialog is not unreferenced. That leads to dialog not being deleted since there is a dangling reference which leads to memory leak.

riccardv:
I confirm the PR broke the dialog as commented by @atanas-wg2

miconda:
The commit message has to start with the component name, not with the file name, respectively ims_qos in this case.

@alexyosifov: is it something that you can check if it is ok?

alexyosifov:
@atanas-wg2 @miconda 
I have checked locally the changes. The P-CSCF doesn't start correctly if trying to load the ims_qos module:
`0(15988) DEBUG: <core> [core/sr_module.c:563]: load_module(): trying to load </usr/local/lib64/kamailio/modules/ims_qos.so>`
`0(15988) ERROR: <core> [core/sr_module.c:570]: load_module(): could not open module </usr/local/lib64/kamailio/modules/ims_qos.so>: /usr/local/lib64/kamailio/modules/ims_qos.so: undefined symbol: unref_dlg`
`0(15988) CRITICAL: <core> [core/cfg.y:3796]: yyerror_at(): parse error in config file /etc/kamailio_pcscf/kamailio.cfg, line 239, column 12-20: failed to load module`

Because of the following warning during compilation:
`Warning: rx_aar.c:184:13: warning: implicit declaration of function 'unref_dlg'; did you mean 'unref_cell'? [-Wimplicit-function-declaration] 
184 | unref_dlg(data->dlg, 1); 
| ^~~~~~~~~ 
`

atanas-wg2:
ops sorry. @alexyosifov  seems that include was missing: 
https://github.com/kamailio/kamailio/pull/3127

PR #3114, opened 2022-05-17
pv: log uninit $var()

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

We've come into the rare issue when a kamailio process used the old $var() value for the current request message and messed some things up.

LOG_WARN when $var() is used uninitialized in config. I can also just LOG_NOTICE this.

Can this be brought to 5.6?

Thank you,
Stefan

### DIFF
--- a/src/modules/pv/pv_core.c
+++ b/src/modules/pv/pv_core.c
@@ -2597,6 +2597,11 @@ int pv_get_scriptvar(struct sip_msg *msg,  pv_param_t *param,
 		res->ri = sv->v.value.n;
 		res->flags = PV_VAL_STR|PV_VAL_INT|PV_TYPE_INT;
 	}
+
+	if (!sv->init) {
+		LM_WARN("Script variable \"$var(%.*s)\" used uninitialized!\n", sv->name.len, sv->name.s);
+	}
+
 	return 0;
 }
 
@@ -2899,6 +2904,10 @@ int pv_set_scriptvar(struct sip_msg* msg, pv_param_t *param,
 		LM_ERR("error - cannot find svar\n");
 		goto error;
 	}
+
+	script_var_t *sv = (param->pvn.u.dname);
+	sv->init = 1;
+
 	if((val==NULL) || (val->flags&PV_VAL_NULL))
 	{
 		if(((script_var_t*)param->pvn.u.dname)->v.flags&VAR_TYPE_NULL)
--- a/src/modules/pv/pv_svar.h
+++ b/src/modules/pv/pv_svar.h
@@ -46,6 +46,7 @@ typedef struct script_var {
 	str name;
 	script_val_t v;
 	struct script_var *next;
+	int init;
 } script_var_t, *script_var_p;
 
 script_var_t* add_var(str *name, int vtype);


miconda:
This does not sound right, by design $var() is initialised to 0 and no reset on every message. It has to be set/reset explicitely in config.

If you need something specific per sip request, the xavp/avp/xavi/xavu should be used.

smititelu:
Hi,

Indeed, from what I've checked so far, the $var() value is initialized to 0. This is done when kamailio starts.

However, the issue we faced is more related to this excerpt from kamailio cookbooks, pseudovariables:
```
Note: A script variable persists over the Kamailio process in which it was initialized, so be sure of giving it a new value before reading it or you'll get the value asigned in any other previous message processed by the same Kamailio process (pid). 
```
Let me know if now its more clear what the PR tries to warn.

Thank you,
Stefan

henningw:
Sorry, I did not understand it as well. This is exactly as $var is supposed to work, it contains the value from previous processing. If no value has been given to it before it returns 0, as commented from Daniel.

miconda:
@smititelu - that note from the docs sounds ok for me (maybe because I wrote it, although I am not sure I did). Practically wants to say that assign explicitly a value before using it, not to get the value set a while ago in another SIP message processing context.

If you do:

```
request_route {
...
  while ($var(i) < 100) {
    ...
    $var(i) = $var(i) + 1;
  }
...
}
```

The while loop gets inside once because  $var(i) is 0 first time, but then remains set to 100 for all next requests. Therefore set it to the value you want before using it:

```
request_route {
...
  $var(i) = 0;
  while ($var(i) < 100) {
    ...
    $var(i) = $var(i) + 1;
  }
...
}
```

Feel free to change the statement in the docs, if you consider more clarifications there, but I don't see any benefit in storing an extra int field in the internal var structure.

Maybe $vn() offers better the kind of variable you need. It is like $var() but can have $null value (state).

smititelu:
Consider this scenario:
1'st part:
```
if (is_present_hf("SOME_HEADER")) {
    $var(data) = _get_data_from_some_header_;
}
```

2'nd part, later:
```
if ($var(data) != "") {
    _do_something_with_data_
}
```

2'nd part gets called even for SIP that does not have  SOME_HEADER header, with value set in previous SIP that had SOME_HEADER.

I wanted to log a warning for this, so if SIP with no SOME_HEADER header comes in first, in one routing process, I will see the warning log and know to set ```$var(data)="";``` before the 1'st part.

Thank you,
Stefan

pkuzak:
Hi @miconda . Is it because of memory consumption that you don't want to store an additional _int_ inside the _var_ structure?
The reason we came up with this PR is a strange incident we had that was rather difficult to debug because we, you could say, wrongly used script variables. 
Reading the docs is important, sure. But I think the project would benefit from this feature because it could help users to debug or avoid unwanted strange scenarios.

pkuzak:
It is not an alarm, it is just a log message warning the user that he uses something that is not initialized. I once used a while(true) statement in the configuration file, which also was a valid statement and intended by me. Back then I saw that there was a warning printed when starting Kamailio. That's where I saw that Kamailio does some checking this and Perls _use strict_ statement kind of motivated us to implement this feature.
Would it be OK if we would make it also configurable?

miconda:
@smititelu, @pkuzak - it is a misunderstanding of what $var() is supposed to do and how it acts. It is not only about an extra int field, it is about the concept behind $var().

What @smititelu gave as an example is what the note in docs try to say and what I gave as an example above. 

It can create big problems no matter is with the warning or not in the C code, because it still keeps the value from previous assignment.

Let's say you have this situation:

- first request has SOME_HEADER (the variable is assigned, and by that no warning log)
- the next request is without SOME_HEADER, but the var still has the value from first request and practically, during processing the 2nd request you use wrong value

Coming back to @smititelu example, you have to set var to a value that you can compare with, like:

```
$var(data)="";
if (is_present_hf("SOME_HEADER")) {
    $var(data) = _get_data_from_some_header_;
}
if ($var(data) != "") {
    _do_something_with_data_
}
```

Which is same concept as with my `while` example above.

Or use $x/avp() without setting a value and compare with $null.

As I said, I don't see any benefit with the PR, irrelevant of being optional or not.

smititelu:
Yes, I agree. The patch is just meant to merely give you a config hint, at best effort. Won't solve the wrong value problem or change $var() behavior in any way.

miconda:
From $var() design point of view, as @henningw said, it is bringing more confusions, because $var() is always initialised.

But you can also add a new variable that you think it meets better your needs.

smititelu:
Ok. I am closing the PR for now. Thanks a lot for the feedback.

PR #3120, opened 2022-05-20
core: const correctness

liviu-rolo:
benign const qualifiers in inline functions

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/core/ip_addr.h
+++ b/src/core/ip_addr.h
@@ -301,7 +301,7 @@ int is_mcast(struct ip_addr* ip);
 
 /* returns 1 if the given ip address is INADDR_ANY or IN6ADDR_ANY,
  * 0 otherwise */
-inline static int ip_addr_any(struct ip_addr* ip)
+inline static int ip_addr_any(const struct ip_addr* ip)
 {
 	int r;
 	int l;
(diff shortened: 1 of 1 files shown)


miconda:
Can you provide more details why these changes are useful or necessary?

liviu-rolo:
They are not necessary.
In fact, I needed const only in `init_dst_from_rcv` for an internal module, but, with the occasion, I added more const for consistency - including existing code like `su_cmp` and `su_getport`.
So, apart from consistency and style, they are neither useful for the community, I guess.

kamailio-sync:
I think they are useful. I have done several changes like this in the past. It allows the Compiler to optimize better and will also prevent unintended modifications which can lead to bugs

miconda:
@henningw - to my knowledge, there is no optimisation that a compiler can do to const parameters of a function.

I asked because the most of the functions there were not touched for ages, so not usual suspects for unintended changes.

Moreover, the commit message is somehow contradictory, first lines suggests is makes something correct (like it was wrong before), but then in the second line says the changes are benign.

But if people thinks it is better aesthetic this way, then ok.

henningw:
Thanks @miconda - you are right the commit message is a bit misleading, I've missed that, my fault.

kamailio-sync:
Its about const correctness, which is a good idea especially for API or core Code. https://isocpp.org/wiki/faq/const-correctness#overview-const

PR #3126, opened 2022-05-31
auth: Add pv_auth_check overload with second passwd parameter

sn4kebite:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This change modifies the `pv_auth_check` function to take an optional second password argument `passwd2`. The motivation for this change is to allow authenticating against two passwords, which is useful if you want to eg. rotate passwords periodically and still allow using the previous password during a transition period. You can technically do this with two calls to `pv_auth_check`, but with `nonce_count` enabled the first call will invalidate nc which will always cause the second call to fail.

### DIFF
--- a/src/modules/auth/auth_mod.c
+++ b/src/modules/auth/auth_mod.c
@@ -84,7 +84,10 @@ static int pv_www_authenticate2(struct sip_msg* msg, char* realm,
 static int fixup_pv_auth(void **param, int param_no);
 static int w_pv_auth_check(sip_msg_t *msg, char *realm,
 		char *passwd, char *flags, char *checks);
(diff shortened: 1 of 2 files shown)


miconda:
Wouldn't be better to add a flag that instructs not to invalidate the nc if authentication fails? In such case, one can even chain more than two calls of `pv_auth_check()`.

sn4kebite:
My coworker actually brought up the same idea after this change was made, but we decided to wait for feedback here first. I will make a separate PR for the added flag, then we can close this one if the flag PR gets accepted.

sn4kebite:
Replaced by #3133

PR #3127, opened 2022-05-31
ims_qos: fix compilation warning

atanasdb:


atanasdb:
Fix for a regression introduced by https://github.com/kamailio/kamailio/pull/3104
Without the patch ims_qos fails to load as described in:
https://github.com/kamailio/kamailio/pull/3104#issuecomment-1141950700

PR #3129, opened 2022-05-31
tm: new option to add additional response codes for trying next destinations when failover with multiple ips

emvondo:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This change allows the tm module to continue to try next destinations on receiving reply codes other than 503 from previous destinations when dns failover resolves to multiple ips.

A new parameter **failover_reply_codes**  added to the module gives hand to add codes or class of code to continue to try 
next destinations.
this parameter has the same format as ds_ping_reply_codes of dispatcher module.

```
modparam("tm", "failover_reply_codes", "code=403;code=488;class=5")
```

### DIFF
--- a/src/modules/tm/doc/params.xml
+++ b/src/modules/tm/doc/params.xml
@@ -1623,4 +1623,29 @@ modparam("tm", "enable_uac_fr", 1)
 		</example>
 	</section>
 
+	<section id="tm.p.failover_reply_codes">
+		<title><varname>failover_reply_codes</varname> (string)</title>
+		<para>
+			This parameter defines the response codes (only codes >= 300 and class >= 3), which enable
+			dns failover  to continue to try next ips when previous one fail. It is a list separated by
+			colons, where you may define either a single code (e.g. "code=408" would accept 408 as an 
+			additional response code) or a class of responses, you want to accept (e.g. "class=3" would accept
+			everything from 300 to 999 as valid response). 
+			(with default behaviour only 503 code or no response from previous one will try next ip).
+		</para>
+		<para>
+		<emphasis>
+			Default value is <quote></quote>
+		</emphasis>
+		</para>
+		<example>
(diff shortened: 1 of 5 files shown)


miconda:
Thanks for the PR!

For reference, the git commit message has to be prefixed with the component name, in this case `tm: `, not the pull request name.

Then, I don't think the list of reply codes and classes has to be in shared memory, they do not change at runtime. I see same is done in dispatcher module, so probably should be changed in both modules later.

I am going to merge it manually.

patrickarton:
> Thanks for the PR!
> 
> For reference, the git commit message has to be prefixed with the component name, in this case `tm: `, not the pull request name.
> 
> Then, I don't think the list of reply codes and classes has to be in shared memory, they do not change at runtime. I see same is done in dispatcher module, so probably should be changed in both modules later.
> 
> I am going to merge it manually.

Thanks for feedback Daniel.

i will pay more attention to git commit message next time.

for list of reply codes in shared memory i wanted also to be able to update it dynamically through kamcmd. 
even if right now i dont see (talking for myself) it happen quite often.

so i started by put them in shared memory to continue after.

Thanks.

emvondo:
> Thanks for the PR!
> 
> For reference, the git commit message has to be prefixed with the component name, in this case `tm: `, not the pull request name.
> 
> Then, I don't think the list of reply codes and classes has to be in shared memory, they do not change at runtime. I see same is done in dispatcher module, so probably should be changed in both modules later.
> 
> I am going to merge it manually.

Thanks for feedback Daniel.

i will pay more attention to git commit message next time.

for list of reply codes in shared memory i wanted also to be able to update it dynamically through kamcmd.
even if right now i dont see (talking for myself) it happen quite often.

so i started by put them in shared memory to continue after.

Thanks.

miconda:
Merged with 062e6daa0f149e01674667f77608afb35b6e099a .

PR #3133, opened 2022-06-03
auth: Add flag for not invalidating nc on auth failure

sn4kebite:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This change modifies `pv_www_authenticate` and related functions to add flag 32. If the flag is set and `nonce_count` is enabled then we skip updating nc in `pre_auth`. On success we call `check_nonce` once more to do the nc update. This can be used to chain calls to authenticate against multiple passwords.

### DIFF
--- a/src/modules/auth/api.c
+++ b/src/modules/auth/api.c
@@ -34,7 +34,7 @@
 #include "rfc2617_sha256.h"
 #include "challenge.h"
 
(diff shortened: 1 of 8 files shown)


miconda:
Thanks!

Issue #2981, opened 2021-12-20
The acc modules writes a random string into the sip_reason field for an ACK message (FAKED_REPLY)


CFrits:
### Description

While upgrading to kamailio version 5.5.2 I noticed that the sip reason is not always filled in correctly. In our current setup with kamailio version 4.4 the acc module writes the INVITE, ACK end BYE messages to a postgres database. Also the missed calls are logged. With version 5.5.2 but also with version 5.5.3 the sip reason seems to contain a random 2 digits.
It gets even worse when a call is missed and kamailio generates a 408 timeout. Then a random string of 15 characters is written to the database. Sometimes this results in a fatal error because the postgres driver can't escape the string.

### Troubleshooting

The relevant part in de code is in the function env_set_code_status in file acc_logic.c

    if (reply==FAKED_REPLY || reply==NULL) {
        /* code */
        acc_env.code_s.s =
            int2bstr((unsigned long)code, code_buf, &acc_env.code_s.len);
        /* reason */
        if (acc_env.reason.len == 0) { /* already extracted in case of locally generated replies */
            acc_env.reason.s = error_text(code);
            acc_env.reason.len = strlen(acc_env.reason.s);
        }

The sip reason is only filled in when the len is 0, but once the length is set it looks like it's not being reset to 0 again somewhere.
I can fix the problem by removing the if statement, but that is probably not the intention of the programmer. 

#### Reproduction

The wrong reason in the ACK can probably be reproduces with any call, but in my test setup an Asterisk system answers the call.
After the "OK" an ACK is received which is seen as a FAKED_REPLY. The reason.len however is still 2 and the reason is not set.
The new function "env_set_reason" is not being called in this case.

The issue with the 408 can be reproduced by calling an endpoint and let kamailio timeout. This has to be done twice to reproduce the error. This time kamailio wants to write the missed INVITE to the database with a random string of 15 characters for the sip reason.
To me it looks like the length of 15 comes from the length of the "Request Timeout" string but the pointer reason.s point to a memory part of a previous transaction which is already gone ?

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`
```
version: kamailio 5.5.2 (x86_64/linux) 55e232
[...]

PR #3137, opened 2022-06-07
acc: Fix reason code for locally generated replies.

CFrits:
- Use heap memory for the reason code instead of the stack memory.
- See issue #2981

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #2981 (replace XXXX with an open issue number)

### DIFF
--- a/src/modules/acc/acc_logic.c
+++ b/src/modules/acc/acc_logic.c
@@ -93,15 +93,18 @@ void env_set_totag(struct cell *t, struct sip_msg *reply)
 		tmb.t_get_reply_totag(t->uas.request, &acc_env.to_tag);
 }
 
-int env_set_reason(struct sip_msg *reply, str *buff) {
+int env_set_reason(struct sip_msg *reply, str *buff, int code) {
+/*
 	int i;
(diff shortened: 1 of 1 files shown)


src/modules/acc/acc_logic.c
```diff
@@ -93,15 +93,18 @@ void env_set_totag(struct cell *t, struct sip_msg *reply)
 		tmb.t_get_reply_totag(t->uas.request, &acc_env.to_tag);
 }
 
-int env_set_reason(struct sip_msg *reply, str *buff) {
+int env_set_reason(struct sip_msg *reply, str *buff, int code) {
+/*
```
linuxmaniac, line 97: please don't comment code, just remove it

miconda:
Indeed, the code that is no longer needed must be removed, otherwise it makes it hard to understand the changes done the patch and in long term the code will become a mess. Comments should be used only to explain the code, if one wants to check a previous version, git history is the right way.

miconda:
Thanks!

PR #3138, opened 2022-06-08
branch_failure route in case 503 and dns failover

vlitvinov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

When we use DNS failover in case of timeout kamailio executes the branch_failure route. In the case of the 503 - on_reply route. This commit is to equate behavior in both cases. So if there are more candidates we will always execute branch_failure route and on_reply route only if this is the last candidate.

### DIFF
--- a/src/modules/tm/config.c
+++ b/src/modules/tm/config.c
@@ -76,6 +76,7 @@ struct cfg_group_tm default_tm_cfg = {
 		500,				 /* default_code */
 		1,					 /* reparse_invite */
 
+		0, /* tm_dns_failover_branch_failure -- if 1 in case 503 will be executed branch_failure route*/
 		0, /* tm_blst_503 -- if 1 blocklist 503 sources, using tm_blst_503_min,
 		 * tm_blst_503_max, tm_blst_503_default and the Retry-After header
 		 * in the 503 reply */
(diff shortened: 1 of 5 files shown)


miconda:
The onreply_route is designed for executing when a sip reply is received and should be kept like that otherwise many configs will get unexpected results. Even without dns failover, the onreply route is not executed on retransmission timeout, but the failure route block.

Branch failure event route was added later and I am not sure how coherent its behaviour is. If it is not executed on 503, the it should be. But also onreply_route should be executed on 503 when the response is received from the network.

From my perspective, if you want to change when the onreply route is executed, you have to make it configurable, via a modparam, with the default behaviour being the one so far.

miconda:
You have to also document the new module parameter in the doc/ folder of tm. PR can be merged once the docs are update and no other comments to PR are made.

henningw:
@vlitvinov any update from your side?

miconda:
This got somehow out of sight for pretty long, but looking now at combined diff it seems that order of code execution changes and can have significant impact. As I understand, the loop with `t_send_branch()` is now executed before `onreply_route` handling. If I got it right, that can have unwanted effects, like sending out new branches before handling the response.

This PR needs proper review and testing before merging.

vlitvinov:
This is the main idea do not execute on_reply route in case 503 and available failover endpoint. The same behavior as with timeout.
There is no possibility to know did we try all candidates. In our case when we use the rtpengine we do delete but after that we can have successful reply from other failover candidate.
As I understand the behavior will change only in case when reply status will be changed in the reply route (to or from 503 in default). So only in case when we want to try or visa-versa don't want to try other candidates, but again we don't know anything about candidates.
This changes were done more than year ago and already works on few thousands installations without any problems (but only with new behavior).

Please let me know if any additional checks can be done from my side.

henningw:
Thank you for providing the update. Right now there seems to be some conflicts probably after some refactoring changes that have been done. Could you maybe re-base your changes and then developers could have another close look.

vlitvinov:
Hello. Thank you for your reply. Re-based.

PR #3143, opened 2022-06-13
pua dialoginfo: improvements

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
Changes from Sipwise flavor:
- refresh_pubruri_avps_flag: if enabled, this message flag indicates whether to refresh R-Uri from avps before sending the PUBLISH requests.
- improvement for use_pubruri_avps: use locking to avoid crashes since we are using shared memory
- local_identity_dlg_var: reporting dialog-information will use the value of the dialog variable if exists


miconda:
Fine to merge from my point of view!

Just a note that there might be some design issue with dialog vars, considering some reports on the tracker, but this has to be analyzed separately, not being something specific to PR, but impacts it.

PR #3144, opened 2022-06-14
tls_wolfssl: new module TLS stack based on wolfSSL

space88man:
New module: add wolfSSL as alternate TLS stack.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
This a new module: an alternate TLS implementation based on wolfSSL. The current tls module based on OpenSSL has many multi-process workarounds and can be quite fragile.

This is the initial code dump which is a copy of `tls/` and edited to compile with wolfSSL by using the OpenSSL compatibility layer. The `doc/` directory has not been changed.

The proposal is to get it into the code base as soon as possible so as to sync up with any ongoing changes in the `tls/` module. Any shared features can be extracted out into a common module: like certificate and configuration.

In the short-term the steps are:
* testing
* use native wolfSSL APIs (remove OpenSSL compatibility layer)
* remove OpenSSL multi-process  hacks in this module

This module is inspired by the `tls_wolfssl` module in the sister SIP project.

### DIFF
--- a/src/modules/tls_wolfssl/Makefile
+++ b/src/modules/tls_wolfssl/Makefile
@@ -1,0 +1,81 @@
+#
+# TLS module makefile
+#
+#
+# WARNING: do not run this directly, it should be run by the main Makefile
+
+include ../../Makefile.defs
+auto_gen=
+NAME=tls_wolfssl.so
+
+# set to yes when wanting to link with static libraries
+LIBSSL_STATIC ?= no
+# set to yes when wanting to link with static libraries compiled from source
+LIBSSL_STATIC_SRCLIB ?= no
+# set to the path of the folder with static libraries compiled from source
+LIBSSL_STATIC_SRCPATH ?= /usr/local/src/openssl
+
+ifeq ($(CROSS_COMPILE),)
+WOLFSSL_BUILDER=$(shell \
+	if pkg-config --exists wolfssl; then \
+		echo 'pkg-config wolfssl'; \
+	fi)
+endif
+
+ifneq ($(WOLFSSL_BUILDER),)
+
+ifneq ($(WOLFSSL_STATIC),yes)
+	DEFS += $(shell $(WOLFSSL_BUILDER) --cflags)
+	LIBS += $(shell $(WOLFSSL_BUILDER) --libs)
+
+else # $(WOLFSSL_STATIC),yes)
+
+	DEFS += -DKSR_WOLFSSL_STATIC
+
+ifneq ($(WOLFSSL_STATIC_SRCLIB),yes)
+	## when static libs (*.a) from packages are compiled with -fPIC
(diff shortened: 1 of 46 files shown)


miconda:
Thanks for this contribution!

Is the module at the phase of "it compiles only" or was it also tested a bit with some tls connections and SIP traffic?

space88man:
It is at "it compiles only" + server starts without segfault.
> 
> Is the module at the phase of "it compiles only" or was it also tested a bit with some tls connections and SIP traffic?

miconda:
Thanks for the details and further work on it! I can merge it all grant you access to the kamailio repo in order to continue the development directly on the main source tree. You can still do pull requests if you want other developers to comment on changes to this module. If you want to push commits to other modules or components, it is recommended to make pull requests, so the developers have a chance to review and avoid conflicts on working on same code part at that moment.

space88man:
> Thanks for the details and further work on it! I can merge it all grant you access to the kamailio repo in order to continue the development directly on the main source tree. You can still do pull requests if you want other developers to comment on changes to this module. If you want to push commits to other modules or components, it is recommended to make pull requests, so the developers have a chance to review and avoid conflicts on working on same code part at that moment.

Yes - I would appreciate access for further development of this module. Thank you.

henningw:
Hello, thanks for the pull request. It was merged really fast. I would appreciate some comments on how to address the code duplication from the existing tls module. Just checked briefly, but several header files are completely identical, in some cases just a few differences (tls_rpc.h, tls_ct_q.h etc..) . Also some implementation is idential (tls_verify.c) The approach from tlsa module to just include the headers if they are not modified seems better to me than to just copy them completely. This is one of the more complicated modules, and now we are having three tls modules.

miconda:
@henningw - if you read the PR description: `Any shared features can be extracted out into a common module: like certificate and configuration.`

But first is to get the module working properly, if proved not feasible at the end, then it can be removed. If ok, then think about next steps.

henningw:
Thanks @miconda for the clarification. If this is work in progress and will be then further improved in the git master, fine with me. I did not got from the description that this extraction or refactoring regarding code duplication was planned from the original author.

miconda:
@space88man: there are many deployments on older systems, so I would not remove anything from tls module right now.

I think it is better to get first the tls_wolfssl working properly as an alternative to the tls (openssl) module. But, being a new module, it can be developed to only support the newer versions of libwolfssl and without being compatible with old versions of libssl (openssl). In other words, you can remove the parts that related to OpenSSL < 1.1.1 in the new module tls_wolfssl.

Once tested and results are satisfactory, then we can look what are the common parts still left between the two modules and decide what would be a best way to deal with. But I won't spend time now to think about nor even attempt to extract code from tls module to another one.

Issue #3168, opened 2022-06-29
Errormessage TLS read:error:0D0E10DF:asn1 encoding routines:asn1_get_uint64:too large


MartinPalmAGDS:
We use Kamilio as a proxy server to connect softphones to our PBX systems.
We have run into some TLS problems. These problems result in a shut down of the TLS connection after the error is encountered.

We have looked into serial numbers of certificates. The current code in kamailio seems to be broken due to changes in the world around... many certificates today do not adhere to the previously assumed behaviour of certificate serial numbers anymore.
The error which causes the connection to drop seems to be related to a comparison of the certificate serial numbers during renegotiation.
tls_err_ret(): TLS read:error:0D0E10DF:asn1 encoding routines:asn1_get_uint64:too large
Kamailio expects serial numbers to fit in a 64 bit unsigned, however during the last years, starting around 2003 more and more systems use automatically generated serial numbers, which are frequently 128 bit long.
The serialnumber of our own Auerswald root certificate looks to be only 64 bit long, but during the negotiation is presented as 9 bytes starting with 00h, which is interpreted as a 72bit value. Trying to put this serial number in a 64 bit long kills the connection.
Due to the different ways that are used to generate serial numbers now, there is no simple algorithm to condense the number into a 64bit value, and the whole number should be compared to see if it is same or higher.
We currently have disabled renegotiation to avoid the connection being dropped during renegotiation. Which however can lead to the connection being dropped by the other side during renegotiation, but this is not a good permanent workaround!

#### Reproduction
This issue happens infrequently

### Possible Solutions
No workaround possible on our side

### Additional Information
Kamailio Version 5.5.4 
LibSSL 1.1.1.n

* **Operating System**:
Linux, Debian oldstable

MartinPalmAGDS:
Hello,

fix has been tested – works as intended. Well done!
Mit freundlichen Grüßen / Best regards
Martin Palm
Project Manager and Agile Coach
Phone +49 5306 9219-573<tel:+49%205306%209219-573> | Fax +49 5306 9219-94<tel:+49%205306%209219-94>
E-Mail ***@***.******@***.***>
[https://assets.auerswald.de/mail/2020/icon_website.png]  auerswald.de<https://www.auerswald.de> | [https://assets.auerswald.de/mail/2020/icon_onlineshop.png]   shop.auerswald.de<https://shop.auerswald.de>

[...]

PR #3171, opened 2022-06-29
tls: cert serial number can exceed uint64

space88man:
#### Pre-Submission Checklist

#### Type Of Change
- [  ] Small bug fix (non-breaking change which fixes an issue)
- [  ] New feature (non-breaking change which adds new functionality)

#### Checklist:
- [X] Tested changes locally
- [X] Related to issue #3168

#### Description
Certificate serial numbers can be large integers (> 2^64). Parse them as BIGNUM and store the string
in pv; we cannot store as integer anymore. The original code handles 2^64 with a static buffer of 19.
We upsize to 2^256 with a static buffer of 80.

**May break configurations** that try to extract the sn pv as an integer
(does anyone actually do this?)

Addresses #3168.

### DIFF
--- a/src/modules/tls/tls_select.c
+++ b/src/modules/tls/tls_select.c
@@ -630,24 +630,35 @@ static int pv_validity(sip_msg_t* msg, pv_param_t* param, pv_value_t* res)
 }
 
 
-static int get_sn(str* res, int* ires, int local, sip_msg_t* msg)
+static int get_sn(str* res, int local, sip_msg_t* msg)
 {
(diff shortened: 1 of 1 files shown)


miconda:
Thanks for this PR! I noticed some functions return pointers that are then used and later freed. I guess that they have to be checked if they are `NULL`, expecting there will be crashes when they are used and they are `NULL`.

miconda:
Thanks, testing and confirming would be useful in order to merge and backport in time for 5.6.1.

PR #3183, opened 2022-07-11
presence: No autocommit+rollback if no active watchers

Ozzyboshi:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
This is a small performance improvement, the goal is to avoid issuing rollbacks if not necessary.
On my setup, Kamailio presence executes a lot of ROLLBACKS in MariaDB since function process_dialogs() runs repeatedly.
What see inside MariaDB logs is the following

```
SET autocommit=0
select `presentity_uri`,`callid`,`to_tag`,`from_tag`,`event` from `active_watchers` where `updated`=3 AND `event`<>'presence.winfo'\G
ROLLBACK
```

Maybe we can improve this behaviour telling presence to do autocommit to do not execute them if the select returns zero elements?

### DIFF
--- a/src/modules/presence/notify.c
+++ b/src/modules/presence/notify.c
@@ -2867,6 +2867,7 @@ int process_dialogs(int round, int presence_winfo)
 	str ev_sname, winfo = str_init("presence.winfo");
 	int now = (int)time(NULL);
 	int updated = 0;
+	int no_active_watchers = 0;
 	db_query_f query_fn = pa_dbf.query_lock ? pa_dbf.query_lock : pa_dbf.query;
 
 	query_cols[n_query_cols] = &str_updated_col;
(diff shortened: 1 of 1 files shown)


miconda:
> Maybe we can improve this behaviour telling presence to do autocommit to do not execute them if the select returns zero elements?

I am not sure I understand what you wanted to ask? Maybe you can elaborate more.

linuxmaniac:
The thing here is to avoid using database transaction to select and just use it if there's going to be update commands later

miconda:
@linuxmaniac: that's about the current patch of this PR, right? I wanted to clarify if @Ozzyboshi wants to propose another improvement that covers the case of current PR as well.

From my point of view, the PR can be merged, if no improvement/alternative is been planned in the near future.

linuxmaniac:
yes, this is about this patch. No other improvements or alternatives were planned

miconda:
It was the question at the end of @Ozzyboshi PR description that I wanted to clarify.

PR #3191, opened 2022-07-19
pv: add monotonic clock to TimeVal pseudovariable

xadhoom:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Sometime is helpful to have a monotonic timer to attach to events in order to discard out-of-order ones on receiving side. Current time values are not ideal since are subject to clock skews, adjustment and so on. Since we have system monotonic clock we can use that value to have a monotonic counter. 

To be 100% fair CLOCK_MONOTONIC can still be skewed only by ntp adjustment, that's why we try to use CLOCK_MONOTONIC_RAW which is linux-only (afaik) that do not suffer from such adjustments.

The new pseudovariable is `$TV(Sm)`, which returns the monotonic counter as a string. Note that even if is based on clock, it starts from an unspecified point in time, so should really be treated as an always increasing counter.

### DIFF
--- a/src/modules/pv/pv_time.c
+++ b/src/modules/pv/pv_time.c
@@ -23,6 +23,8 @@
 #include <ctype.h>
 #include <string.h>
 #include <stdlib.h>
+#include <stdint.h>
+#include <inttypes.h>
 #include <stdio.h>
 #include <time.h>
 #include <sys/time.h>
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the pull-request. Could you also please add the documentation for the new PV in our [wiki](https://github.com/kamailio/kamailio-wiki/blob/main/docs/cookbooks/devel/pseudovariables.md)? Then it could be probably merged, if no other comments come in.

xadhoom:
Ok,

I'm on vacation right now, will do by the end of August.

Il mer 10 ago 2022, 11:28 Henning Westerholt ***@***.***> ha
scritto:

xadhoom:
Done in https://github.com/kamailio/kamailio-wiki/pull/14

henningw:
Thanks, merged

PR #3193, opened 2022-07-21
tls: add timestamp and sni info for a session

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

### DIFF
--- a/src/modules/tls/tls_rpc.c
+++ b/src/modules/tls/tls_rpc.c
@@ -115,6 +115,9 @@ static void tls_list(rpc_t* rpc, void* c)
 	struct tls_extra_data* tls_d;
 	struct tcp_connection* con;
 	int i, len, timeout;
+	struct tm timestamp;
+	char timestamp_s[128];
+	const char* sni;
 
 	TCPCONN_LOCK;
 	for(i = 0; i < TCP_ID_HASH_SIZE; i++) {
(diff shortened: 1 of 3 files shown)


henningw:
@smititelu Thanks for the pull request, did a quick review. It would be great if you could also check for eventual errors at the snprintf call for the time stamp. It might for some reasons fails, as the TLS module might see error conditions during heavy load.

smititelu:
Updated PR. Will print "N/A" for timestamp too, if snprintf call fails.

src/modules/tls/tls_rpc.c
```diff
@@ -132,14 +135,29 @@ static void tls_list(rpc_t* rpc, void* c)
...
+					timestamp.tm_min, timestamp.tm_sec) < 0) {
+				timestamp_s[0] = 'N';
+				timestamp_s[1] = '/';
+				timestamp_s[2] = 'A';
+				timestamp_s[3] = '\0';
+			}
+			rpc->struct_add(handle, "dsdsdsd",
 					"id", con->id,
+					"timestamp", timestamp_s,
 					"timeout", timeout,
```
oej, line 150: Can we get SNI here as well?

oej:
Great addition.

src/modules/tls/tls_rpc.c
smititelu: you mean to print sni right after the "id"?!

oej: Sounds like a good place.

smititelu: :+1: updated PR accordingly

oej:
So what do we do if the SNI doesn't match any known SIP domain we handle? Like you have a wildcard cert in your organisation and someone connects to the wrong DNS name?

smititelu:
Will print that wrong requested name.

henningw:
@smititelu thanks for the updates. @oej This is a generic question, right? Then it should be probably discussed on the mailing list and not in this PR.

oej:
It is related to getting SNI available in the config - but could of course be discussed on the mailing list as well. I was wondering what Stefan thought about this since he added better SNI support :-)

oej:
But it is not a blocker for this PR, not a requirement.

oej:
> Will print that wrong requested name. E.g.

Right, but we should propably think about closing early if that's the case or find a way to handle it in the routing script properly. But at that point, we've already accepted the SIP message. Let's continue this discussion outside of this PR.

miconda:
Thanks, merging!

PR #3206, opened 2022-08-02
siptrace: Trace in-dialog ACK and spiraled dialogs

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

While testing this module, I've seen it doesn't trace in-dialog ACKs.
Found this old maillist discussion: https://lists.kamailio.org/pipermail/sr-users/2015-June/088731.html 
Added a condition to trace also the in-dialog ACKs. Is it ok to keep it as default or should I guard it via a modparam?

Thanks,
Stefan

### DIFF
--- a/src/modules/siptrace/doc/siptrace_admin.xml
+++ b/src/modules/siptrace/doc/siptrace_admin.xml
@@ -651,6 +651,43 @@ modparam("siptrace", "evcb_msg", "ksr_siptrace_msg")
 </programlisting>
                 </example>
         </section>
+
+	<section id="siptrace.p.trace_dialog_ack">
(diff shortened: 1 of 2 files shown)


henningw:
Thanks for the pull-request. This is a small patch and it uses similar dialog callbacks as already used for the other messages. In my opinion we do need a modparam for that as it should trace all messages belonging to the dialog. But lets wait if there are maybe other comments.

smititelu:
Updated PR. Besides in-dialog ack tracing, I've also added tracing of spiraled dialogs. Guarded behavior via 2 new modparams.

Issue #3178, opened 2022-07-08
usrloc keepalive does not work for TLS connections


abrezinsky:
### Description

The keepalive (ka*) features of usrloc does not work for TLS connections (and probably TCP).  If a TLS connection is registered, the keepalive code attempts to send OPTIONS packets to the endpoint using UDP.  Additionally, a TCP/TLS connection is not re-used and the code attempts to create a new one each request.

#### Reproduction

Enable keepalive and register a phone using TLS

#### Log Messages

Kamailio Host: 10.15.5.116
Phone: 10.15.5.117

Phone is registered and seen in the registry

```
kamctl ul show
{
  "jsonrpc":  "2.0",
  "result": {
    "Domains":  [{
        "Domain": {
          "Domain": "location",
          "Size": 1024,
          "AoRs": [{
              "Info": {
                "AoR":  "anonymous",
                "HashID": 3819435025,
                "Contacts": [{
                    "Contact":  {
                      "Address":  "sips:anonymous@10.15.5.117:5061",
                      "Expires":  3531,
                      "Q":  -1,
                      "Call-ID":  "d06e987694afb559",
                      "CSeq": 521595368,
                      "User-Agent": "testclient/1.0",
                      "Received": "[not set]",
                      "Path": "[not set]",
                      "State":  "CS_NEW",
                      "Flags":  0,
                      "CFlags": 0,
                      "Socket": "tls:10.15.5.116:5061",
                      "Methods":  7903,
                      "Ruid": "uloc-62c87f5c-1b6b-2",
                      "Instance": "<urn:uuid:00000000-0000-1000-8000-0010492e1879>",
                      "Reg-Id": 0,
                      "Server-Id":  0,
                      "Tcpconn-Id": 17,
                      "Keepalive":  0,
                      "Last-Keepalive": 1657310753,
                      "KA-Roundtrip": 0,
                      "Last-Modified":  1657310753
                    }
                  }]
              }
            }
  ],
          "Stats":  {
            "Records":  1,
            "Max-Slots":  1
          }
        }
      }]
  },
  "id": 14871
```

When the keepalive code runs it shows the following:

```
Jul  8 14:27:52 localhost kamailio[6976]: 33(7012) DEBUG: usrloc [ul_keepalive.c:112]: ul_ka_urecord(): keepalive for aor: anonymous
[...]

PR #3219, opened 2022-08-16
usrloc: re-use TCP connections for keepalive

abrezinsky:
- Fixes GH #3178. Sets TCP connection id for keepalive based on the value in the database
  If unset, tcpconn_id will remain 0 and keepalives will work the same as they do today
- Update documentation to clarify when it will use received vs AOR for keepalive

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #3178 (replace XXXX with an open issue number)

#### Description

Updated usrloc module to re-use an active TCP session (if it exists).  I marked this breaking only because the current code will use a UDP OPTIONS keepalive method even if the device is registered over TCP.  I doubt that's what anyone would be expecting so I think the change is ok.

### DIFF
--- a/src/modules/usrloc/dlist.c
+++ b/src/modules/usrloc/dlist.c
@@ -114,6 +114,7 @@ int ul_ka_db_records(int partidx)
 	keys2[5] = &ul_ruid_col;
 	keys2[6] = &ul_user_col;
 	keys2[7] = &ul_domain_col;
+	keys2[8] = &ul_con_id_col;
 
 	/* where fields */
 	keys1[0] = &ul_expires_col;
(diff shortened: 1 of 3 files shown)


henningw:
Thanks for the PR. You are introducing a new column in the database. For that it would be necessary to extend the XML database scheme at src/lib/srdb1/schema. Its also necessary to increment the usrloc version number in the XML and also the module source code.

miconda:
There is no new column introduced, the connection id is there for quite some time. As I understand, it just loads the connection id for db-only mode case.

henningw:
@miconda thanks - wrongly conclusion from my side from the new introduced DB API code in the diff.

miconda:
Thanks, merging!

Issue #3216, opened 2022-08-15
topos: Add parameter to disable comma separated values in one Single Via, Record-Route or Route Headers


patrickarton:
### Description
Hello , 
i would like to have parameter in topos to disable comma separated values in one single **Via**, **Record-Route** or **Route** header.

some ua we are interacting with dont handle that (even if its rfc compliant).
and so most scenario calls fail with those particular ua. 

especially for responses we should if possible send the response with the same format of **Request received Via,Route or Record-Route**.

### Expected behavior
topos module should be able to put each Via, Route or Record-Route value in its own header and possibly (not an easy task)
keep the same request received format in response.

#### Actual observed behavior
topos module always combine all values of Via,Record-Route or Route header in a single one.

#### Debugging Data

```
(paste your debugging data here)
```

#### Log Messages

```
(paste your log messages here)
```

#### SIP Traffic

```
(paste your sip traffic here)
```

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
(paste your output here)
```

* **Operating System**:

```
(paste your output here)
```

PR #3220, opened 2022-08-16
topos: disable multiple comma separated values in One Single Via, Record-Route or Route header if needed

emvondo:
Topos module always put a single Via header with multiple comma separated values.
Even if this behaviour is RFC compliant there should a possibilty to disable that behaviour four uas that dont support/handle that.

this pull request adds a new parameter topos module to disable that behaviour if needed otherwise the default topos compact form is used.

```
modparam("topos", "separate_header_values", 1) # for Via only 
modparam("topos", "separate_header_values", 2) # for Record-Route only 
modparam("topos", "separate_header_values", 4) # for Route only 

modparam("topos", "separate_header_values", 7) # for Via, Record-Route  and Route 


The following options are available:

 (1) - disable multiple comma separated values for Via  header
 (2) - disable multiple comma separated values for Record-Route header
 (4) - disable multiple comma separated values for Route header
```

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3216

### DIFF
--- a/src/modules/topos/doc/topos_admin.xml
+++ b/src/modules/topos/doc/topos_admin.xml
@@ -546,6 +546,34 @@ modparam("topos", "methods_nocontact", "CANCEL,PRACK")
 </programlisting>
 		</example>
 	</section>
+	<section id="topos.p.separate_header_values">
+		<title><varname>separate_header_values</varname> (int)</title>
+		<para>
+			List of headers to disable multiple comma separated values inserted in compact form.
+			Altough compact form is RFC compliant this paramter gives possibilty to disable
+			compact form header values for UA that dont support/handle it.
+
+			The following options are available:
+
+			(1) - disable multiple comma separated values for Via  header
+			(2) - disable multiple comma separated values for Record-Route header
+			(4) - disable multiple comma separated values for Route header
+
+		</para>
+		<para>
+		<emphasis>
+			Default value is <quote>0</quote>.
+		</emphasis>
+		</para>
+		<example>
+		<title>Set <varname>separate_header_values</varname> parameter</title>
+		<programlisting format="linespecific">
+...
+modparam("topos", "separate_header_values", 1)
+...
+</programlisting>
+		</example>
+	</section>
 	<section id="topos.p.methods_noinitial">
 		<title><varname>methods_noinitial</varname> (str)</title>
 		<para>
--- a/src/modules/topos/doc/topos_admin.xml
+++ b/src/modules/topos/doc/topos_admin.xml
@@ -546,8 +546,8 @@ modparam("topos", "methods_nocontact", "CANCEL,PRACK")
 </programlisting>
 		</example>
 	</section>
-	<section id="topos.p.separate_header_values">
-		<title><varname>separate_header_values</varname> (int)</title>
+	<section id="topos.p.header_mode">
+		<title><varname>header_mode</varname> (int)</title>
 		<para>
 			List of headers to disable multiple comma separated values inserted in compact form.
 			Altough compact form is RFC compliant this paramter gives possibilty to disable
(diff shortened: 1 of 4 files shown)


patrickarton:
Hello can you please remove the uncessary condition if (c==0 || c== 1) {}  and leave only the body of that condition please and remove also the unused c variable. we dont need it.

Thanks.

emvondo:
Hello can you please remove the uncessary condition if (c==0 || c== 1) {}  and leave only the body of that condition please and remove also the unused c variable. we dont need it.

should be like this 

```
int tps_reappend_via_separate_header(sip_msg_t *msg, tps_data_t *ptsd, str *hbody)
{
        str hname = str_init("Via");
        int i;
        str sb;
        char *p = NULL;

        if(hbody==NULL || hbody->s==NULL || hbody->len<=0 || hbody->s[0]=='\0')
            return 0;

        sb.len = 1;
        p = hbody->s;
        for(i=0; i<hbody->len-1; i++) {
            if(hbody->s[i]==',') {
                if(sb.len>0) {
                    sb.s = p;
                    if(sb.s[sb.len-1]==',') sb.len--;
                    if(tps_add_headers(msg, &hname, &sb, 0)<0) {
                        return -1;
                    }
                }
                sb.len = 0;
                p = hbody->s + i + 1;
            }
            sb.len++;
        }


        if(sb.len>0) {
                sb.s = p;
                if(sb.s[sb.len-1]==',') sb.len--;
                if(tps_add_headers(msg, &hname, &sb, 0)<0) {
                    return -1;
                }
        }


        return 0;
}

```

Thanks.

miconda:
@emvondo: if you want to change the commit of the PR, you can update on your local copy and force push to the same git branch.

emvondo:
> @emvondo: if you want to change the commit of the PR, you can update on your local copy and force push to the same git branch.

Thanks Daniel.

its done.

henningw:
Ok, sounds good. As you probably know, the documentation (XML docbook) needs to be extended as well, to be ready for the merge. But its usually a good idea to do that at the end after all code changes were done.

emvondo:
> Ok, sounds good. As you probably know, the documentation (XML docbook) needs to be extended as well, to be ready for the merge. But its usually a good idea to do that at the end after all code changes were done.

Thanks for feedback.

i pushed updates for Via , Record-Route and Route header.

henningw:
Thank you, lgtm. @miconda any comments from your side? Otherwise I think it could be merged, preferable squashed to one commit.

miconda:
The name of the parameter `separate_header_values` should be changed to `header_mode` to be even more generic and reused if anything else related to header management needs to be added in the future (e.g., add Via before or after Route).

From implementation point of view, the current solution is "good enough" because it is doing a basic split on comma, working with the common format seen for these headers. Otherwise, at least the `Record/-Route` can have comma in its value, being defined as `name-addr` by grammar and in the eventuality of a display name, comma is allowed inside. Not sure if Via is allowed to have quoted value parameter, because I think inside it is also allowed to have comma.

Anyhow, given that this commit adds this behaviour for particular deployments when some UA doesn't support headers with multiple comma separated values, I am fine to merge the implementation as it is, after renaming the parameter.

emvondo:
> The name of the parameter `separate_header_values` should be changed to `header_mode` to be even more generic and reused if anything else related to header management needs to be added in the future (e.g., add Via before or after Route).
> 
> From implementation point of view, the current solution is "good enough" because it is doing a basic split on comma, working with the common format seen for these headers. Otherwise, at least the `Record/-Route` can have comma in its value, being defined as `name-addr` by grammar and in the eventuality of a display name, comma is allowed inside. Not sure if Via is allowed to have quoted value parameter, because I think inside it is also allowed to have comma.
> 
> Anyhow, given that this commit adds this behaviour for particular deployments when some UA doesn't support headers with multiple comma separated values, I am fine to merge the implementation as it is, after renaming the parameter.

Thanks for feedback @miconda  changes done.

henningw:
Thank you all, merged.

Issue #3218, opened 2022-08-16
dialog: db issues in master branch


shaneqld:
### Description

While debugging an issue in Kamailio 5.4, I compiled/installed master to see if the issue persists. However, I immediately ran into multiple issues with dialog in master:

#### Using enable_dmq=1

With dmq enabled, I receive the following log lines which did not appear when running 5.4:

```
CRITICAL: dialog [dlg_timer.c:129]: insert_dlg_timer(): Trying to insert a bogus dlg tl=0x7fa0bc21fc80 tl->next=0x7fa0bbf46ba8 tl->prev=0x7fa0bbf46ba8
CRITICAL: dialog [dlg_dmq.c:314]: dlg_dmq_handle_msg(): Unable to insert dlg timer 0x7fa0bc21fc20 [2687:66928]
```

#### Using database

When using database persistence (sqlite and db_mode=3 (SHUTDOWN)), dialogs are not persisted across restarts (which was working in 5.4). There are no obvious log lines, except when I end the call after restarting Kamailio I receive a "unable to find dialog for BYE" message which would be expected if the dialog did not persist.

### Troubleshooting

- I ran the same configuration using 5.4 and neither issue occurred.
- I attempted to identify the troublesome commit by checking out/compiling older commits based on the commit history of the dialog module, but could only go so far back until builds started failing with `module not found: modules/siprepo`. As such I was not able to identify when the behavior changed.

#### Reproduction

DMQ issue: Use dialog module with dmq enabled and start a call that has `dlg_manage();` in the path. My configuration has the following settings:

```
modparam("dmq", "server_address", SERVER_ADDRESS_DMQ)
modparam("dmq", "notification_address", "redacted")
modparam("dmq", "notification_address", "redacted")
modparam("dmq", "num_workers", 6)
modparam("dialog", "default_timeout", 14520)
modparam("dialog", "profiles_with_value", "group")
modparam("dialog", "enable_dmq", 1)
modparam("dialog", "h_id_start", 0)
modparam("dialog", "h_id_step", 16)
```

DB issue: Use dialog module with db enabled. Start a call that has `dlg_manage();` in the path and keep the call open. Observe `kamcmd dlg.list` shows the dialog. Restart kamailio. Observe `kamcmd dlg.list` no longer lists any dialogs. My configuration has the following additional settings:

```
modparam("dialog", "db_url", "sqlite:////kamailio/kamailio.sqlite")
modparam("dialog", "db_mode", 3)
```

[...]

PR #3221, opened 2022-08-17
dialog: dont reset dlg_db_mode in POSTCHILDINIT for DB_MODE_SHUTDOWN

shaneqld:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #3218

#### Description

Per the DB issue in #3218:

I've determined the commit where dialog with db_mode SHUTDOWN stopped working is 1ff86ffceede46c7a67fec92c8319c34c916a545.

This is the change where  `PROC_POSTCHILDINIT` was enabled/used.

In dialog.c `child_init()`, there is code that resets the DB mode to NONE if not running in the main process:

```c
	/* in DB_MODE_SHUTDOWN only PROC_MAIN will do a DB dump at the end, so
	 * for the rest of the processes will be the same as DB_MODE_NONE */
	if (dlg_db_mode==DB_MODE_SHUTDOWN && rank!=PROC_MAIN)
		dlg_db_mode = DB_MODE_NONE;
```

However, because PROC_POSTCHILDINIT has now been enabled for the module, this code effectively disables DB persistence as PROC_POSTCHILDINIT is called in the main process after PROC_MAIN.

I have modified the rank check to `rank!=PROC_POSTCHILDINIT` and can confirm that DB persistence in SHUTDOWN mode is working once again.

Note that the PR changes the logic to:

```c
	if (dlg_db_mode==DB_MODE_SHUTDOWN && rank!=PROC_POSTCHILDINIT)
		dlg_db_mode = DB_MODE_NONE;
```

Technically dlg_db_mode will be reset to DB_MODE_NONE when called with rank PROC_MAIN, but the value will be restored when PROC_POSTCHILDINIT is called. I assume this behavior is OK, but otherwise I can change it to:

```c
	if (dlg_db_mode==DB_MODE_SHUTDOWN && rank!=PROC_MAIN && rank!=PROC_POSTCHILDINIT)
		dlg_db_mode = DB_MODE_NONE;
```

### DIFF
--- a/src/modules/dialog/dialog.c
+++ b/src/modules/dialog/dialog.c
@@ -824,7 +824,7 @@ static int child_init(int rank)
 
 	/* in DB_MODE_SHUTDOWN only PROC_MAIN will do a DB dump at the end, so
 	 * for the rest of the processes will be the same as DB_MODE_NONE */
-	if (dlg_db_mode==DB_MODE_SHUTDOWN && rank!=PROC_MAIN)
+	if (dlg_db_mode==DB_MODE_SHUTDOWN && rank!=PROC_POSTCHILDINIT)
 		dlg_db_mode = DB_MODE_NONE;
 	/* in DB_MODE_REALTIME and DB_MODE_DELAYED the PROC_MAIN have no DB handle */
 	if ( (dlg_db_mode==DB_MODE_REALTIME || dlg_db_mode==DB_MODE_DELAYED) &&


henningw:
Thanks for the PR. This commit was done as part of several changes related to the POSTCHILDINIT topic. Maybe @miconda can have a look, having done the original change to make sure we do not introduce other side-effects.

miconda:
@shaneqld: thanks for tracking it and troubleshooting details!

If `dlg_db_mode = DB_MODE_NONE` is done on `PROC_MAIN`, then on `PROC_POSTCHILDINIT` is not restored, I don't see how that could happen. Therefore it should better be:

```
if (dlg_db_mode==DB_MODE_SHUTDOWN && rank!=PROC_MAIN && rank!=PROC_POSTCHILDINIT)
		dlg_db_mode = DB_MODE_NONE;
```

shaneqld:
Hi @miconda 

> If dlg_db_mode = DB_MODE_NONE is done on PROC_MAIN, then on PROC_POSTCHILDINIT is not restored, I don't see how that could happen. 

It's restored as the value is reset at the start of `child_init()` as so:

```c
static int child_init(int rank)
{
	dlg_db_mode = dlg_db_mode_param;
```

Happy to update the PR if checking against both `PROC_MAIN` and `PROC_POSTCHILDINIT` is preferred.

miconda:
Indeed, I didn't spot that line, thinking that `dlg_db_mode` is the modparam variable.

PR #3228, opened 2022-08-30
Add More Logging Option

webstean:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x ] Tested changes locally

#### Description

Will you accept this? I find this extremely useful for troubleshooting. If accepted I'd like to add to most of the routing blocks.

### DIFF
--- a/etc/kamailio.cfg
+++ b/etc/kamailio.cfg
@@ -589,6 +589,10 @@ request_route {
 # Wrapper for relaying requests
 route[RELAY] {
 
+#!ifdef WITH_MORELOGGING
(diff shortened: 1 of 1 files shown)


miconda:
Thanks for the PR, but I don't think this is suitable for default kamailio.cfg, it is more a customisation you would want. There is log_prefix that makes all log message to include cseq (with method), call id, src ip, ...

For the future, note that you have to prefix the first line of the commit message with the component you change, see the contributing guidelines:

  - https://github.com/kamailio/kamailio/blob/master/.github/CONTRIBUTING.md#commit-message-rules

PR #3249, opened 2022-09-21
Update sca.c - Fix done for Appearance purge stale

bensly-shanmugam:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

We have observed that when the call is answered and it is in active state more than 30 seconds, kamailio sends the notify messages with the "Idle" appearance to all the subscribed extensions. Practically it should not send the idle notification when the call is in active state.
	We have tried to change the mod parameter "purge_expired_interval" to 3600 seconds. but still idle notification transmit in 30 seconds. To overcome this issue, we have commented below line in the sca.c file.
	
		register_timer( sca_appearance_purge_stale, sca, sca->cfg->purge_expired_interval);

### DIFF
--- a/src/modules/sca/sca.c
+++ b/src/modules/sca/sca.c
@@ -381,8 +381,8 @@ static int sca_mod_init(void)
 
 	register_timer(sca_subscription_purge_expired, sca,
 			sca->cfg->purge_expired_interval);
-	register_timer(
-			sca_appearance_purge_stale, sca, sca->cfg->purge_expired_interval);
+	//register_timer(
+	//		sca_appearance_purge_stale, sca, sca->cfg->purge_expired_interval);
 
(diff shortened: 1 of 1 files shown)


miconda:
If some code is not needed, then it should be removed, not commented.

However, I am not sure if this is a proper fix. I am not familiar with the module and its code, so I may be wrong, but feels like removing a callback that removing a callback for cleanup state is somehow not right. Maybe it should be investigated why changing purge_expired_interval has no effect.

@linuxmaniac - do you have any comment here?

miconda:
Closing it, now follow up for more than 3 months after expressing concerns it is not a good fix. PR can be opened when new details are added or a new PR can be made with a different fix.

PR #3255, opened 2022-09-28
speeddial: allow search uri of speed dial by group

Tomohare:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This new feature provides a way to have multiple users sharing the same speed dial number, if they belong to the same group.

A new function is add: "sd_lookup_group". It finds the URI for a specific sd_username and group_id (and sd_domain if is_domain value is defined) in the database, returning its URI.

### DIFF
--- a/src/lib/srdb1/schema/entities.xml
+++ b/src/lib/srdb1/schema/entities.xml
@@ -11,6 +11,7 @@
 <!ENTITY xcap_uri_len "255">
 <!ENTITY id_len "64">
 <!ENTITY table_id_len "10">
+<!ENTITY group_len "128">
 <!ENTITY hf_len "255">
 <!ENTITY contact_len "512">
 <!ENTITY long_hf_len "512">
--- a/src/lib/srdb1/schema/speed_dial.xml
+++ b/src/lib/srdb1/schema/speed_dial.xml
@@ -9,7 +9,7 @@
 
 <table id="speed_dial" xmlns:db="http://docbook.org/ns/docbook">
     <name>speed_dial</name>
-    <version>2</version>
+    <version>3</version>
     <type db="mysql">&MYSQL_TABLE_TYPE;</type>
     <description>
 		<db:para>
@@ -98,6 +98,14 @@
         <size>&user_len;</size>
         <default/>
         <description>Description</description>
+    </column>
+
+    <column>
+        <name>group_id</name>
+        <type>string</type>
+        <size>&group_len;</size>
+        <default/>
+        <description>Group ID</description>
     </column>
 
     <index>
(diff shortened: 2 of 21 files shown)


henningw:
Thanks for the pull requests. Maybe I missed something, but it seems that the actual implementation of the new functions is missing? if its still work in progress, no problem just let us known.

henningw:
Ok, thanks - now its more clear. There might be more review comments from people more familiar about this module.
One remark about the SQL changes - this needs to be done not in the actual SQL files, but in the XML template available in lib/srdb1/schema. You should probably also increment the table version as you are adding a new field to the table.

Tomohare:
> Ok, thanks - now its more clear. There might be more review comments from people more familiar about this module. One remark about the SQL changes - this needs to be done not in the actual SQL files, but in the XML template available in lib/srdb1/schema. You should probably also increment the table version as you are adding a new field to the table.

Yes, updated the commit and added a new one for the generated files.

miconda:
Wouldn't it work in the old version by providing `sip:group_id@domain`  as the second parameter of `sl_lookup()`, with the `username` set to `group_id` inside database table?

Tomohare:
> Wouldn't it work in the old version by providing `sip:group_id@domain` as the second parameter of `sl_lookup()`, with the `username` set to `group_id` inside database table?

Yes, I failed to communicate the idea:  that a group could have multiple domains. For example: group A has speed dial 123@domain1 (returning sip:999999@domain1) while 123@domain2 from the same group returns a different URI.

miconda:
If I am not misunderstanding something, i see no problem for such situation: the domain can also be set to a certain value `domainA`, it is not required to be the same to sd_domain

Tomohare:
> If I am not misunderstanding something, i see no problem for such situation: the domain can also be set to a certain value `domainA`, it is not required to be the same to sd_domain

Sorry for the delay, I finally understood what you meant, but as I see it, passing `sip:group_id@domain` as the second parameter could have clashes with other users, and that was my concern.

However, having the domain fixed to be used only to speed dial in the system, something like `sip:group_id@local.sd.domain`, would work making this modification useless. With this in mind I close this PR. Thanks and, again, sorry for the time wasted.

henningw:
@Tomohare thanks for the feedback!

miconda:
@Tomohare: maybe, if it suits your use case better, you can try another approach to make the query only on user column even when the `use_domain` is set, via new function `sd_lookup_group()` or a new optional parameter to `sd_lookup()`. So there will be one function to do the query on both user/domain and the other one only on user column.

The function take the db table name as a parameter, so you can use two db tables to separate the group and user-specific records (`sd_lookup_group(groupdial)` and `sd_lookup(speedial)`) . You can also set the name of the column to something that suits better both cases (e.g., `owner`).

From my point of view it felt that the module offers what you need, but it is fine to add new function/functionality for more flexibility, but I didn't want to change the database structure unless really needed, because that has impact on upgrades (even on major release upgrade, but better to avoid when possible).

PR #3261, opened 2022-10-13
mqtt: Add tls_alpn module option

agranig:
This is used when a service (like AWS IoT Core) uses one TLS port for multiple services (like https and mqtt), so you have to set the ALPN to 'mqtt' to be able to connect kamailio.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Some services use ALPN to have one single TLS port serving multiple protocols, in order to avoid additional round-trips for negotiating TLS connections. https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation has the details.

One of those services using ALPN is AWS IoT Core, which allows components to send messages both via HTTP and MQTT to AWS for further processing (e.g. logging, analysis, storage to db etc). This patch enables kamailio to communicate with this service by allowing the kamailio admin to set the mqtt.so module mod-param *tls_alpn*  to *mqtt*.

### DIFF
--- a/src/modules/mqtt/doc/mqtt_admin.xml
+++ b/src/modules/mqtt/doc/mqtt_admin.xml
@@ -328,6 +328,27 @@ modparam("mqtt", "tls_method", "tlsv1.3")
 		</programlisting>
 		</example>
 	</section>
+	<section id="mqtt.p.tls_alpn">
+		<title><varname>tls_alpn</varname> (str)</title>
+		<para>
+        Used to set the TLS ALPN option for cases where one TLS port is used to share multiple services. Prominent
+        example is AWS IoT, where you have to set the <varname>tls_alpn</varname> to "mqtt" to be able to connect
+        via MQTT.
+		</para>
+		<para>
+		<emphasis>
+			Default value is NULL, where no ALPN is set on the TLS connection.
+		</emphasis>
+		</para>
+		<example>
+		<title>Set <varname>tls_alpn</varname> parameter</title>
+		<programlisting format="linespecific">
+...
+modparam("mqtt", "tls_alpn", "mqtt")
+...
+		</programlisting>
+		</example>
+	</section>
 	<section id="mqtt.p.certificate">
 		<title><varname>certificate</varname> (str)</title>
 		<para>
--- a/src/modules/mqtt/mqtt_dispatch.c
+++ b/src/modules/mqtt/mqtt_dispatch.c
@@ -224,6 +224,17 @@ int mqtt_run_dispatcher(mqtt_dispatcher_cfg_t* cfg)
 			LM_ERR("mosquitto_tls_set() failed: %d %s\n",errno, strerror(errno));
 			return -1;
 		}
+        if (cfg->tls_alpn != NULL) {
+#if LIBMOSQUITTO_VERSION_NUMBER >= 1006000
+            res = mosquitto_string_option(_mosquitto, MOSQ_OPT_TLS_ALPN, cfg->tls_alpn);
+            if (res != MOSQ_ERR_SUCCESS) {
+                LM_ERR("mosquitto_string_option() failed setting TLS ALPN: %d %s\n",errno, strerror(errno));
+                return -1;
+            }
+#else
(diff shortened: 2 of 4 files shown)


src/modules/mqtt/mqtt_dispatch.c
```diff
@@ -224,6 +224,17 @@ int mqtt_run_dispatcher(mqtt_dispatcher_cfg_t* cfg)
 			LM_ERR("mosquitto_tls_set() failed: %d %s\n",errno, strerror(errno));
 			return -1;
 		}
+        if (cfg->tls_alpn != NULL) {
+#if LIBMOSQUITTO_VERSION_NUMBER >= 1006000
+            res = mosquitto_string_option(_mosquitto, MOSQ_OPT_TLS_ALPN, cfg->tls_alpn);
+            if (res != MOSQ_ERR_SUCCESS) {
+                LM_ERR("mosquitto_string_option() failedi setting tls alpn: %d %s\n",errno, strerror(errno));
```
oej, line 231: "failedi" ?

oej: TLS upper case

src/modules/mqtt/doc/mqtt_admin.xml
```diff
@@ -324,6 +324,27 @@ modparam("mqtt", "ca_path", "/etc/ssl/certs")
...
+	<section id="mqtt.p.tls_alpn">
+		<title><varname>tls_alpn</varname> (str)</title>
+		<para>
+        Used to set the TLS ALPN option for cases where one TLS port is used to share multiple services. Prominent
+        example is AWS IoT, where you have to set the <varname>tls_alpn</varname> to "mqtt" to be able to connect
+        via MQTT.
+		</para>
+		<para>
+		<emphasis>
+			Default value is NULL, where no ALPN is set on the TLS connection.
```
oej, line 340: You can't set NULL in a mod-adam. I would say that "Default value is empty string, which means that no TLS ALPN is used.

oej: Yes, I think we have to change the others. NULL is in the code, not in the configuration.

agranig: I'd propose to do this in a second step to not mix commits. So I'm going to amend the current one, and in a second pull request change the wording of all mod params regarding NULL

henningw: Yes, lets not mix this PR with other enhancements.

oej: Fully agree, my intention was to change the others separately.

agranig: Is there actually good practice on how to document a param which has no default value (aka is NULL)? I'm finding quite some old modules having it documented like this, see https://www.kamailio.org/docs/modules/devel/modules/auth_db.html#auth_db.p.load_credentials for example.

agranig: It's a bit of a rabbit hole. https://www.kamailio.org/docs/modules/devel/modules/dispatcher.html#dispatcher.p.db_url is one of the most famous ones I guess. Any good example how to properly document a non-set behavior, so it's consistent?

oej: That's a bad one with "NULL" in quotes like it's a settable value. Let's focus now on make this new option better, and then we'll work on improving the existing docs to something non-developers may understand :-)

agranig: Alright. So, I will keep the docs for this commit as-is with the NULL, so it's consistent with the documentation of the other params in this module. In parallel, I've prepared another PR on top of this one, which changes the docs of all module parameters at once, changing the wording from NULL to `Default value is that option is not set ([description of implication this has, e.g. random id or no auth or whatever])`. Agreed?

oej: Thank you for doing that. Agreed!

PR #3279, opened 2022-11-09
ims_usrloc_pcscf:   implementation of db_mode DB_ONLY (value 3)

petermarianF:
In order to support a redundant PCSCF configuration - i.e. a logical PCSCF consists of 2 physical nodes (node1 and node2) some enhancements were introduced for handling of contacts. Redundancy means that SIP messages for a client are normally handled by node1 but in case node1 is not reachable SIP messages are redirected to node2. Of course the DB_ONLY mode must work also for single PCSCF node configuration. Important aspects of this implementation are database integrity - i.e. avoid invalid table entries (for example data which are expired long time ago or have invalid states) - and keeping PCSCF cache in sync with database tables. A wrapper was built for method get_pcontact which tries to find the pcontact in cache and if search is not successful tries to download and insert from db location table - also some effort is added here to find the pcontact if it exists in cache. The contact expiry handler was modified to sync contact expiry in cache with db location entry and in case of real contact expiry sends PUBLISH to SCSCF to let NOTIFY finally delete the contact. An audit for older expired pcontacts was introduced which cares for getting rid of these contacts.
So db queries for this db mode are only supported for Mysql.
Some code was introduced to help registering callbacks for contacts which are inserted into cache when being downloaded from database - for example ims_qos callback as at the time of insertion the message that triggered the original callback registering is long gone.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally in local branch based on 5.5.4

### DIFF
--- a/src/modules/ims_usrloc_pcscf/udomain.c
+++ b/src/modules/ims_usrloc_pcscf/udomain.c
@@ -57,12 +57,14 @@
 #include "usrloc.h"
 #include "utime.h"
 #include "usrloc.h"
+#include "ul_callback.h"
 #include "usrloc_db.h"
 #include "../../core/parser/parse_uri.h"
 
 #include "../../lib/ims/useful_defs.h"
 #include "../../modules/presence/presence.h"
-
+extern int expires_grace;
+extern int audit_expired_pcontacts_timeout;
 extern int db_mode;
 extern int db_mode_ext;
 extern int match_contact_host_port;
@@ -406,9 +408,11 @@ int update_pcontact(struct udomain* _d, struct pcontact_info* _ci, struct pconta
 
 	//TODO: update path, etc
 
-	if (db_mode == WRITE_THROUGH && db_update_pcontact(_c) != 0) {
-		LM_ERR("Error updating record in DB");
-		return -1;
+	if ((db_mode == WRITE_THROUGH) || (db_mode == DB_ONLY)){
+        if (db_update_pcontact(_c) != 0) {
+	        LM_ERR("Error updating record in DB");
+	        return -1;
+        }
 	}
 
 	run_ul_callbacks(PCSCF_CONTACT_UPDATE, _c);
@@ -428,7 +432,7 @@ int insert_pcontact(struct udomain* _d, str* _contact, struct pcontact_info* _ci
 		run_ul_create_callbacks(*_c);
 	}
 
-	if (db_mode == WRITE_THROUGH && db_insert_pcontact(*_c) != 0) {
+	if (((db_mode == WRITE_THROUGH) || (db_mode == DB_ONLY)) && db_insert_pcontact(*_c) != 0) {
 		LM_ERR("error inserting contact into db");
 		goto error;
 	}
@@ -447,7 +451,7 @@ error:
  * @int reverse_search - reverse search for a contact in the memory
  * @return 0 if found <>0 if not
  */
-int get_pcontact(udomain_t* _d, pcontact_info_t* contact_info, struct pcontact** _c, int reverse_search) {
+int get_pcontact_from_cache(udomain_t* _d, pcontact_info_t* contact_info, struct pcontact** _c, int reverse_search) {
 	unsigned int sl, i, j, aorhash, params_len, has_rinstance=0;
 	struct pcontact* c;
 	struct sip_uri needle_uri;
@@ -468,7 +472,7 @@ int get_pcontact(udomain_t* _d, pcontact_info_t* contact_info, struct pcontact**
 		LM_DBG("Have an AOR to search for\n");
 		if (parse_uri(contact_info->aor.s, contact_info->aor.len, &needle_uri) != 0) {
 			LM_ERR("Unable to parse contact aor in get_pcontact [%.*s]\n", contact_info->aor.len, contact_info->aor.s);
-			return 0;
+			return 1;
 		}
 		LM_DBG("checking for rinstance");
 		/*check for alias - NAT */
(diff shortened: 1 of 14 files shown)


carstenbock:
Thanks for the contribution.

At first review, I did see a few topics:
- please update the XML for the documentation (in the doc/ folder) not the README directly - it would get overwritten, since it is autogenerated
- please make sure to add documentation (in the XML) for the new parameters introduced, in the README only the DB-Mode was added, not the other two new parameters
- instead of commenting code, please remove code. GIT will keep the history, no need to comment

I am no angel myself regarding documentation, but please make sure to keep up with certain standards.

christoph-v-kontron:
@carstenbock , @petermarianF 
As far as I know, Peter will issue a few additional PRs soon:
1) db_mode DB_ONLY for S-CSCF
2) Bugfixes for bugs that occur, when db_mode DB_ONLY is active

So I guess, it would be a good idea to issue a separate PR for documentation, as last PR in this series.

Sounds like a plan?

Kr,
Christoph

src/modules/ims_usrloc_pcscf/udomain.c
```diff
@@ -1077,91 +1113,408 @@ int preload_udomain(db1_con_t* _c, udomain_t* _d)
...
-	columns[14] = &path_col;
+	int i, len, querylen;
+        str aor, query_location, query_buffer, port={0,0};
+        db1_con_t* db_handle = 0; 
+        char *p_location;
+
+        if (_aor->len>0 && _aor->s){
+                LM_INFO("Querying database for P-CSCF contact [%.*s]\n", _aor->len, _aor->s);
+
+    p_location = "SELECT  domain,aor,host,port,protocol,received,received_port,received_proto,rx_session_id,reg_state,expires,socket,service_routes,public_ids,path from location WHERE aor=\"%.*s\""; // AND reg_state=%d";
```
henningw, line 1186: This is not the way you should interact with the DB API. There are actually methods to build a select query. You can find some examples e.g. in the usrloc/dlist.c or in usrloc_db.c files.
This way the module will work with all database drivers that implement the proper functions.

src/modules/ims_usrloc_pcscf/udomain.c
```diff
@@ -1077,91 +1113,408 @@ int preload_udomain(db1_con_t* _c, udomain_t* _d)
...
+                querylen = _aor->len;
+        } else {
+                LM_INFO("Querying database for P-CSCF received_host [%.*s] and received_port [%d]\n", contact_info->received_host.len, contact_info->received_host.s, contact_info->received_port);
+                p_location = "SELECT domain,aor,host,port,protocol,received,received_port,received_proto,rx_session_id,reg_state,expires,socket,service_routes,public_ids,path from location WHERE received=\"%.*s\" AND received_port=\"%.*s\"";
```
henningw, line 1190: See above

src/modules/ims_usrloc_pcscf/udomain.c
```diff
@@ -1077,91 +1113,408 @@ int preload_udomain(db1_con_t* _c, udomain_t* _d)
...
+int audit_usrloc_expired_pcontacts(udomain_t* _d) {
+db1_res_t* location_rs = NULL;
+
+pcontact_info_t *ci;
+db_row_t *row;
+int i;
+pcontact_t* c;
+time_t expired_50secs;
+
+char *p_location = "SELECT  domain,aor,host,port,protocol,received,received_port,received_proto,rx_session_id,reg_state,expires,socket,service_routes,public_ids,path from location WHERE expires < \"%.*s\"";
```
henningw, line 1418: This is not the way you should interact with the DB API. There are actually methods to build a select query. You can find some examples e.g. in the usrloc/dlist.c or usrloc_db files.
This way the module will work with all database drivers that implement the proper functions.

henningw (changes requested):
Thank you for the PR. I've did longer review today and added several comments.

Many comments are rather small, like removing some commented out code or review and adapt the log levels (which probably are set higher from development). The documentation was already mentioned from Carsten. I've added also some suggestions for naming of parameter.

There is probably one condition regarding memory allocation which needs to be fixed.

The major issue is that you use the DB API incorrect. You should not use raw queries to execute statements on the database. Instead the appropriate select and delete functions should be used. I've added some notes also with pointers to examples.

Just let me know if there are any questions.

src/modules/ims_usrloc_pcscf/ul_callback.c
```diff
@@ -98,6 +98,39 @@ void destroy_ul_callbacks_list(struct ul_callback* cb) {
...
+        /* build a new callback structure */
+        if ( types & PCSCF_CONTACT_UPDATE){
+            if (!(cbp_registrar=(struct ul_callback*)shm_malloc(sizeof( struct ul_callback)))) {
+                LM_ERR("no more share mem\n");
+                return E_OUT_OF_MEM;
+            }
+            cbp_registrar->callback = f;
+        }
+        else{
+            if (!(cbp_qos=(struct ul_callback*)shm_malloc(sizeof( struct ul_callback)))) {
```
henningw, line 125: This variable is causing a conflict with ims_qos module:
/usr/bin/ld: ims_qos_mod.o:./src/modules/ims_qos/./../ims_usrloc_pcscf/../ims_usrloc_pcscf/usrloc.h:246: multiple definition of `cbp_qos'; cdpeventprocessor.o:./src/modules/ims_qos/./../ims_usrloc_pcscf/../ims_usrloc_pcscf/usrloc.h:246: first defined here

Should be renamed

petermarianF:
Hi @henningw,
I added one new commit which should resolve all your comments.
In a few minutes I will write an email to ask for further proceeding.

petermarianF:
I resolved conflict in src/modules/ims_usrloc_pcscf/doc/ims_usrloc_pcscf_admin.xml
by editing this file and clicking "Mark as resolved"

petermarianF:
My resolve action probably did not work as it seems file does not have been saved.
But I copied the changes to my local branch and the README build was fine including Hennings last change. So probably I could add another commit with admin xml

petermarianF:
I finally did a rebase to resolve the conflict in the admin xml.
To summarize my changes after first review:
1. removed unnecessary  comments
2. changed INFO/CRIT logs into DBG/ERR logs
3. removed my changes to db_insert_pcontact and db_update_pcontact
4. removed method which deleted pua entry in pua db
5. changed definition of cpb_qos and cpb_registrar to declaration in usrloc.h and moved definition to usrloc.c
6. changed the new db queries to the correct generic approach
7. documented 2 new modparams
8. Free cpb_qos and cpb_registrar  in destroy method

src/modules/ims_usrloc_pcscf/ul_callback.c
```diff
@@ -98,6 +98,39 @@ void destroy_ul_callbacks_list(struct ul_callback* cb) {
...
+                LM_ERR("no more share mem\n");
+                return E_OUT_OF_MEM;
```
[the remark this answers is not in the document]
henningw: I made a mistake, its if/else case so its not executed both

src/modules/ims_usrloc_pcscf/udomain.h
```diff
@@ -82,5 +83,7 @@ int update_security(udomain_t* _d, security_type _t, security_t* _s, struct pcon
 int update_temp_security(udomain_t* _d, security_type _t, security_t* _s, struct pcontact* _c);
 
 int preload_udomain(db1_con_t* _c, udomain_t* _d);
-
+int audit_usrloc_expired_pcontacts(udomain_t* _d);
+int db_load_pcontact(udomain_t* _d, str *_aor, int insert_cache, struct pcontact** _c, pcontact_info_t* contact_info);
+int db_delete_presentityuri_from_pua(str *presentity_uri);
```
henningw, line 88: This probably can be removed as well, as there is no implementation.

src/modules/ims_usrloc_pcscf/usrloc_db.c
```diff
@@ -301,7 +307,7 @@ int db_insert_pcontact(struct pcontact* _c)
 	SET_PROPER_NULL_FLAG(_c->rinstance, values, LP_RINSTANCE_IDX);
 	SET_PROPER_NULL_FLAG(_c->rx_session_id, values, LP_RX_SESSION_ID_IDX);
 
-	VAL_DOUBLE(GET_FIELD_IDX(values, LP_REG_STATE_IDX)) = _c->reg_state;
+	VAL_INT(GET_FIELD_IDX(values, LP_REG_STATE_IDX)) = _c->reg_state;
```
henningw, line 310: Why you are changing this from double to int?

henningw (commented):
@petermarianF Thank you for integrating all the comments and sorry for the delay. I've added two more comments, which are probably some minor things.

src/modules/ims_usrloc_pcscf/usrloc_db.c
henningw: You are right, it seems to be an error in the module. It probably works because the database converts it automatically.

henningw:
Thank you. As there seems to be no more comment from other developers, squashed and merged it. Further changes can be done if necessary directly in the repository.

PR #3286, opened 2022-11-28
Added asynchronous offer, answer and delete rtpe functions

adityagupta-9900:
New feature 

Description:
Added functions that provide asynchronous functionality of the standard rtpe functions; Offer, Answer, Delete.

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -226,7 +226,7 @@ static int fixup_free_rtpengine_query_v(void **param, int param_no);
 
 static int parse_flags(struct ng_flags_parse *, struct sip_msg *, enum rtpe_operation *, const char *);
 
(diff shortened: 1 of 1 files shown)


miconda:
At first look, this is not about async processing, but only suspending the transaction before doing the rtpproxy control command. There is no passing of a task to some async worker, nor resuming the transaction.

Practically your code does something like:

```
t_suspend();
rtpengine_offer()/_answer()/_delete();
```

Maybe you can just to use `async_task_route()` for INVITE:

  - https://www.kamailio.org/docs/modules/stable/modules/async.html#async.f.async_task_route

And use the rtpengine functions in the target routing block. It looks like what you want to implement inside rtpengine module, but you can achieve that by using async module.

henningw:
If it can be implemented with another module, this is of course better and easier then changing the module code, which needs maintenance in the end.

miconda:
Closing this PR, because it is not a real async implementation for rtpengine functions.

If further work is done, just make a new PR with updated patches.

PR #3296, opened 2022-12-07
pdb: change log level for pdb query response time from LM_INFO to LM_DBG

mtryfoss:
- using LM_INFO cause a lot of log events on a busy server with normal log level

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
Proposal to change log level of the query response time.
If this is of interest, it should probably be monitored in some other metric and not flood the syslog.

### DIFF
--- a/src/modules/pdb/pdb.c
+++ b/src/modules/pdb/pdb.c
@@ -388,7 +388,7 @@ static int pdb_query(struct sip_msg *_msg, struct multiparam_t *_number, struct 
 		timeoutlogs=-10;
 	}
 	if (gettimeofday(&tnow, NULL) == 0) {
-		LM_INFO("got an answer in %f ms\n", ((double)(tnow.tv_usec-tstart.tv_usec+(tnow.tv_sec-tstart.tv_sec)*1000000))/1000);
+		LM_DBG("got an answer in %f ms\n", ((double)(tnow.tv_usec-tstart.tv_usec+(tnow.tv_sec-tstart.tv_sec)*1000000))/1000);
 	}
 	avp_val.n=carrierid;
 	/* set avp ! */


henningw:
Thanks for the PR. Yes, it can be probably a bit excessive on a busy server. One could of course reduce the log level from INFO to NOTICE to silence it (and others). Lets wait if there are other comments regarding this change before merging.

mtryfoss:
Yeah, I'm aware of the possibility to reduce the log level.
However, one might want to have INFO from other modules.
I feel like using DBG here is more harmonised with the verbosity of most other modules.

miconda:
It looks like we are going to end up with two used levels for log messages, ERROR and DEBUG, because, more or less based on personal needs/preferences, INFO tends to be replaced by DEBUG. What is considered useful info for someone can be too verbose for another one.

At the end INFO is intended to provide feedback about what is executed and how, not content useful for troubleshooting.

There is also the possibly to set debug level per module using debugger module, but considering its target use for debugging, it might not be something considered suitable for production.

Maybe we should try to find a common approach for the future for such changes, e.g., a module parameter is introduced to be able to specify the log level for "info" messages.

mtryfoss:
A module parameter for setting the log level of "request info" messages pr modules sound like a good idea to me.

However, in this particular case it would be much more useful to have the response time set to var which could be used in metrics, used in some other logging related to that call, or even selective logging if latency is above a certain value.

miconda:
@mtryfoss: having the value in a variable can be useful to push to specific monitoring systems, however many built monitoring systems that are based on digesting logs (e.g., ELK stack) and extracting the information from there.

mtryfoss:
Yes, of course - but you could easily log that the same way using xlog? :)

miconda:
_Yes and no_ - because it is not the same, the source of the log message is going to be the xlog, not the original the module, it requires changes to existing systems and how monitoring is handling.

Then, log levels are designed for controlling the verbosity, by setting debug core parameter. For a similar approach, it would mean to add a parameter to control when variables are set or not.

miconda:
I have just pushed a more generic way to allow remapping log levels for local purpose of the module when using LLM_XYZ() log macros instead of LM_XYZ() and updated the pdb module to offer a modparam for INFO level to be remapped. The most relevant commit:

  - https://github.com/kamailio/kamailio/commit/03203d1baa5f185db8743a90f889e5483e90e911
  - https://github.com/kamailio/kamailio/commit/e818baa1b21249529c93990ef0340dd8b56b26cf

To get the equivalent of what you wanted to change, set ll_info modparam to 3.

If there are problems, open a bug in the issue tracker.

PR #3299, opened 2022-12-09
misc/examples: add a space between comma and quote

dilyanpalauzov:


henningw:
We do not have an "official" format for that kind of syntax, but the format with the whitespace after the comma is mostly used e.g. in default configuration and also module READMEs. So it make sense to unify it. If you are looking to it, maybe you can also submit a PR for the default cfg, there are a few functions could be also unified.

PR #3300, opened 2022-12-09
misc/examples: change tls_method SSLv23 → TLSv1.2+

dilyanpalauzov:


henningw:
Thank you, this makes sense, we should not use obselete algorithms in our examples.

PR #3307, opened 2022-12-15
sanity: Prevent segfault related to contact URI check

btriller:
For star Contacts, there are no URIs that can be checked.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

If Contact URI should be checked, there is no URI if the Contact is a star Contact [1].

[1] https://www.rfc-editor.org/rfc/rfc3261#section-10.2.2


henningw:
Thanks for the PR, merged and will also backport. I also noted that the Contact URI check is also not documented in the module readme, will add it as well.

PR #3308, opened 2022-12-16
app_lua: Document that Lua versions 5.2, 5.3 and 5.4 can also be used.

dilyanpalauzov:
since commits 2e85bb541ff19e4b007 and 673dab16593d9a70b .


miconda:
Thanks!

PR #3320, opened 2023-01-02
Replace kamailio.org/wiki with kamailio.org/wikidocs

dilyanpalauzov:


miconda:
Thanks! I did the changes manually to commit per component.

PR #3323, opened 2023-01-05
topos: added Call-ID mask support

toharish:
Call ID mask with a key similar to topoh insted of swaping with new ID. Call ID masking is dome when a message is received from the upstream and unmasked before sending towords upstream.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally


henningw:
Thanks for the fixed pull request. @miconda probably want to review it. 
I just noticed two copied files from topoh regarding the call-id masking. Maybe it make sense to discuss if this could be avoided, but there might be no easy solution.

toharish:
> Thanks for the fixed pull request. @miconda probably want to review it. I just noticed two copied files from topoh regarding the call-id masking. Maybe it make sense to discuss if this could be avoided, but there might be no easy solution.

It will be hard to replace the CallID in the same data storage of other headers as CallID is used as the key. hence it will be a chicken-egg problem if we replace the callID. Hence I found it better to mask it with a key so that it can be unmasked and used as key. Unlike other headers, Call-ID will not make overhead strings and make the header bigger like via in topoh. Masking and unmasking were already available in Topoh hence it was better to use the same algorithm for masking which will be better during the maintenance process.

miconda:
The approach to mask call id in topos like is done by topoh could be a good enough compromise for some specific needs. However, I think it would be better to avoid cloning the code from one module to another.

A way for it would be to make topoh work in library mode with a new modparam, in the way that is not registering the internal callback for updating the headers when loaded. Then the functions to encode/decode strings should be made available via inter-module API. Actually the decode function is already exported, as I could spot quickly, for sipcature needs.

@toharish: I can eventually take care of making topoh run in lib mode, then you make use of it and update topos patches. Also, do not make PRs with commented code, if that was useful during the developement/testing, remove them before making the PR -- I see that the bot `github-code-scanning` scans automatically and reported above.

toharish:
> The approach to mask call id in topos like is done by topoh could be a good enough compromise for some specific needs. However, I think it would be better to avoid cloning the code from one module to another.
> 
> A way for it would be to make topoh work in library mode with a new modparam, in the way that is not registering the internal callback for updating the headers when loaded. Then the functions to encode/decode strings should be made available via inter-module API. Actually the decode function is already exported, as I could spot quickly, for sipcature needs.
> 
> @toharish: I can eventually take care of making topoh run in lib mode, then you make use of it and update topos patches. Also, do not make PRs with commented code, if that was useful during the developement/testing, remove them before making the PR -- I see that the bot `github-code-scanning` scans automatically and reported above.

@miconda thank you for your time,
 Can I close this PR and open a new PR when you have completed topos as lib mode or wait and make the necessary changes in this PR.

henningw:
@toharish You can close this PR and create a new one, or "force-push" the new code to the existing PR. Creating a new PR is usually a bit easier. Regarding the library changes for topoh, @miconda needs to comment of course.

miconda:
@toharish: just added the code to topoh module to offer the mask call id function, you have to set its use_mode to 1, then bind from topos to its inter-module api (probably an example is in sipcapture, if not see how register module loads the sl module api).

PR #3334, opened 2023-01-21
Topos: Added Call-ID mask

toharish:
Added Call-ID mask for Topos using API call of Topoh
Call ID mask using topoh API call instead of swapping with new ID. Call ID masking is done when a message is received from the upstream and unmasked before sending it to the upstream.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally


src/modules/topos/topos_mod.c
```diff
@@ -490,10 +503,9 @@ int tps_msg_received(sr_event_param_t *evp)
 			}
 		}
 		dialog = (get_to(&msg)->tag_value.len>0)?1:0;
-		if(dialog) {
-			/* dialog request */
-			tps_request_received(&msg, dialog);
-		}
+		
+		tps_request_received(&msg, dialog);
+
```
miconda, line 508: Is the above change necessary? Should tps_request_received() be executed for initial (non-in-dialog) requests?

miconda: During the kamailio config processing, the call-id must be the original value, not masked. The mask has to be done when the message is sent out, unmask has to be done when the message is received. You can compare with what's done by topoh for call-id masking/unmasking.

toharish: To mask the Call-ID after the processing will be a challenge as TOPOS is using Call-ID as the key and further it is identifying from the request/response as UPSTREAM / DOWNSTREAM after tps_storage_load_dialog function call.
which means when there is a message from Downstream we will not be able to UnMask first before loading the dialog from the database as we are not knowing the direction.
To solve this we will have to add additional cookies and headers similar to Topoh which will be too much additional processing instead this approach of masking when receiving from Upstream and hiding other headers before sending downstream and Unmasking before sending to Upstream and hiding other headers when received from downstream will be better.

miconda: The config has to process the SIP message as it would be without topos (or topoh), otherwise things become more complex to track (accounting, troubleshooting, sipdump, ...). Masking the call id comes with a prefix:

  - https://www.kamailio.org/docs/modules/stable/modules/topoh.html#topoh.p.callid_prefix

If the prefix is not matched, unmasking is not done, see th_unmask_callid() from topoh.

At the end, even it has to be more complexity added to topos module, it is the way to go, instead of affecting everything else.

PR #3341, opened 2023-01-25
topoh: uses socket IP when no mask_ip is defined

TorPetterson:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

topoh: uses socket IP when no mask_ip is defined

If the parameter mask_ip is not defined the module finds the socket IP
and uses that as mask IP for the message.
If the socket has an advertised IP it is used, otherwise the socket IP is used.


miconda:
The patch seems to be pretty consistent in size, it will need some time to be reviewed.

src/modules/topoh/doc/topoh_admin.xml
```diff
@@ -92,11 +92,13 @@ modparam("topoh", "mask_key", "some secret here")
 		SIP URIs. Can be any IP address, even a private-space or non-existing
 		IP address (e.g., 192.168.1.1, 127.0.0.2), including the SIP server
 		address, but must not be an address potentially used by clients.
+		If not set, the advertised IP of the, incoming or outgoing, socket is used.
```
dilyanpalauzov, line 95: The comma is not properly used.

TorPetterson: Is the new version better?

dilyanpalauzov: Yes

TorPetterson:
I think I resolved the conflicts but the notice is still there?

src/modules/topoh/topoh_mod.c
```diff
@@ -179,82 +192,228 @@ static int mod_init(void)
...
+			goto error;
+		}
+	} else {
+		th_socket_hash_table = pkg_malloc(sizeof(struct str_hash_table));
+		if(th_socket_hash_table == NULL){
+			PKG_MEM_ERROR_FMT("th_socket_hash_table\n");
+			goto error;
+		}
+		if(str_hash_alloc(th_socket_hash_table, TH_HT_SIZE))
+			goto error;
```
henningw, line 231: We probably should free the th_socket_hash_table in error block.

src/modules/topoh/topoh_mod.c
```diff
@@ -179,82 +192,228 @@ static int mod_init(void)
...
+
+		str_hash_init(th_socket_hash_table);
+		if(th_parse_socket_list(*get_sock_info_list(PROTO_UDP)) != 0 ||
+		   th_parse_socket_list(*get_sock_info_list(PROTO_TCP)) != 0 ||
+		   th_parse_socket_list(*get_sock_info_list(PROTO_TLS)) != 0 ||
+		   th_parse_socket_list(*get_sock_info_list(PROTO_SCTP)) !=0)
+			goto error;
```
henningw, line 238: We should probably free the th_socket_hash_table and the str_hash_alloc list in error block.

henningw (commented):
The stale flag got now removed from the bot. I think this is a useful addition, even if it changes the default behaviour of the module. For people want the previous behavior back, they can easily configure 127.0.0.1 manually. I did a quick review, just added two comments regarding memory management in mod_init. Other developers want probably also review it, as its a larger patch.

TorPetterson:
Is there anything else I can to to move this pull request forward?

TorPetterson:
Thanks!

I have no problem with the config change.

PR #3346, opened 2023-01-26
nathelper: ping_nated_only=1 sends SIP-request-based pings

dilyanpalauzov:
My experiments show that with
```
modparam("nathelper", "ping_nated_only", 1)
modparam("nathelper", "sipping_bflag", FLB_NATSIPPING)
modparam("usrloc", "nat_bflag", FLB_NATB)
```
when both FLB_NATB and FLB_NATSIPPING branch flags are set, no OPTIONS from Kamailio to UAC is generated, so I assume that nathelper.ping_nated_only=1 does force 4-bytes UDP keepalive.

This change spells the above, as the behaviour currently is not specified.

### DIFF
--- a/src/modules/nathelper/doc/nathelper_admin.xml
+++ b/src/modules/nathelper/doc/nathelper_admin.xml
@@ -27,12 +27,12 @@
 		header field with request's source address:port pair. The function
 		<function>fix_nated_sdp()</function> adds the active direction indication
 		to &sdp; (flag 0x01) and updates the source &ip; address too (flag 0x02). The function
-		<function>fix_nated_register()</function> exports exports the request's source
+		<function>fix_nated_register()</function> exports the request's source
 		address:port into an AVP to be used during <function>save()</function> and should
 		be used for <quote>REGISTER</quote> requests.
 	</para>
         <para>
-		Note: <function>fix_nated_contact</function> changes the <quote>Contact</quote>
+		Note: <function>fix_nated_contact()</function> changes the <quote>Contact</quote>
 		header, thus it breaks the RFC. Although usually this is not an issue, it may
 		cause problems with strict SIP clients.  An alternative is to use
 		<function>add_contact_alias()</function> that together with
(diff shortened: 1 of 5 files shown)


henningw:
Regarding the uac nat ping method, I think its documented here:
https://kamailio.org/docs/modules/5.5.x/modules/nathelper.html#nathelper.p.ping_nated_only

dilyanpalauzov:
The documentation says for nathelper.ping_nated_only=1 that only contacts that have the behind NAT “nat_bflag” flag in user location database set will get ping.

The current wording does not specify, if the latter ping is 4-byted UPD-packet ping, or SIP-request-based ping.

Yesterday when I submitted the change I was behing SIP ALG.  Today, without SIP ALG, testing again, I came to the opposite conclusion compared to yesterday:

nathelper.ping_nated_only=1 does send SIP-request-based pings.  It does not send 4-byte UDP-packet pings.

dilyanpalauzov:
To be precise, the current wording:
>  If this parameter is set to 1 then only contacts that have the behind NAT “nat_bflag” flag in user location database set will get ping.
> 
> If it is 0 and sipping_bflag is not set, then the 4-bytes UDP ping is sent to all contacts. If it is 0 and sipping_bflag parameter is set, then SIP-request-based pinging is sent to all contacts. 

for me implies, that  sipping_bflag is relevant when ping_nated_only=1, and irrelevant when ping_nated_only=0.

dilyanpalauzov:
> for me implies, that sipping_bflag is relevant when ping_nated_only=1, and irrelevant when ping_nated_only=0.
READ:

for me implies, that sipping_bflag is relevant when ping_nated_only=0, and irrelevant when ping_nated_only=1.

miconda:
Merging because of the typos, then I will try to adjust the section for pinging.

miconda:
Updated docs, it is supposed to send OPTIONS requests when `nat_bflag`+`sipping_bflag` branch flags are both set for the case when `ping_nated_only=1`.

PR #3347, opened 2023-01-26
Topos: Added Call-ID mask support for Topos

toharish:
Added Call-ID mask Support for Topos with API call from Topoh to mask the callID
Call-ID mask happens before the request is sent downstream and unmasked when received from downstream. 

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally


miconda:
The lump operations do not change the Call-Id immediately, I have to think about what could be a good solution, eventually applying changes as an intermediary step or do a custom in-place replacement of Call-Id.

toharish:
> The lump operations do not change the Call-Id immediately, I have to think about what could be a good solution, eventually applying changes as an intermediary step or do a custom in-place replacement of Call-Id.

Thankyou, However The same methord of lump operation is ued in Topoh and this code was working in our VoIP network when locally tested.

miconda:
Topoh does not use call-id value for finding records, like it is done by topos. I will have to go deeper in the code to review properly and be sure it works as expected. There are some tricks that a partial change can be discovered, but I am not sure that it applies to call-id.

toharish:
> Topoh does not use call-id value for finding records, like it is done by topos. I will have to go deeper in the code to review properly and be sure it works as expected. There are some tricks that a partial change can be discovered, but I am not sure that it applies to call-id.

when lump is removed from the momory i am getting error in releasing the memory as below. Is the allocation and release of memory release of memory for the leg is related to call-ID. If the Call-ID mask is removed this error is not available.  The error is as below:

9(3259925) CRITICAL: {1 20 INVITE 1307026397@xx.xxx.xx.xx} <core> [core/mem/q_malloc.c:501]: qm_free(): BUG: bad pointer 0x7faec4a78500 (out of memory block!) called from core: core/data_lump.c: free_lump(470) - ignoring

miconda:
I merged it, going to update some parts. Thanks!

PR #3349, opened 2023-01-27
websocket: typos

dilyanpalauzov:
I run Kamailio with
```
log_stderror=yes
/* LOG Levels: 3=DBG, 2=INFO, 1=NOTICE, 0=WARN, -1=ERR, ... */
debug=-1
```
and inspect the output.  Some messages from the websocket module are logged repeatedly at ERROR-level, but they are in no way errors.  This changes moves such periodially logged no-error messages to debug-level INFO.

### DIFF
--- a/src/modules/websocket/doc/websocket_admin.xml
+++ b/src/modules/websocket/doc/websocket_admin.xml
@@ -488,7 +488,7 @@ end
 		</para>
 		<para><emphasis>Default value is 1.</emphasis></para>
 		<example>
-		<title>Set <varname>timer_interval</varname>parameter</title>
+		<title>Set <varname>timer_interval</varname> parameter</title>
 		<programlisting format="linespecific">
 ...
 modparam("websocket", "timer_interval", 5)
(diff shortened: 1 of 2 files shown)


lazedo:
if you're changing this maybe set to `L_DEBUG` instead ?  or add a mod param for this ?

dilyanpalauzov:
I have changed the logging level to DEBUG (`LM_DBG()`).

henningw:
Thanks for the PR. It probably needs to be applied manually, as it also contains some (unrelated) documentation changes.

miconda:
This falls again in the debate of how far we go changing levels for log messages, because the importance of them is relative from person to person.

The designed way  was that if someone considers the logging too excesive for a particular deployment, then the `debug` core parameter has to be decreased.

Similar to #3296.

On the changes themselves, the messages were about errors for sending messages or keepalives, not when sending was successful, they should not be periodical logs if the connections are ok.

I also see an INFO changed to DEBUG when the connection was unexpectedly lost.

dilyanpalauzov:
> I also see an INFO changed to DEBUG when the connection was unexpectedly lost.

This change was not intended, I have reverted it.

I agree the remaining changes are related to error conditions, not just to basic logging.  I have removed them, leaving in the changeset only typos.

PR #3352, opened 2023-01-28
rr: doc — replace enable_username with add_username

dilyanpalauzov:
The documentation of rr.custom_user_avp speaks about option `enable_username`, but there is no such option.

I assume the documentation means `add_username`.


miconda:
Thanks, split and pushed manually.

PR #3353, opened 2023-02-01
doc/tutorials/dns: USE_NAPTR is on by default

dilyanpalauzov:
My reading of the original text:

>     USE_NAPTR - if defined the naptr lookup support will be compiled in.
>      NAPTR support still has to be enabled from Kamailio's config file (it's
>      off by default).

was that USE_NAPTR was by default off (undefined).  This change clarifies that by default USE_NAPTR is defined, but naptr support is by default disabled from the cofiguration file.


henningw:
Thank you

PR #3356, opened 2023-02-03
core: include DBG_SYS_MEMORY and MEM_JOIN_FREE on flags: output

dilyanpalauzov:
E.g. when calling `core.info` RPC.


dilyanpalauzov:
Include `, DBG_SYS_MEMORY` in flags: output, too.

dilyanpalauzov:
> Include `, DBG_SYS_MEMORY` in flags: output, too.

and MEM_JOIN_FREE

Issue #3210, opened 2022-08-08
sanity: do not generate waring when used UDP protocol and no "Content-Length" header


sergey-safarov:
### Description
According to [RFC3261](https://datatracker.ietf.org/doc/html/rfc3261#section-20.14)

>    The Content-Length header field indicates the size of the message-
>    body, in decimal number of octets, sent to the recipient.
>    Applications SHOULD use this field to indicate the size of the
>    message-body to be transferred, regardless of the media type of the
>    entity.  If a stream-based protocol (such as TCP) is used as
>    transport, the header field MUST be used. 

When UDP protocol is used then this header is optional.
When Kamailio receives an OPTIONS request like in the example, then it floods Kamailio logs with messages like
```
sanity [sanity.c:612]: check_cl(): content length header missing in request
```

OPTIONS message example
```
OPTIONS sip:sbc-0.example.com:5060 SIP/2.0
Via: SIP/2.0/UDP 64.58.61.151:5060;branch=z9hG4bKl7oo3f30a0som61aeu00
Call-ID: 49bc1fcac14b779958dad4b9c43954b400149n2@64.58.61.151
To: sip:ping@sbc-0.example.com
From: <sip:ping@64.58.61.151>;tag=8fbe5af9178677655bc3a5388c2a8b8c00149n2
Max-Forwards: 70
CSeq: 299694 OPTIONS
Route: <sip:3.236.25.4:5060;lr>
```

### Expected behavior
Do not generate warnings about SIP messages that are allowed according to RFC.

#### Actual observed behavior
Generated warning when "Content-Length" header missing and used UDP protocol.

henningw:
You can switch off this check by using the appropriate flags: [link](https://kamailio.org/docs/modules/5.5.x/modules/sanity.html#sanity.overview)

Its probably a good idea to warn about the missing header when the user activated this check. But it could be of course changed as part of this feature request.

miconda:
From my point of view, having a condition in the config file is fine, considering that most UAs add Content-Lenght -- it is up to config write to adjust the options/parameters for own eco-system. In many cases lack of Content-Length happens more often due to broken UA processing than UAs not adding it on purpose.

If one wants to add a new flag (or a modparam) to skip this check for UDP (and probably SCTP), I am fine. But if nobody commits soon to do it, then should be closed. A PR can be made at any time.

sergey-safarov:
I will try to create PR.
If it is not done next two months then I will close the request.

PR #3359, opened 2023-02-05
No warning generated when UDP and no Content-Length header

sergey-safarov:
GH #

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [ ] Related to issue #3210

#### Description

Added 'always_log_missed_content_length' mode for sanity module.
If 'always_log_missed_content_length' is 1 then the SIP message
with no 'Content-Length' header will cause warning message in
the logs. Otherwise warning message is not generated for cases
when SIP message with no 'Content-Length' header is sent over UDP.

### DIFF
--- a/src/modules/sanity/doc/sanity_admin.xml
+++ b/src/modules/sanity/doc/sanity_admin.xml
@@ -252,6 +252,27 @@ modparam("sanity", "noreply", 1)
 	    </programlisting>
 	</example>
 	</section>
+
+	<section id="sanity.p.always_log_missed_content_length">
+	<title><varname>always_log_missed_content_length</varname> (integer)</title>
+	<para>
+		If the value is 1 then the SIP message with no 'Content-Length' header
+		will cause warning message in the logs. If the value is 0 then the warning
+		message is not generated for cases when SIP message with no 'Content-Length'
+		header is sent over UDP or SCTP.
+	</para>
+	<para>
+		Default value is 1.
+	</para>
+	<example>
+	    <title>Set <varname>always_log_missed_content_length</varname> parameter</title>
+	    <programlisting>
+...
+modparam("sanity", "always_log_missed_content_length", 0)
+...
+	    </programlisting>
+	</example>
+	</section>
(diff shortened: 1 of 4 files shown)


miconda:
Content-Length is not mandatory for UDP, the sip message is required to be in a single datagram.

Even that one can check with `is_present_hf("Content-Length")` and write a log message from config, I am fine to add an option to log from the module, but I would prefer to be a more generic approach, in order to reuse the same mechanism for similar needs. I would go with a new parameter like `log_filter` and then use flags (bits) to control what is logged or not (e.g., bit 1 set, the log missing content length).

As an common recommendation, parameter names should be kept short, ideally one-two words, not to have a very long snake-case name. The name should be suggestive (related to its purpose), but `not_the_full_description_of_the_parameter`.

miconda:
And do not forget the commit message format, check the CONTRIBUTING guide for it.

dilyanpalauzov:
This should apply not only to UDP and SCTP, but also to WebSockets: https://www.rfc-editor.org/rfc/rfc7118.html#section-5 

> Because the WebSocket transport preserves message boundaries, the use of the Content-Length header in SIP messages is not necessary when they are transported using the WebSocket subprotocol.

> This simplifies the parsing of SIP messages for both clients and servers.  There is no need to establish message boundaries using Content-Length headers between messages.  Other SIP transports, such as UDP and the Stream Control Transmission Protocol (SCTP) [[RFC4168](https://www.rfc-editor.org/rfc/rfc4168)], also provide this benefit.

henningw:
Thanks for the PR. Any plans to integrate the remarks made from Daniel some weeks ago?

sergey-safarov:
yes, @henningw we will make these changes.

miconda:
Closing this one for aging, just reopen or make a new PR when ready.

dilyanpalauzov:
Implementing this is a really good idea and I am looking forward to see progress on this improvement.

PR #3369, opened 2023-02-12
rtpengine: document options SDES-pad and SDES-nonew

dilyanpalauzov:


henningw:
Thanks for the PR. It needs probably applied manually, as it fixes also additional spelling errors in the docs and C code.

miconda:
Thanks, split and applied manually.

PR #3382, opened 2023-02-27
doc/htable: fix documentation for rpc commands

igorolhovskiy:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Commands for get/set described in documentation are not working giving 500 errors. If attach `s:` to key name it all work as expected (as was advised in [the mailing list](https://www.mail-archive.com/sr-users@lists.kamailio.org/msg16436.html) by Fred Posner)


miconda:
The prefix `s:` is needed only if your key is a number, because kamcmd does auto convertion of numbers to integer and the key has to be string. If the key is string, then `s:` is not needed and all examples were correct. Probably it is better that you add examples where the key is a number and it is prefixed `s:`, the existing examples with string keys should stay the same.

igorolhovskiy:
Make sense, will do!

PR #3387, opened 2023-03-03
utils/kamctl: allow the definition of a specific startup file at star…

bundasmanu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
Added the possibility to specify an alternative file different from the default: `kamailio.cfg`, when starting `kamailio`, using the `kamctl` tool.
This makes it possible to start `kamailio` by passing a startup file located in a specific folder (via `ETCDIR` env variable) and with a specific name, instead of using the default name: `kamailio.cfg`.

Several local tests have been performed and it has been verified that it is now possible to perform the basic start/stop/restart operations of `kamailio` via `kamctl`, considering the use of a startup file other than the default: `kamailio.cfg`.

Example of `.kamctlrc` used:

```sh
ETCDIR="/etc/my-component/confs"
RPCFIFOPATH="/run/my-component/my-component_rpc.fifo"
PID_FILE="/run/my-component/my-component.pid"
STARTUP_CONFIG_FILE="starter.cfg"
STARTOPTIONS="-m 512 -M 64 -u user -g group"
```

Tested on `CentOS Linux release 7.9.2009 (Core)`.

I think this change will be interesting for many users.

Thanks

### DIFF
--- a/utils/kamctl/kamctl
+++ b/utils/kamctl/kamctl
@@ -73,6 +73,10 @@ if [ -z "$MYLIBDIR" ] ; then
 	fi
 fi
 
+if [ -z "$STARTUP_CONFIG_FILE" ]; then
+	STARTUP_CONFIG_FILE="kamailio.cfg"
+fi
+
 ##### ------------------------------------------------ #####
 ### load base functions
 #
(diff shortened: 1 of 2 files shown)


dilyanpalauzov:
If this can be managed over an environment variable, then the latter shall include all the directories.  As far as I see with this change, the file must still be in /etc (or $ETCDIR, but I assume both are almost thr same).

bundasmanu:
No, with this change we can use any directory and any startup config filename to manage `kamailio` via `kamctl`. `ETCDIR` in the example provided uses `/etc`, but we can use any directory to accomplish that.

miconda:
Thanks!

PR #3390, opened 2023-03-09
auth_ephemeral: fix authentication with secrets added at runtime

Den4t:
- proper way to work with shm pointer to secret list

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
In case of adding secrets at runtime with autheph.add_secret rpc, authentication with new secrets not work,
this is because rpc process modify local pointer to secret list, instead shm.

### DIFF
--- a/src/modules/auth_ephemeral/auth_ephemeral_mod.c
+++ b/src/modules/auth_ephemeral/auth_ephemeral_mod.c
@@ -236,6 +236,7 @@ static inline int add_secret(str _secret_key)
 		if (secret_list == NULL)
 		{
 			LM_ERR("unable to allocate shared memory\n");
+			shm_free(secret_struct);
 			return -1;
 		}
 		*secret_list = NULL;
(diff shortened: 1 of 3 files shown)


src/modules/auth_ephemeral/auth_ephemeral_mod.c
```diff
@@ -226,12 +230,22 @@ static inline int add_secret(str _secret_key)
...
 	secret_struct->secret_key = _secret_key;
 	SECRET_LOCK;
-	if (secret_list != NULL)
+	if (secret_list == NULL)
+	{
+		secret_list = (struct secret **) shm_malloc(sizeof(struct secret *));
+		if (secret_list == NULL)
+		{
+			LM_ERR("unable to allocate shared memory\n");
+			return -1;
```
henningw, line 239: You probably want to free the previously allocated secret_struct

henningw (commented):
Thanks for the PR. I just added one remark related to the freeing of the previous allocated memory in an error condition. Maybe others like to comment as well, as I do not use this module a lot.

src/modules/auth_ephemeral/auth_ephemeral_mod.c
Den4t: Fixed.

miconda:
Thanks!

Issue #3372, opened 2023-02-13
Statistic counters for LCR module


Ozzyboshi:
### Description

I am in need to get some statistic counters related to each lcr gw currently loaded in Kamailio.
Kamailio already provides the counters I need inside the KEX module with the following command:
```
kamcmd> stats.get_statistics core:
core:bad_URIs_rcvd = 69
core:bad_msg_hdr = 2
core:drop_replies = 0
core:drop_requests = 8499
core:err_replies = 0
core:err_requests = 0
core:fwd_replies = 4
core:fwd_requests = 65
core:rcv_replies = 502088
...
```
Here kamailio doing exactly what i want, it's exposing the counter data via xmlrpc which is great for attaching an external tool to build statistics.
Unfortunately Kamailio KEX module is counting every single request/reply hitting the proxy server, the idea is to have the above stats related to the gateways currently loaded by the lcr module.
In this way, for example I can tell how many request/replies I got from gw id X or gw id Y and produce gw based stats.
Unfortunately the KEX module has no "concept" of "GW" and "LCR", so the idea is to implement the same counting logic inside LCR module.
The counting should not be persistent (counters must go to zero every time kamailio restarts).
When adding or deleting a new gateway, the counting in memory must reflect the status of the internal state or lcr gatewas, so, for example, after adding a new gateway and reloading lcr tables, the new counters must be immediately available through xmlrpc without a new kamailio restart. The same concept applies when the gateway is removed from the list.

### Expected behavior
A possible implementation should give the user to query counters in a way similar to this:
```
# kamcmd lcr.stats
{
	gw: {
		{
			gw_id: <id of lcr mysql table here>
			gw_name: <description of lcr mysql table here>
			requests: 0
			requests_invite: 0
			requests_cancel: 0
			...
		}
		{
			gw_id: <id of lcr mysql table here>
			gw_name: <description of lcr mysql table here>
			requests: 0
			requests_invite: 0
			requests_cancel: 0
			...
		}
	}
}
```

Additionally, it would be helpful to set a gw id as argument in order to filter, for example:

```
# kamcmd lcr.stats 95
{
	gw: {
		{
			gw_id: 95
			gw_name: <description of gw id 95>
			requests: 222
			requests_invite: 1
			requests_cancel: 1
			requests_ack: 1
			...
		}
	}
}
```

#### Actual observed behavior

To my knowledge, nothing related to this issue has been implemented in Kamailio so far

oej:
[...]

PR #3391, opened 2023-03-10
lcr: add counters per gateway

Ozzyboshi:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3372

#### Description
Implementing the suggesting lcr counters per gateway

### DIFF
--- a/src/modules/lcr/doc/lcr_admin.xml
+++ b/src/modules/lcr/doc/lcr_admin.xml
@@ -1258,6 +1258,30 @@ modparam("lcr", "ping_socket", "192.98.102.10:5060")
 		</example>
 	</section>
 
+	<section id="lcr.p.lcr_stats">
+		<title><varname>lcr_stats</varname> (int)</title>
+		<para>
+		Enables call counters per gateway.
+		If the value is set to 1, the call counting gets enabled.
+		A gateway matching is only done using the ip or/and port comparison,
+		for this reason the call counters don't get incremented for a gateway, which misses this data.
+		</para>
+		<para>
+		<emphasis>
+			Default value is
+			<quote>0</quote>.
+		</emphasis>
+		</para>
+		<example>
+		<title>Setting <varname>lcr_stats</varname> module parameter</title>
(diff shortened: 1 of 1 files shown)


miconda:
The patch seems substantial in size, probably @juha-h, as main author of lcr module, should look over it before merging.

juha-h:
I did some reading of the code and looks like it does not affect current behavior nor have any performance impact if the the stats are off.

But if I understood correctly, in the PR, matching of a SIP request/reply to an lcr gw is done based on IP address/port (`fetch_gw_id` function).  On my LCR module web page, I have:
```
LCR Instance 'default' Gateway 'test'

Inbound Properties
Common Name or IP Address | Transport

Outbound Properties
IP Address or Domain Name | Port | Transport
```
That is, it is possible that LCR  module does not have any knowledge of the IP address/port of a gateway.

The only way I can find if a request comes from a gateway specified by Common Name is in the script like this:
```
        $var(gw_in_id) = $sht(htable=>cn::$tls_peer_subject_cn::gw_id);
```
That is, looking a common name match from htable.

So unless I didn't read the code correctly, I don't consider it a good idea to have a stats feature that applies only to some gateways.

Ozzyboshi:
Yes,counters will be only incremented if ip and port are maching as you said.
This is because ip and port is how I usually store in the lcr table and my commit is just based on my needs.
Maybe someone in the future can improve my commit to cover all cases.
For now I just added a notice on the docs describing this behaviour.

juha-h:
Alessio Garzi writes:

> Yes,counters will be only incremented if ip and port are maching as you said.
> This is because ip and port is how I usually store in the lcr table and my commit is just based on my needs.
> Maybe someone in the future can improve my commit to cover all cases.
> For now I just added a notice on the docs describing this behaviour.

As I wrote, I don't consider it a good idea to include a feature that
only works in some situations.

Is there some reason why you can't implement the same in the script
without modifying lcr module?

PR #3403, opened 2023-03-23
tm: memory leak in case dropping messages in local-request event route

sem32:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
I faced an issue with memory leak in case drop NOTIFY messages in the "local-request" event route in `mem_copy_subs` and `build_uac_req`

analysis of mem leak:
kamcmd corex.shm_summary
<code>...
Mar 23 12:00:28 pbx kamailio[72308]: NOTICE: fm_status: fm_sums():  count=    11 size=      7520 bytes from presence: hash.c: mem_copy_subs(141)<br>             
Mar 23 12:00:28 pbx kamailio[72308]: NOTICE: fm_status: fm_sums():  count=    23 size=     22536 bytes from tm: t_msgbuilder.c: build_uac_req(1618)
...</code>

a lot of dropped messages in local-request event route

<code>...
Mar 23 12:01:09 pbx kamailio[72308]: NOTICE: fm_status: fm_sums():  count=    83 size=     51256 bytes from presence: hash.c: mem_copy_subs(141)
Mar 23 12:01:09 pbx kamailio[72308]: NOTICE: fm_status: fm_sums():  count=    98 size=    116864 bytes from tm: t_msgbuilder.c: build_uac_req(1618)
...</code>

no messages, no activities

<code>...
Mar 23 12:02:34 pbx kamailio[72308]: NOTICE: fm_status: fm_sums():  count=    83 size=     51256 bytes from presence: hash.c: mem_copy_subs(141)
Mar 23 12:02:34 pbx kamailio[72308]: NOTICE: fm_status: fm_sums():  count=    96 size=    115432 bytes from tm: t_msgbuilder.c: build_uac_req(1618)
...</code>

In the code, we have a comment like "never free cbp here because if t_uac_prepare fails, cbp is not freed and thus caller has no chance to discover if it is freed or not", but we'll free the cbp in two cases:
1. in case of error, but we have if (ret == E_DROP) then ret = 0; (so, no error result)
2. in case call insert_tmcb for cbp, but we don't do it for E_DROP

### DIFF
--- a/src/modules/tm/uac.c
+++ b/src/modules/tm/uac.c
@@ -540,6 +540,7 @@ static inline int t_uac_prepare(uac_req_t *uac_r,
 	if (unlikely(goto_on_local_req>=0 || tm_event_callback.len>0)) {
 		refresh_shortcuts = t_run_local_req(&buf, &buf_len, uac_r, new_cell, request);
 		if (unlikely(refresh_shortcuts==E_DROP)) {
+			shm_free(buf);
 			ret=E_DROP;
 			goto error1;
 		}
@@ -653,6 +654,10 @@ int prepare_req_within(uac_req_t *uac_r,
 	
 	if (unlikely(ret < 0 && ret == E_DROP)) {
 		ret = 0;
+		if(uac_r->cbp) {
+			shm_free(uac_r->cbp);
+		}
+
 	}
 
  err:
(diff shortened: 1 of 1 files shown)


sem32:
@grumvalski , probably you could review my changes.

henningw:
@sem32 Any feedback to the earlier comment from Daniel? There is now also one conflict, would be great if you could rebase it.

PR #3404, opened 2023-03-23
core: add socket timestamp of received packets and keep track of receive delay

nchaigne:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description

This is a proposed evolution that adds socket timestamping of received UDP packets (using SO_TIMESTAMPNS).
It computes the delay between this timestamp and the time at which the packet is read by Kamailio.
This can then be accessed using pv "$TV(RD)" (RD for "receive relay").

Rationale :
The Kamailio server is handling traffic that is irregular and can temporarily spike to a point where all children are still busy when new packets are incoming on the network interface.
Consequently, these packets have to wait until a worker is available. The client will see a higher response time, that currently cannot be monitored by Kamailio itself.
This evolution allows to make this wait time observable by Kamailio, allowing to log or write the information to CDR for example.

If this is possible, I'd like to merge this feature in Kamailio.
Of course, I'm ready to discuss the PR and make any necessary changes.
All comments and reviews are most welcome.

I've tested this locally on Linux (3.10.0-957.10.1.el7.x86_64) with the latest Kamailio master as of 2023/03/23.

Thanks!

### DIFF
--- a/src/core/ip_addr.h
+++ b/src/core/ip_addr.h
@@ -174,6 +174,7 @@ typedef struct receive_info {
 	struct socket_info* bind_address; /* sock_info structure on which
 										* the msg was received */
 	recv_flags_t rflags; /* flags */
+	struct timespec ts; /* network timestamp of incoming packet (using CLOCK_REALTIME) */
 	char proto;
 #ifdef USE_COMP
 	char proto_pad0;  /* padding field */
--- a/src/core/parser/msg_parser.h
+++ b/src/core/parser/msg_parser.h
@@ -296,6 +296,7 @@ typedef struct sip_msg {
 	unsigned int id;               /*!< message id, unique/process*/
 	int pid;                       /*!< process id */
 	struct timeval tval;           /*!< time value associated to message */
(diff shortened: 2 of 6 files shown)


oej:
Interesting. Have you tested this on TCP/TLS too?

nchaigne:
> Interesting. Have you tested this on TCP/TLS too?

No, this only works for UDP.
I didn't try to implement it for TCP, not sure how to do it. I looked a bit at the code but it seems more complicated...

miconda:
I can't have an opinion right away for merging, the changes are on a critical part of the core code and I am not sure it is portable for bsd/macos/*unix. I think it needs further analysis, it is also only for udp.

nchaigne:
> @nchaigne are you interested in working further on this? Right now its also in a conflict state due to other changes in the last weeks in the core.

Hello Henning,
Unfortunately I don't think I will be able to continue working on this.

As Daniel pointed out, it is currently too limited in scope (only for udp, maybe not portable).
So it's probably better to close the PR.

henningw:
Thanks for the feedback, closing it then.

PR #3407, opened 2023-03-29
userloc: change data type of id field to BIGINT

juliusf:
As discussed on the mailinglist [1]  the `id` field of the location database in the `userloc` module has the data type `INT(10)`. This can be too small for large long running installations of kamailio. This changes the data type of the `id` field from `INT(10)` to `BIGINT`

[1]: https://lists.kamailio.org/mailman3/hyperkitty/list/sr-dev@lists.kamailio.org/thread/TKSUP4ZRLXAQ6P5ZNP2KCQBBJPFYVHCF/


miconda:
The database structure should be defines via xml file in `src/lib/srdb1/schema/`, but I guess that is not done for p_usrloc.

While I am not the maintainer of the p_usrloc module, so I won't going to make the decision for it, maybe it would be good to have an uniform approach for all tables.

henningw:
@miconda You are right, the p_usrloc DB scheme should be also moved to the src/lib/srdb/schema (similar to others IMS modules). It already got out of sync with the main usrloc module, some colums were not increased in the last years.

The p_usrloc is a special module, and most of its users are using their own scheme anyway. So I do not see a big value of adapting the ID column in this scheme. 

As mentioned already in the list discussion, the id column is not used from Kamailio, so you can just adapt it as you wish. My recommendation would be to to close this PR.

henningw:
Close this one - I will make a dedicated commit moving the p_usrloc.sql to the XML schema directory.

PR #3430, opened 2023-04-28
registrar: adding tcpconn_id to xavp_cfg

jchavanton:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description

This is adding the variable `tcpconn_id` to `xavp_cfg`

When using Kamailio as a "mid registrar", we need to save the contact in the reply-route (once we have the confirmation from the registrar), one problem I faced was that the tcpconid was the one of the response message coming from the registrar.
I needed a way to save the one of the original request message.

## mid registrar config example where we need/use the tcpconnid

```
modparam("registrar", "xavp_cfg", "reg")

route[REGISTER] {
    ...
    t_on_reply("ON_REPLY_REGISTER");
    $xavp(reg=>tcpconn_id) = $conid; // set the TCP connection id to be saved on the reply
   ....
}

onreply_route[ON_REPLY_REGISTER] {
        if (t_check_status("200")) {
                if (is_present_hf("Expires") && $hdr(Expires) == "0") {
                        xinfo("[ON_REPLY_REGISTER][$rm][$ci] registrar remove[$var(ret)]\n");
                        $var(ret) = unregister("location", "$tu");
                } else {
                        xinfo("[ON_REPLY_REGISTER][$rm][$ci] registrar save[$var(ret)]\n");
                        $var(ret) = save("location", "0x02");
                }
        }
}

route[INVITE] {
    ...
    if(reg_fetch_contacts("location", "$ru", "caller"))  {
         if(!tcp_conid_alive("$var(conid)")) {
                 xinfo("[RELAY][$ci] using tcp connection inactive [$var(conid)] suspend/push\n");
                 route(PUSH_SUSPEND);
         } else {
                  xinfo("[RELAY][$ci] using tcp connection [$var(conid)]\n");
         }
     }
     ...
}
```

### DIFF
--- a/src/modules/registrar/save.c
+++ b/src/modules/registrar/save.c
@@ -71,6 +71,27 @@ extern sruid_t _reg_sruid;
 static int q_override_msg_id;
 static qvalue_t q_override_value;
 
+int reg_get_cfg_tcpconnid(void)
+{
+       int n;
+       sr_xavp_t *vavp=NULL;
+       str vname = {"tcpconn_id", 10};
+
+       n = 0;
+
+       if(reg_xavp_cfg.s!=NULL)
+       {
+               vavp = xavp_get_child_with_ival(&reg_xavp_cfg, &vname);
+               if(vavp!=NULL)
+               {
+                       n = vavp->val.v.i;
+                       LM_DBG("using tcpconn_id value from xavp: %d\n", n);
+               }
+       }
+
+       return n;
+}
+
 /*! \brief
  * Process request that contained a star (*) as a contact, in that case,
  * we will remove all bindings with the given username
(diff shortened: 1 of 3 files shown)


src/modules/registrar/save.c
```diff
@@ -71,6 +71,27 @@ extern sruid_t _reg_sruid;
...
+       str vname = {"tcpconn_id", 10};
+
+       n = 0;
+
+       if(reg_xavp_cfg.s!=NULL)
+       {
+               vavp = xavp_get_child_with_ival(&reg_xavp_cfg, &vname);
+               if(vavp!=NULL)
+               {
+                       n = vavp->val.v.i;
```
btriller, line 87:
```suggestion
                       n = (int)vavp->val.v.l;
```

miconda:
Thanks!

Issue #3440, opened 2023-05-08
pv_cache gets filled if dynamic names are used


linuxmaniac:
### Description

_**Sipwise** hat on_

We noticed that pkg memory was getting exhausted and we found out that pv_cache_add() was the origin of the allocations.
We are using ``$xavp(user_<UUID>)`` variables that, for sure, are kind of dynamic.

There is a pv_cache_drop() that tries to remove ``$sht()`` occurrences when the cache is about to be filled but in our case that's not helping.

#### Log Messages

```
WARNING: ROUTE_SET_CALLEE_DIALOG_TOTAL <core> [core/pvapi.c:340]: pv_cache_add(): pv cache limit is going to be exceeded - pkg memory may get filled with pv declarations
WARNING: dialog:failed <core> [core/pvapi.c:340]: pv_cache_add(): pv cache limit is going to be exceeded - pkg memory may get filled with pv declarations
```

### Possible Solutions

Possible solutions that came on the top of my head:

- add a datetime member to struct _pv_cache so we can keep the last time it was accessed and search for the oldest and remove them in pv_cache_drop()
- add pv module parameter to define a magic prefix, for instance '__' that will mark those vars as "temporal" should remove it from cache when needed.
- add pv module parameter to control what types of variables will be added to cache

Any thoughts about this?

miconda:
How do you use the `$xavp(...)`? Do you need to work with them using the kamailio-PV-style name? Otherwise it might be better to use the dedicated functions exported to kemi from pv module:

  - https://www.kamailio.org/docs/tutorials/devel/kamailio-kemi-framework/modules/#pvx 

Then, yes, we can also think of ways to make the pv cache more flexible for cleaning, ...

miconda:
It might be a bit tricky to really make it generic, because some variables keep pointer to shared memory, definitely is not going to work for `$shv(...)` to be (re-)initialized after forking. So if a shv is used not very often and destroyed, then it becomes unusable for ever, (re-)initialisation will not make it properly visible to all processes. Then I am not sure that every PV does a full clean on destroy of their internal structure, so it can still leak. That's also a reason now is done for sht, because its destroy was reviewed to be clean.

So, likely you have to combine with a list of variable types that are ok for last-access-based destroy and (re-)initialise.

PR #3448, opened 2023-05-10
core: pv_cache_dump

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3440

#### Description

implemented a way to dump the names of the variables stored in pv_cache this helped me to find out what was filling the pv_cache

### DIFF
--- a/src/core/cfg_core.c
+++ b/src/core/cfg_core.c
@@ -55,6 +55,7 @@
 #include "sock_ut.h"
 #include "cfg/cfg.h"
 #include "cfg_core.h"
+#include "pvapi.h"
 
 struct cfg_group_core default_core_cfg = {
 		L_WARN,		/*!<  print only msg. < L_WARN */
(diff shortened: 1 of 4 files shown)


miconda:
Shouldn't the value changed back to 0 on dump callback? To allow setting it back to pid later ...

linuxmaniac:
Uhm... I don't see that in mem_dump_pkg_cb

https://github.com/kamailio/kamailio/blob/6b4e3b4b227381bfe896b23af5c4ce996c95f6e8/src/core/pt.c#L536-L562

miconda:
Probably it is going to require two changes of the value for another dump, I guess the same happens for similar cases.

Likely they were considered like one time use for troubleshooting, so it is fine for me to merge and eventually improve later.

linuxmaniac:
Ah, I see. So like this? Change already pushed
```
diff --git a/src/core/pvapi.c b/src/core/pvapi.c
index b36c92b2be..c184db28d9 100644
--- a/src/core/pvapi.c
+++ b/src/core/pvapi.c
@@ -119,6 +119,7 @@ void pv_cache_dump_cb(str *gname, str *name)
 {
        if (cfg_get(core, core_cfg, pv_cache_dump) == my_pid()) {
                pv_cache_dump(cfg_get(core, core_cfg, memlog));
+               cfg_get(core, core_cfg, pv_cache_dump) = 0;
        }
 }
```

miconda:
Fine for me. Thanks.

PR #3469, opened 2023-05-24
src/Makefile.defs: recognize GCC 13 as recent ⇔  not too old compiler

dilyanpalauzov:


dilyanpalauzov:
I have the trouble with v5.7 and I assume the same applies for the `master` branch.

miconda:
Thanks!

PR #3473, opened 2023-05-31
rtpengine: support receiving events from rtpengine and raise an event

joeygo:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -111,7 +111,7 @@ MODULE_VERSION
 
 #define DEFAULT_RTPP_SET_ID 0
 
-#define RTPENGINE_DTMF_EVENT_BUFFER	32768
+#define RTPENGINE_DTMF_EVENT_BUFFER 32768
 
 enum {
 	RPC_FOUND_ALL = 2,
@@ -246,8 +246,8 @@ static int build_rtpp_socks(int lmode, int rtest);
 static char *send_rtpp_command(struct rtpp_node *, bencode_item_t *, int *);
 static int get_extra_id(struct sip_msg *msg, str *id_str);
 
-static int rtpengine_set_store(modparam_t type, void * val);
-static int rtpengine_set_dtmf_events_sock(modparam_t type, void * val);
+static int rtpengine_set_store(modparam_t type, void *val);
+static int rtpengine_set_dtmf_events_sock(modparam_t type, void *val);
 static int rtpengine_add_rtpengine_set(char * rtp_proxies, unsigned int weight, int disabled, unsigned int ticks);
 
 static int mod_init(void);
@@ -263,7 +263,7 @@ static int add_rtpp_node_info(
 static int rtpp_test_ping(struct rtpp_node *node);
 
 static void rtpengine_dtmf_events_loop(void);
-static int rtpengine_raise_dtmf_event(char* buffer, int len);
+static int rtpengine_raise_dtmf_event(char *buffer, int len);
 
 /* Pseudo-Variables */
 static int pv_get_rtpestat_f(struct sip_msg *, pv_param_t *, pv_value_t *);
@@ -498,7 +498,7 @@ static param_export_t params[] = {
 	{"mos_average_jitter_pv",     PARAM_STR, &global_mos_stats.average.jitter_param      },
 	{"mos_average_roundtrip_pv",  PARAM_STR, &global_mos_stats.average.roundtrip_param   },
 	{"mos_average_roundtrip_leg_pv", PARAM_STR, &global_mos_stats.average.roundtrip_leg_param },
-	{"mos_average_samples_pv",    PARAM_STR, &global_mos_stats.average.samples_param     },	
+	{"mos_average_samples_pv",    PARAM_STR, &global_mos_stats.average.samples_param     },
 
 		/* designated side A */
 		{"mos_A_label_pv", PARAM_STR, &side_A_mos_stats.label_param},
@@ -1251,12 +1251,12 @@ error:
 	return -1;
 }
 
-static int rtpengine_set_dtmf_events_sock(modparam_t type, void * val)
+static int rtpengine_set_dtmf_events_sock(modparam_t type, void *val)
 {
-	char * p;
-	p = (char* )val;
+	char *p;
+	p = (char *)val;
 
-	if(p==0 || *p=='\0'){
+	if(p == 0 || *p == '\0') {
 		return 0;
 	}
 
@@ -1272,11 +1272,13 @@ static void rtpengine_dtmf_events_loop(void)
 	char *p;
 	str s_port;
 	unsigned int port;
+	unsigned int socket_len;
 	union sockaddr_union udp_addr;
 	char buffer[RTPENGINE_DTMF_EVENT_BUFFER];
 
 	p = q_memchr(rtpengine_dtmf_event_sock.s, ':', rtpengine_dtmf_event_sock.len);
-	if (!p) {
+
+	if(!p) {
 		LM_ERR("failed to initialize dtmf event listener because no port was specified %.*s!\n", rtpengine_dtmf_event_sock.len, rtpengine_dtmf_event_sock.s);
 		return;
 	}
@@ -1284,7 +1286,7 @@ static void rtpengine_dtmf_events_loop(void)
 	s_port.s = p + 1;
 	s_port.len = rtpengine_dtmf_event_sock.s + rtpengine_dtmf_event_sock.len - s_port.s;
 
-	if (s_port.len <= 0 || str2int(&s_port, &port) < 0 || port > 65535) {
+	if(s_port.len <= 0 || str2int(&s_port, &port) < 0 || port > 65535) {
 		LM_ERR("failed to initialize dtmf event listener because port is invalid %.*s\n", rtpengine_dtmf_event_sock.len, rtpengine_dtmf_event_sock.s);
 		return;
 	}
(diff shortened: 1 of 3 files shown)


src/modules/rtpengine/rtpengine.c
```diff
@@ -1240,8 +1250,208 @@ static int rtpengine_add_rtpengine_set(char *rtp_proxies, unsigned int weight,
...
+	}
+
+	rtpengine_dtmf_event_sock.s = p;
+	rtpengine_dtmf_event_sock.len = strlen(rtpengine_dtmf_event_sock.s);
+
+	return 0;
+}
 
-static int fixup_set_id(void **param, int param_no)
+static void rtpengine_dtmf_events_loop(int rank)
```
rfuchs, line 1268: The `rank` argument seems unused?

src/modules/rtpengine/rtpengine.c
```diff
@@ -2147,6 +2401,9 @@ static int child_init(int rank)
 		/* probe rtpengines only in first worker */
 		if(build_rtpp_socks(0, 1))
 			return -1;
+			
+		if (rtpengine_dtmf_event_sock.len > 0)
+			rtpengine_dtmf_events_loop(rank);
```
rfuchs, line 2406: This enters into an infinite read loop during child_init(). Doesn't this block the startup of all child processes?

rfuchs (commented):
A bit unsure about how this is supposed to work, maybe I'm just dense?

src/modules/rtpengine/rtpengine.c
miconda: It seems to block the child process with rank 1 (PROC_SIPINIT), which should not be done, because it can create problems for other modules such as usrloc.

A new process has to be started -- in mod init, it has to be registered with:

```
		register_procs(1);
		cfg_register_child(1);
```

And in child init created with fork_process(), something like:

```
	if(rank == PROC_MAIN) {
	    /* fork worker process */
	    newpid = fork_process(PROC_RPC, "RTPENGINE DTMF WORKER", 1);
	    if(newpid < 0) {
		    LM_ERR("failed to fork worker process %d\n", i);
		    return -1;
	    } else if(newpid == 0) {
		    if(cfg_child_init())
			    return -1;
		    /* child - this will loop forever */
		    rtpengine_dtmf_events_loop();
	    } else {
		    /* parent process*/
	    }
		return 0;
	}
```

src/modules/rtpengine/rtpengine.c
joeygo: You are right. I removed it.

src/modules/rtpengine/rtpengine.c
joeygo: Thanks for the example.
I fixed the code.

src/modules/rtpengine/rtpengine.c
```diff
@@ -1240,8 +1250,208 @@ static int rtpengine_add_rtpengine_set(char *rtp_proxies, unsigned int weight,
...
+	rtpengine_dtmf_event_sock.len -= s_port.len + 1;
+	trim(&rtpengine_dtmf_event_sock);
+	rtpengine_dtmf_event_sock.s[rtpengine_dtmf_event_sock.len] = '\0';
+
+	memset(&udp_addr, 0, sizeof(udp_addr));
+
+	if (rtpengine_dtmf_event_sock.s[0] == '[') {
+		udp_addr.sin6.sin6_family = AF_INET6;
+		udp_addr.sin6.sin6_port = htons(port);
+		ret = inet_pton(AF_INET, rtpengine_dtmf_event_sock.s, &udp_addr.sin6.sin6_addr);
```
rfuchs, line 1300: Must be AF_INET6 here

src/modules/rtpengine/rtpengine.c
```diff
@@ -1240,8 +1250,208 @@ static int rtpengine_add_rtpengine_set(char *rtp_proxies, unsigned int weight,
...
+	}
+
+	rtpengine_dtmf_event_fd = socket(udp_addr.s.sa_family, SOCK_DGRAM, 0);
+
+	if(rtpengine_dtmf_event_fd < 0) {
+		LM_ERR("can't create socket\n");
+		return;
+	}
+	
+	if (bind(rtpengine_dtmf_event_fd, (struct sockaddr*)&udp_addr.s, sizeof(udp_addr.s)) < 0) {
```
rfuchs, line 1319: sizeof(udp_addr.s) is sizeof(struct sockaddr) which I think may be too small for INET6. sizeof(udp_addr) would be a better option

src/modules/rtpengine/rtpengine.c
```diff
@@ -1240,8 +1250,208 @@ static int rtpengine_add_rtpengine_set(char *rtp_proxies, unsigned int weight,
...
+
+		p = shm_malloc(ret + 1);
+		if (!p) {
+			LM_ERR("could not allocate %d for buffer %.*s\n", ret, ret, buffer);
+			goto end;
+		}
+		memcpy(p, buffer, ret);
+		p[ret] = '\0';		
+
+		if (rtpengine_raise_dtmf_event(p, ret) < 0) {
```
rfuchs, line 1349: I think this leaks the `p` buffer if the return value == 0

src/modules/rtpengine/rtpengine.c
```diff
@@ -1240,8 +1250,208 @@ static int rtpengine_add_rtpengine_set(char *rtp_proxies, unsigned int weight,
...
+	srjson_doc_t jdoc;
+	srjson_t *it = NULL;
+	struct sip_msg *fmsg = NULL;
+	struct run_act_ctx ctx;
+	int rtb;	
+
+	LM_DBG("executing event_route[rtpengine:dtmf-event] (%d)\n", dtmf_event_rt);
+	LM_DBG("dispatching buffer: %s\n", buffer);
+
+	srjson_InitDoc(&jdoc, NULL);
```
rfuchs, line 1369: jdoc needs to be freed somewhere

src/modules/rtpengine/rtpengine.c
```diff
@@ -261,7 +262,7 @@ static int add_rtpp_node_info(
 		void *ptrs, struct rtpp_node *crt_rtpp, struct rtpp_set *rtpp_list);
 static int rtpp_test_ping(struct rtpp_node *node);
 
-static void rtpengine_dtmf_events_loop(int rank);
+static void rtpengine_dtmf_events_loop();
```
rfuchs, line 265: nit: for better standards compliance, be explicit with `(void)`

src/modules/rtpengine/rtpengine.c
```diff
@@ -2377,12 +2382,35 @@ static int mos_label_stats_parse(struct minmax_mos_label_stats *mmls)
...
+			mypid = fork_process(PROC_RPC, "RTPENGINE DTMF WORKER", 1);
+			if(mypid < 0) {
+				LM_ERR("failed to fork RTPENGINE DTMF WORKER process %d\n", mypid);
+				return -1;
+			} else if(mypid == 0) {
+				if(cfg_child_init())
+					return -1;
+				/* this will loop forever */
+				rtpengine_dtmf_events_loop();
+			}
```
rfuchs, line 2405: I suppose there should be an exit() at the end here? Although I'm not too familiar with how the Kamailio internals are handling this

src/modules/rtpengine/rtpengine.c
joeygo: Fixed.

src/modules/rtpengine/rtpengine.c
joeygo: Fixed.

src/modules/rtpengine/rtpengine.c
joeygo: Fixed.

src/modules/rtpengine/rtpengine.c
joeygo: Maybe @miconda can comment on that?

miconda: I think we do not have explicitly `exit` after the functions doing infinite loops in the special child process. But it could be a good practice to kill the process (and by that kamailio instance) if the function that should do infinite loop exits due to some errors before it enters the infinite loop.

src/modules/rtpengine/rtpengine.c
```diff
@@ -1240,8 +1251,215 @@ static int rtpengine_add_rtpengine_set(char *rtp_proxies, unsigned int weight,
...
+	set_route_type(rtb);
+	if(ctx.run_flags & DROP_R_F) {
+		LM_ERR("exit due to 'drop' in event route\n");
+		goto error;
+	}
+
+	return 0;
+
+error:
+	srjson_DestroyDoc(&jdoc);
```
rfuchs, line 1458: This also needs to be done on success, no?

src/modules/rtpengine/rtpengine.c
joeygo: I don't think it is right to kill the entire kamailio process due to an exit in such function.
I print error logs in case of failure in the function. isn't it enough?
If you prefer that I'll kill the process just let me know and I'll do that.

rfuchs:
LGTM now

joeygo:
Hi @miconda,
Is there anything else I should do or join wait for someone to merge the PR?

Thanks

miconda:
Done, thanks!

joeygo:
Hi @miconda,
I was wondering, if I wish to back-port this to 5.7.1, what is the right way of doing so?

Thanks,
Joey.

miconda:
This is a new feature, the stable branches can get backports only for bug fixes or documentation improvements.

PR #3475, opened 2023-06-01
Silence “make: --libs: No such file or directory” warnings

dilyanpalauzov:
With GNU make 4.4 building Kamailio prints many “make: --libs: No such file or directory” and similar warnings.  This changeset utilizes immediate variables and silences the warnings on my systems.


miconda:
Afaik `:=` evaluates the variable expression only once, does it evaluate it for each module Makefile, or only for the first time that variable is assigned? I mean, if module `a` and `b` have something like `BUILDER=...` in their Makefile, a `make all` is going to evaluate `BUILDER` variable each time, or the one from module `a` propagate to `b` compilation?

On the other hand `=` evaluates every time the variable is used.

GCC 4.4 is really old, I would avoid fixing warning for it if it can create side effects by propagating values when not expecting.

dilyanpalauzov:
When the file `a` contains `BUILDER:=a`, the file `b` contains `BUILDER:=b`, and `Makefile` contains
```make
include a
include b

all:
        @echo $(BUILDER)
```
(there is a TAB before @), then typing `make` prints `b`.

I do not understand the question and I do not understand the build system of Kamailio.

dilyanpalauzov:
The proposed here change does fix problems (warnings).  I object that this PR is closed, instead of fixing the underlying warnings by the suggested or other ways.

miconda:
The automatic actions have the purpose to keep the reporter, and devs when it is the case, engaged. An issue or PR marked `stale` can have the label removed by reporter if it is still interested in it. Also, a closed issue or PR can be reopened. Note also that the `stale` label is added based on inactivity period, comments are considered activity, so the inactivity is reset with each of them..

Have in mind that this is a collaborative project, so everyone have to stay engaged, nobody pays for time spent by developers on reviewing commits. From that point of view, I think it is fair that the reporter reacts on a stale or close notification and indicate that the issue is still there or the PR is still relevant. Just coming to blame, it is by far not a nice approach. Some of developers here spent significant resources and time without any direct benefits over the past 20 years to make this application available, that should be appreciated. This kind of automatisation tries to make the life easier for everyone, feel free to propose tools or processes that can help improving and moving forward.

dilyanpalauzov:
> From that point of view, I think it is fair that the reporter reacts on a stale or close notification and indicate that the issue is still there or the PR is still relevant.

Writing repeatedly any text just for the sake of keeping a ticket open does not help anyhow.

miconda:
And, furthermore, I guess the option of the reporter to reopen the issue is still available.

linuxmaniac:
I've implemented a way to manage ``stale`` and re-open ``expired`` issues and pull requests.
So, if the comment as /notexpired it will be re-opened and labels ``expired`` and ``stale`` will be removed.

Hope this helps

PR #3477, opened 2023-06-02
Adding TLS support to db_redis (also password support) and ndb_redis

joelbax:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
A couple of commits into db_redis and ndb_redis adding TLS support and also password support to db_redis.

This mainly includes checking if proper parameter is provided (for ndb_redis is `tls` option in the DB URL and, for db_redis, a new `opt_tls` parameter) and creates a temporary SSL context that is used to initialise the redis context.

Also added `ca_path` parameter to both modules to be able to define a valid folder containing the root certificates used to validate TLS' certificate chain.

db_redis is also updated with a `db_pass` parameter to provide a DB access password.

TLS support is automatically enabled by checking libhiredis_ssl.so existence in each Makefile and defining a `WITH_SSL` flag that enables all the corresponding code lines.

### DIFF
--- a/src/modules/db_redis/db_redis_mod.c
+++ b/src/modules/db_redis/db_redis_mod.c
@@ -31,9 +31,9 @@
 
 #ifdef WITH_SSL
 int db_redis_opt_tls = 0;
-char *ca_path = 0;
+char *db_redis_ca_path = 0;
 #endif
-char *db_pass = 0;
+char *db_redis_db_pass = 0;
 
 MODULE_VERSION
 
@@ -60,9 +60,9 @@ static param_export_t params[] = {
 		{"verbosity", PARAM_INT, &db_redis_verbosity},
 #ifdef WITH_SSL
 		{"opt_tls", PARAM_INT, &db_redis_opt_tls},
-		{"ca_path", PARAM_STRING, &ca_path},
+		{"ca_path", PARAM_STRING, &db_redis_ca_path},
 #endif
-		{"db_pass", PARAM_STRING, &db_pass}, {0, 0, 0}};
+		{"db_pass", PARAM_STRING, &db_redis_db_pass}, {0, 0, 0}};
 
 
 struct module_exports exports = {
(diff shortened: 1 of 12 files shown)


joelbax:
@henningw  This is the new PR for adding TLS support to Redis. Once again, I'm sorry for all the nuisance.
@oej Also fyi (as you commented on the original PR).
Thank you very much in advance!

src/modules/db_redis/db_redis_mod.c
```diff
@@ -29,6 +29,12 @@
 #include "redis_dbase.h"
 #include "redis_table.h"
 
+#ifdef WITH_SSL
+int db_redis_opt_tls = 0;
+char *ca_path = 0;
+#endif
+char *db_pass = 0;
```
miconda, line 36: Add module name prefix to the global variables ca_path and db_pass, like it is db_redis_opt_tls.  The are also declared with extern and being prefixed reduce the chances of conflicts with similar variables (e.g., ca_path can be quite common with some libs).

joelbax: Very good point! Thank you very much. 

Done.

henningw:
Or maybe just as other modules are doing (like ndb_redis), assume a few default paths if there is no pkg-config available.

joelbax:
@miconda I've added logic to check gcc search paths for libraries in case `pkg-config` returns nothing.

I've also fixed som formatting issues but  the check-format task seems to complain about some lines but , I'm really sorry but, I can't see the difference between the wrong line and the new one.

miconda:
Thanks!

miconda:
@joelbax: I merged and then updated a bit the find command in the Makefile for n/db_redis modules to silent some warning about missing directories and use `-name` for finding file expression. Now compiles fine on ubuntu 22.04, before I got errors. Maybe you can test on your system to be sure it didn't break something else.

PR #3485, opened 2023-06-16
pdt: make add_to_tree() idempotent

vcill:
- the function call will suceed even if the prefix being inserted already exists in the tree

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

PDT module initialization currently fails if you try to insert a prefix multiple times. We use kamailio as a loadbalancer and have had this change out in the field for quite some time now as our database often contains a prefix multiple times (associated with the same domain). Would it make sense to change the behavior in this case ?

Thanks,
Vladimir.


vcill:
Yeah, a modparam would be even better!

Sent from Proton Mail mobile

-------- Original Message --------

miconda:
@vcill: are you planning on working further on this to add modparam?

miconda:
Alternative implementation pushed to git repo.

Issue #3486, opened 2023-06-18
 wrong source address usage


sergey-safarov:
### Description
When Kamailio listening IP address is assigned to a secondary network interface and this card does not use for the default route, then at the L3-network level selected wrong IP address and in the "Recrod-Route" and "Via" headers used correct Kamailio IP.

For example in the attached file, frame 5 and fragments in frames 3,4.  [invite-from-wrong-ip.pcap.gz](https://github.com/kamailio/kamailio/files/11781289/invite-from-wrong-ip.pcap.gz)

![image](https://github.com/kamailio/kamailio/assets/2562241/68696b29-b37c-4d0a-b349-97054db14727)

#### Reproduction
Our use case.
1. create a new virtual machine with a random IP address in the AWS cloud;
2. create a new network interface card with a static IP address;
3. attach a new network interface to a virtual machine, in Linux should be configured only one default route via the first network interface with a dynamic IP address;
4. start Kamailio daemon with configured listening TCP/socket with the IP address from the secondary IP card.
5. make a call via Kamailio.

miconda:
I don't think this has to do anything with kamailio source code. It is about IP routing rules in the OS/kernel.

If the sip headers have the expected addresses based on listen socket, then kamailio selected what it should.

As noted on other issues here and mailing list discussions, for tcp is not really guaranteed that the source address of a tcp connection can be enforced, ultimately is the kernel/os stack making the decision.

If you want to discuss more, probably sr-users mailing list is more appropriate. If it results in discovering some code issue, then an issue can be created and it can be patched.

sergey-safarov:
The network source address is important because the Kamailio address is whitelisted in the ACL.
For me easier to manage this at the socket level, than manage it on the network OS routing rules.

We now working pull request for this issue.
So we will make a new attempt to discuss via PR.

sergey-safarov:
As a hotfix we used options
```
tcp_source_ipv4 = IPv4 address
tcp_source_ipv6 = IPv6 address
```
https://www.kamailio.org/wiki/cookbooks/5.4.x/core#tcp_source_ipv4_tcp_source_ipv6

This works if all Kamailio sockets are bound to one IP.
To get it working for multiple IPs we try to make such a change
```diff
diff --git a/src/main.c b/src/main.c
index 4ee5afe9d0..3670a8e826 100644
--- a/src/main.c
+++ b/src/main.c
@@ -1600,6 +1601,7 @@ int main_loop(void)
[...]

PR #3490, opened 2023-06-20
core: fixed wrong network interface selection.

ivanuschak:
fixed incorrect source IP address selection for the SIP messages sending procedure when TCP transport is used or for UDP with the 'mhomed' setting set as 'mhomed=1'.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3486(replace XXXX with an open issue number)

#### Description

The current service behavior is as follows: if OS contains several network interfaces and the service has a listening TCP socket (or UDP with 'mhomed=1' setting) on the second (third, fourth and so  on) interface, then the outbound SIP messages will be always sent from the first (default) interface.
These changes make the outbound SIP messages always going through the network_interface/address, which is taken for listening by service.

### DIFF
--- a/src/core/tcp_main.c
+++ b/src/core/tcp_main.c
@@ -1240,6 +1240,49 @@ error:
 	return 0;
 }
 
+inline static int find_listening_sock_info(
+		int s, union sockaddr_union **from, int type)
+{
+	struct ip_addr ip;
+	struct socket_info *si = NULL;
+	su2ip_addr(&ip, *from);
+
+	si = find_si(&ip, 0, type);
+
+	if(unlikely(si == 0)) {
+		si = find_sock_info_by_address_family(type, ip.af);
+		if(si) {
+			int optval = 1;
+			su2ip_addr(&ip, &si->su);
+			*from = &si->su;
+			if(setsockopt(s, SOL_SOCKET, SO_REUSEADDR, (void *)&optval,
+					   sizeof(optval))
+					== -1) {
+				LM_ERR("setsockopt SO_REUSEADDR %s\n", strerror(errno));
+				/* continue, not critical */
+			}
+			optval = 1;
+			if(setsockopt(s, SOL_SOCKET, SO_REUSEPORT, (void *)&optval,
+					   sizeof(optval))
+					== -1) {
+				LM_ERR("setsockopt SO_REUSEPORT %s\n", strerror(errno));
+				/* continue, not critical */
+			}
+			if(unlikely(bind(s, &si->su.s, sockaddru_len(si->su)) != 0)) {
+				LM_WARN("binding to source address %s failed: %s [%d]\n",
+						su2a(&si->su, sizeof(si->su)), strerror(errno), errno);
+				return -1;
+			}
+		}
+	} else {
+		if(unlikely(bind(s, &(*from)->s, sockaddru_len(**from)) != 0)) {
+			LM_WARN("binding to source address %s failed: %s [%d]\n",
+						su2a(&si->su, sizeof(si->su)), strerror(errno), errno);
+			return -1;
+		}
+	}
+	return 0;
+}
 
 /* do the actual connect, set sock. options a.s.o
  * returns socket on success, -1 on error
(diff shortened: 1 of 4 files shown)


miconda:
There is a note that tcp_set_src_addr() should be used only on init:

  - https://github.com/kamailio/kamailio/blob/master/src/core/tcp_main.c#L217

Are you sure it does not have side effects, when there is a mix of forcing and not-forcing send socket?

Is it compatible with the old behaviour when the global parameters tcp_source_ipv4/tcp_source_ipv6 were supposed to be global for all traffic?

ivanuschak:
_tcp_set_src_addr()_/_udp_set_src_addr()_ function calls have been moved to the _main()_ after _fix_all_socket_list()_ invocation. This place is where service is being initiated and not yet entered into the main loop.

_udp_source_ipv6_/_tcp_source_ipv6_/_udp_source_ipv4_/_tcp_source_ipv4_ are global variables and they are initiated on the service init step and then there are no changes for this variables.

Regarding
> Is it compatible with the old behaviour when the global parameters tcp_source_ipv4/tcp_source_ipv6 were supposed to be global for all traffic?
Is there some old behavior, which can be enabled by some option? If yes, then the compatibility with this mode should be kept, right?

I didn't find these variables (_tcp_source_ipv4_/_tcp_source_ipv6_) are initialized somewhere in the source code, but they are used, so it looks like its initialization is needed.

miconda:
The `tcp_source_ipv4` and `tcp_source_ipv6` are set by config parser:

  - https://github.com/kamailio/kamailio/blob/master/src/core/cfg.y#L1203-L1222

Your commit likely is overwriting them.

src/main.c
```diff
@@ -2864,6 +2864,16 @@ int main(int argc, char** argv)
 		fprintf(stderr,  "failed to initialize list addresses\n");
 		goto error;
 	}
+	for(struct socket_info* si=udp_listen; si; si=si->next){
+		udp_set_src_addr(&si->address);
+	}
```
miconda, line 2869: This loop results is setting the global variables to the last IPv4 or IPv6 listen addresses in the list, right?

Same happens with the loop for tcp below.

ivanuschak:
Yes, you're correct the `tcp_source_ipv4` and `tcp_source_ipv6` are set by config parser. My commit replaces this logic. This replacing is not correct, since these vars are configurable from the config file. I'll remove the logic for TCP.

But I don't see similar logic for UDP, i.e. there is no default IP addresses.
Is there similar source IP addresses for UDP?

src/main.c
ivanuschak: Yes, correct. The loop result is setting the global variables to the last IPv4 or IPv6 listen addresses in the list. The same for UDP.
But these loops are not correct since these source IP addresses are set up by config file (at least for TCP), these loops just replace the source IP addresses. I will change it.

miconda:
I haven't added `tcp_source_ipv4` or `tcp_source_ipv6`, neither used them, so I am not sure what was the main purpose for them. I also don't know why there is no equivalent for udp, maybe one should be introduced. But more or less from the beginning of the project, the rule to reuse the incoming socket for outgoing traffic was in place.

Maybe the next steps should be:

  - add also for udp the corresponding core parameters to be set from config
  - add a new parameter like `proto_source_mode` which when set to 0 (the default) the current behaviour is preserved and when set to 1 the one you want is done, if that is solving issues you are facing

In this way we have backward compatibility by default and you have an option to change that behaviour for your specific needs.

ivanuschak:
Implementation was re-worked. Now for _tcp_do_connect_ function if _from_ parameter is null then the sending socket is bound on the first listening address/port, which matches the required address family (IPv4/IPv6).
The same approach is implemented for UDP sockets when 'mhomed=1'.

src/core/tcp_main.c
```diff
@@ -1240,6 +1240,49 @@ struct tcp_connection *tcpconn_new(int sock, union sockaddr_union *su,
...
+
+	si = find_si(&ip, 0, type);
+
+	if(unlikely(si == 0)) {
+		si = find_sock_info_by_address_family(type, ip.af);
+		if(si) {
+			int optval = 1;
+			su2ip_addr(&ip, &si->su);
+			*from = &si->su;
+			if(setsockopt(s, SOL_SOCKET, SO_REUSEADDR, (void *)&optval,
```
sergey-safarov, line 1258: Current implementation requires kernel with `SO_REUSEADDR` and `SO_REUSEPORT`.
I have checked sources and found several places where this features also required.

I have a question.
Can we relay on statement "all kernels provide  `SO_REUSEADDR` and `SO_REUSEPORT`" functionality?

PR #3506, opened 2023-07-06
geoip2: added distance function

sergey-safarov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
Added distance function which allows to calculate the distance between the geocoordinates of the the IP-address passed as incoming parameter (coordinates for the IP are determined inside the function) and the geocoordinates passed as parameters for the function.
Functions can be used to evaluate the RTPengine server closest to a user.

### DIFF
--- a/src/modules/geoip2/doc/geoip2_admin.xml
+++ b/src/modules/geoip2/doc/geoip2_admin.xml
@@ -149,6 +149,29 @@ if(geoip2_match("$si", "src"))
 </programlisting>
 	    </example>
 	</section>
+	<section>
+		<title>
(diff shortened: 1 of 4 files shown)


sergey-safarov:
This function does not added to `geoip` module because:
1) `geoip` does not supports IPv6 addresses;
2) after IPv6 support was added to `geoip` found two databases (IPv4 and IPv6) cannot be added as module config parameters;
3) ASAN reported a memory leak.

Think `geoip2` module prefered to extend then `geoip`.

miconda:
Note that the config function was renamed to geoip2_distance() to be coherent with the other exported function and reflect the module name.

Issue #3502, opened 2023-07-04
OpenSSL depreacation warning


sergey-safarov:
**crypto_uuid.c**
```
CC (gcc) [M crypto.so]		api.o
CC (gcc) [M crypto.so]		crypto_aes.o
CC (gcc) [M crypto.so]		crypto_evcb.o
CC (gcc) [M crypto.so]		crypto_mod.o
CC (gcc) [M crypto.so]		crypto_uuid.o
crypto_uuid.c: In function 'crypto_generate_callid':
crypto_uuid.c:184:9: warning: 'SHA1_Init' is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  184 |         SHA1_Init(&crypto_ctx);
      |         ^~~~~~~~~
In file included from crypto_uuid.c:37:
/usr/include/openssl/sha.h:49:27: note: declared here
   49 | OSSL_DEPRECATEDIN_3_0 int SHA1_Init(SHA_CTX *c);
      |                           ^~~~~~~~~
crypto_uuid.c:185:9: warning: 'SHA1_Update' is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  185 |         SHA1_Update(&crypto_ctx, crypto_callid_seed, SEED_LEN);
      |         ^~~~~~~~~~~
In file included from crypto_uuid.c:37:
/usr/include/openssl/sha.h:50:27: note: declared here
   50 | OSSL_DEPRECATEDIN_3_0 int SHA1_Update(SHA_CTX *c, const void *data, size_t len);
      |                           ^~~~~~~~~~~
crypto_uuid.c:186:9: warning: 'SHA1_Update' is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  186 |         SHA1_Update(&crypto_ctx, crypto_callid_counter, CTR_LEN);
      |         ^~~~~~~~~~~
In file included from crypto_uuid.c:37:
/usr/include/openssl/sha.h:50:27: note: declared here
   50 | OSSL_DEPRECATEDIN_3_0 int SHA1_Update(SHA_CTX *c, const void *data, size_t len);
      |                           ^~~~~~~~~~~
crypto_uuid.c:187:9: warning: 'SHA1_Final' is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  187 |         SHA1_Final(crypto_buf, &crypto_ctx);
      |         ^~~~~~~~~~
[...]

PR #3518, opened 2023-07-20
crypto: SHA1_Init deprecated at openssl 3.0

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [x] Related to issue #3502

#### Description
From https://www.openssl.org/docs/man3.0/man7/migration_guide.html

> Use of low-level digest functions such as SHA1_Init(3) have been informally
> discouraged from use for a long time. Applications should instead use the
> high level EVP APIs EVP_DigestInit_ex(3), EVP_DigestUpdate(3) and
> EVP_DigestFinal_ex(3), or the quick one-shot EVP_Q_digest(3).

### DIFF
--- a/src/modules/crypto/crypto_uuid.c
+++ b/src/modules/crypto/crypto_uuid.c
@@ -177,14 +177,33 @@ static inline int crypto_format_rfc4122_uuid(
  */
 void crypto_generate_callid(str *callid)
 {
+#if OPENSSL_VERSION_NUMBER > 0x030000000L
+	EVP_MD_CTX *crypto_ctx = NULL;
+#else
 	static SHA_CTX crypto_ctx = {0};
+#endif
 	static unsigned char crypto_buf[SHA_DIGEST_LENGTH] = {0};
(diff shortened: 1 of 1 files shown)


miconda:
Thanks!

I wonder if it would work to have:

```
static EVP_MD_CTX crypto_ctx = {0};
```

instead of:

```
EVP_MD_CTX *crypto_ctx = NULL;
```

And then get rid of using EVP_MD_CTX_new()/_free() functions. That makes it look very similar like the old variant.

linuxmaniac:
I don't think that would fly looking into the definition of the structure...
https://github.com/openssl/openssl/blob/7551264186f176ca5801aa84d60c7b91d8fba31f/crypto/evp/evp_local.h#L16-L34

miconda:
It may require to add `EVP_MD_CTX_reset()` ... anyhow, you can merge as it is and later when I get some time I can look a bit more at libssl and see if we can avoid use of new/free for each execution.

PR #3521, opened 2023-07-24
ims_qos: doc update and warn logs

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Primary goal is to update doc, for Rx_AAR() function, to mention is used only for replies.

I've also changed some LM_DBG to LM_WARN but I can revert that on request (don't want to overflow noone with warnings)

### DIFF
--- a/src/modules/ims_qos/doc/ims_qos_admin.xml
+++ b/src/modules/ims_qos/doc/ims_qos_admin.xml
@@ -29,7 +29,7 @@
 
       <itemizedlist>
         <listitem>
-          <para>Dialog</para>
+          <para>IMS Dialog</para>
         </listitem>
 
         <listitem>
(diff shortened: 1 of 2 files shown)


smititelu:
@carstenbock what do you think of this?

miconda (commented):
Removing the condition `if(msg->first_line.type != SIP_REPLY) {` seems legit from native kamailio.cfg scripting language as the exports limits to ONREPLY_ROUTE and the internal interpreter can detect it, but for KEMI (which might not be the case here, however, making it as a general comment), it is not possible to detect at startup where a function is going to used. Therefore, thinking of thee future, it is probably better to have runtime checks about the type of message or route block. Even for the native scripting language, there are workarounds in some cases to execute a function in another route block (e.g., execute request-specific exported functions inside a `route[x]` block that is called from on/reply route block.

From that perspective, I would not remove this check. Over all, I am fine to merge it with this change as well. Nowadays the config writer should be more careful anyhow, especially if using KEMI, to be sure functions are not used in a wrong place.

smititelu:
Thanks for the feedback! Will update PR soon. Did not know about the ```route[x]``` stuff.

carstenbock:
@smititelu : Looks good to me, sorry was on holiday in the last two weeks.... :-)

Issue #3443, opened 2023-05-09
tcp_send_timeout not used on reused socket


schoberw:
### Description

Having clients connected via TCP doing only passive connections in the kamailio server it seems the tcp_send_timeout is not used on messages going out to the clients. While tcp_keepalive is doing well as long as there was not sent any packet to the client tcp_send_timout does nothing. The underlying Linux tries to send the SIP message for ages (15 minutes in my case).

Do I have a wrong understanding of tcp_send_timeout?

I tried with tcp_async yes and no with same results.

### Troubleshooting

These are my tcp settings:

```
# def = 10 Time in seconds after a TCP connection will be closed if it is not available for writing in this interval (and Kamailio wants to send something on it).
# Lower this value for faster detection of broken TCP connections.
tcp_send_timeout=3

tcp_connect_timeout=10  # def = 10
tcp_connection_lifetime=3605
tcp_rd_buf_size=32768
tcp_max_connections=2048
tcp_connection_match=1  # match also local port

tcp_keepalive=yes       # Enables keepalive for tcp (sets SO_KEEPALIVE socket option)
tcp_keepcnt=3           # Number of keepalives sent before dropping the connection (TCP_KEEPCNT socket option). Linux only.
tcp_keepidle=30         # Time before starting to send keepalives, if the connection is idle (TCP_KEEPIDLE socket option). Linux only.
tcp_keepintvl=10        # Time interval between keepalive probes, when the previous probe failed (TCP_KEEPINTVL socket option). Linux only.

tcp_async=yes           # alias tcp_buf_write, default=yes
tcp_crlf_ping=no
```

I also tried to set the tcp_send_timeout to much too high value 333666 to trigger this line of code in tcp_options.c:

```
WARN("%s: timeout too big (%u), the maximum value is %u\n",
                                name, *to, max_val);
```

But even in -d -d -d debugging that line never hits.

I do **not** make any connections outbound (senseless on NAT). Instead I'd like to reject this:

```
                if ($dP =~ "tcp") {
                        set_forward_no_connect();
                        xlog("L_INFO", ">> checking tcp connection to $dd:$dp");
                        if(tcp_get_conid("$dd:$dp", "$var(conid)")) {
                                xlog("L_INFO", ">> connection id to $dd:$dp is: $var(conid)");
                        } else {
                                xlog("L_ERR", ">> connection to $dd:$dp is no longer connected. Client must reregister!");
                                sl_send_reply("480", "TCP not connected");
                                exit;
[...]

PR #3528, opened 2023-07-31
core: Add TCP_USER_TIMEOUT socket option on listening socket.

schoberw:
Use tcp_send_timeout config option also on listening socket to timeout outbound messages sent on passive inbound connections.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3443

#### Description
Based on the description of core parameter "tcp_send_timeout" the timeout should also work for sending on forked incoming tcp connections. But sending on a broken connection causes the kernel to use the default values of `tcp_retries1` and `tcp_retries2` , leading to try to send a SIP message for 15 minutes. This makes absolutely no sense in a real time kamailio application.

Following man 7 tcp TCP_USER_TIMEOUT can be used on recent Linux kernels to utilize tcp_send_timeout:

       TCP_USER_TIMEOUT (since Linux 2.6.37)
              This  option  takes  an  unsigned  int as an argument.  When the value is greater than 0, it specifies the maximum
              amount of time in milliseconds that transmitted data may remain unacknowledged before TCP will forcibly close  the
              corresponding connection and return ETIMEDOUT to the application.  If the option value is specified as 0, TCP will
              use the system default.

              Increasing user timeouts allows a TCP connection to survive extended periods without end-to-end connectivity.  De‐
              creasing  user  timeouts  allows applications to "fail fast", if so desired.  Otherwise, failure may take up to 20
              minutes with the current system defaults in a normal WAN environment.

              This option can be set during any state of a TCP connection, but is effective only during the synchronized  states
              of  a  connection  (ESTABLISHED,  FIN-WAIT-1, FIN-WAIT-2, CLOSE-WAIT, CLOSING, and LAST-ACK).  Moreover, when used
              with the TCP keepalive (SO_KEEPALIVE) option, TCP_USER_TIMEOUT will override keepalive to determine when to  close
              a connection due to keepalive failure.

              The option has no effect on when TCP retransmits a packet, nor when a keepalive probe is sent.

              This option, like many others, will be inherited by the socket returned by accept(2), if it was set on the listen‐
              ing socket.

              Further details on the user timeout feature can be found in RFC 793 and RFC 5482 ("TCP User Timeout Option").

Having a tcp connection break by firewall or network breakdown the retransmits to this destination are now aborted after `tcp_send_timeout` seconds with a 

    NOTICE: <core> [core/tcp_read.c:267]: tcp_read_data(): error reading: Connection timed out (110) ([1.2.3.4]:51151 ->

### DIFF
--- a/src/core/tcp_main.c
+++ b/src/core/tcp_main.c
@@ -3172,6 +3172,18 @@ int tcp_init(struct socket_info *sock_info)
 	}
 #endif
 	init_sock_keepalive(sock_info->socket);
+#ifdef HAVE_TCP_USER_TIMEOUT
+	if((optval = TICKS_TO_S(cfg_get(tcp, tcp_cfg, send_timeout)))) {
+		optval *= 1000;
+		if(setsockopt(sock_info->socket, IPPROTO_TCP, TCP_USER_TIMEOUT, &optval,
+				   sizeof(optval))
+				< 0) {
+			LM_WARN("failed to set TCP_USER_TIMEOUT: %s\n", strerror(errno));
+		} else {
+			LM_INFO("Set TCP_USER_TIMEOUT=%d ms\n", optval);
+		}
+	}
+#endif
 	if(bind(sock_info->socket, &addr->s, sockaddru_len(*addr)) == -1) {
 		LM_ERR("bind(%x, %p, %d) on %s:%d : %s\n", sock_info->socket, &addr->s,
 				(unsigned)sockaddru_len(*addr), sock_info->address_str.s,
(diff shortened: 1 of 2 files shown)


miconda:
Thanks! Linux 2.6.37, which added TCP_USER_TIMEOUT, is old enough to just merge this PR. If there are be reports related to its unavailability, we will find a solution at that moment.

schoberw:
Is there a reason why it didn't make it into 5.7.2?

schoberw:
Anything I need to do?

PR #3529, opened 2023-08-01
statsd: Add labels to metrics.

eloycoto:
A user reached me to add a way to add the custom labels into the statsd
module, so a better way to report metrics to the observability platform.

I keep the same old functions and add a new parameter to the statsd
modules to both interfaces cfg and kemi.

The full documentation can be found here:
https://docs.datadoghq.com/developers/dogstatsd/datagram_shell/?tab=metrics

So, each function can be used like this: ``` statsd_set("fooo", 1,
"inbound"); statsd_gauge("NotFound", "+1", "outbound,carrierFoo");
statsd_gauge("AuthFailed", "+1", "carrier=foo,priority=10"); ```

Signed-off-by: Eloy Coto <eloy.coto@acalustra.com>
Tested-by: Alex Antonevych <alex.antonevych@replicant.ai>

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

I'm waiting until first review on function arguments to create the docs
changes.

### DIFF
--- a/src/modules/statsd/statsd.c
+++ b/src/modules/statsd/statsd.c
@@ -34,12 +34,22 @@ MODULE_VERSION
 static int mod_init(void);
 void mod_destroy(void);
 static int func_gauge(struct sip_msg *msg, char *key, char *val);
+static int func_gauge_with_labels(
+		struct sip_msg *msg, char *key, char *val, char *labels);
 static int func_histogram(struct sip_msg *msg, char *key, char *val);
+static int func_histogram_with_labels(
+		struct sip_msg *msg, char *key, char *val, char *labels);
 static int func_set(struct sip_msg *msg, char *key, char *val);
+static int func_set_with_labels(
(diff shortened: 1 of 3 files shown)


miconda:
Thanks for this feature! I understand that the PR keeps the old functions and functionality, therefore the it can be merged.

@eloycoto you already have developer access.

src/modules/statsd/statsd.c
linuxmaniac, line 1: For the KEMI part: I think you can't use the same name for the function even with different number of parameters. Is that correct @miconda?

eloycoto: @miconda any hint here?

miconda: Indeed, there has to be different function names, not all scripting languages support variadic number of parameters and those that do may have different approach of working with them. Therefore a function name must be unique, exported only once for a module.

PR #3531, opened 2023-08-07
dispatcher: fix weighted sort for n<100 calls

devopsec:
- update weighted/relative weighted algos to use probability based sort
- calls from 0-99 now follow the weighted distributions
- removed the now unused shuffle_uint100array() function

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
The previous random sort for weights in a dispatcher set distributed correctly only after a minimum of n=100 calls.  
This PR fixes the sort algorithm to distribute calls based on those weights, even for n<100 calls.  
The real benefit here is customers can see the weighted distribution occur sooner now, when testing.


miconda:
Thanks for this PR! However, it brings significant changes to the exiting behaviour/design of the weight distribution, which was done in purpose to be randomised along the 100 calls.

I think your PR should be changed to add a new dispatching, algorithm, leaving the existing one as it is.

A few more comments for coding style, to keep it as much as possible coherent: declare variables at the beginning of function or block; one line comments should be in a single line `/* ... */`, not three lines.

devopsec:
@miconda thanks for reviewing this.  
I chose to go with replacement since I believe this better represents the original intention of the weighted / relative weighted algorithms.

Now to be honest, I could be completely wrong in thinking that.  
At n>100 calls; or simply when scaling, the new algorithm distributes the same percentage of calls to each destination.  
The only difference being the ordering the destinations are selected.

If you think that is incorrect then I can just move it to a new dispatching algorithm.

And i'll update the formatting on those comments as well.

henningw:
If I understand it correctly, then going for a new algorithm makes only sense if we consider that the algorithm seems to be broken for less than 100 calls a feature of this algorithm. ;-) Its probably a better idea to fix the bug, instead of offering two algorithms for our user, a broken and a fixed one.

devopsec:
@miconda updated formatting per https://github.com/kamailio/kamailio/pull/3531#issuecomment-1668288141

@henningw I agree. I think it is a bug that calls are not distributed per the weights, until a minimum of 100 calls are made.

miconda:
For the records, variable length arrays seems to be no longer mandatory to be supported from C11 on, likely the modern compilers support them, though -- here is a good summary comparing with malloc:

  - https://stackoverflow.com/questions/58163105/do-i-really-need-malloc/58163652#58163652

devopsec:
@miconda yes the new code is not using randomized selection for the destination. It is selecting based on probability throughout (recalculate the percentage of calls to that destination each iteration, out of the total calls made). Think of the sorting loop as if it were "processing a new call" per each iteration of `i`. Maybe an explanation based on your above example; weight1=75, weight2=25, would clarify.

First iteration of `i` no calls have been processed, so the call goes to dst1. On the second iteration of `i` dst1 has 1/1 calls, 100% is greater than the intended weight, so we distribute the call to dst2. for `i=3` both dsts have 50% of the calls, but dst1 is intended to have 75% of the calls, so the call goes to dst1. This repeats until 100 "calls" have been processed.

The effect is the exact same in terms of distribution of calls as a percentage, i.e. the intention of the algorithm, but the "bug fix" is that distribution is equivalent at high call volumes and low call volumes. Thus, it is true that the new algorithm does not use random sorting under 100 calls like the last, but I believe that is a bug fix, not a new feature. I think that was a decent explanation, but if not please ask for more clarification.

On the topic of memory allocation, thanks for the reference, I am always happy to learn! In my live tests the compiler happily accepted my code and gdb gave me good output back... I don't tout to be a master of memory management so I will defer to you on whether `pkg_malloc/free()` is required here to be compliant with all the intended C specs.

miconda:
@devopsec: at the end you get more or less to a specific order of selecting the addresses to route to. For example, with two addresses one 99 and the other 1, the one with weight 1 is going to be placed the 2nd (or 1st depending on the initial order). If weights are 90 and 10, it should end up like `1st, 2nd, 1st, 1st, 1st, 1st, 1st, 1st, 1st, 1st, 1st, 2nd, 1st, 1st...`. For calls of same type (e.g., codecs, audio/video) and duration, could bring some benefits, but for calls of random type and duration the random ordering proved to give pretty good load of resources.

With the above remarks, considering that random ordering was done on purpose at the time of implementation, I don't consider this PR a bug fix, but a different distribution algorithm (for example it can be called percentage ordered distribution).

miconda:
Is any work still planned on refactoring to a new algorithm? Maybe this can be closed and a new PR created when it is done.

devopsec:
@miconda i'll open this as a new PR when I can get back around to it, thanks!

PR #3534, opened 2023-08-09
dispatcher: add option to retain existing latency stats when reloading destinations

charlesrchance:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Added new module param to enable retention of call stats for existing destinations when reloading the list from file or database.

Also updated to retain stats for other destinations when adding/removing individual ones via RPC.

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -109,6 +109,7 @@ static int *_ds_ping_active = NULL;
 extern int ds_force_dst;
 extern str ds_event_callback;
 extern int ds_ping_latency_stats;
+extern int ds_retain_latency_stats;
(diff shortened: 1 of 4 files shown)


charlesrchance:
Thanks!

I agree...happy to make it the default if others think the same, but we never quite know how it's already being used and who relies on a reload to actually reset the stats.

miconda:
Fine to merge from my point of view, if no other comments.

henningw:
I have no strong opinion on the question regarding the default value. Merging it, thank you. If somebody thinks that it should be configured differently w/r to the default value, it can be changed later as well.

charlesrchance:
Thanks @henningw!

Issue #2946, opened 2021-11-22
kamailio depends on obsolete pcre3 library


linuxmaniac:
From https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1000134:
```
Source: kamailio
Severity: important
User: matthew-pcredep@debian.org
Usertags: obsolete-pcre3

Dear maintainer,

Your package still depends on the old, obsolete PCRE3[0] libraries
(i.e. libpcre3-dev). This has been end of life for a while now, and
upstream do not intend to fix any further bugs in it. Accordingly, I
would like to remove the pcre3 libraries from Debian, preferably in
time for the release of Bookworm.

The newer PCRE2 library was first released in 2015, and has been in
Debian since stretch. Upstream's documentation for PCRE2 is available
here: https://pcre.org/current/doc/html/

Many large projects that use PCRE have made the switch now (e.g. git,
php); it does involve some work, but we are now at the stage where
PCRE3 should not be used, particularly if it might ever be exposed to
untrusted input.

This mass bug filing was discussed on debian-devel@ in
https://lists.debian.org/debian-devel/2021/11/msg00176.html

Regards,

Matthew [0] Historical reasons mean that old PCRE is packaged as
pcre3 in Debian 
```

miconda:
We've done some bad decisions in naming as well, but naming `pcre2` a newer version of `pcre3` is pretty counterintuitive.

Anyhow, from what I remember, next modules are using pcre: dialplan, lcr and regex.

linuxmaniac:
I forgot about this :/

linuxmaniac:
And now we're out of Debian testing. 

```
Testing: Excuse for kamailio
Migration status for kamailio (- to 5.7.0-1): BLOCKED: Rejected/violates migration policy/introduces a regression
Issues preventing migration:
Removal request by auto-removals
Trying to remove package, not update it
Updating kamailio would introduce bugs in testing: [#1000134](https://bugs.debian.org/1000134)
Additional info:
Piuparts tested OK - https://piuparts.debian.org/sid/source/k/kamailio.html
74 days old (needed 5 days)
Excuses generated on: Tue Aug 1 09:07:08 2023 UTC
```

PR #3535, opened 2023-08-09
regex: migration to pcre2

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [x] Related to issue #2946

#### Description
migrate modules to pcre2

### DIFF
--- a/pkg/kamailio/deb/bionic/control
+++ b/pkg/kamailio/deb/bionic/control
@@ -31,6 +31,7 @@ Build-Depends:
  libmono-2.0-dev [amd64 armel armhf i386 mipsel kfreebsd-amd64 kfreebsd-i386 ppc64 ppc64el s390x],
  libmosquitto-dev,
  libncurses5-dev,
+ libpcre2-dev,
  libpcre3-dev,
  libperl-dev,
  libphonenumber-dev (>= 7),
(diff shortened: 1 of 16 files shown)


sergey-safarov:
I have checkd `pcre2-config` config presence in the CentOS dists.
Looks like utility present in the CentOS 7, 8, 9 but does not exist.

Should you we drop CentOS 6 support?

SirDMZ:
CentOS dropped CentOS 6 support years ago.

Anyone running Kamailio on CentOS 6 is most likely not updating their Kamailio either.
If you pay Redhat for extended support, you've got another year of RHEL6. Those users can run a mildly older branch of Kamailio until that day comes. Kamailio 5.7.x

I am saying this as someone this would affect. Drop us.

linuxmaniac:
@miconda should I try to keep pcre3 and pcre2 at the same time or just drop cpre3? Is is worth the effort?

miconda:
Fine for me to drop it and go now for the simpler way. If later proves to be high demand for the old variant, we can find solutions at that time.

linuxmaniac:
Any regex user willing to check this? It seems to work

miconda:
I don't think there are many users of this particular module. I would merge it and then see if later are reports about it.

PR #3569, opened 2023-09-14
New record-route is not updating when the device network is broken/power-off & on (sca_subscribe.c).

schiramchetty:
After initial subscription by the SCA device, if we power off and on again, we won't get an unsubscribe notification and we are getting a new record-route with sca-subscription. In this case kamailio not updating the new record-route for the SCA Device.

#### Pre-Submission Checklist

- [ *] Commits are split per component (core, individual modules, libs, utils, ...)
- [ *] Each component has a single commit (if not, squash them into one commit)
- [ *] No commits to README files for modules (changes must be done to docbook files

#### Type Of Change
- [* ] Small bug fix (non-breaking change which fixes an issue)

#### Checklist:

- [ *] Tested changes locally

#### Description

When we initially subscribe to the SCA (Shared call appearance) device, we receive a new subscription notification. However, if we subsequently power off and then power on the SCA device, we will no longer receive an unsubscribe notification; instead, we are receiving a new subscription notification with a new record-route. In this scenario, kamailio is not updating the new record-route and sca notification is working after the device is powered on. In two different scenario this issue is occuring(1.Device power off & on 2. Device network broken).

### DIFF
--- a/src/modules/sca/sca_subscribe.c
+++ b/src/modules/sca/sca_subscribe.c
@@ -906,6 +906,10 @@ static int sca_subscription_update_unsafe(sca_mod *scam,
 
 		SCA_STR_COPY(&update_sub->rr, &saved_sub->rr);
 	}
+	else{
+     
+         SCA_STR_COPY(&saved_sub->rr, &update_sub->rr);
+     }
 
 	rc = 1;
 


miconda:
Thanks for the PR!

A few remarks though:

 - the commit has to be for development version (git master branch), not 5.5 branch. It will be then backported if it is a bug.
 - 5.5 series is actually end of maintenance (its last release was out a while ago: https://www.kamailio.org/w/2023/07/kamailio-v5-5-7-released/)
 - you have to format the commit message as per contributing guidelines in order to facilitate merging from the web interface, see:
   - https://github.com/kamailio/kamailio/blob/master/.github/CONTRIBUTING.md#commit-message-rules

In short, you have to do a new PR for the git master branch.

@linuxmaniac: any comment related to the change proposed here? Patch looks pretty minimal.

linuxmaniac:
@miconda I think the fix is not really good since is not taking care of memory and if the previous value is equal. Created a new PR with a possible solution. @schiramchetty please take a look and test it

schiramchetty:
> @miconda I think the fix is not really good since is not taking care of memory and if the previous value is equal. Created a new PR with a possible solution. @schiramchetty please take a look and test it

Hi @linuxmaniac, Thanks for your feedback.
We are not allocating any new memory for the requested PR and we are just updating the existing variable value using 

     SCA_STR_COPY(&saved_sub->rr, &update_sub->rr);
     
Also we have tested with this PR and we are not finding any memory-related issues and all the scenarios are working as expected. Please let us know do we need to create a new PR as proposed by @miconda. or do you need any clarifications on this PR. Please suggest.

linuxmaniac:
Hi, what is going to happen if the length of updated RR is bigger than the reserved memory at saved RR? And in the case of RR being the same value?

schiramchetty:
> Hi, what is going to happen if the length of updated RR is bigger than the reserved memory at saved RR? And in the case of RR being the same value?

Hi @linuxmaniac, The updated RR and saved RR variables are updated by kamailio itself when receiving any sca subscription message. This is already handled by kamailio existing source code, also both variables are using the same "_sca_subscription" structure. We are only copying values between these variables. Hope it will clarify.

linuxmaniac:
@schiramchetty 

when you say you're only copying ... are you assuming that the length is always the same? 

```
#define SCA_STR_COPY(str1, str2)               \
	memcpy((str1)->s, (str2)->s, (str2)->len); \
	(str1)->len = (str2)->len;
```

SCA_STR_COPY doesn't do any magic. You need to check that you have enough memory reserved for that. And on top of that, why should I copy a value that is exactly the same? So before copying we need to check the value is different.

schiramchetty:
> @schiramchetty
> 
> When you say you're only copying ... are you assuming that the length is always the same?
> 
> ```
> #define SCA_STR_COPY(str1, str2)               \
> 	memcpy((str1)->s, (str2)->s, (str2)->len); \
> 	(str1)->len = (str2)->len;
> ```
> 
> SCA_STR_COPY doesn't do any magic. You need to check that you have enough memory reserved for that. And on top of that, why should I copy a value that is exactly the same? So before copying we need to check the value is different.

Hi @linuxmaniac, Please refer to line number 907 in sca_subscribe.c file, The existing code used to copy the RR in the below line  ...

SCA_STR_COPY(&update_sub->rr, &saved_sub->rr);

The same code we are used to update the new RR. Please note that in our scenario the updated RR and saved RR are different. Still, if you feel the code needs validation can you please add the validation code here. So that I will be able to raise the PR with the modified code.

linuxmaniac:
I already created the code at https://github.com/kamailio/kamailio/pull/3571
Please test it

schiramchetty:
> I already created the code at #3571 Please test it

Thanks for the update, We will test this code and will get back to you.

miconda:
Closing this one, further discussions should be done to:

  - https://github.com/kamailio/kamailio/pull/3571

PR #3581, opened 2023-09-20
ims_qos: Add suspend_transaction parameter

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Currently, when making an AAR, transaction is suspended and then continued. This is because one may want to block call upon AAR failure.

In our use case, we don't want to block any calls, regardless of AAR return code. This makes suspend/continue of the transaction an overhead. Moreover, if the PCRF is slow, transaction blocks untill PCRF responds.

Add a module parameter to control if one wants to suspend or not the tm transaction (default is the current behavior). Local basic tests work ok, more testing will follow.

### DIFF
--- a/src/modules/ims_qos/doc/ims_qos_admin.xml
+++ b/src/modules/ims_qos/doc/ims_qos_admin.xml
@@ -517,6 +517,25 @@ modparam("ims_qos", "rr_default_bandwidth", 2000)
         </programlisting>
       </example>
     </section>
+
+    <section>
+      <title><varname>suspend_transaction</varname> integer</title>
+
+      <para>If eneabled, suspends the tm transaction while doing the AAR.</para>
+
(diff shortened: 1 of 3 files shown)


miconda:
From my point of view after a very brief code review, it is fine to merge.

I do have a comment, though. I would suggest to prefix global variables with some value specific to the module, like (`ims_qos_` or `_ims_qos_`, e.g., `_ims_qos_suspend_transaction`) because there might be cases when global variables become exposed across modules or core and when they have rather common words then it can result in duplicates and conflicts. IMS modules might not have such common prefix to global variables yet, but it would be good to start doing it, at least for the newly added variables. This can be done as a follow up commit after merging this PR.

The name of the exposed modparam can stay the same, it is only about the C variable name.

smititelu:
Ok, will push a new commit to rename the C var, directly on master.

PR #3587, opened 2023-10-01
examples/kamailio-basic-kemi-lua.lua: Add example for rtpengine

dilyanpalauzov:
A while ago I wanted to adjust my Lua configuration to work with rtpengine and I have adjusted this example accordingly.  I have not tried the example, but if there are no obvious bugs with it, I think it will be useful.


miconda:
The sample kemi configs were intended to offer the same functionality (the native cfg, lua, python ... files in that folder) to be able to compare and learn easier migration from one scripting language to another one. But probably we can diverge. I am merging it and try to add to the other files soon, if time allows, if not, they are going

PR #3588, opened 2023-10-03
Fix wss forwarding

joelbax:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/core/parser/parse_uri.c
+++ b/src/core/parser/parse_uri.c
@@ -172,8 +172,8 @@ int parse_uri(char *buf, int len, struct sip_uri *uri)
 		/* ws */
 		VW_W,
 		VW_S_FIN,
-        /* wss */
-        VW_SS_FIN
+		/* wss */
+		VW_SS_FIN
 	};
 	register enum states state;
 	char *s;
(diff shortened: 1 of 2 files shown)


miconda:
The rfc7118 (https://www.rfc-editor.org/rfc/rfc7118.html) defines only `transport=ws`, where have you met `transport=wss`?

joelbax:
@miconda  SIP.js is including it in the contact header, then, the register module keeps it. I think this is also happening with pjsip but I haven't had a chance to properly verify it. 
Also, IMHO, I think that this is still a misleading subject because, referring to Iñaki's message in jssip: https://groups.google.com/g/jssip/c/tWhZla-3Ktk, he says that it's like RFC3261 not defining `transport=tls` (that is actually defined in RTC5630) but we use it.
Maybe time to do a RTC5630 equivalent for 7118 and define some uses for wss transport? :)

miconda:
@joelbax: I think RFC3261 mention the deprecation of `transport=tls` for `sips` scheme (ie., end to end secure transmission), but it is defined in the grammar (e.g., for hop-by-hop tls usage). See RFC3261 page 223:

```
transport-param   =  "transport="
                     ( "udp" / "tcp" / "sctp" / "tls"
                     / other-transport)

```

  - https://datatracker.ietf.org/doc/html/rfc3261

joelbax:
@miconda  our bad, SIP.js was configured to use transport=wss. Closing this PR. 

I'm sorry for the inconvenience.

miconda:
OK, though somehow I agree that not having wss in the transport is confusing, because it is in Via, but I wanted to be sure I haven't missed any new specs introducing it.

PR #3594, opened 2023-10-05
kamctl: Source only one kamctlrc

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

kamctl now sources the kamctlrc from only one location if available in this order: 

1. same folder as kamctl,
2. {intstall_prefix}/etc/kamailio/,
3. ~/.kamctlrc
4. /etc/kamailio/. 

If no kamctlrc is found, it continues just like before. Should it stop?

#### Issue description
`kamctl` uses env variables from all the config files above, and retains the env value of the latest config that declares it.

Let's say we have 3 kamctlrc files right now:

```
# first config file (same folder that kamctl resides)
# {prefix_install}/sbin/kamctlrc
...
SIP_DOMAIN=kamailio.org
DBENGINE is commented
...
```

```
# second config file 
# /etc/kamailio/kamctlrc
...
SIP_DOMAIN=kam0.XXXX.net
DBENGINE=MYSQL
...
```

```
# third config file 
# {prefix_install}/etc/kamailio/kamctlrc
...
SIP_DOMAIN is commented
DBENGINE=MYSQL
...
```

Running `{install_prefix}/sbin/kamctl` (before changes in order) it prints :

```
Loading config file /home/xenofon/kamailio-source-install/sbin/kamctlrc
Loading config file /etc/kamailio/kamctlrc
Loading config file /home/xenofon/kamailio-source-install/etc/kamailio//kamctlrc
SIP_DOMAIN env var is kam0.XXXX.net
DBENGINE env var is MYSQL
```
suggesting that env variables from multiple sources are gathered and overwritten in cases.

### DIFF
--- a/utils/kamctl/kamctl
+++ b/utils/kamctl/kamctl
@@ -18,6 +18,7 @@ else
 fi
 
 ### include config files
+RC_FILE_SOURCED="false"
 
 # check for rc file at same location with kamctl
 which greadlink >/dev/null 2>&1
@@ -39,11 +40,13 @@ if [ -n "$KAMCTLFULLPATH" ] ; then
 fi
 
 # check for rc file at standard locations
-if [ -f /etc/kamailio/kamctlrc -a -r /etc/kamailio/kamctlrc ]; then
+if [ $RC_FILE_SOURCED = "false" ] && [ -f /etc/kamailio/kamctlrc -a -r /etc/kamailio/kamctlrc ]; then
 	. /etc/kamailio/kamctlrc
+	RC_FILE_SOURCED="true"
 fi
-if [ -f /usr/local/etc/kamailio/kamctlrc -a -r /usr/local/etc/kamailio/kamctlrc ]; then
+if[ $RC_FILE_SOURCED = "false" ] &&  [ -f /usr/local/etc/kamailio/kamctlrc -a -r /usr/local/etc/kamailio/kamctlrc ]; then
 	. /usr/local/etc/kamailio/kamctlrc
+	RC_FILE_SOURCED="true"
 fi
 if [ -f ~/.kamctlrc -a -r ~/.kamctlrc ]; then
 	. ~/.kamctlrc
@@ -3255,4 +3258,3 @@ case $1 in
 		exit 1
 		;;
 esac
-


miconda:
The scope of loading all configs in cascade is to allow overwriting on specific options with home config, not to have the entire config there. Practically the /etc config is the one with all options and then add a user home config only to change specific options (e.g., have a different mysql user/password in the home config, but all the other options of kamctl to be from /etc/kamailio/kamctlrc).  To my knowledge, sourcing many shell (config) files is common, like bash loading /etc/profile and then ~/.profile.

I am fine to have a new option to load only one file, the first found, but current behaviour has to be kept the default.

henningw:
Many linux software packages uses a different behaviour, if one configuration file is found, its using this and no merge of the values is done. Bash profile loading does it differently indeed, but I would argue that kamctl is more like a tool. But if people depend on this current behaviour, fine.

The main issue that motivated the change is that, if you are executing the Kamailio as non-root user, it will fail to start as it can't read the /etc/kamailio/kamctlrc. This should surely be fixed.

miconda:
This behaviour is likely there from the beginning, or at least more than 10-15 years old. I do use it, although not very often, to put in home kamctlrc some new queries that can be run via db command, relying on etc file for the rest of the options.

I pushed a commit to check if kamctlrc is readable.

henningw:
Thank you

xkaraman:
Hello @miconda,

Indeed this script goes way back and i am still getting familiar with the codebase. You are suggesting that `/etc/kamailio/kamctl` should be and is the base config file and that is fine. 

I see some flows as it right now.

1) What if user has installed kamailio not from official packages but another way in a non standard location. That file is nowhere to be found. Hence, all the other locations will be used to source with order described already (messy i believe). Even `bash` that you have mention, has the base `/etc/profile`  and then **one of** `~/.bash_profile`,` ~/.bash_login`,`~/.profile` sourced according to [man bash](https://linux.die.net/man/1/bash ), `invocation section`.
 
2) The current code has the base of kamctl to be the one residing in the **same folder as kamctl**, ie `{install_prefix}/sbin/kamctlrc`, when `/etc/kamailio/kamctl` should be the base. If right now , I have a config in the same folder that kamctl is (the user config as you say and none of the other two exist, `{install_prefix}/etc/kamilio/kamctlrc` and `~/.kamctlrc`), my options would be overridden by /etc/ one when in fact i placed this to overwrite it. This is due to the search order of kamctl in the script.

Just my observations and thank you for the readable commit!

miconda:
@xkaraman: usually the applications are designed to work with standard installation from a package or source tree. That's what kamctl tries to look after as well. It is somehow expected that when one does own custom installation may have to tune it properly in many places.

I am not against improving things of about how is done now, I said that your PR breaks existing behaviour, which is common to other pieces of linux software, but also used as it is for many years.

Then, likely a typo of yours, but `/etc/kamailio/kamctlrc` instead of `/etc/kamailio/kamctl` is the expected base config file for kamctl, being the one deployed with package-based installation -- just for sake of clarity.

To your point 1. I somehow responded above, at least partially -- now how many files one app looks for and loads is probably specific to it, I wanted to highlight that sourcing many files is not uncommon. If you want to elaborate more I can do it. 

To your 2., can you point where `{install_prefix}/sbin/kamctlrc` is taken as base config?

xkaraman:
> It is somehow expected that when one does own custom installation may have to tune it properly in many places.

 Surely, i understand that. This specific script `kamctl` offers an easy way to accomplish that by searching some standard paths it may find the `kamctlrc`. Maybe we can document, that it searches all the paths and uses all of the config files found and in which order.

> I said that your PR breaks existing behaviour

Yes, unfortunately, this breaks the existing behaviour of users and completely understand that. If we can't find a way to preserve the previous behaviour. We can close this altogether.

> Then, likely a typo of yours, but `/etc/kamailio/kamctlrc` instead of `/etc/kamailio/kamctl` is the expected base config file for kamctl, being the one deployed with package-based installation -- just for sake of clarity.

True that was a type. I meant `/etc/kamailio/kamctlrc`.

> To your 2., can you point where {install_prefix}/sbin/kamctlrc is taken as base config?
 
I based the above finding on these lines. https://github.com/kamailio/kamailio/blob/18b23187dd89e3b36c0dd95d4796484bef3d2e62/utils/kamctl/kamctl#L22C1-L39C3

The first path that is searched and used (sourced) is the `kamctlrc` found in the same folder as `kamctl` that runs (either in the standard location or not).  As you can see in my description of this pr, 
```
# Running {install_prefix}/sbin/kamctl (before changes in order) it prints :

Loading config file /home/xenofon/kamailio-source-install/sbin/kamctlrc
Loading config file /etc/kamailio/kamctlrc
Loading config file /home/xenofon/kamailio-source-install/etc/kamailio//kamctlrc
SIP_DOMAIN env var is kam0.XXXX.net
DBENGINE env var is MYSQL
```

miconda:
I missed that part, but indeed is looking for a kamctlrc in the same folder as kamctl. Digging in memory, the actual purpose for it is to allow running/operating kamailio from source directory (just compile, not install), which is convenient when developing. That's also why loadmodule looks not only for `modulename.so` file in `mpath` folders, but also for `modulename/modulename.so`.

If you are in kamailio sources folder, you can run `./utils/kamct/kamctl` and then, as running in a devel environment, the latest `kamctlrc` of the source tree is expected to be the base config. However, the order can be changed, if most of the people find it  differently to be more natural. Practically the scope it is to have a base config and a way for each user to amend only specific options, not to have full a config again.

Usually is expected to have the default config as deployed by install (in /etc/ or /usr/local/etc) when running a globally deployed kamctl or by presence in source three when running a devel environment, plus user specific options (e.g., user/password for database) in a user home config.

miconda:
As we kind of try to move from shell-based kamctl/kamdbctl to kamcli, I looked at it and it does a similar approach, relying on the fact that Python's `configparser.read()` has support to get a list of (config) files:

  - https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read

Just to give example of a pretty established API for sourcing many files to config.

xkaraman:
> Usually is expected to have the default config as deployed by install (in /etc/ or /usr/local/etc) when running a globally deployed kamctl or by presence in source three when running a devel environment, plus user specific options (e.g., user/password for database) in a user home config

I can understand that. I'm not against that of course. Maybe better documentation of how `kamctl` loads its configuration and order would be the best then. ie [(main documentaion](https://www.kamailio.org/wikidocs/tutorials/getting-started/main/#tools) of `kamctl` does not provide any information other than `kamctl` loads its config from `kamctlrc` found in same folder of `kamailio.cfg`). When first read that, I made the assumption that only one file is loaded, but after checking the source code that is not the case. 

I may also have missed something in other documentation pages, so please point them to me if you happen to have one.

miconda:
I see that the patch introduced a bunch of `echo` lines, which could be good for debugging but can be a pain when using kamctl many times, specially when one wants to divert the output to a file or pipe for further processing (e.g., output of a kamctl rpc command which should be just the jsonrpc output that then can be processed with `jq`).

Either remove them or replace with `mdbg` (an internal function for debug messages).

xkaraman:
Removed them as suggested.

miconda:
@xkaraman: the PR has conflicts with the master branch, which introduced a while ago checks for access rights of the rc files.

Issue #3602, opened 2023-10-11
ims_dialog: Support for db storage


mtryfoss:
### Description

I've been looking at the ims_dialog module and notice that the documentation states that db storage is not yet supported.
According to the source code that only partial true. Not sure if it's residues from the other dialog module, or some unfinished work.

After some initial testing and small adjustments, the main issue seems to that for my use the "Dialog-iD"/did is not filled. I can not see this being set for calls that are not "concurrently confirmed".

The field in the database for this id is NON NULL, so the insert fails. It's also used as WHERE-criteria when restarting and fetching the dialog_out entries, belonging to each dialog_in, so it needs to be set for this to work.

This is not really a feature request. It's more about a confirmation that my assumptions are correct, before trying to do the proposed changes.

### Expected behavior

I do assume that this "did"-field should be or be linked to the same did added as Record-Route parameter, identifying each dialog.

#### Actual observed behavior

kamcmd dlg2.list displays NULL for Dialog-ID.

```
kamcmd dlg2.list
{
	Size: 4096
	Dialogs: {
		Dialog: {
			Entry: 1096
			Id: 8069
			RURI: sip:+4791500025@10.111.64.16;user=phone
			From: sip:+4746180445@10.111.64.16;user=phone
			Call-ID: 34c5d9f15ff42a0564f809cb35e19c75@10.111.64.16:5060
			Caller Contact: sip:+4746180445@10.111.64.16:5060
			Caller Route Set: <null string>
			Dialog-ID: <null string>
			From Tag: as4df55540
			State: Confirmed
			Ref: 2
			dlg_outs: {
				dlg_out: {
					Entry: 2680
					Id: 0
				}
			}
		}
	}
}

```

### Possible Solutions

Always set this did field when a new dialog is created.

I've also been studying the DB schema. All operations for the dialog_in table are done using hash_entry and hash_id as keys. Why are they not used as primary key instead, which would be much more efficient?

Also, having an integer with auto_increment as primary key would make this stop when the integer hits max value, if it's not reset regularly. I can not see this field used anywhere in the code either, so it could probably just be removed.

The same applies to dialog_out, but this one should have an index on the did field used for looking up the corresponding entries too.

Altering this could have negative impact if there are old entries from a previous running process laying around. A duplicate key on an insert should be handled, and updated instead. The current behaviour would mean that Kamailio instead updates two records with the same values.

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
master
```

[...]

PR #3603, opened 2023-10-13
ims_dialog: Add support for database backend

mtryfoss:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3602 

#### Description
Added missing code for database backend to work. Tested basic scenarios locally for all db modes.
DB schema could be tweaked according to my thoughts in issue #3602, but that also applies to several other modules which are done in the same matter. Because of that, I didn't consider such a change now.

### DIFF
--- a/src/modules/ims_dialog/dlg_db_handler.c
+++ b/src/modules/ims_dialog/dlg_db_handler.c
@@ -643,6 +643,13 @@ static int load_dialog_info_from_db(int dlg_hash_size, int fetch_num_rows)
 
 	} while(nr_rows > 0);
 
+	if(dlg_db_mode == DB_MODE_SHUTDOWN) {
+		if(dialog_dbf.delete(dialog_db_handle, 0, 0, 0, 0) < 0) {
+			LM_ERR("failed to clear dialog in table\n");
+			goto error;
+		}
+	}
+
 	if(dlg != NULL) {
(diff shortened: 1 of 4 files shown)


mtryfoss:
Seems like there's some issue with the CodeQL Analyzer (runner image out of disk space)?

miconda:
Thanks! CodeQL can indeed fail from other reasons than the code, can be ignored if that happens.

Regarding the patch of the PR, I could spot only a database delete operation being added. The commit message says `Add support for database backend`. Was there code for database support but not activated? If yes, only the db delete operation was missing?

mtryfoss:
Like I wrote in the ticket related to this task, most of the stuff was there already - but it both caused a crash and was not fully functional. Entries were being written to the database, but was not restored correctly on startup.

I assume the old dialog module was used as a basis and that the code for db operations was not finalized with the new structure.

The missing delete was probably a result of that, since this new module got two tables ("in" and "out") compared to only one in the old module.

mtryfoss:
My plan is to have this merged and then continue working on db operations for the ims_charging module. I just wanted to have this completed first, since the other part will not work without it.

miconda:
Thanks for the details, I am not using the module, I wanted to be sure nothing was omitted -- I am going to merge it.

PR #3608, opened 2023-10-19
Make tcp_check_timer default to depend on tcp_msg_data_timeout and tcp_msg_read_timeout values

juha-h:
If  `tcp_check_timer` is not set, use default value that is half of `tcp_msg_data_timeout` or `tcp_msg_read_timeout` value depending on which one is smaller and not zero.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1729,16 +1729,19 @@ int main_loop(void)
 		if(!tcp_disable) {
 			if(ksr_tcp_check_timer == -1) {
 				if(ksr_tcp_msg_data_timeout > 0 && ksr_tcp_msg_read_timeout > 0)
-					ksr_tcp_check_timer =
-						MIN(ksr_tcp_msg_data_timeout, ksr_tcp_msg_read_timeout) / 2;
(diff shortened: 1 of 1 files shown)


juha-h:
Don't know why `check-format (pull_request)` failed again, because latest commit 38a1c736fa3ebef8827a0f3ef645e7e4fa165b5a... PASSED.

miconda:
I am going to apply it manually, because the commit message is not following the contributing guidelines.

miconda:
I guess clang format may differ based on clang version and maybe global defaults on OSes -- the check here should be taken as a soft recommendation, but if it looks pretty ok, it can be merged.

Anyhow, in this particular case I merged manually, closing this one..

linuxmaniac:
> Don't know why `check-format (pull_request)` failed again, because latest commit [38a1c73](https://github.com/kamailio/kamailio/commit/38a1c736fa3ebef8827a0f3ef645e7e4fa165b5a)... PASSED.

yes, the action checks if *all* commits follows the clang-format not the final result. That is why rebase and squash is a better idea in this cases when the first commit is not well formated.

PR #3609, opened 2023-10-19
http_client: Add parameter connect_timeout_ms / timeout_ms

nchaigne:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description

The new parameter connect_timeout_ms (global) / timeout_ms (in httpcon) allows to specify a timeout in milliseconds on curl requests.
If this parameter is defined (non zero), then the timeout in seconds is ignored.

See https://curl.se/libcurl/c/CURLOPT_TIMEOUT_MS.html

### DIFF
--- a/src/modules/http_client/curlcon.c
+++ b/src/modules/http_client/curlcon.c
@@ -5,9 +5,9 @@
  * Copyright (C) 2008 Elena-Ramona Modroiu (asipto.com)
  *
  * This file is part of kamailio, a free SIP server.
- * 
+ *
  * SPDX-License-Identifier: GPL-2.0-or-later
- * 
+ *
  * Kamailio is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation; either version 2 of the License, or
@@ -69,6 +69,7 @@ typedef struct raw_http_client_conn
 	int verify_host;
 	int tlsversion;
 	int timeout;
+	int timeout_ms;
 	int maxdatasize;
 	int http_follow_redirect;
 	int authmethod;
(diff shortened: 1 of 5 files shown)


oej:
Thank you for contributing. I would prefer if there's only one variable and that we use the milliseconds always internally and just keep the old setting for backwards compatibility.

nchaigne:
> Thank you for contributing. I would prefer if there's only one variable and that we use the milliseconds always internally and just keep the old setting for backwards compatibility.

What do you mean exactly ?
A single parameter "timeout" which could store seconds and milliseconds, e.g. "1.5" ?
Or two parameters "timeout" and "timeout_ms" but only store a value in milliseconds internally (with priority to "timeout_ms" if defined) ?

oej:
I think we need two config options - but they set the same variable internally and only call ONE curl interface - the one you add here. We need also to verify the http_client API if we need to add something there. Having two variables internally is just overhead.

nchaigne:
> I think we need two config options - but they set the same variable internally and only call ONE curl interface - the one you add here. We need also to verify the http_client API if we need to add something there. Having two variables internally is just overhead.

Alright, I'll try something like you propose.

I had thought of this initially, but it seemed too convoluted because the configuration can come from 3 different sources : global parameter, httpcon parameter, or file configuration... so each time we need to "fixup" the actual timeout, depending on whether we have the ms timeout or not.

I'll try a new pull request, and then you can check if you think it is better.
Thanks for the feedback!

nchaigne:
Regarding the http_client API, there is no impact, because it uses the name of a preset http connection.

miconda:
It might add more complexity to try to work on the same variable, as I quickly looked over the commit of the PR, one field is inside a cfg-param reload structure, which is trickier to figure out if it was set because of a modparam or because of a rpc command and then know which parameter name was used to set it. I think it will bring more overhead in code maintenance than two variables (or two fields in the cfg-param reload structure).

If two variables/fields is considered overhead, I would rather suggest to remove the old second-based timeout completely and add the new one with a different name, so on upgrading to the next release, the config has to be updated as well.

nchaigne:
> It might add more complexity to try to work on the same variable, as I quickly looked over the commit of the PR, one field is inside a cfg-param reload structure, which is trickier to figure out if it was set because of a modparam or because of a rpc command and then know which parameter name was used to set it. I think it will bring more overhead in code maintenance than two variables (or two fields in the cfg-param reload structure).
> 
> If two variables/fields is considered overhead, I would rather suggest to remove the old second-based timeout completely and add the new one with a different name, so on upgrading to the next release, the config has to be updated as well.

I'm almost done with the second pull request :(

miconda:
@nchaigne: ok, let's see it.

nchaigne:
> It might add more complexity to try to work on the same variable, as I quickly looked over the commit of the PR, one field is inside a cfg-param reload structure, which is trickier to figure out if it was set because of a modparam or because of a rpc command and then know which parameter name was used to set it. I think it will bring more overhead in code maintenance than two variables (or two fields in the cfg-param reload structure).
> 
> If two variables/fields is considered overhead, I would rather suggest to remove the old second-based timeout completely and add the new one with a different name, so on upgrading to the next release, the config has to be updated as well.

Regarding the RPC commands, I see only "httpclient.listcon" for this module.
I don't think the module can be reconfigured through RPC, or am I missing something ?

nchaigne:
Alright, the new pull request is done:
https://github.com/kamailio/kamailio/pull/3611

I have tested locally with all possible combinations (I hope).
Yes, it's a bit complicated... having just a timeout in ms would have been much simpler. But I wanted to avoid breaking existing configurations.

Anyway... let me know what you prefer to do!

miconda:
Closing this one, considering the related discussion in the other PR.

PR #3628, opened 2023-11-03
kamcmd: don't clash with ENV NAME or ctl.so module

Ansuel:
NAME variable might be set to the current HOSTNAME in some shell and also clash with the value set by the module calling MOD_INSTALL_UTILS by passing a NAME variable.

With commit 1774cee62098 ("kamcmd: allow defining the name of the application from command line") this resulted in the kamcmd bin being renamed to all kind of name from hostname to ctl.so.

Fix this by checking the variable to a more safe name that is not already defined in shell by default and also that doesn't clash with module variables, use UTIL_NAME as an alternative to NAME.

UTIL_NAME now needs to be used to create kamcmd with custom name.

Fixes: 1774cee62098 ("kamcmd: allow defining the name of the application from command line")

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

### DIFF
--- a/utils/kamcmd/Makefile
+++ b/utils/kamcmd/Makefile
@@ -8,10 +8,15 @@ include $(COREPATH)/Makefile.targets
 auto_gen=
 RELEASE=1.5
 UTIL_SRC_NAME=kamcmd
+# Pass CUSTOM_NAME to overwrite the kamcmd/sercmd bin name
+ifeq ($(CUSTOM_NAME),)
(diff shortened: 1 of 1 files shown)


miconda:
Am I misunderstanding the patch or is it actually `CUSTOM_NAME` to be used for defining the application name instead of `UTIL_NAME`, which is mentioned in the commit comments?

PR #3629, opened 2023-11-06
tm: T_ASYNC_SUSPENDED flag not removed when cancelling a suspension

mtryfoss:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
I've been testing the ims_charging module which relies on a diameter request and async reply. If there's no available diameter server, Ro_CCR() calls t_cancel_suspend() to cancel suspension of the transaction and return an error.

Based on my understanding, it should now be possible to continue script logic as if the transaction was not suspended at all.

In our case we want the call to go through anyway, as a fallback.
The INVITE is being sent to the destination, but all replies are filtered since the suspended flag is still set.

Tested this change locally, and it's now working according to our needs.


miconda:
Thanks!

miconda:
Thanks!

PR #3633, opened 2023-11-09
cdp: Support for diameter routing agent (DRA) / relaying

mtryfoss:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
If an endpoint responds with a CEA (Capabilities-Exchange Answer) indicating support for relaying (application-id 0xffffffff), let this endpoint be used for any application. The DRA will then be responsible for routing the request to the correct destination.

To fully support this, there might be a need for additional changes in some of the modules. That is to make sure subsequent requests for a given session is always routed to the same destination, indicated by Origin-Host in answers from a diameter server. Destination-Host of the new request should then be filled with this value.

A change for the ims_charging module is on the way.

### DIFF
--- a/src/modules/cdp/diameter.h
+++ b/src/modules/cdp/diameter.h
@@ -87,6 +87,8 @@
 #define to_32x_len(_len_) ((_len_) + (((_len_)&3) ? 4 - ((_len_)&3) : 0))
 
 
+#define RELAY_APP_ID 0xffffffff
+
 /* AAA TYPES */
 
 #define AAA_NO_VENDOR_ID 0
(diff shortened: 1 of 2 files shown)


miconda:
Thanks! The patch is rather minimal, so I am merging it. Any updates that one may find useful can be made as a new PR.

mtryfoss:
Thanks for merging!

Sorry for not clarifying.
Additional changes in the ims_charging module will be with a separate PR, containing several improvements for better compatibility against other charging server vendors.

PR #3639, opened 2023-11-13
topos: Add db version check

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Add a database version check and fail gracefully if it does not match.

### DIFF
--- a/src/modules/topos/doc/topos_admin.xml
+++ b/src/modules/topos/doc/topos_admin.xml
@@ -600,6 +600,23 @@ modparam("topos", "methods_noinitial", "OPTIONS,NOTIFY")
 </programlisting>
 		</example>
 	</section>
+	<section id="topos.p.version_table">
+		<title><varname>version_table</varname> (integer)</title>
+		<para>
+		If set to 0, the module will skip checking the version of the tables.
+		</para>
+		<para>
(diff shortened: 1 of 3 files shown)


miconda:
If I haven't overlooked something, it seems to be missing closing db connection on mod_init, which results in inheriting it on children after fork and conflicts during runtime.

Maybe it is better to move it in child_init with rank 1, we discussed a bit about that during K Devs meeting.

Furthermore, I recommend to add a module parameter for being able to disable this check, some other modules do it:

  - https://www.kamailio.org/docs/modules/stable/modules/auth_db.html#auth_db.p.version_table
  
It was working fine without it for many releases and it should keep a way for same behaviour like in the past. Disabling version table check is useful during upgrades, when new database table structure includes the old one, in case of needing to revert the upgrade for a while, running the old version is fine.

xkaraman:
Updated to close the db connection and added a module parameter for being able to disable the check.

miconda:
You also need to add documentation for the new module parameter in the xml file from doc/ subfolder.

To avoid globals conflict, add `_tps_` prefix to the new global variable in the C code, to be similar to the other variables that are made modparms by topos.

miconda:
Thanks, merging!

PR #3640, opened 2023-11-14
ims_charging: Various improvements to make module compatible to other charging server vendors

mtryfoss:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
**Use origin-host of CCR response as destination-host of subsequent CCR requests**
When connecting to a diameter node which using relay, it may load balance requests to multiple servers for handling. This change ensures that subsequent requests are sent to the same server.

**Honour AVP_Time_Quota_Threshold in CCR response**
Charging server indicates it want an interim update X number of seconds before all quota has been used.

**Add new modparam «strip_plus_from_e164»**
The existing formatitng routine check if it's a tel: or sip: uri, and sets E.164 or SIP according to that. If it's E.164, it strips "tel:" too. A subscription id with leading "+" is actually of type SIP.

I didn't want to make a breaking change here, so I introduced a new setting which fixes the behaviour.

**Moved subscription id formatting into a common function**
For more clean code, and one place to make further adjustments.

**Handle CCR response without AVP_Validity_Time**
If missing, various places in the module reported immediate expire.

**Handle IMS_RAR (Re-auth) request from charging server**
Typical scenario is that there's a missing interim update, and the server want to check if the session is still alive. At reception we try a lookup and if found, schedule a new request.

**Handle IMS_ASR (Abort) request from charging server**
Similar handling as for re-auth, but for server-initiated termination of the call.

**Fixed issue with signed/unsigned int related to check for Expires header**
cscf_get_expires_hdr() returns signed (-1 for not found), while that value was added
to the outgoing diameter request unsigned. That again resulted in an Expires AVP with
a really large value, and was rejected by the charging server.

Most probably cscf_get_expires_hdr() should be changed to return unsigned since the value theoretical can be that big, but since that is a common function for the ims modules it would require more testing than I'm capable of right now.

**Use str_dup() and str_free() from ims_getters instead of locally defined**
Those were needed more places.

-- 

I'm a little bit unsure about updating a string allocated in shm, where the length is changed (look in ims_ro.c, credit_control_session_callback()). Is the correct way to do a shm_free() and then reallocate, or is there a more simple way of doing it?

### DIFF
--- a/src/modules/cdp/acctstatemachine.c
+++ b/src/modules/cdp/acctstatemachine.c
@@ -176,6 +176,8 @@ int cc_acc_client_stateful_sm_process(
 					//nothing we can do here, we have sent callback, client needs to send CCR Update
 					LM_DBG("Reservation close to expiring\n");
 					break;
+				case ACC_CC_EV_SESSION_MODIFIED:
+					break;
 				default:
 					LM_ERR("Received unknown event [%d] in state [%d]\n", event,
 							x->state);
--- a/src/modules/cdp/diameter_code_avp.h
+++ b/src/modules/cdp/diameter_code_avp.h
@@ -266,6 +266,8 @@ typedef enum
 	AVP_Service_Selection = 493, //RFC5778
 	AVP_Call_Id = 494,
 
+	AVP_Time_Quota_Threshold = 868, // 3GPP TS 32.299
+
 } AAA_AVPCodeNr;
 
 enum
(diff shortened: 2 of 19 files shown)


mtryfoss:
I notice check-format fails, but that doesn't seem to be from this PR?

henningw:
Yes, I don't know why it fails for older commits. If its not related to changes you've done, it can be ignored.

henningw:
Thanks for the PR. Regarding your questions about the shared memory management, there is a function to re-alloc an existing memory segment for SHM and PKG pool [link](https://www.asipto.com/pub/kamailio-devel-guide/#c04shm_realloc)

mtryfoss:
Thanks for the tip! Updated with shm_realloc() now. Also rebased with master, and the check seems to be successful.

I'm currently also investigating a deadlock issue (#3641), but it doesn't seem to be related to changes in this PR. Any feedback there would be helpful too.

miconda:
I assume this can be merged soon if no other comments.

Issue #3641, opened 2023-11-14
Deadlock in ims_charging / cdp modules on connection issues


mtryfoss:
### Description

I've been simulating calls while there's (faked) issues against the diameter server using TCP.
Ro_CCR() uses t_suspend() to suspend the SIP transaction while waiting for a response from the charging server.
A cdp transaction is also created with a callback to resume_on_initial_ccr() in the ims charging module.

When a response to a charging request is received, this is handled in cdp/api_process.c where cdp_take_trans() find the given transaction and returns it without a lock held. The callback is then executed, and everything is fine.

The problem occurs when a CCR times out, and the callback is done by the cdp_trans_timer() function. This hold a lock for the entire transaction list while executing callbacks for every timed out entry. In my case, the charging module does t_continue and the execution of that routing block is started again.

Here we reply a cause indicating it was an issue with the charging server. This will trigger a new callback from the ims_dialog module which again will try to start a new cdp transaction (CCR Stop) - still while the cdp_trans_timer function holds a lock.

Since the cdp timers now are blocked, no new connection attempts to the failed diameter server is done either.

#### Reproduction

```
Ro_CCR("RO_CONTINUE_DISPATCHER", "orig", 30, "", "")
this times out, and will do continue in the routing block specified (based on the cdp transaction timeout timer)

route[RO_CONTINUE_DISPATCHER] {
send_reply("500", "Internal error");
exit;
}
```

### Debugging Data

cdp_add_trans() on line 2 is blocked while waiting for lock held by cdp_trans_timer() on line 26 to be freed. The line numbers are a bit off because of some debugging.

```
#0  0x00007f6d9a3e0e29 in syscall () from /lib64/libc.so.6
#1  0x00007f6d979622f6 in futex_get (lock=0x7f6d706622a0) at ../../core/mem/../futexlock.h:121
#2  0x00007f6d979632cd in cdp_add_trans (msg=0x7f6d70921280, cb=0x7f6d9084caf5 <resume_on_termination_ccr>, ptr=0x0, timeout=5, auto_drop=1) at transaction.c:126
#3  0x00007f6d97980740 in AAASendMessage (message=0x7f6d70921280, callback_f=0x7f6d9084caf5 <resume_on_termination_ccr>, callback_param=0x0) at diameter_comm.c:158
[...]

PR #3643, opened 2023-11-15
cdp: Avoid deadlock for callback on diameter timeout

mtryfoss:
Callback function might do an operation which creates a new transaction, which will result in a deadlock.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3641

#### Description
Problem in detail described in ticket.

Summary is that the lock used to iterate expired transactions is also needed when creating a new transaction.
If the callback function (directly or indirectly) do something that creates a new transaction, it will result in a deadlock.

In my case, this was the ims_charging module where a start charging request times out. The callback does t_continue() where the routing logic then rejects the call. This will create a stop request to the charging server, and it will block.

### DIFF
--- a/src/modules/cdp/transaction.c
+++ b/src/modules/cdp/transaction.c
@@ -206,6 +206,15 @@ void cdp_free_trans(cdp_trans_t *x)
 int cdp_trans_timer(time_t now, void *ptr)
 {
 	cdp_trans_t *x, *n;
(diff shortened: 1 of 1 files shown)


miconda:
I am not familiar with the code of the module for a proper review, but if no other comments, the PR could be merged soon.

Issue #3570, opened 2023-09-14
ims_ipsec_pcscf: add AoR parameter to ipsec_destroy()


jbipre2:
### Description
Have two 4G UE's registered (over Open5GS CN / Amarisoft eNB): UE1 (OnePlus) connects to P-CSCF over UDP, UE2 (Huawei) connects using IPSec SA.
Immediately after registration, VoLTE calls work from UE1 to UE2 and vice-versa; but after some time, only from UE2 to UE1. Calls initiated from UE1 fail with a "network error" message.

Following Kamailio logfile extract shows periodic OPTIONS processing failing (cause unknown to me) leading to de-registration:
```
Sep 14 00:56:07 corsa03 p-cscf[8138]: INFO: <script>: OPTIONS to sip:67435055-f201-4c46-a31f-10b7f30c9574@10.46.0.11:5060 via sip:10.46.0.11:5060;transport=tcp...
Sep 14 00:56:07 corsa03 p-cscf[8138]: INFO: <script>: OPTIONS to sip:001010000051194@10.46.0.6:31907 via sip:10.46.0.6:31907;transport=tcp...
Sep 14 00:57:07 corsa03 p-cscf[8138]: INFO: <script>: OPTIONS to sip:67435055-f201-4c46-a31f-10b7f30c9574@10.46.0.11:5060 via sip:10.46.0.11:5060;transport=tcp...
Sep 14 00:57:07 corsa03 p-cscf[8138]: INFO: <script>: OPTIONS to sip:001010000051194@10.46.0.6:31907 via sip:10.46.0.6:31907;transport=tcp...
Sep 14 00:58:01 corsa03 p-cscf[8124]: INFO: cdp [authstatemachine.c:270]: auth_client_statefull_sm_process(): after callback of event 1
Sep 14 00:58:01 corsa03 p-cscf[8134]: INFO: cdp [authstatemachine.c:270]: auth_client_statefull_sm_process(): after callback of event 17
Sep 14 00:58:01 corsa03 p-cscf[8134]: INFO: cdp [authstatemachine.c:425]: auth_client_statefull_sm_process(): state machine: AUTH_EV_RECV_STA about to clean up
Sep 14 00:58:07 corsa03 p-cscf[8138]: INFO: <script>: OPTIONS to sip:67435055-f201-4c46-a31f-10b7f30c9574@10.46.0.11:5060 via sip:10.46.0.11:5060;transport=tcp...
[...]

PR #3645, opened 2023-11-17
Add AoR argument to ims_ipsec_pcscf ipsec_destroy()

jbipre2:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [X] Related to issue #3570

#### Description

Added optional AoR argument to ipsec_destroy(); to permit insertion of a real AoR from a cfg script. Otherwise, ipsec_destroy() uses the dummy AoR you@kamailio.org. The dummy AoR does not identify the IPSec SA of any UE present in the registrar, so no IPSec SA can be torn down.
The real AoR, if present as an argument to ipsec_destroy() replaces the dummy AoR in m->first_line.u.request.uri.s, before calling fill_contact() (see changes to cmd.c / cmd.h).
All other changes address ims_ipsec_pcscf_mod.c only, as required to implement the ipsec_destroy() optional argument.

### DIFF
--- a/src/modules/ims_ipsec_pcscf/cmd.c
+++ b/src/modules/ims_ipsec_pcscf/cmd.c
@@ -1041,7 +1041,7 @@ cleanup:
 }
 
 
-int ipsec_destroy(struct sip_msg *m, udomain_t *d)
+int ipsec_destroy(struct sip_msg *m, udomain_t *d, str *uri)
 {
 	struct pcontact_info ci;
 	pcontact_t *pcontact = NULL;
@@ -1051,6 +1051,12 @@ int ipsec_destroy(struct sip_msg *m, udomain_t *d)
 	if(m->first_line.type == SIP_REPLY) {
 		t = tmb.t_gett();
 	}
+
+        // Insert URI in SIP message
+        if(uri != NULL) {
+                m->first_line.u.request.uri.s = uri->s;
+                m->first_line.u.request.uri.len = uri->len;
+        }
 
 	// Find the contact
 	if(fill_contact(&ci, m, t, 0) != 0) {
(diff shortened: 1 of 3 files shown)


miconda:
Thanks, I will merge it manually because the commit message does not follow the format from contributing guidelines, use it for next commits that you want to push in order to be able to do the merging from PR portal:

miconda:
Merged manually. However, as I did the merge, I noticed:

  - the docs were not updated to reflect the new parameter for ipsec_destroy() - you have to update the xml files inside `doc/` subfolder when adding/changing module parameters or functions. This time I did it, you can review to see if you want to add more there
  - the change of msg structure ruri is not safe by just changing the string pointer and the length, you have to use C function `rewrite_uri(...)`  from the core because old value might need to be freed and r-uri-parsed structure invalidated. But such change then it is permanet, if you need it to stay like that, then use `$ru = ...;` in config, functions like `ipsec_destroy()` must not change it permanently (at least not without documenting such behaviour). Therefore I pushed another commit to the module which is no longer replacing msg structure r-uri field, but uses the config parameter to build the contact info.

I also changed the fixup code to use an existing helper function from core, which simplifies that part.

You have to test with latest git master branch and see if it works as expected, if not, then open an issue.

jbipre2:
Thank you for taking the time to correct my input and provide this guidance. I hope to make better contributions in future.
I knew my “solution” was a quick hack; but as I couldn’t find out (quickly) where the dummy message was instantiated, I couldn’t work out what its side-effects might be.
What matters is that the ipsec_destroy() function will now actually tear down an IPSec SA, because none were being removed before. Output from “ip xfrm state list” would indicate
perhaps 10 SA’s while only one ipsec-3gpp capable UE was active (and unreachable in a VoLTE call).

PR #3649, opened 2023-11-21
core: timer_proc don't execute timers on shutdown phase

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

If kamailio is in the shutdown phase proc timers should not be executed since memory objects are getting destroyed and cores could happen

### DIFF
--- a/src/core/timer_proc.c
+++ b/src/core/timer_proc.c
@@ -29,6 +29,7 @@
 #include "pt.h"
 #include "ut.h"
 #include "mem/shm_mem.h"
+#include "sr_module.h"
 
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


miconda:
I think `ksr_shutdown_phase()` has to be used in this case, because the timers are different processes, `destroy_modules_phase()` is for main/parent process only.

linuxmaniac:
Nothing explodes in my tests

linuxmaniac:
@miconda do you think this can be merged?

miconda:
Yes, thanks!

PR #3650, opened 2023-11-21
pv: Add support to $hfl $hflc for Diversion header 

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
This PR adds support for $hfl(Diversion)[] and $hflc(Diversion). 

This parse_diversion.c was inspired by [parse_pai_ppi.c](https://github.com/kamailio/kamailio/blob/master/src/core/parser/parse_ppi_pai.c). I am not sure if it's breaking the existing implementation and usage. I believe it does not. 

During some experimentation with the hfl(name)[-1], I encountered weird behavior with negative indexing for the already supported headers like Via, Route, and Contact.  -1 yields null and -2 yields the last one and so on. Separate PR can be introduced if necessary.

Any feedback is welcome.

### DIFF
--- a/src/core/parser/parse_diversion.c
+++ b/src/core/parser/parse_diversion.c
@@ -31,6 +31,7 @@
 #include "../dprint.h"
 #include "../ut.h"
 #include "../mem/mem.h"
+#include "parse_diversion.h"
 #include "parse_from.h"
 #include "parse_to.h"
 #include "msg_parser.h"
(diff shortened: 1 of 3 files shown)


miconda:
Just adding the note, no review was actually done: changing the structure of a parsed header can impact cloning of the sip msg in transaction and how it is freed.

Then, on a quick look, it seems the patch for hfl has also conditions on p-asserted-identity and p-preferred-identity header types, not only on diversion, as the commit message says.

xkaraman:
Yeah something went wrong on my branches and the formatting and messed up. I will create another pull request for both (Diversion and PPI).

Regarding the changes on the header parsing, of course, there must be a review since it's probably used in other places as well.  I only checked for hfl and hflc usage.

PR #3671, opened 2023-12-11
utils: Prefer `printf` over `echo -e`.

seanbright:
On Ubuntu, `/bin/sh` defaults to `dash` whose built-in `echo` command does not support the `-e` flag. `printf`, on the other hand, supports backslash escapes by default so we use that instead.

In passing, also clean up some messy ANSI escape codes.


seanbright:
Here's a before-and-after. This is on a default install of Ubuntu 22.04:

image

linuxmaniac:
Thanks

Issue #1968, opened 2019-05-30
dmq_usrloc not replicating contact attributes


pwakano:
### Description

The dmq_usrloc module is not replicating the extra attributes saved in the xavp for the contact. So remote nodes don't receive these values, being unable to access it properly.
It is expected that the attributes are replicated as well.

#### Reproduction

Setup two nodes using the dmq_usrloc and usrloc in db_mode=0.
When receiving a REGISTER, before saving the contact add some attribute in the xavp_contact. In a subsequent message to this contact (e.g.: INVITE) use a lookup() or registered() and try to access the desired xavp_contact attribute. If the request is received in the local node (which received the original REGISTER), the attribute will be valid, but if the request is received in the remote node, the attribute is empty.

#### Log Messages

```
May 29 16:06:18 kamailio-2 /usr/sbin/kamailio[23230]: DEBUG: registrar [save.c:410]: pack_ci(): generated ruid is: uloc-5cee2114-5abe-2
May 29 16:06:18 kamailio-2 /usr/sbin/kamailio[23230]: DEBUG: usrloc [ucontact.c:73]: ucontact_xavp_store(): trying to clone per contact xavps
May 29 16:06:18 kamailio-2 /usr/sbin/kamailio[23230]: DEBUG: <core> [core/xavp.c:697]: xavp_clone_level_nodata(): cloned root xavp [ulattrs]
May 29 16:06:18 kamailio-2 /usr/sbin/kamailio[23230]: DEBUG: <core> [core/xavp.c:721]: xavp_clone_level_nodata(): cloned inner xavp [cluster_node]
May 29 16:06:18 kamailio-2 /usr/sbin/kamailio[23230]: DEBUG: usrloc [ucontact.c:1701]: update_ucontact(): exists callback for type= UL_CONTACT_UPDATE
May 29 16:06:18 kamailio-2 /usr/sbin/kamailio[23230]: DEBUG: usrloc [ul_callback.h:84]: run_ul_callbacks(): contact=0x7f51b624d418, callback type 2/15, id 1 entered
May 29 16:06:18 kamailio-2 /usr/sbin/kamailio[23230]: DEBUG: dmq_usrloc [usrloc_sync.c:776]: dmq_ul_cb_contact(): Callback from usrloc with type=2
[...]

PR #3679, opened 2023-12-18
dmq_usrloc: DMQ also copies and sends attributes related to contacts

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue https://github.com/kamailio/kamailio/issues/1968. 

#### Description

This PR fixes the issue in https://github.com/kamailio/kamailio/issues/1968. 

Any feedback is welcome.

### DIFF
--- a/src/modules/dmq_usrloc/usrloc_sync.c
+++ b/src/modules/dmq_usrloc/usrloc_sync.c
@@ -51,6 +51,10 @@ int usrloc_dmq_send_multi_contact(
 		ucontact_t *ptr, str aor, int action, dmq_node_t *node);
 void usrloc_dmq_send_multi_contact_flush(dmq_node_t *node);
 
+static int traverse_avp_tree(
+		sr_xavp_t *avp_list, srjson_doc_t *jdoc, srjson_t *root);
+static void srjson_to_xavp(srjson_t *json, sr_xavp_t **xavp);
+
 #define MAX_AOR_LEN 256
 
 extern int _dmq_usrloc_sync;
@@ -340,6 +344,8 @@ static int usrloc_dmq_execute_action(srjson_t *jdoc_action, dmq_node_t *node)
 	static str host = STR_NULL, c = STR_NULL, callid = STR_NULL,
 			   path = STR_NULL, user_agent = STR_NULL;
 
+	sr_xavp_t *xavp = NULL;
+
 	action = expires = cseq = flags = cflags = q = last_modified = methods =
 			reg_id = server_id = port = proto = 0;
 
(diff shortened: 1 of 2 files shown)


src/modules/dmq_usrloc/usrloc_sync.c
```diff
@@ -825,6 +837,16 @@ int usrloc_dmq_send_contact(
 	srjson_AddNumberToObject(&jdoc, jdoc.root, "reg_id", ptr->reg_id);
 	srjson_AddNumberToObject(&jdoc, jdoc.root, "server_id", ptr->server_id);
 
+	/*  Loop through Χavp attributes of the contact and and create a json object */
+	srjson_t *jdoc_xavp = srjson_CreateObject(&jdoc);
```
miconda, line 841: Variable should be declared at the beginning of the block to be coherent and also keep compliant with older C standards.

miconda:
The commit for usrloc #3d79568 does not only change the destroy order, is it also fixing something?

src/modules/dmq_usrloc/usrloc_sync.c
xkaraman: Fixed on some other variables too on the newly static functions.

xkaraman:
Hey Daniel,
 
Besides the order change of destroy, it copies to the new contact the `xavp` from incoming `ucontact_t` argument just like the rest of the fields. 

If the config variable `xavp_contact` is found and not empty, then `ucontact_xavp_store` destroy this xavp and fills it up with whatever is contained inside. 

Some doubts i had, is whether i shall `c->xavp = _ci->xavp;` or `	_c->xavp = xavp_clone_level_nodata(_ci->xavp);` or `xavp_add(_ci->xavp, &c->xavp)`. What is the difference between all these methods?

henningw:
Are there more comments or objections? If not I would merge it in the next days.

miconda:
I haven't gotten the time to review if the list of XAVPs can be just linked or needs to be cloned. If it is just used temporarily in a function, then linking should be fine, but if it needs to be kept for long, then it could be that clone has to be done, otherwise the list stays linked from two places and attempted to be destroyed twice.

If the above is clarified, I am fine to merge it.

henningw:
Thanks, yes this was basically the question also @xkaraman had earlier. The XAVPs will be used in the cfg script context later on I think. He will do another review.

xkaraman:
Hi @miconda,

What does long mean in this case? What does `xavp_clone_level_nodata` achieve in this regard?

Thanks!

henningw:
The PR was reviewed and discussed together with Xenofon, some comments have been added and it was pushed in the referenced commits. Close this PR.

PR #3687, opened 2023-12-22
tm: Reset local_response_sent_lookup

### LINKED ISSUE #3064: callout from event_route [tm:local-response] only occurs once
version: kamailio 5.5.2 (arm6/linux) 55e232
raspberry pi 3b

I initially stated that I was running on 5.5.3 but I was mistaken.  However I checked the 5.5.3 source and it appears to be the same.

  I think that there is a bug in the tm module with respect to the "tm:local-response" event route.  In the "t_reply.c" file, there is a static variable called '_tm_local_response_set_lookup'.  This variable is initialized at load time to zero.  It is checked in the '_reply_light()' routine and will initiate a local callback from the config script if "armed".  The problem as I see it is that if the callback is readied, the variable is set to one.  But it is never reset.

So the observed behavior is as follows.  A REGISTER is received and the request_route arms the callback.  The REGISTER requires an authorization (local database sqlite).  After the 401 is sent back, the callback via the  event route is called as expected.  All good except that the event route script fragment is never executed again after the first call...  ever, even if it's a new REGISTER request.
I would think that somewhere in the tm module, the variable should be reset to zero so that subsequent transactions can initiate the event route again.  Or maybe the variable ought to live somewhere in the transaction cell???

. . .
    t_on_reply ("MY_FRAG");
    t_on_failure ("MY_FRAG");
. . .
event_route [tm:local-response] {
    xlog ("L_NOTICE", IN tm:local-response\n");
    my_function();
}
. . .

So 'my_function()' is only called once.

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3064  (replace XXXX with an open issue number)

#### Description

This PR fixes #3064. 

It resets the check variable back to 0 after sending a tm:local-response reply.

As of currently, event_route[tm:local-response] will be called only max `children` times (one for each) and then won't process any more.

### DIFF
--- a/src/modules/tm/t_reply.c
+++ b/src/modules/tm/t_reply.c
@@ -618,6 +618,7 @@ static int _reply_light(struct cell *trans, char *buf, unsigned int len,
 					}
 					set_route_type(backup_rt);
 					p_onsend = 0;
+					_tm_local_response_sent_lookup = 0;
 
 					free_sip_msg(&pmsg);
 				} else {


miconda:
@xkaraman: thanks for looking into #3064 and proposing a PR!

As I looked at the commit of this PR and reviewed the code, the purpose was to avoid doing lookup of the event route every time, but do it only once per process. The commit in this PR resets the global variable so lookup is done again and should make a fix for #3064, but still doing the lookup every time.

I pushed an alternative commit to master branch that does the caching of the first lookup in the global variable, so it is going to be done only once, avoiding the search in the hash table of event route blocks each time.

Should commit 1f5444fd2e25d227ffd70f1087d057a3b8002558 (now in master) fail to fix #3064, then we can look further into it.

This PR should be closed if you think the above referenced commit fixes it.

xkaraman:
Hey @miconda,

Great, now it makes sense on the usage of that check too. :)

I think new commit https://github.com/kamailio/kamailio/commit/1f5444fd2e25d227ffd70f1087d057a3b8002558 works as expected and i will close the PR. 

Thanks

Issue #3653, opened 2023-11-22
pv: hfl(name)[N] for negative index bug


xkaraman:
### Description

I am working on extending $hfl and $hflc to support some more headers. 

During some experimentation, I stumbled into some weird behavior regarding $(hfl(Route)[-1]) and negative indexing of $hfl pseudo-variable (also for Via, Contact).

Per the [docs ](https://www.kamailio.org/wikidocs/cookbooks/devel/pseudovariables/#hflname-header-field-with-list-of-bodies ): -1 index should print the last header found and so on (as stated by $hdr $hfl).

Right now, -1 produces null, and from -2 and onwards it prints correctly from the last one to the first one.

Contacts have also the weird behavior of having the contacts from last to first for positive indexing and vice-versa. Is this expected?

#### Reproduction
```
#kamailio.cfg
...
request_route {
    ...
    xlog("L_INFO", "Route headers hdr: $hdr(Route)\n");
    xlog("L_INFO", "Route headers hfl: $hfl(Route)\n");
    xlog("L_INFO", "Route headers count: $hflc(Route)\n");
    xlog("L_INFO", "Route headers 0: $(hfl(Route)[0])\n");
    xlog("L_INFO", "Route headers 1: $(hfl(Route)[1])\n");
    xlog("L_INFO", "Route headers 4: $(hfl(Route)[4])\n");
    xlog("L_INFO", "Route headers -1: $(hfl(Route)[-1])\n");
    xlog("L_INFO", "Route headers -2: $(hfl(Route)[-2])\n");
    xlog("L_INFO", "Route headers -3: $(hfl(Route)[-3])\n");
    ...
}
```

Seng a SIP message containing Route (but also applicable for VIA, Contact headers):
```
OPTIONS_MSG_2CONTACTS="OPTIONS sip:example.com SIP/2.0
Via: SIP/2.0/UDP 192.168.1.100:5060;branch=z9hG4bK123456789
Max-Forwards: 70
From: <sip:bob@example.com>;tag=123456789
To: <sip:alice@example.com>
Call-ID: 987654321@192.168.1.100
CSeq: 1 OPTIONS
Contact: <sip:bob0@example.com>;expires=3600,<sip:bob1@example.com>;expires=3600,<sip:bob2@example.com>;expires=3600
Route: <sip:server1.example.com;lr;ftag=1234>, <sip:server2.example.com;lr;ftag=5678>, <sip:server3.example.com;lr;ftag=9012>
[...]

PR #3697, opened 2024-01-04
pv_core: Fix negative index bug for hfl()

xkaraman:
Before: -1 yielded null and -2 the last element of a header

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3653

#### Description

This PR fixes a bug referenced in #3653.

Before fix `$(hfl(HEADER)[-1]` where `HEADER='Via' | Contact | Route | Record-Route ` produced `null` instead of the last element of header.


miconda:
Thanks!

I thought it was based on the same approach as for `$(hdr(name)[index])`, but that works fine.

PR #3718, opened 2024-01-18
core: Fix check to also work for KEMI framework

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR fixes an issue regarding KEMI framework for onsend_route.

KSR.set_drop() in [ksr_onsend_route](https://www.kamailio.org/wikidocs/cookbooks/5.7.x/core/#onsend_route) does not make kamailio dropping the message when using kemi (python3) (probably holds true for other engines too):

This is exactly the example from [documentation](https://kamailio.org/docs/tutorials/devel/kamailio-kemi-framework/kemimods/#ksrxdrop):
```
def ksr_onsend_route(self, msg):
  KSR.set_drop()
  exit()
```

The issue arises because we are not setting `ret = 0` when the drop flag is set, but instead check if it's not set to return `ret=1` (The case for native that returns `0` but no `DROP_R_F` was set).

**Note**:
KEMI Case: `ret = 1` always See [app_python3/apy_kemi.c](https://github.com/kamailio/kamailio/blob/79697b4d565665b696ee39021e27d82e716bc176/src/modules/app_python3/apy_kemi.c#L88-L92)
Native case: `ret` can be `0` for drop, or `1` for success

### DIFF
--- a/src/core/onsend.c
+++ b/src/core/onsend.c
@@ -87,7 +87,18 @@ int run_onsend(sip_msg_t *orig_msg, dest_info_t *dst, char *buf, int len)
 			orig_msg->fwd_send_flags = fwd_snd_flags_bak;
 			orig_msg->rpl_send_flags = rpl_snd_flags_bak;
 			exec_post_script_cb(orig_msg, ONSEND_CB_TYPE);
-			if((ret == 0) && !(ra_ctx.run_flags & DROP_R_F)) {
+			/*
+			 * KEMI case (python,lua,jsdt)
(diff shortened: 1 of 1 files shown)


miconda:
Thanks, I will try to review a bit deeper soon to try to see if there are side effects, not to impact the native scripting.

miconda:
I have just merged it manually to change a bit the commit message and comments.

Issue #3720, opened 2024-01-19
ASYNC functions in event_route[xhttp:request] don't work


vsaranov:
### Description
When calling the `async_route()` from `event_route[xhttp:request]`, an error occurs.

```
ERROR: tm [t_lookup.c:769]: t_lookup_request(): too few headers
ERROR: async [async_sleep.c:238]: async_sleep(): cannot create the transaction
```

#### Reproduction
The problem is reproducible. Below is a fragment of the configuration file.
...
```
listen=tcp:192.168.3.115:8080
loadmodule "xhttp.so"
loadmodule "async.so"
```
...

```
event_route[xhttp:request] {
        async_route("TEST", "2");
    exit;
}

route[TEST] {
        xhttp_reply("200", "OK", "text/html", "<html><body>URL $hu</body></html>\r\n");
        exit;
}
```

Making a HTTP-request:

```
curl -X POST http://192.168.3.115:8080/test \
  -H "Content-Type: application/json" \
  -d '{"test": 1}'
```

#### Log Messages

[messages.zip](https://github.com/kamailio/kamailio/files/13994226/messages.zip)

### Additional Information

```
#kamailio -v
version: kamailio 5.7.3 (x86_64/linux) e4d96b
flags: USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, MEM_JOIN_FREE, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLOCKLIST, HAVE_RESOLV_RES, TLS_PTHREAD_MUTEX_SHARED
ADAPTIVE_WAIT_LOOPS 1024, MAX_RECV_BUFFER_SIZE 262144, MAX_URI_SIZE 1024, BUF_SIZE 65535, DEFAULT PKG_SIZE 8MB
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
id: e4d96b
compiled on 12:59:36 Nov 17 2023 with gcc 4.8.5
```

* **Operating System**:

```
#cat /etc/redhat-release
CentOS Linux release 7.9.2009 (Core)

```

[...]

PR #3723, opened 2024-01-22
xhttp: fix execution of async functions

vsaranov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3720

#### Description
Fixing RURI and adding necessary headers when creating fake_msg in xhttp_mod.c for successful message validation and the ability to perform various kamailio functions from event_route[xhttp:request].

### DIFF
--- a/src/modules/xhttp/xhttp_mod.c
+++ b/src/modules/xhttp/xhttp_mod.c
@@ -58,6 +58,12 @@ static int fixup_xhttp_reply(void **param, int param_no);
 
 static int pv_get_huri(struct sip_msg *msg, pv_param_t *param, pv_value_t *res);
 
+#define FAKE_RURI "sip:127.0.0.1:9;uri="
+#define FAKE_RURI_LEN (sizeof(FAKE_RURI) - 1)
+#define FAKE_URI "<sip:127.0.0.1>"
+#define FAKE_HDR "From: " FAKE_URI "\r\nTo: " FAKE_URI "\r\nCall-ID: 1\r\nCSeq: 1 POST\r\n"
+#define FAKE_HDR_LEN (sizeof(FAKE_HDR) - 1)
+
 static int xhttp_route_no = DEFAULT_RT;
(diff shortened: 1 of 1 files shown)


miconda:
First, this is not a fix, the http traffic was never supposed to be fully handled as sip traffic. Kamailio is not a full HTTP server/proxy. The workaround for creating a fake sip message was done for local processing, mainly to be able to use authentication and authorization functions as well as config variables.

The async framework is purely designed for received SIP traffic, creating a SIP transaction that stays in memory, runs various timer callbacks and sip-specs specific tasks.

You patch creates an internal sip message that does not even follow the strict requirements for a sip requests, like tags, via branch, ... CSeq has an hardcoded method even there can be different ones even in http. Also it breaks the $hu. These are just a few remarks upon a quick look. This PR is not appropriate to merge.

If you need to do some async processing in Kamailio, you have to do it with sip traffic. You can generate from cli with sipexer or sipsak tools.

vsaranov:
You are right, there are many ways to solve the problem. The presence of an HTTP server in kamailio allows you to implement microservices and I suggested expanding its use. I don’t think I can convince you otherwise, but kamailio has long become a multifunctional tool necessary for processing SIP calls. Our company was faced with the need for this functionality to build a system for testing, I think that this improvement would allow other clients to use Kamailio more efficiently. I think that issuing $hu is easy to fix, since the original URI value is not lost :)
Thank you.

miconda:
Even adding the http url value as parameter to r-uri is not safe, because its value can have special characters that have different meaning.

If you need to do something async in xhttp, then you have to add something specific for http requests, knowing the complexity of processing for SIP transaction, leveraging its suspend()/continue() does not feel the right way.

Note also that a new module was added, microhttpd, that is supposed to handle better http traffic than the old custom xhttp. It is at the beginning, not many features, but I would recommend building on top of it if you want support also for modern http protocol, not only http 1.x.

Issue #3667, opened 2023-12-07
Redefine Advertised socket URI when used haproxy protocol


sergey-safarov:
### Description
When used haproxy protocol then from client perspetive TCP connection established to loadbalancer IP address.
Real Kamailio listener may be IPv6 and expected to see `$RAut` value with IP of load balancer IP.
Really I can see for first message from the client
```
sip:2600::6:45060;transport=tcp
```
For second message
```
Advertised socket URI: sip:[2600::6]:5060
```
For first case missed `[]` in the IP and expected to see loadbalancer IPv4 because client use IPv4 to establish connection to Kamailio.
 
### Expected behavior
Used loadbalancerserver IP from haproxy protocol for `$RAut` value and for `Record-Route` headers generation.

miconda:
I don't really understand what you describe and what you want. The development version has the option to set also the protocol for advertised address in listen parameter, no longer taking it from local socket address. So you can even have `listen=udp.1.2.3.4:5060 advertise tls:3.4.5.6:8000`, you just have to take care that the network layer does the proper transmission.

I think this is covering everything needs in terms of listen socket and advertised address.

If you don't get what you need with master branch, then open a new issue where you put the relevant snippets of config, with some logging from config with xlog() to show what is printed for variables and what you expect, plus pcap file with corresponding traffic.

PR #3731, opened 2024-01-28
extented haproxy protocol support

sergey-safarov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #3667

#### Description
Allow creating Record-Route and Via header using destination address and port in the haproxy protocol header.
 
PR created behalf @ivanuschak

### DIFF
--- a/src/core/msg_translator.c
+++ b/src/core/msg_translator.c
@@ -111,6 +111,7 @@
 
 extern char version[];
 extern int version_len;
+extern int ksr_tcp_accept_haproxy;
 
 str _ksr_xavp_via_params = STR_NULL;
 str _ksr_xavp_via_fields = STR_NULL;
@@ -558,6 +559,9 @@ static inline int lumps_len(
 	int recv_proto_id = PROTO_NONE;
 	int recv_af = 0;
 	struct socket_info *send_sock;
+	char ip_buf[MAX_URI_SIZE] = {0};
+	str _recv_address_str = STR_NULL;
+	str _recv_port_str = STR_NULL;
 
 
 #ifdef USE_COMP
@@ -828,7 +832,15 @@ static inline int lumps_len(
 			send_proto_id = send_sock->proto;
 	}
 	/* init recv_address_str, recv_port_str & recv_port_no */
-	if(msg->rcv.bind_address) {
+	if(ksr_tcp_accept_haproxy && msg->rcv.proto_reserved2) {
+		ip_addr2sbuf(&msg->rcv.dst_ip, ip_buf, MAX_URI_SIZE);
+		recv_address_str = &_recv_address_str;
+		recv_address_str->s = ip_buf;
+		recv_address_str->len = strlen(ip_buf);
+		recv_port_str = &_recv_port_str;
+		recv_port_str->s = int2str(msg->rcv.dst_port, &recv_port_str->len);
+		recv_port_no = msg->rcv.dst_port;
+	} else if(msg->rcv.bind_address) {
 		if(msg->rcv.bind_address->useinfo.name.len > 0) {
 			recv_address_str = &(msg->rcv.bind_address->useinfo.name);
 			recv_af = msg->rcv.bind_address->useinfo.af;
@@ -959,6 +971,9 @@ void process_lumps(struct sip_msg *msg, struct lump *lumps, char *new_buf,
 	int recv_proto_id = PROTO_NONE;
 	int recv_af = 0;
 	struct socket_info *send_sock;
+	char ip_buf[MAX_URI_SIZE] = {0};
+	str _recv_address_str = STR_NULL;
+	str _recv_port_str = STR_NULL;
 
 #ifdef USE_COMP
 #define RCVCOMP_PARAM_ADD                                             \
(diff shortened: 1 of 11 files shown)


sergey-safarov:
We work to clean warning
```
CC (gcc) [kamailio]    core/daemonize.o
CC (gcc) [kamailio]    core/msg_translator.o
core/msg_translator.c: In function ‘lumps_len’:
core/msg_translator.c:558:6: warning: variable ‘recv_port_no’ set but not used [-Wunused-but-set-variable]
  int recv_port_no = 0;
      ^~~~~~~~~~~~
core/msg_translator.c: In function ‘process_lumps’:
core/msg_translator.c:974:6: warning: variable ‘recv_port_no’ set but not used [-Wunused-but-set-variable]
  int recv_port_no = 0;
      ^~~~~~~~~~~~
CC (gcc) [kamailio]    core/fmsg.o
```

src/core/forward.h
```diff
@@ -204,6 +204,10 @@ static inline int msg_send_buffer(
 			goto error;
 		}
 
+		if(!dst->id) {
+			dst->id = con->id;
+		}
+
```
miconda, line 210: Any more details about why this is needed? I cannot easy relate it to the commit message.

src/core/msg_translator.c
```diff
@@ -828,7 +829,22 @@ static inline int lumps_len(
 			send_proto_id = send_sock->proto;
 	}
 	/* init recv_address_str, recv_port_str & recv_port_no */
-	if(msg->rcv.bind_address) {
+	if (ksr_tcp_accept_haproxy && msg->rcv.proto_reserved2) {
+		if (msg->rcv.bind_address->proto == PROTO_TLS && msg->rcv.bind_address != NULL && msg->rcv.bind_address->useinfo.address_str.len > 0) {
+			recv_address_str = &msg->rcv.bind_address->useinfo.address_str;
```
miconda, line 834: It should be formatted with clang-format in order to match the coding style.

src/core/msg_translator.c
```diff
@@ -2845,6 +2876,47 @@ int branch_builder(unsigned int hash_index,
 	return size;
 }
 
+static int is_haproxy(struct dest_info *send_info, struct receive_info *haproxy_rcv) {
```
miconda, line 2879: Can you add some comment to describe the purpose of the function? The name is rather generic, from C code point of view it looks like it tries to figure out if the target via via haproxy...

src/core/parser/msg_parser.h
```diff
@@ -396,6 +396,7 @@ typedef struct sip_msg
 	char *unparsed; /*!< here we stopped parsing*/
 
 	struct receive_info rcv; /*!< source & dest ip, ports, proto a.s.o*/
+	struct receive_info haproxy_rcv; /*!< source & dest ip, ports, proto a.s.o for the message from haproxy*/
```
miconda, line 399: Traffic via haproxy is not very common, does it make sense to add a full value-structure for every message instead of a pointer that is allocated only when it is the case. Practically the size of message increases for common traffic over direct UDP/TCP/TLS/SCTP/WebSocket. On the other hand, cloning in shm has to be done if allocated. This is just to ponder on what would be better...

src/core/tcp_read.c
```diff
@@ -1843,13 +1843,17 @@ inline static int handle_io(struct fd_map *fm, short events, int idx)
...
 			tcpconn_listadd(tcp_conn_lst, con, c_next, c_prev);
-			t = get_ticks_raw();
-			con->timeout = t + S_TO_TICKS(TCP_CHILD_TIMEOUT);
-			/* re-activate the timer */
-			con->timer.f = tcpconn_read_timeout;
-			local_timer_reinit(&con->timer);
-			local_timer_add(&tcp_reader_ltimer, &con->timer,
-					S_TO_TICKS(TCP_CHILD_TIMEOUT), t);
+			if(con->rcv.proto_reserved2 && con->type == PROTO_WSS) {
+				//not setting up timers for haproxy WSS connections
```
miconda, line 1847: The patch is quite consistent to important parts of the core. It has to be reviewed properly, so far I did a quick walk through, other developers are welcome to jump in.

src/core/forward.h
ivanuschak: This change is needed for siptrace. Below in this file at the line 334 `dst` is passed as an argument to the `SREV_NET_DATA_SENT` related function call.

If the tcp connection ID is assigned to `dst` param then the tcp_connection search is correct and we have a proper siptrace with HAPROXY IP addresses (IP address of client and IP address of HAPROXY) , otherwise tcp connection search is based on real IP addresses (IP addresses of haproxy and IP address of kamailio) and in this case the siptrace contains not the correct IP-addresses.

src/core/parser/msg_parser.h
ivanuschak: Thanks for highlighting that.
We will check what would be less expensive in terms of resource usage - to have a full value-structure, or to have a pointer that needs to be cloned in shm.

src/core/tcp_read.c
ivanuschak: Please pay attention at this particular change - from line 1846 to line 1856.

The purpose of this change is to keep the WSS connection connected. When testing this haproxy change without this particular piece of change (from line 1846 to line 1856) we observed that kamailio was dropping WSS connection by timer, and this drop looked strange, because it happened in about 5-7 seconds after the connection was established. One of the requirement for haproxy was to keep TLS/WSS connection established during all the dialog and not to drop it without any important reason. I tried to find how to make it not to be dropped by timer, but found the only way is to not to set the timer up for WSS. Also the drop cause is still not cleat for me.
I don't think this is a good solution, but I was not able to find a better one. I would appreciate if you check this change attentively. If needed I can provide with the logs where kamailio drops WSS connection.

linuxmaniac:
I was thinking. Could we have something generic that abstracts the rcv parameter of the message?

henningw:
@sergey-safarov do you like to continue to work on this? I think there have been already several discussions about the PR.

sergey-safarov:
Yes, we prefer to merge this.
@ivanuschak continues to work on this to allow dynamic memory allocation, but this has not been implemented.

PR #3734, opened 2024-01-30
db_mysql: enable TLS when building with mariadb-connector-c

space88man:
#### Pre-Submission Checklist
- [X]  Commits are split per component (core, individual modules, libs, utils, ...)

#### Type Of Change

#### Checklist:
- [X] Tested changes locally
- [X] Related to issue #3735

#### Description
When db_mysql is built with mariadb-connector-c it does not use TLS and fails if the server requires TLS. It seems to be related to the fact that the MYSQL object needs to have a minimal non-NULL configuration (unlike MySQL Connector/C).

This is a minimal setting to allow TLS. A more full-featured solution would be to enable db_mysql to read from an external configuration file.

@linuxmaniac, kindly take a look

UPDATE: the flag is MYSQL_OPT_SSL_ENFORCE which uses TLS if server advertises it. It does not make TLS mandatory (i..e doesn't fail if server has no TLS)

UPDATE: reuse module param `opt_ssl_mode` to configure MariaDB Connector/C.

Use existing `opt_ssl_mode` and map as follows.:

1. 0(off), 1(SSL_MODE_DISABLED): mariadb-connector-c defaults
2. 2(SSL_MODE_PREFERRED)  3(SSL_MODE_REQUIRED) 4(SSL_MODE_VERIFY_CA): `MYSQL_OPT_SSL_ENFORCE`
3. 5(SSL_MODE_VERIFY_IDENTITY): `MYSQL_OPT_SSL_VERIFY_SERVER_CERT`

### DIFF
--- a/src/modules/db_mysql/doc/db_mysql_admin.xml
+++ b/src/modules/db_mysql/doc/db_mysql_admin.xml
@@ -46,6 +46,7 @@
 			<listitem>
 			<para>
 				<emphasis>mysql</emphasis> - the development libraries for the MySQL database. In some Linux distributions named "libmysqlclient-dev".
+				<emphasis>MariaDB</emphasis> - the development libraries for the MariaDB database. In some Linux distributions named "libmariadbclient-dev".
 			</para>
 			</listitem>
 			</itemizedlist>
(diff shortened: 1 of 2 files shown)


linuxmaniac:
But then, is the connection still possible without TLS or are we forcing that the server would support TLS?

space88man:
> But then, is the connection still possible without TLS or are we forcing that the server would support TLS?

Yes connection is possible if the server doesn't enforce TLS —  the connection will use plain TCP.

linuxmaniac:
Much better now @space88man . Thanks for taking care.

PR #3736, opened 2024-01-30
core: infra to run functions in threads

space88man:

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
Introduce utility functions that run module functions in a run-once thread —- primary use case is to init libssl in a thread. Also to avoid a lot of duplicate pthread_XXXX code.

Does not affect core as it is just a header file, however it could be put into core or a utility object.

This infra is applied to a few modules to see the code savings.

### DIFF
--- a/src/core/rthreads.h
+++ b/src/core/rthreads.h
@@ -1,0 +1,99 @@
+/*
+ * Copyright (C) 2024 Chan Shih-Ping
+ *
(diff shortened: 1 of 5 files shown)


henningw:
Thanks for the PR. I just had a quick look, please add a licence header to the new core file (BSD, similar to other core parts).

Issue #3738, opened 2024-02-01
5.7 database modules: init libssl in a thread


space88man:
### Description

OpenSSL integration update was back-ported to 5.7 fixing tls and outbound.

The commonly used modules db_mysql, db_unixodbc, db_postgres - fixed on master - did not receive thread-local fixes in time for 5.7.4.

Kamailio configurations with tls +(database module) may see shm errors

Some database modules are observed to initialise OpenSSL even if  the actual db connections are in clear.

### Troubleshooting

- run the main kamailio process under gdb.
- set breakpoint on CRYPTO_THREAD_set_local (an OpenSSL internal thread-local function)
- observe if any calls happen in Thread#1—this is the failure case

### Workaround 

When combining tls + database module switch to tls_wolfssl instead

Note: master branch has thread-local fixes for the db modules

Issue #3737, opened 2024-01-31
TLS Module Regression (ee key too small)


devopsec:
### Description

The TLS module seems to have some regression from 5.7.3 to 5.7.4 causing `tls.reload` to fail loading certificates.

#### System Info

- kamailio version: 5.7.4 (from official kamailio repos)
- distro version: debian 12
- OS/kernel version: Linux ip-172-31-30-183 6.1.0-17-cloud-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.69-1 (2023-12-30) x86_64 GNU/Linux
- openssl version: 3.0.11 (from official debian repos)

#### Reproduction

On a fresh install of debian 12:
- install openssl from debian repos
- install kamailio 5.7.4 from kamailio repos
- install kamailio-tls-modules from kamailio repos
- create a self signed cert with 4096 bit rsa key
- create a basic tls.cfg to load those files on the client/server default domain
- reload kamailio
- try running `kamcmd tls.reload`

#### Debugging Data

Example tls.cfg

```
[server:default]
method = TLSv1.2+
verify_certificate = yes
require_certificate = yes
private_key = /etc/dsiprouter/certs/dsiprouter-key.pem
certificate = /etc/dsiprouter/certs/dsiprouter-cert.pem
ca_list = /etc/dsiprouter/certs/ca-list.pem
ca_path = /etc/dsiprouter/certs/ca

[client:default]
method = TLSv1.2+
verify_certificate = yes
require_certificate = yes
private_key = /etc/dsiprouter/certs/dsiprouter-key.pem
certificate = /etc/dsiprouter/certs/dsiprouter-cert.pem
ca_list = /etc/dsiprouter/certs/ca-list.pem
ca_path = /etc/dsiprouter/certs/ca
```

Example tls cert:

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            7a:43:b8:fa:df:c9:ed:a7:d6:ab:bb:9a:89:c0:8e:95:fd:62:de:26
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: C = US, ST = MI, L = Detroit, O = dSIPRouter, CN = ec2-34-224-90-100.compute-1.amazonaws.com
        Validity
            Not Before: Jan 30 15:28:11 2024 GMT
            Not After : Jan 29 15:28:11 2025 GMT
[...]

Issue #3727, opened 2024-01-26
TLS: Recent fixes for for OpenSSL 3.X are causing consistent startup failures when unixodbc is used


kdanderso:
### Description

Kamailio fails to start after recent changes were made to improve support for OpenSSL 3.X. The failure occurs for both OpenSSL 3.X and OpenSSL 1.1.X, and are the same as failures I had been seeing on previous Kamailio versions when using OpenSSL 3.X. These errors were not occurring for OpenSSL 1.1.X prior to commits on Jan 4, 2024.

My configuration is using the unixodbc module to communicate with a Microsoft SQL Server database. Furthermore, the dispatcher module is configured to use the database.

With this configuration, Kamailio always fails at startup. It may log a "freeing already free pointer bug" or sometimes indicates pointer corruption. In either case, it fails to connect to the database.

#### Reproduction

Start Kamailio with configuration files similar to the following. For the log file shown below, Kamailio was started with "kamailio -dd -E --atexit=no"

[kamailio.txt](https://github.com/kamailio/kamailio/files/14068346/kamailio.txt)

[tls.txt](https://github.com/kamailio/kamailio/files/14068347/tls.txt)

#### Log Messages

[kamailio.log](https://github.com/kamailio/kamailio/files/14068278/kamailio.log)

#### SIP Traffic

None. Problem occurs during initialization.

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

version: kamailio 5.8.0-dev2 (x86_64/linux) 4db5d7-dirty
flags: USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, MEM_JOIN_FREE, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLOCKLIST, HAVE_RESOLV_RES, TLS_PTHREAD_MUTEX_SHARED
[...]

Issue #3115, opened 2022-05-18
[bullseye] Database TLS connection through cnf file: TLS error followed by memory corruption messages


liviu-rolo:
### Description
Using kamailio 5.5.4 from the official Docker image (created like `create_dockerfile.sh bullseye 5.5.4`, but with no ENTRYPOINT).
Doing TLS listening like:
```
socket_workers=2
tls_max_connections=20000
listen=tls:192.168.0.11:5061
enable_tls=yes
dns_tls_pref=50
tls_port_no=5061
loadmodule "tls.so"
```
It is the second loaded module after tcpops.
Have to connect to the database through cnf files for the SSL support:
`modparam("sqlops","sqlcon","euler=>mysql://[euler]/euclid")`
Relevant section in the cnf file:
```
[euler]
port = 3306
host = 127.0.0.1
protocol = tcp
user = gauss
password = fermat
default-character-set = utf8
ssl-ca = /SSL/riemann/ca-cert.pem
ssl-cert = /SSL/riemann/client-cert.pem
ssl-key = /SSL/riemann/client-key.pem
#ssl-verify-server-cert
```

When starting kamailio:
```
 8(18) ERROR: db_mysql [km_my_con.c:163]: db_mysql_new_connection(): driver error: SSL connection error: PEM lib
 8(18) ERROR: <core> [db.c:319]: db_do_init2(): could not add connection to the pool
 8(18) ERROR: sqlops [sql_api.c:165]: sql_connect(): failed to connect to the database [euclid]
 8(18) ERROR: <core> [core/sr_module.c:874]: init_mod_child(): error while initializing module sqlops (/home/euler/lib64/euler/modules/sqlops.so) (idx: 8 rank: -2 desc: [Async Task Worker - default])
 8(18) ERROR: <core> [core/pt.c:338]: fork_process(): init_child failed for process 8, pid 18, "Async Task Worker - default"
 8(18) ERROR: <core> [main.c:1850]: main_loop(): error in init_child
12(22) CRITICAL: <core> [core/mem/q_malloc.c:501]: qm_free(): BUG: bad pointer 0x7fafa23e6270 (out of memory block!) called from tls: tls_init.c: ser_free(323) - ignoring
```

Then kamailio just hangs on in an incomplete state:

```
/ # ps auxww --forest
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
[...]

PR #3740, opened 2024-02-01
5.7 db modules: init libssl in a thread

space88man:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [X] Related to issue #3738 (replace XXXX with an open issue number)

#### Description
5.7.4 has OpenSSL fixes for tls and outbound but missed out the db modules—the most common users of libssl

This PR updates db_mysql, db_unixodbc, db_postgres to init libssl in a thread to handle more configurations (tls + db module)

This PR syncs all the OpenSSL thread-local fixes from master to 5.7

### DIFF
--- a/src/core/rthreads.h
+++ b/src/core/rthreads.h
@@ -1,0 +1,100 @@
+/*
+ * Copyright (C) 2024 Chan Shih-Ping
+ *
+ * This file is part of Kamailio, a free SIP server.
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+/*
+ * A set of helpers to run functions in threads.
+ *
+ * This is not a thread pool implementation -
+ * - it runs functions in a run-once thread to avoid
+ * creating thread-locals in the calling thread.
+ *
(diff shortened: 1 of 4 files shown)


henningw:
Thanks for the PR. It adds a new core implementation, which are usually not introduced in stable releases. As most people are not using databases with SSL, I am not sure if we really need to introduce this in the stable branch. We already got some regressions (e.g. #3737) in the TLS module in the stable branch from this type of changes. Lets see what other developers think about it.

miconda:
Unfortunately adapting to work properly with changes brought by libssl3+ was not easy, but it might not be a simpler way for avoiding those random crashes. I really appreciate the time and effort @space88man spent on it.

This backported PR is not doing any fundamental change, it just makes common wrappers that can be used from the modules directly, instead of having those wrappers in each module. Those functions are not used in core, just placed there to be available everywhere. For consistency with devel version (upcoming 5.8 as well) , I think it should be merged.

henningw:
I of course also appreciate the work @space88man put here into that. Maybe I am wrong, but right now the db_* modules in stable branches are not using the pthread approach. My question was about if its really necessary to apply this pthread logic to the DB modules. If the db_* modules are not using TLS, I would expect no problems, or I am wrong?

miconda:
Well, that's the problem actually, because the db module can use tls if, for example, connection to MySQL/MariaDB is done over TLS. That was reported for db_unixodbc when connecting to Microsoft SQL Server via TLS. Any other module that eventually ends up using libssl3 though chain of dependencies can have similar impact. Also in the past there were reports that can be related:

  - https://github.com/kamailio/kamailio/issues/3115
  - https://github.com/kamailio/kamailio/issues/2560

Solutions were found at the times for respective versions, but with libssl3 this might be the right approach, or at least the one we could explore now.

Anyhow, I expected those patches to be backported to have the fix completed, right now at least the db_unixodbc is reported to be broken.

space88man:
> I of course also appreciate the work @space88man put here into that. Maybe I am wrong, but right now the db_* modules in stable branches are not using the pthread approach. My question was about if its really necessary to apply this pthread logic to the DB modules. If the db_* modules are not using TLS, I would expect no problems, or I am wrong?
When troubleshooting #3727 the behaviour of 
the MySQL and MariaDB ODBC connectors is that they will initialise libssl/libcrypto even if not using TLS subsequently.  However this will cascading effects on `tls.so` and other modules that use TLS since the thread-locals will be init'ed in process_no = 0.

I.e. the database modules, while not using SSL themselves, can contaminate the thread-locals of main process and the workers will be affected.

space88man:
Thank you  for your comments — I will proceed with these commits as they are necessary for client TLS (any module using libssl3) coexistence, i.e., even if not using kamailio as TLS server, the various client modules can contaminate the main thread-local state.

The commits are non-intrusive — they run the db_XXXX_init functions in a thread.

I have validated: 5.7.3/5.7.4 with tls.so and client SSL will fail intermittently using libssl1; using libssl3 the config will not startup (race condition GH #3727).

With these patches and on master GH #3727 (tls.so  + dispatcher/unixodbc+SSL) now works cleanly (no more shm double free errors).

(Unfortunately the libssl3 db patches did not make it into 5.7.4 so this is to catch-up with the fixes on master.)

PR #3741, opened 2024-02-01
New module/file_out: Module to log custom strings to file

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This pull request is intended to introduce a new, straightforward module that can efficiently stream output to files while also handling changes based on a specific interval. The module contains only one function that streams a chunk of text to the current output file handle associated with the given file index.

It's not 100% completed but the basic features are there. Any reviews and of course naming of the module, much appreciated.

---

How the module works:

- Set parameter `base_folder` to define where you want your logs to be saved.
- Set `base_filename` (up to 10) to define the names of your files. Each is associated with an index according to the order you defined it. First `base_filename`'s index is 0, second's index is 1 and so on. This is appended with a new timestamp when `interval_seconds` are passed.
- Set `interval_seconds` to define how much time before the file is closed and a new one is opened. Right now shared for all files.
- Set `extension` to define what should be the extension of files.  Right now shared for all files. 

Use the provided function `file_out` to write any text (including pvs) to a specific file you want.

---
**Example:** 

Define parameters such as:
```
# kamailio.cfg
...
loadmodule "file_out.so"
modparam("file_out", "base_folder", "/tmp/kamailio/file_out/")
modparam("file_out", "base_filename", "accounting")
modparam("file_out", "base_filename", "missed_calls")
modparam("file_out","interval_seconds", 600)
modparam("file_out","extension", ".txt")
...

request_route {
...
	file_out("0", "Writing  to accounting.txt file $rm from $fu (IP:$si:$sp)");
	file_out("1", "Writing  to missed_calls.txt file $rm from $fu (IP:$si:$sp)");
...
}
```

### DIFF
--- a/src/modules/file_out/Makefile
+++ b/src/modules/file_out/Makefile
@@ -1,0 +1,8 @@
+# WARNING: do not run this directly, it should be run by the main Makefile
+
+include ../../Makefile.defs
+auto_gen=
+NAME=file_out.so
+LIBS=
+
+include ../../Makefile.modules
--- a/src/modules/file_out/doc/Makefile
+++ b/src/modules/file_out/doc/Makefile
@@ -1,0 +1,4 @@
+docs = file_out.xml
+
+docbook_dir = ../../../../doc/docbook
+include $(docbook_dir)/Makefile.module
--- a/src/modules/file_out/doc/file_out.xml
+++ b/src/modules/file_out/doc/file_out.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding='ISO-8859-1'?>
+<!DOCTYPE book PUBLIC "-//OASIS//DTD DocBook XML V4.4//EN"
+"http://www.oasis-open.org/docbook/xml/4.4/docbookx.dtd" [
+
+<!-- Include general documentation entities -->
+<!ENTITY % docentities SYSTEM "../../../../doc/docbook/entities.xml">
+%docentities;
+
+]>
+
+<book xmlns:xi="http://www.w3.org/2001/XInclude">
+	<bookinfo>
+		<title>file_out Module</title>
+		<productname class="trade">&kamailioname;</productname>
+		<authorgroup>
+			<author>
+				<firstname>Xenofon</firstname>
+				<surname>Karamanos</surname>
+				<affiliation>
+					<orgname>GILAWA Ltd</orgname>
+				</affiliation>
+				<email>xk@gilawa.com</email>
+				<address>
+					<otheraddr>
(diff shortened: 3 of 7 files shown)


henningw:
@miconda any suggestions regarding the name? Maybe going for the existing "log_" prefix, like log_file, log_file_out or similar?

miconda:
I have no suggesting about a better name right now, I can think about. Probably `log_` prefix is not suitable, because it does not divert all the log messages like log_custom.

Otherwise upon a quick look, some files are missing the copyright at the top. And I would suggest to add prefix (rename) to the structures used by the module, because Node, Queue ... are too common names.

ovidiusas:
How about `f_log` or `file_log` instead of `file_out`?
Also, the first parameter of `file_out` function seems to be an index to the defined filenames. If the order of the `file_out` modparams is changed, then the script has to be updated. It would be more intuitive to use the actual filename instead of the index.

xkaraman:
@miconda Thanks for the feedback, fixed.

henningw:
Thanks for the feedback as well @ovidiusas. The name file_log sounds also good. Regarding the use of an actual file name, its a good idea and can be probably done without performance impact with a fixup function. We will look into that. 

I would merge the module manually in some hours (to do the quick rename), and then the further changes can be done directly in the repository.

miconda:
I would not recommend using the word `log` in the name. It has nothing to do wit the internal logging frawork. The module is purely writings text to a file. The core module has basic similar functionality, those functions are not related to logging and this module should also keep away from bringing any confusion with the prefix or suffix. The current name is better for the purpose of the module.

henningw:
Ok, keep it like this then, no special preference from our side. :-)

ovidiusas:
How about `f_print` or `file_print`?
In the end, it doesn't really matter, but `file_out` doesn't seem to be very intuitive ... or maybe that's just me :)
Thanks for contributing this, it is a nice addition!

miconda:
Adding a few more details, previous comment was done on mobile, typing there on the web form is not the easiest.

I think that log_file would be a candidate for the name of a module that writes all the logs in a file, similar to log_custom that sends all the log to an udp target. Then having a module named log_file and another one file_log would looks rather confusing.

Renaming can still be done till the branch 5.8 is created, even if it is past the freezing time. First part of testing phase is to tune accordingly the new features to be more suitable for long term maintenance.

Names that came meanwhile in my mind would be: fqwrite (file-queued-write); fqout (fout being rather commonly used to name output file variables, adding q for queuing). This just ideas if renaming is still considered, of course other variant can be used.

PR #3745, opened 2024-02-06
DRAFT: new global config var to constrain thread wrapping

space88man:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
Add global config var

```
enable_tls_thread=1 #0 | 1 | 2
```

RFC: @miconda @henningw

### DIFF
--- a/src/core/rthreads.h
+++ b/src/core/rthreads.h
@@ -27,6 +27,7 @@
  */
 #include <pthread.h>
 
+#include "./globals.h"
 /*
  * prototype: void *fn(void *arg) { ... }
  */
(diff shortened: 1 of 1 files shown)


miconda:
Looks like we were working in parallel and I pushed a commit to add the global parameter with the name `tls_threads_mode` (internal global variable `ksr_tls_threads_mode`). I also replied to the email on sr-dev where you asked for assistance on it. Maybe you can amend your code to use the global parameter that I added because I consider it to reflect better its scope.

space88man:
> Looks like we were working in parallel

Done - please check

henningw:
@space88man thanks for the work on this topic. It is great that there is then a way from version 5.7.4 and forward to restore the previous behaviour, especially in the stable branch.

PR #3747, opened 2024-02-07
ldap: option to start even when connecting to ldap server fails

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR aims to give the option for ldap to allow kamailio to start-up even if server is not reachable/available when starting.
It follows similar logic to https://github.com/kamailio/kamailio/commit/2df43b992b1e6932974bd0c928c0465a839b75da for SQL.

A new config parameter for `ldap` module is defined, `connect_mode`, that defaults to 0, preserving the old behavior. If it's set `1`, `kamailio` will not exit and allow `ldap` to try and reconnect when necessary.

### DIFF
--- a/src/modules/ldap/doc/ldap_admin.xml
+++ b/src/modules/ldap/doc/ldap_admin.xml
@@ -455,6 +455,23 @@ modparam("ldap", "config_file", "/usr/local/etc/&kamailiobinary;/ldap.ini")
 		  </programlisting>
         </example>
       </section>
+
+      <section>
+        <title><varname>connect_mode</varname> (int)</title>
+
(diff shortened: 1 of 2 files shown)


henningw:
Its not strictly a bug fix, but without this change the ldap module (and kamailio) will not start when the ldap server is not available, leading to an unavailable service. The module already has logic in place to deal with the reconnect case, but the startup was missing. Therefore it would be great to add for upcoming 5.8.0 release.

miconda:
It is rather the common to stop starting kamailio if connection to backend fails, that's the common afaik, although couple of modules allow to start without being able to connect to backend. Not having the backend available is very likely to have the routing fail as well. Anyhow, overall, I am not against to merge it till the 4.8 branch is created.

henningw:
Thanks for the feedback, merged

PR #3788, opened 2024-03-18
rtpengine: add flags processing on the daemon side

zenichev:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Introduce a versatile behavior of the rtpengine module
in terms of ability to parse flags on rtpengine side,
instead of module. Previous behavior is also kept (so backwards compatibility).

General points:
 - rtpengine daemon supports rtpp flags processing from now on
 - module still provides in the bencode (when calling daemon):
   call-id, to/from tags, viabranch (so identification call data)
 - even though the module's interface is updated,
   a backwards compatibility is given, so no obligatory changes
   from kamailio script users required
 - each rtpengine module's function where it's reasonable to use rtpp flags
   as a parameter, now is able to get a third parameter `viabranch`,
   which is used to detect, which approach to use (older/newer):
   - without the viabranch - older one used
   - with the viabrnach - new one used, so rtpp flags parsing on
     rtpengine side

The reason why the `via-branch` has been selected as a point of behavior
differentiation is that currently it's only given via option flags list (raw string),
meanwhile with a newer behavior option flags will not be parsed by the module.
Since the module still has to provide all the basic identifiers, such as:
call-id, From/To tags and via-branch, via-branch now is moved to a separate parameter,
and gives to the module a clue a newer behavior is to be applied.

The goal (for the future) is to deprecate processing of option flags
on the module side and only parse them using rtpengine.
This brings a list of benifits, such as:
 - no need to keep in sync rtpengine and module (for specific flags)
 - support of different rtpp flag string formats (raw), so that,
   for example, kamailio script users can use plain text or
   bencode dictionary like format, when providing flags from
   the kamailio script

Current change is only applicable with rtpengine versions equal or later than mr12.3
Backwards compatibility provided, so users are not forced to change anything.

### DIFF
--- a/src/modules/rtpengine/doc/rtpengine.xml
+++ b/src/modules/rtpengine/doc/rtpengine.xml
@@ -74,6 +74,14 @@
 		</address>
 		</editor>
 		<editor>
+		<firstname>Donat</firstname>
+		<surname>Zenichev</surname>
+		<affiliation><orgname>Sipwise GmbH</orgname></affiliation>
+		<address>
+			<email>dzenichev@sipwise.com</email>
+		</address>
+		</editor>
+		<editor>
 		<firstname>Joey</firstname>
 		<surname>Golan</surname>
 		<address>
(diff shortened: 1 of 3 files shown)


rfuchs:
I'll merge this if there are no objections @miconda

miconda:
Fine from my point of view. Thanks!

rfuchs:
Merged as 5cd02f56 and c6fff3a531

PR #3792, opened 2024-03-25
tcp/docs: Added tcp_accept_iplimit config sample

arnivoit:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
We noticed that configuration option tcp_accept_iplimit was included in newer kamailio versions (most notably 5.7.4), and it was a breaking change for us. Added sample configuration value in /etc/kamailio.cfg file - already tested in our production deployment.

### DIFF
--- a/etc/kamailio.cfg
+++ b/etc/kamailio.cfg
@@ -211,6 +211,9 @@ tcp_connection_lifetime=3605
 /* upper limit for TCP connections (it includes the TLS connections) */
 tcp_max_connections=2048
 
+/* upper limit for TCP connections for one ip address - default 1024 */
+#tcp_accept_iplimit=1024
+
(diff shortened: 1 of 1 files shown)


henningw:
Looks good to me, its also just a comment in the end. Lets wait a bit more for feedback and then merge it.

miconda:
Fine for me to add it to the default config file `etc/kamailio.cfg`, but the commit message should be updated to reflect that, not the prefix `tcp/docs:` is not suggesting at all it is about `etc/kamailio.cfg`.

henningw:
Thanks, it was merged manually due to a wrong commit message.

henningw:
commit 30a0fe2ea4b1b5

PR #3793, opened 2024-03-25
xhttp_prom: export pkg mem statistics

ovidiusas:
#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
Enhance xhttp_prom module to export pkgmem stats.
The new stats can be enabled via a new module parameter: `xhttp_prom_pkg_stats`.
Default value for the new module parameter: 0 (no pkg mem stats are generated).
Any value different then 0 will export the pkg mem statistics.

### DIFF
--- a/src/modules/kex/api.c
+++ b/src/modules/kex/api.c
@@ -1,0 +1,53 @@
+/**
+ * Copyright (C) 2024 Ovidiu Sas (VoIP Embedded, Inc)
+ *
+ * This file is part of Kamailio, a free SIP server.
+ *
+ * This file is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version
+ *
+ *
+ * This file is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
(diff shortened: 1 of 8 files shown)


henningw:
Thanks for the PR. Two comments:
- the code format checker seems to fail, could you please check with clang-format and update if necessary?
- you introduced a new parameter "xhttp_prom_pkgmem_stats_enabled", what about just checking during startup if the kex module is available, if yes enable the statistics, if not log a INFO message about it. In the end there is already a parameter to specify the statistics to show: "xhttp_prom_stats", we probably should re-use it.

You are probably aware of this, but after changes just force-push to this PR is ok, no need to create a new one.

ovidiusas:
I think it's better to control the pkg mem statistics via a parameter.
If the admin want's the pkg mem stats to be available but the kex module is loaded after the xhttp_prom, the pkg mem stats won't be available which will lead to confusion. Also, this will keep the existing behavior untouched.

ovidiusas:
Another reason for implementing the pkg mem via a new parameter was related to the fact the pkg mem stats are not available via the `get_statistics` mi command. The idea was to mirror the mi layout over the xhttp_prometheus control.

miconda:
I haven't looked at code, but if `xhttp_prom_stats` cannot be easily reused, then at least a shorter name for modparam has to be found, `xhttp_prom_pkgmem_stats_enabled` is really long and looking rather odd. Usually private memory is referred simply as `pkg`, not `pkgmem`, also `enabled` should be the state/meaning for the value of parameter, not something useful to be included in the name of the parameter.

ovidiusas:
Are we good to go with this one?
The name of the parameter has been changed to `xhttp_prom_pkg_stats`.

miconda:
I am fine to merge if there are no other comments.

ovidiusas:
OK!. Then please merge it ... or I can do it. Is it just pressing the `Squash and merge` button?
After that I will add the documentation.

PR #3795, opened 2024-03-27
core: dprint - include time in stderr json log

alexbakker:
#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This patch introduces a new "time" field for JSON logging to stderr. The CEE format already includes this field, but the Kamailio-specific format doesn't. If you need timestamps in your structured logs, but don't want to use the CEE format due to its limitations, this patch'd give you the best of both worlds.

I know this comes a little out of the blue, so any feedback is welcome.


miconda:
I am fine with it, should it be no other comments.

PR #3802, opened 2024-04-02
ipops: Add PTR query support

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR aims to add support for PRT queries in ipops module.

Some functions were modified to accommodate and reuse the structures already defined for DNS.

Any feedback is appreciated!

### DIFF
--- a/src/modules/ipops/doc/ipops_admin.xml
+++ b/src/modules/ipops/doc/ipops_admin.xml
@@ -883,6 +883,93 @@ if(dns_query("test.com", "xyz"))
 
     </section>
 
+    <section id="ipops.f.ptr_query">
+      <title>
+        <function moreinfo="none">ptr_query(ip, pvid)</function>
+      </title>
+
+      <para>
+		  Store the hostname that correspond to an IP address (IPv4 or IPv6)
+		  in a pseudo-variable variable $ptrquery(pvid=>hostname).
+      </para>
+
+      <para>
+      <emphasis>Note: Results will be in in $ptrquery(pvid=>key).</emphasis>
+      key can be:
+      <itemizedlist>
+        <listitem>
+          <para>
+            <emphasis>addr</emphasis> - IP address
+          </para>
+        </listitem>
+        <listitem>
+          <para>
+            <emphasis>type</emphasis> - type of IP address
+          </para>
+        </listitem>
+        <listitem>
+          <para>
+            <emphasis>ipv4</emphasis> - 1 if IPv4 address
+          </para>
+        </listitem>
+        <listitem>
+          <para>
+            <emphasis>ipv6</emphasis> - 1 if IPv6 address
+          </para>
+        </listitem>
+        <listitem>
+          <para>
+            <emphasis>count</emphasis> - number of hostnames found
+          </para>
+        </listitem>
+        <listitem>
+          <para>
+            <emphasis>hostname</emphasis> - hostname
+          </para>
+        </listitem>
+      </itemizedlist>
+      </para>
+
+      <para>Parameters:</para>
+
+      <itemizedlist>
+        <listitem>
+          <para>
+			  <emphasis>ip</emphasis> - string or pseudo-variable containing the ip.
+          </para>
+        </listitem>
+        <listitem>
+          <para>
+            <emphasis>pvid</emphasis> - AVP id for script variable.
+          </para>
+        </listitem>
+      </itemizedlist>
+
+      <para>
+        This function can be used from ANY_ROUTE.
+      </para>
+
+      <example>
+        <title>
+          <function>ptr_query</function> usage
+        </title>
+        <programlisting format="linespecific">
+...
+if(ptr_query("8.8.8.8","ip4"))
+{
+		xlog(" number of names: $ptrquery(ip4=>count)\n");
+		$var(j) = 0;
+		while($var(j)&lt;$ptrquery(ip4=>count)) {
+			xlog(" #[$var(j)] name [ $ptrquery(ip4=>hostname) ]\n");
+			$var(j) = $var(j) + 1;
+		}
+}
+...
+        </programlisting>
+      </example>
+
+    </section>
+
     <section id="ipops.f.srv_query">
       <title>
         <function moreinfo="none">srv_query(srvcname, pvid)</function>
--- a/src/modules/ipops/ipops_mod.c
+++ b/src/modules/ipops/ipops_mod.c
@@ -109,6 +109,7 @@ static int w_dns_int_match_ip(sip_msg_t *, char *, char *);
 static int fixup_detailed_ip_type(void **param, int param_no);
 static int fixup_free_detailed_ip_type(void **param, int param_no);
 static int w_dns_query(sip_msg_t *msg, char *str1, char *str2);
+static int w_ptr_query(sip_msg_t *msg, char *str1, char *str2);
 static int w_srv_query(sip_msg_t *msg, char *str1, char *str2);
 static int w_naptr_query(sip_msg_t *msg, char *str1, char *str2);
 static int w_dns_set_local_ttl(sip_msg_t *, char *, char *);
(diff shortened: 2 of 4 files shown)


oej:
The topic contains a spelling error - you mean PTR and not PRT :-)

Please add documentation as well

xkaraman:
Oh good catch! 

Yep, will do!

src/modules/ipops/ipops_mod.c
```diff
@@ -1292,6 +1297,31 @@ static int ki_dns_query(sip_msg_t *msg, str *naptrname, str *pvid)
 	return dns_update_pv(naptrname, pvid);
 }
 
+/**
+ *
```
oej, line 1301: There is room for a description of the function here - that's why we have comments before functions.

src/modules/ipops/ipops_mod.c
```diff
@@ -1292,6 +1297,31 @@ static int ki_dns_query(sip_msg_t *msg, str *naptrname, str *pvid)
...
+	str name;
+
+	if(msg == NULL) {
+		LM_ERR("received null msg\n");
+		return -1;
+	}
+
+	if(fixup_get_svalue(msg, (gparam_t *)ip, &ip_address) < 0) {
+		LM_ERR("cannot get the IP address\n");
+		return -1;
```
oej, line 1315: Just thinking loud - any reason you want different returns in different situations in your scripts? All errors return exactly the same value here.

xkaraman: I followed the `w_dns_query` as a base and assumed this is how error reporting is done. How would you like me to change them?

xkaraman:
Docs added regarding `ptr_query`.

src/modules/ipops/ipops_mod.c
xkaraman: Fixed

src/modules/ipops/doc/ipops_admin.xml
```diff
@@ -883,6 +883,57 @@ if(dns_query("test.com", "xyz"))
 
     </section>
 
+    <section id="ipops.f.ptr_query">
+      <title>
+        <function moreinfo="none">ptr_query(ip, pvid)</function>
+      </title>
+
+      <para>
+		  Store the hostname that correspond to ip
```
oej, line 892: s/ip/an IP address/

Also document if you support both IPv4 and IPv6 clearly. Thanks!

src/modules/ipops/doc/ipops_admin.xml
```diff
@@ -883,6 +883,57 @@ if(dns_query("test.com", "xyz"))
...
+		  in a config variable $ptrquery(pvid=>hostname).
```
oej, line 893: You mean a "pseudo variable" not a config variable.

src/modules/ipops/doc/ipops_admin.xml
```diff
@@ -883,6 +883,57 @@ if(dns_query("test.com", "xyz"))
...
+      </para>
+
+      <para>Parameters:</para>
+
+      <itemizedlist>
+        <listitem>
+          <para>
+			  <emphasis>ip</emphasis> - string or pseudo-variable containing the ip.
+			  The resulting IP addresses from DNS query are compared with ipaddr.
```
oej, line 902: In line 901 you just say "ip" and here "ipaddr". I suggest you write clear text on both cases. Like "containing the IP address" and "compared with the IP address"

src/modules/ipops/doc/ipops_admin.xml
```diff
@@ -883,6 +883,57 @@ if(dns_query("test.com", "xyz"))
...
+          </para>
+        </listitem>
+        <listitem>
+          <para>
+            <emphasis>pvid</emphasis> - container id for script variable.
```
oej, line 907: I don't think we call a pseudo variable a "container" anywhere. If so, I consider it a bug :-)

src/modules/ipops/ipops_mod.c
oej: Depends if you think there's any point of separating the various errors from each other in the script of if it doesn't matter - the coder just want to know that it failed, not caring about how. There are several functions in other modules, like TM, that has multiple return values. It's really up to you, just a suggestion that you consider this.

src/modules/ipops/ipops_pv.c
```diff
@@ -418,6 +413,91 @@ int dns_update_pv(str *hostname, str *name)
 	return 1;
 }
 
+/*
+*
```
oej, line 417: Add a comment here

src/modules/ipops/ipops_pv.h
```diff
@@ -24,13 +24,46 @@
...
+#define PV_DNS_ADDR 64
+#define PV_DNS_RECS 32
+#define SR_DNS_PVIDX 1
+
+typedef struct _sr_dns_record
+{
+	int type;
+	char addr[PV_DNS_ADDR];
+} sr_dns_record_t;
+typedef struct _sr_dns_item
```
oej, line 36: Empty line missing

oej (commented):
You're doing good. Just a few small nits after reading it again.

src/modules/ipops/doc/ipops_admin.xml
henningw: Actually the module uses already the term "container" in 3 places. So its probably fine in this regards.

src/modules/ipops/doc/ipops_admin.xml
henningw: The term "config variable" is also already used in the module, but I agree its probably more clear to spell it explictly.

src/modules/ipops/doc/ipops_admin.xml
xkaraman: These were mostly taken from dns_query docs and just altered it where it made sense.

src/modules/ipops/doc/ipops_admin.xml
oej: You are right. But it doesn't make it more correct... :-) 

xkaraman - you decide if you want to improve the docs or keep it at the current state.

src/modules/ipops/doc/ipops_admin.xml
xkaraman: What is the correct term for the `member/container` inside a pv that has multiple keys to get info from?

src/modules/ipops/doc/ipops_admin.xml
```diff
@@ -883,6 +883,93 @@ if(dns_query("test.com", "xyz"))
...
+        <listitem>
+          <para>
+            <emphasis>ipv4</emphasis> - 1 if IPv4 address
+          </para>
+        </listitem>
+        <listitem>
+          <para>
+            <emphasis>ipv6</emphasis> - 1 if IPv6 address
+          </para>
+        </listitem>
```
xkaraman, line 919: For these i am not exactly sure their purpose. They were accesible before these commits. Are they for counting IPv4/6 in dns_query or just to check if at least one IPv4/6 is associated with hostname?

src/modules/ipops/ipops_pv.c
xkaraman: Fixed

src/modules/ipops/doc/ipops_admin.xml
xkaraman: Removed altogether. there is no comparison done at all. It must be a leftover from DNS query.

src/modules/ipops/doc/ipops_admin.xml
xkaraman: Fixed

src/modules/ipops/doc/ipops_admin.xml
xkaraman: Fixed

src/modules/ipops/ipops_mod.c
xkaraman: Main reason i see for separating errors is for debugging purposes. 

Another thing, that might be good, would be to separate the failure over bad input and successful call with no hostname associated ([getnameinfo](https://man7.org/linux/man-pages/man3/getnameinfo.3.html)). Do you think this makes sense to separate and document it?

Other than that, as you have said the code just want to know it had failed and not how.

src/modules/ipops/doc/ipops_admin.xml
miconda: I use both terms config variables and pseudo variables, and I am referring to the same concept. Historically the term pseudo-variables was used from developer's perspective, trying to differentiate them from C-style variables. Over the time, I noticed that config-variables are easier to "understand" what one refers to, specially by those coming new to use Kamailio.

In other words, pseudo-variables tries to define them from technical point of view (i.e., no declaration, no explicit type association, automatically available by loading various modules, some read-only, other read-write). The config-variables indicates where these variables are available/can be used.

src/modules/ipops/ipops_mod.c
oej: Seems like you have reasons for multiple return values, so go ahead.

xkaraman:
So i've rebased to master and squashed some commits to make it more manageable. I have also add some more in-code comments/doc.

If there are no other suggestions/review, I will merge it soon.

xkaraman:
Thanks! Merged!

PR #3808, opened 2024-04-10
cfgutils lock

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

expose in config via cfgutils ``core_case_hash`` optional second string key for hashing


miconda:
I have nothing explicitly against this enhancement, but I guess that if one wanted to have two strings to compose the lock key,  then could use `lock("$var(str1)$var(str2)")` as alternative to new variant `lock("$var(str1)", "$var(str2)")`. The hashing behind is probably not the same result, but from the perspective of taking in consideration two strings for hashing, should be a similar behaviour.

PR #3815, opened 2024-04-17
usrloc: check on db delete the return value of memchr

dunst0:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

When inserting into the database the AOR is split at the @ sign and if there is no @ sign in the AOR only the domain part is filled and the user part is left empty.

But for deleting this is not done and the query failed to be executed and the AOR is not deleted. This PR add this behaviour of only comparing against the domain part if the AOR doesn't contain a @ sign.


linuxmaniac:
@dunst0 please use clang-format and force push the updated commit

linuxmaniac:
+2 from me

PR #3819, opened 2024-04-23
modules/statsd: ensure statsd failures do not stop execution

ostap0207:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

2 issues:

**Issue 1:**
Currently statsd functions return bool: `true` or `false` depending when
the function succeeds or fails respectively.

This value gets implicitly converted to `int`: `true` -> `1`, `false` ->
`0`.

For Kamailio `1` means succesfull execution, but `0` means to stop
processing messages, which is not what we want as statsd should not impact
flow execution. Instead we want to return `-1` which signifies error,
but the flow continues.

**Issue 2:**
statsd_init executes `statsd_connect` which tries to connect to statd
server.

If connection fails then kamailio fails to start.
This is not the desired behaviour as:
1. Kamailio should continue working even if statsd server is down,
   metrics should not impact runtime.
2. `statsd_connect` is also re-executed each time we try to send the metric https://github.com/salemove/kamailio/blame/master/src/modules/statsd/lib_statsd.c#L76,
so it's initial result is not essential.

Note, that before 5.8 the result of init was already ignored due to
implicit conversion of `false` to `0`
until after
https://github.com/kamailio/kamailio/commit/0186246fce8f0e4bb46b30c05174983cd957a3ba
was introduced (which could be considered a breaking change even if it
seemingly fixes a bug in conversion).

### DIFF
--- a/src/modules/statsd/statsd.c
+++ b/src/modules/statsd/statsd.c
@@ -50,6 +50,7 @@ static int func_incr(struct sip_msg *msg, char *key);
 static int func_incr_with_labels(struct sip_msg *msg, char *key, char *labels);
 static int func_decr(struct sip_msg *msg, char *key);
 static int func_decr_with_labels(struct sip_msg *msg, char *key, char *labels);
+static int convert_result(bool result);
 static char *get_milliseconds(char *dst);
 
 typedef struct StatsdParams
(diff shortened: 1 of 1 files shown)


src/modules/statsd/statsd.c
```diff
@@ -278,42 +279,51 @@ static int ki_statsd_stop_with_labels(sip_msg_t *msg, str *key, str *labels)
...
+	return convert_result(statsd_count(key->s, "-1", labels->s));
+}
+
+static int convert_result(bool result)
+{
+	if(result == false) {
+		return -1;
+	}
+
+	return 0;
```
miconda, line 325: Shouldn't the above line be `return 1`? Or I misunderstood the purpose.

ostap0207: Yes, correct, we noticed this in our local build but haven't updated here yet, will fix

ostap0207: Updated https://github.com/kamailio/kamailio/compare/3bbed6d62a2ef3a3fbcbf4d8ba125ab95aec28dd..ee772af9e7acbdac14247010e1e7ecf539d759ba

miconda:
Thanks!

PR #3830, opened 2024-04-25
utils/kamctl: make root host/port configurable

devopsec:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
- add support for configuring the DB root host/port connection parameters separate from the ro/rw settings. This allows the user to authenticate over the specific socket configured (tcp/udp/unix) for each connection type (root/ro/rw).
- note that support is added for mysql and postgresql connections. no changes to the other DB engine connection parameters.

### DIFF
--- a/utils/kamctl/kamctlrc
+++ b/utils/kamctl/kamctlrc
@@ -44,6 +44,14 @@
 ## database access host (from where is kamctl used)
 # DBACCESSHOST=192.168.0.1
 
+## database host for super user (useful for specifying a local socket or virtual hostname)
+# defaults to value of DBHOST when not set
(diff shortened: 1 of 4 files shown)


devopsec:
## background

We found that mariadb ver >= 10.6.1 --port= now defaults to transport=tcp.  
Using mysql client to connect to a unix socket --port= should **not** be provided.  
To support this in kamdbctl the port / host is only added to the command if *set* in kamctlrc.

## tests

Tested with mariadb 10.11..  
Simply run `kamdbctl create` with akmctlrc configured.

miconda:
Thanks!

PR #3845, opened 2024-05-10
p_usrloc: use a global p_usrloc_columns variable in db_load* functions

lbalaceanu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Proposing a small refactoring around code that creates columns parameter for queries in the database when all/most columns are used. If considered useful, this can be adapted for udomain.c file in usrloc module.

### DIFF
--- a/src/modules/p_usrloc/udomain.c
+++ b/src/modules/p_usrloc/udomain.c
@@ -39,6 +39,47 @@
 #include "utime.h"
 #include "ul_db_layer.h"
 
+enum col_index
+{
+	CONTACT_COL,
+	EXPIRES_COL,
+	Q_COL,
+	CALLID_COL,
+	CSEQ_COL,
+	FLAGS_COL,
+	CFLAGS_COL,
+	USER_AGENT_COL,
+	RECEIVED_COL,
+	PATH_COL,
+	SOCK_COL,
+	METHODS_COL,
+	LAST_MOD_COL,
+	RUID_COL,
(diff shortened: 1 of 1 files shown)


henningw:
Looks good to me. It similar code is used in usrloc udomain.c, the same refactoring could be of course also done there.

lbalaceanu:
Thank you Henning, let's see if there are other comments.

miconda:
If the array is used only in that file, I would suggest to make it static and maybe it would be also good rename it to p_usrloc_columns (or similar) to have the module prefix in the name, being a global variable.

lbalaceanu:
Hi @miconda,

Do I "Squash and merge" or "Rebase and merge" this PR? Do you consider worthwhile to do a similar PR for usrloc?

Thank you

miconda:
Hello, squashing is for combining moro commits of the same PR in a single one (if I haven't gotten it wrong), but here is only one, so rebase and merge should do it.

Even with more commits in same PR, rebasing and merge should be done when the commits are sort of independent. Squashing would be suitable when a commit amends another one (e.g., fixing fomratting of a previous commit).

henningw:
I think it would be also a good addition for the usrloc module. It certainly makes the code better understandable (some integer IDs vs. easy names).

lbalaceanu:
Hi @henningw, 

Will make a similar PR for usrloc.

Thank you

PR #3846, opened 2024-05-10
permissions: introduce func `allow_register_include_port()`

zenichev:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
Introduce new function: `allow_register_include_port()` to be able to check the whole Contact header including port.

Example, register.deny content is:
```
	ALL : "^sip:.*192.168.0.101:5062"
```

If the Contact is: "Contact: <sip:testuser1004@192.168.0.101:5062>" then this will check the Contact hf including port of it.

Otherwise if usual `allow_register()` function is used, then only the "testuser1004@192.168.0.101" will be taken into account, which will lead the regex to be failing and letting the check to pass through.

The func `allow_register_include_port()` works similarly as `allow_register()` except it checks Contact's port.

A separate function was created in order to not complicate things by introducing one more parameter to the
existing function `allow_register()`, which already takes a variable amount of parameters, so 1 or 2 parameters
(depending on if it is a "basename" or "allow-file, deny-file").

Documentation updated accordingly.
Additionally, "Register File Format" section has been added to the doc (to provide allow/deny file examples).

Full backwards compatibility is kept in place,
no need for users of the module to change anything in their configuration or kamailio script itself.

### DIFF
--- a/src/modules/permissions/doc/permissions.xml
+++ b/src/modules/permissions/doc/permissions.xml
@@ -48,6 +48,13 @@
 			<email>eschmidbauer@voipxswitch.com</email>
 		</address>
 		</editor>
+		<editor>
+		<firstname>Donat</firstname>
+		<surname>Zenichev</surname>
+		<address>
+			<email>dzenichev@sipwise.com</email>
+		</address>
+		</editor>
 	</authorgroup>
 	<copyright>
 		<year>2003</year>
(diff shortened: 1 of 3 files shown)


zenichev:
@mtirpak can you pleae have a look?

zenichev:
@henningw maybe you could have a look?
We've tested that on our systems, and actually use it already in our latest master.

Regards,
Donat

PR #3858, opened 2024-05-24
core/resolve: Check dns_cache_init and choose appropriate functions

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3350

#### Description

This PR fixes a crash when `dns_cache_init=off`.

The issue is described in #3350 and the crash is caused when trying to search in `dns_cache.c` [LOCK_DNS_HASH();](https://github.com/kamailio/kamailio/blob/d4629be286fc6d3cf61574e34ee877b2c5e9ee4a/src/core/dns_cache.c#L671). 

When `dns_cache_init=off`, the lock is never initialized and when trying to lock it crashes. 

Per my understanding, when `dns_cache_init=off`, no calls to `dns_cache.c` should be made and instead call the alternative ones from `resolve.c`, since the whole functions are compiled only when USE_DNS_CACHE definition is present.

Just a question to clarify, what is the reasoning behind USE_DNS_CACHE definition and a separate `dns_cache_init` parameter?

### DIFF
--- a/src/core/resolve.c
+++ b/src/core/resolve.c
@@ -1623,7 +1623,11 @@ struct hostent *no_naptr_srv_sip_resolvehost(
 			srv_name.s = tmp_srv;
 			srv_name.len = strlen(tmp_srv);
 #ifdef USE_DNS_CACHE
-			he = dns_srv_get_he(&srv_name, port, dns_flags);
+			if(dns_cache_init) {
+				he = dns_srv_get_he(&srv_name, port, dns_flags);
+			} else {
+				he = srv_sip_resolvehost(&srv_name, 0, port, proto, 1, 0);
+			}
 #else
 			he = srv_sip_resolvehost(&srv_name, 0, port, proto, 1, 0);
 #endif
@@ -1660,6 +1664,7 @@ struct hostent *naptr_sip_resolvehost(
 	struct rdata *naptr_head;
 	char n_proto;
 	str srv_name;
+	str *name_copy = 0;
 	naptr_bmp_t tried_bmp; /* tried bitmap */
 	char origproto = PROTO_NONE;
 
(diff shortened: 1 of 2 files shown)


src/core/resolve.c
```diff
@@ -1623,7 +1623,12 @@ struct hostent *no_naptr_srv_sip_resolvehost(
 			srv_name.s = tmp_srv;
 			srv_name.len = strlen(tmp_srv);
 #ifdef USE_DNS_CACHE
-			he = dns_srv_get_he(&srv_name, port, dns_flags);
+			if(dns_cache_init) {
+				he = dns_srv_get_he(&srv_name, port, dns_flags);
+			} else {
+				LM_WARN("USE_DNS_CACHE is defined but dns_cache_init=off\n");
```
dilyanpalauzov, line 1629: Why is this a run-time warning, depending on how Kamailio is compiled and current settings.  I mean the distributions compile Kamailio and let the users select, `dns_cache_init` ON or OFF.  With this change users will get a warning, they do not to fix, as it will be up to the distribution to fix the warning, by preventing the users to toggle `dns_cache_init`.

src/core/resolve.c
```diff
@@ -1833,6 +1847,23 @@ ip_addr_t *str2ip(str *st)
...
+	} else {
+		return _resolvehost(name);
+	}
+}
+
+struct hostent *__sip_resolvehost(str *name, unsigned short *port, char *proto)
+{
+	if(dns_cache_init) {
+		return dns_sip_resolvehost(name, port, proto);
+	} else {
```
dilyanpalauzov, line 1863: This `else` can be skipped:
```c
if (x) return y;
/*skipped else*/ return z;
```

src/core/resolve.c
xkaraman: You are right, removed.

src/core/resolve.c
xkaraman: True, both way are valid though. I think it's clearer with the `else`, but if maintainers have a preference I am fine to change that.

xkaraman:
Thanks, merging!

PR #3863, opened 2024-05-29
fixed handle_ruri_alias_mode(1) cannot proper handle multi alias

codenot:
when receive BYE sip:xxx@xxxx;alias=ip1;alias=ip2 SIP/2.0, handle_ruri_alias_mode(1) doesn't work as documented

### DIFF
--- a/src/modules/nathelper/nathelper.c
+++ b/src/modules/nathelper/nathelper.c
@@ -1160,8 +1160,6 @@ static int ki_handle_ruri_alias_mode(struct sip_msg *msg, int mode)
 				/* use first alias parameter */
 				break;
 			}
-			rest = rest + _ksr_contact_alias.len;
-			rest_len = rest_len - _ksr_contact_alias.len;
 		}
 		sep = memchr(rest, 59 /* ; */, rest_len);
 		if(sep == NULL) {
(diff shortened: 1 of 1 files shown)


miconda:
Thanks for proposing the fix, it will be reviewed. Note that you have to format the commit message as per contributing guidelines:

xkaraman:
Updated the commit's message to match the guidelines and will be merged soon if no more comments.

xkaraman:
@miconda @henningw 
Indeed the patch seems to resolve the problem. The issue seems to be happening due to the line here. https://github.com/kamailio/kamailio/blob/51dcab019a42e5b1f3b81247ca71ffcaf46180de/src/modules/nathelper/nathelper.c#L1180-L1184

We use the `rest=start;` to search for the parameters of `alias`, but in the while above `rest` and `rest_len` are changed for `mode=1`. Specifically, the problem is in `rest_len` in 1184 which does not represent the length of `start` anymore , but still used.

Removing the two lines as suggested, fixes the problem since the search is still done correctly. 

Also, I believe the `+4` is not needed anymore. https://github.com/kamailio/kamailio/blob/51dcab019a42e5b1f3b81247ca71ffcaf46180de/src/modules/nathelper/nathelper.c#L1156-L1156

miconda:
@xkaraman: I am fine to merge it if you reviewed it and fixes it.

xkaraman:
@miconda Yeap, i did some experiments with the patch and it works as intended now. 

Rebased to master, removed the `+4` magic constant and will merge when checks pass.

PR #3877, opened 2024-06-14
siptrace: handle trace_to_database param at runtime

bundasmanu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This change allows to enable/disable the insertion of SIP traces into the DB at runtime (i know we could do this with `siptrace.status`, but this is a different case).
This feature is strictly important if you start `Kamailio` with `trace_to_database` parameter disabled, and you want to enable it later and at runtime.
Nowadays, if you try to start `Kamailio` with `trace_to_database` enabled, but without a connection to the DB, the service won't start (`child_init` logic).
In this type of situation, it makes sense to start `Kamailio` disabling the `trace_to_database` parameter, and when the DB is ok, at runtime, we enable `trace_to_database` parameter. This prevents the need to restart the service, and also we can start `Kamailio` without problems, if DB of SIP Traces is NOK at that time.

My PR may need some improvements, especially on the enable side, so feel free to make any changes to improve that.

### DIFF
--- a/src/modules/siptrace/siptrace.c
+++ b/src/modules/siptrace/siptrace.c
@@ -164,6 +164,8 @@ int *trace_on_flag = NULL;
 int trace_sl_acks = 1;
 
 int trace_to_database = 1;
+int *trace_to_database_flag = NULL;
+
 int trace_db_delayed = 0;
 int trace_db_mode = 0;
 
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the PR. I did not reviewed the code yet, but noticed that the format check is failing. Could you please fix the format by executing clang-format on the file(s) and do a force-push?

miconda:
I am not sure if this is actually working, or at least some operations to be done in the rpc command callback. Have you tested it?

Checking the DB capability and opening the DB connection inside the rpc command callback is not enough, because the module write from SIP workers not from RPC worker process. Actually, the RPC worker process might not need database connection at all, because it might never handle SIP traffic.

Are the SIP workers also opening connections after the flag is enabled via RPC? The PR does not seem to have code for it.

miconda:
Somehow this got out of focus, but reviewing agan, I can't see how this is working unless kamailio is started with:

```
modparam("siptrace", "trace_to_database", 1)
```

so every Kamailio worker process get the connection to database initialized at startup. Then tracing to database can be disabled/(re-)enabled during runtime.

With this PR, if trace_to_database=0 at startup, then the global static db_con is NULL for worker processes and the RPC command initializes it only for the RPC process, it cannot do it for all the other Kamailio processes. There, db_con stays NULL, eventually resulting in crashes or not writing to database.

bundasmanu:
> Somehow this got out of focus, but reviewing agan, I can't see how this is working unless kamailio is started with:
> 
> ```
> modparam("siptrace", "trace_to_database", 1)
> ```
> 
> so every Kamailio worker process get the connection to database initialized at startup. Then tracing to database can be disabled/(re-)enabled during runtime.
> 
> With this PR, if trace_to_database=0 at startup, then the global static db_con is NULL for worker processes and the RPC command initializes it only for the RPC process, it cannot do it for all the other Kamailio processes. There, db_con stays NULL, eventually resulting in crashes or not writing to database.

Hi @miconda,

From the tests i made, i don't have problems, but i see your point.
What do you suggest, to improve? Can you help here?

Thanks

PR #3891, opened 2024-06-24
[WIP] ims_registrar_pcscf,ims_qos: added a parameter for trusting the bottom Via header on requests

vingarzan:
#### Pre-Submission Checklist

  - Not yet - first let's see if the work is valid, then I'll recompose the whole work to satisfy this. Otherwise... if I need to fix something, it's too hard to work like this...
  - ditto

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Normally, the IMS P-CSCF should identify the clients (UEs) by the received IP address and ports on Rx. The current code is using a mix of that, plus using Contact and Via headers, with arguable potential security issues.

This patch adds a new parameter to `ims_registrar_pcscf` and `ims_qos` modules, allowing for an optional outsource of the IPsec functionality to another element, which is also in charge of checking/enforcing correct UE Via header. The existing code is allowed to work as before, with the default value of the flag being towards that.

List of functional changes:
- `ims_qos`
  - added `trust_bottom_via` parameter
  - used it on `w_rx_aar_register()`
- `ims_registrar_pcscf`
  - added `trust_bottom_via` parameter
  - used it on `update_contacts()`, `save_pending()`, `check_contact()`, `getContactP()`, `check_service_routes()`, `enforce_service_routes()`
  - made `ims_ipsec_pcscf` dependency optional, with checks when used
  - skipped checks of `port-uc` in `checkcontact()` if the `ims_ipsec_pcscf` module was not loaded

List of indirect changes:
- default I-CSCF config example contained a questionable line which adds a `+` as a prefix in Request-URI. After way too much time wasted to figure out why the Diameter LIR has bogus SIP or TEL URI values in UserName AVP, I have discovered this. Seems like someone had just tel-URIs in their network, but otherwise the blind addition of this prefix makes no sense to me.
- `lib/ims`: added a `str2ushort()` macro, since code was using some dangerous casting and macros with a larger type
- `ims_registrar_pcscf`: added `Route` headers in `SUBSCRIBE` to `reginfo`, with values from `Service-Route`s, as per 3GPP specs.

List of non-functional fixes:
- (automatic) code formatting on files that I have touched
  -  in array initialization, if the last element (typically a `{0}`) has a seemingly useless `,` after, clang-format does a better job at aligning things for readability
- spelling in comments
- comments at the end of line moved above the line they refer to; with just 80 columns code-formatting, commenting on the same line provides for some super weird and hard to read code, so IMHO should not be allowed (or ... much harder now... increase to 120 columns)

Note: this was cleaned-up in https://github.com/kamailio/kamailio/pull/3901

### DIFF
--- a/src/modules/ims_registrar_pcscf/service_routes.c
+++ b/src/modules/ims_registrar_pcscf/service_routes.c
@@ -27,6 +27,7 @@
 #include "../../core/parser/msg_parser.h"
 #include "../../core/data_lump.h"
 #include "../../lib/ims/ims_getters.h"
+#include "../ims_ipsec_pcscf/cmd.h"
 
 #define STR_APPEND(dst, src)                             \
 	{                                                    \
@@ -40,7 +41,9 @@ static unsigned int current_msg_id = 0;
 static pcontact_t *c = NULL;
 
 extern usrloc_api_t ul;
+extern ipsec_pcscf_api_t ipsec_pcscf;
 extern int ignore_contact_rxport_check;
+extern int trust_bottom_via;
 static str *asserted_identity;
 static str *registration_contact;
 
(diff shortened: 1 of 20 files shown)


vingarzan:
@herlesupreeth Would appreciate your review here please. We can talk on IM of your choice also if you wish. Thanks!

vingarzan:
I think this security point should be fixed even outside my use-case and optional flag. But... since I can't see tests around this code, I'm a bit afraid to go in and do changes which would change behaviors for others.

herlesupreeth:
Thanks for the explanation. 

> I'm actually offloading the IPsec work to an external entity, which guarantees that the bottom Via is not spoofed. So I'm adding an optional "trust-the-bottom-Via" flag.

I didnt understand what you mean by offloading the IPsec work to an external entity. You mean IPSec creation is done by other entity (other than ims_ipsec_pcscf module)?

Also, the naming "trust-the-bottom-Via" doesnt go well in my opinion ("bottom-via" part I mean). While handling SIP REQUEST the first Via needs to be considered and in case SIP REPLY I think last Via needs to be considered

vingarzan:
> I didnt understand what you mean by offloading the IPsec work to an external entity. You mean IPSec creation is done by other entity (other than ims_ipsec_pcscf module)?

Yup. That whole thing through the kernel is not giving me enough flexibility for the scenario that I'm going after, so I had to do another one.
 
> Also, the naming "trust-the-bottom-Via" doesnt go well in my opinion ("bottom-via" part I mean). While handling SIP REQUEST the first Via needs to be considered and in case SIP REPLY I think last Via needs to be considered

The existing functions use "first" and "last". "Top"/"bottom" seemed better to me, since "last" could mean last added, which would be the "top" one. Or "first" could be interpreted as first added, which would be "top".  My version seems to be less controversial maybe?

Naming things it's hard :stuck_out_tongue_closed_eyes: ... let's just pick whatever we'd agree is less confusing for everyone.

I get the argument for first (a.i. top) Via on requests. Yet from the perspective of 3GPP security, there should be a single Via in any requests originating from the UE. Otherwise there is something spying between the UE and the security gateway of the P-CSCF. So I'd argue that the last (a.i. bottom) would be sufficient for all cases, with an extra check, based on local policy, to reject requests from the UE with more than 1 Via. Anyway, this is an ideal case, IRL one could argue that the P-CSCF has multiple parts and could merge Vias internally, etc, so I'd keep it flexible.

tl;dr - if you look from the UE's perspective, a P-CSCF using the bottom Via is protecting you better than one which uses the top one and as such might let someone do a man-in-the-middle attack on you. Makes sense?

herlesupreeth:
Thanks again!!

Here are my further comments. Sorry for nitpicking, but here is my two cents, I would probably issue a PR addressing the `List of non-functional fixes:` mentioned above first and keep the implementation of `trust_bottom_via` in another PR so that its easier to review and rollback if we face issues at a later point.

There are some doubts related to some of the changes in this PR which I will raise shortly

src/modules/ims_registrar_pcscf/service_routes.c
```diff
@@ -117,43 +120,81 @@ int checkcontact(struct sip_msg *_m, pcontact_t *c)
...
+					security_server_port, c->received_port, _m->rcv.src_port);
+			return 1;
 		}
 	}
 
 	if((ignore_reg_state || (c->reg_state == PCONTACT_REGISTERED))
+			// Weird... this condition is for rxport, not rxproto. If it was intentional, a comment would be nice,
+			// otherwise I'm thinking it's an unintended effect.
 			&& (ignore_contact_rxport_check
-					|| (c->received_port == _m->rcv.src_port)
```
herlesupreeth, line 150: I am not sure about this change.. why was the check for SIP message received source port removed?

vingarzan:
> Here are my further comments. Sorry for nitpicking, but here is my two cents, I would probably issue a PR addressing the `List of non-functional fixes:` mentioned above first and keep the implementation of `trust_bottom_via` in another PR so that its easier to review and rollback if we face issues at a later point.

Absolutely not a problem! Will do! Fully appreciate your review and time here! I did a draft since I wasn't sure if it all makes sense, but also didn't want to lose whatever smaller fixes I did.

src/modules/ims_registrar_pcscf/service_routes.c
vingarzan: It's moved to lines 183->184 in the additions.

I wish I hadn't, but since in my scenario the `port-uc` wasn't stored anywhere in Kamailio (`ims_ipsec_pcscf` not loaded), I had to put it behind an if for that module loaded. I'll make further improvement and also skip on `ignore_contact_rxport_check`. But it's still checked, just above.

vingarzan: With `ignore_contact_rxport_check=1`  the protocol (not port! :upside_down_face:) hit me hard today, when REGISTER happened over TCP and then an MO MESSAGE kept being rejected because ... UDP.

I added a parameter `ignore_contact_rxproto_check` with default `1` (so changing behavior!). My opinion is that in IMS the IPsec SA is negotiated for all transport protocols (so in practice UDP and TCP), hence if a UE managed to correctly encrypt whatever UDP/TCP packet correctly and send it to us on the correct Security-Association flows, we should allow it.

src/modules/ims_registrar_pcscf/service_routes.c
```diff
@@ -117,43 +121,79 @@ int checkcontact(struct sip_msg *_m, pcontact_t *c)
...
+			return 1;
 		}
 	}
 
 	if((ignore_reg_state || (c->reg_state == PCONTACT_REGISTERED))
-			&& (ignore_contact_rxport_check
-					|| (c->received_port == _m->rcv.src_port)
-					|| (security_server_port == _m->rcv.src_port))
-			&& (ignore_contact_rxport_check
-					|| (c->received_proto == _m->rcv.proto))) {
```
vingarzan, line 153: For context - the `ignore_contact_rxproto_check` parameter fixes this logical condition group. I don't know if this was on purpose, or a copy-paste mistake (checking proto for port flag).

src/modules/ims_registrar_pcscf/service_routes.c
herlesupreeth: > My opinion is that in IMS the IPsec SA is negotiated for all transport protocols (so in practice UDP and TCP), hence if a UE managed to correctly encrypt whatever UDP/TCP packet correctly and send it to us on the correct Security-Association flows, we should allow it.

I agree

vingarzan:
Closing as replaced by https://github.com/kamailio/kamailio/pull/3901

PR #3892, opened 2024-06-25
db_redis: change logs on delete

smititelu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Issue: when using usrloc with db_redis, I get constant WARNING logs for 'delete' via usrloc timer. The real WARNING, in my opinion, is when doing full table scans(which is already logged as so).

Make logs less important when performing table scans using a match key.

### DIFF
--- a/src/modules/db_redis/redis_dbase.c
+++ b/src/modules/db_redis/redis_dbase.c
@@ -1827,13 +1827,13 @@ static int db_redis_perform_delete(const db1_con_t *_h, km_redis_con_t *con,
 					"performing delete\n",
 					CON_TABLE(_h)->len, CON_TABLE(_h)->s);
 		else
-			LM_WARN("performing table scan on table '%.*s' while performing "
-					"delete using match key "
-					"'%.*s' at offset %llx\n",
+			LM_DBG("performing table scan on table '%.*s' while performing "
+				   "delete using match key "
+				   "'%.*s' at offset %llx\n",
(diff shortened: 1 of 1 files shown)


smititelu:
I never get clang formatting right, unless using the clang-format tool :D

henningw:
If there is indeed a performance implication in this table scans with key, maybe we should not silence it completely by using DBG and using INFO instead. But maybe @agranig like to comment as well here.

miconda:
Fine to merge it from my perspective, I guess that some full table scans can't be avoided always.

smititelu:
Thanks for reviews. Merging this.

Actually if one designs the 'keys' right for a db_redis table, should get *no* full table scans (I think, at least from my tests so far). So then will get no warnings after this merge.

Full table scans are actually dangerous because consume alot of kamailio PKG memory. So if one's db_redis table is big, and full scans are performed => kamailio processes won;t be able to load it in the first palce. So the WARNING for performing full table scan is still there, because of this.

Tip: For location table in db_redis i got no full table scans anymore (e.g. when contact expires or when user contacts is queried) when using this key definition:
```
modparam("db_redis", "keys", "location=entry:ruid&usrdom:username,domain&timer:expires")
```
... because index info is added in redis for 'user@domain' or for user's contact 'expires' so only those ones need to be scanned and iterated => no full scans needed

PR #3895, opened 2024-06-26
ims_registrar_scscf: drop subscription on notification failure

vingarzan:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

https://datatracker.ietf.org/doc/html/rfc3265#section-3.2.2 says that if a NOTIFY fails, the subscription should be removed. There is one exception only, when a Retry-After is specified.

This is done in order to cleanup state and not keep it around for huge periods of time. This is preferred and the client is the one that should re-subscribe at intervals to fix the situation.

Anyway, I guess I was super annoyed by the huge number of useless notifications that I'm getting for gone IMS clients.

### DIFF
--- a/src/modules/ims_registrar_scscf/registrar_notify.c
+++ b/src/modules/ims_registrar_scscf/registrar_notify.c
@@ -1394,7 +1394,7 @@ int subscribe_to_reg(struct sip_msg *msg, char *_t, char *str2)
 		res = ul.get_impurecord(
 				domain, &presentity_uri, &presentity_impurecord);
 		if(res != 0) {
-			LM_DBG("usrloc does not have imprecord for presentity being "
+			LM_DBG("usrloc does not have impurecord for presentity being "
 				   "subscribed too, we should create one.... TODO\n");
 			ul.unlock_udomain(domain, &presentity_uri);
 			goto doneorerror;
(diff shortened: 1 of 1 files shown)


vingarzan:
Hey @jaybeepee! Since you authored most of this, if you haven't retired :stuck_out_tongue_closed_eyes:, I'd appreciate your review! Thanks!

src/modules/ims_registrar_scscf/registrar_notify.c
```diff
@@ -2489,15 +2607,18 @@ void send_notification(reg_notification *n)
...
 		set_uac_req(&uac_r, &method, &h, 0, td, TMCB_LOCAL_COMPLETED,
-				uac_request_cb, 0);
+				uac_request_cb, watcher_contact);
 		tmb.t_request_within(&uac_r);
 	}
-	if(h.s)
-		pkg_free(h.s);
-	if(content.s) {
-		pkg_free(content.s);
+	watcher_contact = 0;
```
herlesupreeth, line 2613: isnt this leaking memory here?

vingarzan: Shouldn't - both branches of the `if` before have the watcher_contact saved on the transaction. I think that freeing would result in a double-free, so instead we leave it for the `uac_request_cb` to clean up after the response or timeout is reached.

herlesupreeth (approved):

PR #3899, opened 2024-07-01
rtpengine: show errno and error string in more places where it is useful

guss77:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [ ] Related to issue 

#### Description

In `send_rtpp_command()` if socket operations fail with an error other than `EINTR` or `ENOBUFS`, the actual error is not reported - we just get a generic "Can't send command to/read reply from RTPEngine".

Other places in the module report `errno` and `strerror(errno)` using the log format "`(%s:%d)`" so added that to the error log.

Also fixed `bind_force_send_ip()` error log to use the same semantic as the other places.

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -830,7 +830,7 @@ static int bind_force_send_ip(int sock_idx)
 						sizeof(ip6addr)))
 					< 0) {
 				LM_ERR("can't bind socket to required ipv6 interface\n");
-				LM_ERR("ret=%d errno=%d\n", ret, errno);
+				LM_ERR("ret=%d (%s:%d)\n", ret, strerror(errno), errno);
 				return -1;
 			}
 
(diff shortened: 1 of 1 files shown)


guss77:
I tried to match the line indentation formatting to what the CI action is showing, but it doesn't show the difference between tabs and spaces - and now I have the feeling that it showed extra indentation using spaces while all other indentation is tabs, and I overshot.

rfuchs:
You should be able to just run `clang-format` over your edited code and that should reformat it to make the format checker happy.

There are some function calls (`close()` for example) between testing the `errno` value and printing out the log line. These functions might reset `errno` to something else. You should either reorder the code so that the log printing happens immediately after testing `errno`, or save the value to an intermediary variable and then print that.

guss77:
> You should be able to just run `clang-format` over your edited code and that should reformat it to make the format checker happy.

I figured it out afterwards :sweat_smile: 

> There are some function calls (`close()` for example) between testing the `errno` value and printing out the log line. These functions might reset `errno` to something else. You should either reorder the code so that the log printing happens immediately after testing `errno`, or save the value to an intermediary variable and then print that.

I'll fix.

guss77:
A heavier touch, but we can postpone `close(fd)` to the end and it also doesn't matter if we close needlessly - it is almost a no-op with `fd = -1`. If you disagree, I can try something else.

rfuchs:
LGTM, thanks

PR #3901, opened 2024-07-02
ims_registrar_pcscf,ims_qos: support for trusting bottom via

vingarzan:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to draft-PR https://github.com/kamailio/kamailio/pull/3891

#### Description

Normally, the IMS P-CSCF should identify the clients (UEs) by the received IP address and ports on Rx. The current code is using a mix of that, plus using Contact and Via headers, with arguable potential security issues.

This patch adds a new parameter to `ims_registrar_pcscf` and `ims_qos` modules, allowing for an optional outsource of the IPsec functionality to another element, which is also in charge of checking/enforcing correct UE Via header. The existing code is allowed to work as before, with the default value of the flag being towards that.

List of functional changes:
- `ims_qos`
  - added `trust_bottom_via` parameter
  - used it on `w_rx_aar_register()`
- `ims_registrar_pcscf`
  - added `trust_bottom_via` parameter
  - used it on `update_contacts()`, `save_pending()`, `check_contact()`, `getContactP()`, `check_service_routes()`, `enforce_service_routes()`
  - made `ims_ipsec_pcscf` dependency optional, with checks when used
  - skipped checks of `port-uc` in `checkcontact()` if the `ims_ipsec_pcscf` module was not loaded
  - added `ignore_contact_rxproto_check` parameter - IMS devices open IPsec Security Associations just between IPs and ports, with both UDP and TCP protocols allowed. The default then was set here to always ignore protocol checks. Before, the `ignore_contact_rxport_check` was used to skip this and still make it work, but that seemed like a typo/mistake with hidden effects.

List of indirect changes:
- `core/ut.h`: added a `str2ushort()` macro, since code was using some dangerous casting and macros with a larger type
- `ims_registrar_pcscf`: added `Route` headers in `SUBSCRIBE` to `reginfo`, with values from `Service-Route`s, as per 3GPP specs.
- `ims_usrloc_pcscf`: small log fixes

For even more context, please see the discussion in https://github.com/kamailio/kamailio/pull/3891

### DIFF
--- a/src/core/ut.h
+++ b/src/core/ut.h
@@ -700,6 +700,14 @@ static inline int str2int(str *_s, unsigned int *_r)
 	str2unval(_s, _r, int, UINT_MAX);
 }
 
+/*
+ * Convert a str to unsigned short
+ */
+static inline int str2ushort(str *_s, unsigned short *_r)
+{
+	str2unval(_s, _r, short, USHRT_MAX);
+}
+
 
 #define str2snval(_s, _r, _vtype, _vmin, _vmax)                                \
 	do {                                                                       \
--- a/src/lib/ims/ims_getters.c
+++ b/src/lib/ims/ims_getters.c
@@ -1638,7 +1638,7 @@ str *cscf_get_service_route(struct sip_msg *msg, int *size, int is_shm)
 			}
 			x = pkg_reallocxf(x, (*size + k) * sizeof(str));
 			if(!x) {
-				LM_ERR("Error our of pkg memory");
+				LM_ERR("Error out of pkg memory");
 				return 0;
 			}
 			r2 = r;
--- a/src/modules/ims_qos/doc/ims_qos.xml
+++ b/src/modules/ims_qos/doc/ims_qos.xml
@@ -55,6 +55,12 @@
 
         <email>carsten@ng-voice.com</email>
       </author>
+
+      <author>
+        <firstname>Dragos</firstname>
+        <surname>Vingarzan</surname>
+        <email>vingarzan -at- gmail dot com</email>
+      </author>    
     </authorgroup>
 
     <copyright>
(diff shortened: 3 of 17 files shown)


herlesupreeth (approved):
Looks good to me. 

A question though:

> ims_registrar_pcscf: added Route headers in SUBSCRIBE to reginfo, with values from Service-Routes, as per 3GPP specs.

Is it only for SUBSCRIBE or is it required for INVITE as well or UE remembers it from the response to SUBSCRIBE?

vingarzan:
Many thanks for the review and help with this!

> Is it only for SUBSCRIBE or is it required for INVITE as well or UE remembers it from the response to SUBSCRIBE?

This was just for the SUBSCRIBE which the P-CSCF itself originates, in UAC mode (so not the proxied SUBSCRIBE from a UE). I think this could also be routed through the I-CSCF, but I found that 3GPP said to skip that extra "work" (at HSS also) and target directly the same S-CSCF which indicated itself in the "Service-Route" header during registration.

---
Otherwise, for UE originated requests:
- The UE also might SUBSCRIBE to the same `reginfo` event at the S-CSCF, but that would be another dialog. And all the initial dialog requests (a.i. first SUBSCRIBE, first INVITE), or standalone ones (a.i. MESSAGE) should also use the Service-Route, including, based on the UE's behavior, also the P-CSCF in the route (some UEs might just force the outbound proxy without a Route, some add a Route and do loose route).
- All the subsequent dialog methods from the UE should use the saved Record-Routes for that dialog.

And in case the UE misbehaves, the P-CSCF must enforce all of these routes. (e.g. the `pcscf_force_service_routes()` from `ims_registrar_pcscf` should do it, respectively something similar from `dialog`/`ims_dialog`)

PR #3903, opened 2024-07-02
sqlops: Don't truncated sql command when logging.

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR Fixes a bug where logging shows a truncated command instead of the original one.

Original query in .cfg:  `sql_query("ca", "UPDATE dialog SET timeout = timeout + 1 WHERE callid = '$ci' AND timeout > 0;`

logged error before fix:

sql_do_query(): cannot do the query [UPDATE dialog SET timeout = timeout + 1 WHERE callid = 'al@there]

logged error after fix:

sql_do_query(): cannot do the query [UPDATE dialog SET timeout = timeout + 1 WHERE callid = 'al@there.com AND timeout > 0;]

### DIFF
--- a/src/modules/sqlops/doc/sqlops_admin.xml
+++ b/src/modules/sqlops/doc/sqlops_admin.xml
@@ -194,6 +194,28 @@ modparam("sqlops", "tr_buf_size", 4096)
 </programlisting>
 		</example>
 	</section>
+	<section id="sqlops.p.log_buf_size">
+		<title><varname>log_buf_size</varname> (int)</title>
+		<para>
+		The size of the buffer (characters) when logging raw SQL operations.
+		</para>
+		<para>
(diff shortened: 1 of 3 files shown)


miconda:
It is not a bug, it was done on purpose to limit the size of logging message, because there can be rather long queries some time and can fill the logs or slow down. If you find 64 too short, you can eventually double it. Alternative you can add a modparam to set the max length to be printed.

xkaraman:
Hmm, ok. Seems fair but it's not documented anywhere and it can lead to misunderstanding that the query itself is truncated and producing the error.

Maybe the modparam limit is the best way to go but is it worth it to be introduced just for this logging?

xkaraman:
I added a modparam to control the max length printed when logging raw queries. I also updated the docs.

miconda:
It can be merged.

PR #3904, opened 2024-07-03
pkg: Add systemd template for starting multiple kamailio services

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR aims to allow multiple kamailio instances with different configs to be managed by systemd. 

The new `kamailio@.service` is the systemd template file, where one can provide an instance name after `systemctl start kamailio@dev`, to start a new kamailio instance with name `kamailio-dev` and use config file found in `/etc/kamailio/kamailio-dev.cfg`. 

I tested that this file and its content can work by placing the file in `/etc/systemd/system/kamailio@.service` and using the above `dev` instance name as an example. Kamailio loads and handles any new messages along with the default kamailio service.

My question right now, as I am not really familiar with the packaging workflow, is this an appropriate way to accomplish it. If yes what should be modified to install this alongside the normal `kamailio.service`. 

Any feedback and review is appreciated and greatly needed! 

Thanks!

### DIFF
--- a/pkg/kamailio/deb/debian/kamailio@.service
+++ b/pkg/kamailio/deb/debian/kamailio@.service
@@ -1,0 +1,26 @@
+[Unit]
+Description=Kamailio - the Open Source SIP Server (instance %i)
+Documentation=man:kamailio(8)
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+Type=forking
+User=kamailio
+Group=kamailio
+Environment='CFGFILE=/etc/kamailio/kamailio-%i.cfg'
+Environment='SHM_MEMORY=64'
+Environment='PKG_MEMORY=8'
+EnvironmentFile=-/etc/default/kamailio-%i
+EnvironmentFile=-/etc/default/kamailio-%i.d/*
+# PIDFile requires a full absolute path
+PIDFile=/run/kamailio-%i/kamailio.pid
+# ExecStart requires a full absolute path
+ExecStart=/usr/sbin/kamailio -P /run/kamailio-%i/kamailio.pid -Y /run/kamailio-%i -f $CFGFILE -m $SHM_MEMORY -M $PKG_MEMORY
+Restart=on-failure
+# /run/kamailio in tmpfs
+RuntimeDirectory=kamailio-%i
+RuntimeDirectoryMode=0770
+
+[Install]
+WantedBy=multi-user.target


sergey-safarov:
I understand it may not be easy, but is possible to specify control sockets for `/var/run/kamailio/kamailio_ctl`, `/var/run/kamailio/kamailio_rpc.sock`, `/var/run/kamailio/kamailio_rpc.fifo`?

xkaraman:
Hey @sergey-safarov,

As this PR is concerned, each new kamailio instance creates it's own set of these 3 files/sockets in `/run/kamailio-{name}`, ie `/run/kamailio-dev/` has these 3 files.

Then one can issue rpc command with

`kamcmd -s /run/kamailio-dev/kamailio_ctl corex.debug 3`.

Also, modparams control this in each `kamailio-{name}.cfg` (unless I am misunderstanding the params) such as 

```
/* set the path to RPC fifo control file */
# modparam("jsonrpcs", "fifo_name", "/run/kamailio/kamailio_rpc.fifo")
/* set the path to RPC unix socket control file */
# modparam("jsonrpcs", "dgram_socket", "/run/kamailio/kamailio_rpc.sock")

# ----- ctl params -----
/* set the path to RPC unix socket control file */
# modparam("ctl", "binrpc", "unix:/run/kamailio/kamailio_ctl")
```

`kamctl` and `kamctlrc` have some variables regarding these files/sockets like

```
## control engine: RPCFIFO
## - default RPCFIFO
# CTLENGINE="RPCFIFO"

## path to FIFO file for engine RPCFIFO
# RPCFIFOPATH="/run/kamailio/kamailio_rpc.fifo"
``` 

but not sure how to handle this for the template as kamctl can't take arguments I think and we can't create a second `kamctlrc-dev` to pass it on.

Does this make sense in any way for your question?

sergey-safarov:
I am fine with this PR.
Later I will package similar files for RPM packages.

@xkaraman, a question for discussion (no need to update the unit file).
You have to create a dedicated config file for each Kamailio instance.
```
Environment='CFGFILE=/etc/kamailio/kamailio-%i.cfg'
```
But have you tried to pass socket values as `define` values during the Kamailio process?
```
-A define    Add config pre-processor define (e.g., -A WITH_AUTH,
                  -A 'FLT_ACC=1', -A 'DEFVAL="str-val"')
```

xkaraman:
Hmm, i don't think it will be difficult to implement if I understand correctly what you want.

If we define those extra vars in file though, shouldn't our config file check for the existence of these defines and then change the port or whatever they are for? In this case, it can introduce a more complex config base file, but yeah probably just this one. 

Moreover in the unit service file when we call

`ExecStart=/usr/sbin/kamailio -P /run/kamailio/kamailio-%i.pid -Y /var/run/kamailio-%i -f $CFGFILE -m $SHM_MEMORY -M $PKG_MEMORY` 

we should know the name of the variable beforehand? 

What I mean by this, one can define in the file 

```
PORT=5061
```

sergey-safarov:
I found an interesting URL
https://bandonga.com/kamailio/directives/

For me, it looks like we can read environment variables from the Kamailio config file
```
#!trydefenv PORT
#!trydefenvns FOO
#!trydefenvns BAR
```
And do not need adjust `ExecStart` command.

xkaraman:
Hmm, sounds promising indeed! 

I tried to make some playground files using those directives ( [official docs](https://www.kamailio.org/wikidocs/cookbooks/devel/core/#trydefenv) ) but they seem broken.

-  `#!defenv PORT`: 
   - This should error if PORT env var is not defined. It does not.
   - If it reads something, it reads it as a string (there is `#!defenvs` for that) because config complaints about the line
    `port = PORT` with `column 8-13: number expected`, while I have exported with `export PORT=5080`.
- same goes for `trydefenv`

I will try and check if these are indeed bugs. 

If they work as intended, your template idea should work just fine. But then again, a config file must be edited accordingly to try and check for these vars!

What would be the benefit over a separate config for each new service started?

linuxmaniac:
I have no issue to add the systemd template

xkaraman:
@linuxmaniac great!

Quick question, we would expect that each instance should have a separate run folder right? 

Each contains the socket files and pid right? Maybe a small adjustment must be made so `.pid` is in the correct folder and not on the base `kamailio` unlike what we have now

```
sudo ls --color=tty -la /run/kamailio/
total 8
drwxrwx---  2 kamailio kamailio  140 Jul 23 10:09 .
drwxr-xr-x 39 root     root     1340 Jul 23 10:09 ..
srw-------  1 kamailio kamailio    0 Jul 23 10:09 kamailio_ctl
-rw-r--r--  1 kamailio kamailio    8 Jul 23 10:09 kamailio-dev.pid
-rw-r--r--  1 kamailio kamailio    8 Jul 23 10:09 kamailio.pid
prw-rw----  1 kamailio kamailio    0 Jul 23 10:09 kamailio_rpc.fifo
srw-rw----  1 kamailio kamailio    0 Jul 23 10:09 kamailio_rpc.sock
```
```
sudo ls --color=tty -la /run/kamailio-dev
total 0
drwxrwx---  2 kamailio kamailio  100 Jul 23 10:09 .
drwxr-xr-x 39 root     root     1340 Jul 23 10:14 ..
srw-------  1 kamailio kamailio    0 Jul 23 10:09 kamailio_ctl
prw-rw----  1 kamailio kamailio    0 Jul 23 10:09 kamailio_rpc.fifo
srw-rw----  1 kamailio kamailio    0 Jul 23 10:09 kamailio_rpc.sock
```

pkg/kamailio/deb/debian/kamailio@.service
```diff
@@ -0,0 +1,26 @@
...
+Type=forking
+User=kamailio
+Group=kamailio
+Environment='CFGFILE=/etc/kamailio/kamailio-%i.cfg'
+Environment='SHM_MEMORY=64'
+Environment='PKG_MEMORY=8'
+EnvironmentFile=-/etc/default/kamailio
+EnvironmentFile=-/etc/default/kamailio.d/*
+# PIDFile requires a full absolute path
+PIDFile=/run/kamailio/kamailio-%i.pid
```
linuxmaniac, line 17: you need to use /run/kamailio-%i/ here too

linuxmaniac:
> @linuxmaniac great!
> 
> Quick question, we would expect that each instance should have a separate run folder right?
> 
> Each contains the socket files and pid right? Maybe a small adjustment must be made so `.pid` is in the correct folder and not on the base `kamailio` unlike what we have now

yes, ``/run/kamailio-%i/`` for instance

> Apart from this change, should I integrate the template somehow with the pkg workflow in this PR? Currently, it's just a file added to repo, nothing more.

Just add the file and I will take over from there for deb packaging

pkg/kamailio/deb/debian/kamailio@.service
xkaraman: Ahh good catch, also modified on the argument of kamailio.

xkaraman:
Updated as you noted.

sergey-safarov:
> ExecStart=/usr/sbin/kamailio -P /run/kamailio-%i/kamailio-%i.pid -Y /var/run/kamailio-%i -f $CFGFILE -m $SHM_MEMORY -M $PKG_MEMORY

I think the template for `/run` folder is enough and we can use
```
-P /run/kamailio-%i/kamailio.pid
```

xkaraman:
Sure, keeping `kamailio.pid` for all instances in each `run` folder makes sense.

linuxmaniac:
FTR, refreshed deb packaging files after merge at 097cd7353e

PR #3921, opened 2024-07-16
ims_qos_npn: forked module from ims_qos, for better rx support

vingarzan:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

The `ims_qos` module had a few limitations. This fork of that module adds:
- AAR for the registration (signaling path monitoring)
- subsequent AARs in calls
- support for STR
- extra event_routes on Rx authorization session events
- extra AVPs decoding and use

The dependency and use of the `ims_dialog` was removed. We think that the respective module has fallen behind `dialog` (although it should've replaced it? don't know...). Cleaning/upgrading that module too might be too much effort.

Since these changes might be breaking things for other folks using IMS, we are pushing these as an alternative module. We hope that we'll find though consensus in the future and either remove or merge with the older `ims_qos` module. Same would be valid for `ims_dialog` and `dialog`, yet those also might need some upgrades.


henningw:
This was already merged after 2 days without anybody commenting, but nevertheless my comments now.

I don't think its a good idea to create a fork of this module because we are afraid of breaking things. IMHO the better approach would be to work on the existing module and then figure out if things really break, and add some compabilitity functions later. As we have seen in the ims_dialog module case, this forked modules seems to live a long time as nobody has time to do a later merging.

vingarzan:
Hey Henning!

sorry, will wait more in the future. For IMS modules, the interest seemed to be low and the state kind of experimental, so I pulled the trigger on the last day before going 1 week in vacation, reasoning that at least it has no breaking changes.

I fully agree with you. But also this module breaks away and there is further work required to make it really good and worthy of replacing the older module and dropping `ims_dialog`. I would rather surface WIP, than come-up with a huge PR after a long time.

For context: we're working at Neat Path Networks on a slightly different approach, where the IPsec part is external to a Kamailio P-CSCF. Hence some modules need less functionality, or we're trying to upgrade some to rather use the routing script and better-maintained/generic modules, instead of IMS-specific modules and opaque functions. We still hope though to share efforts with the bigger Kamailio IMS community, hence we'd rather avoid keeping our modules on a fork.

P.S. the IMS modules have no tests that I'm aware of, so trying to merge into the `ims_qos` module was way harder. If anyone is using Kamailio IMS into production, let's bundle up and write some, so that we can merge&deploy with confidence.

PR #3925, opened 2024-07-18
core: local TCP socket is bound on listening address

sergey-safarov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
As a local IP address for TCP sending operation the Kamailio service is taking the same network_interface/IP_address, which is used by the service for TCP listening.

### DIFF
--- a/src/core/tcp_main.c
+++ b/src/core/tcp_main.c
@@ -1274,6 +1274,7 @@ inline static int find_listening_sock_info(
 {
 	struct ip_addr ip;
 	struct socket_info *si = NULL;
+	socklen_t from_len;
 	su2ip_addr(&ip, *from);
 
 	si = find_si(&ip, 0, type);
@@ -1283,7 +1284,7 @@ inline static int find_listening_sock_info(
 		if(si) {
 			int optval = 1;
 			su2ip_addr(&ip, &si->su);
-			*from = &si->su;
+			memcpy(*from, &si->su, sockaddru_len(si->su));
 #if !defined(TCP_DONT_REUSEADDR)
 			if(setsockopt(s, SOL_SOCKET, SO_REUSEADDR, (void *)&optval,
 					   sizeof(optval))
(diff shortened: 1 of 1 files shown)


miconda:
Can you explain in more details what is happening now and how it changes with this PR?

Also, I noticed that instead of a reference in an output parameter, now it makes a hard copy:

```
-			*from = &si->su;
+			memcpy(*from, &si->su, sockaddru_len(si->su));
```

A reference has the size of a pointer, but the commit is copying data of size `sockaddru_len(si->su)`. Is it really correct and what you wanted to do?

sergey-safarov:
This PR is for a use case when Kamailio is listening `dummy` network interface and `tcp_reuse_port` is not used.
In this case, a new TCP connection will be sent from an IP address not listened by Kamailio.

But if you enable `tcp_reuse_port=yes` in this case new outbound TCP connection will be from the correct IP address.

This PR allows create a new TCP connection from the IP address listened by the Kamailio. In case when Kamailio listens on "dummy" interface.

@ivanuschak could comment about 
```
-			*from = &si->su;
+			memcpy(*from, &si->su, sockaddru_len(si->su));
```

ivanuschak:
@miconda regarding this change:
```
-			*from = &si->su;
+			memcpy(*from, &si->su, sockaddru_len(si->su));
```

At [this line](https://github.com/kamailio/kamailio/pull/3925/files#diff-aee8567648ae9f54eaf1f7a0feb3302d981966a749b16710e2b457da7c22c412R1306) the port is changed for the `*from`, which is the pointer to the object of `union sockaddr_union`.
If we don't use a hard copy and just use pointers then the port for `si->su` will be re-written to zero, and then at [this line](https://github.com/kamailio/kamailio/pull/3925/files#diff-aee8567648ae9f54eaf1f7a0feb3302d981966a749b16710e2b457da7c22c412R1318) it will be re-written with new port, which linux kernel has allocated for this sockaddr_union as a result of [this bind invocation](https://github.com/kamailio/kamailio/pull/3925/files#diff-aee8567648ae9f54eaf1f7a0feb3302d981966a749b16710e2b457da7c22c412R1309), but `si->su` is the `sockaddr_union` object for kamailio listening socket. So if we don't use hard copy here then we will replace `sockaddr_union` object for listening socket with the data for another socket.

As for `sockaddru_len` usage - If I'm not wrong this macro reflects the size of `union sockaddr_union`
```
/* len of the sockaddr */
#ifdef HAVE_SOCKADDR_SA_LEN
#define sockaddru_len(su) ((su).s.sa_len)
#else
#define sockaddru_len(su)                                         \
	(((su).s.sa_family == AF_INET6) ? sizeof(struct sockaddr_in6) \
					: sizeof(struct sockaddr_in))
#endif /* HAVE_SOCKADDR_SA_LEN*/
```
Please let us know if you suggest to get the size by some other approach.

miconda:
@ivanuschak - related to:

```
-			*from = &si->su;
+			memcpy(*from, &si->su, sockaddru_len(si->su));
```

What I wanted to comment about is that the size are different between the pointer to a structure and the structure itself. A pointer is 4 bytes on 32bit CPU and 8 bytes on 64bit CPU. So the `&si->su` is a pointer (address), being 4/8bytes, `*from` is written with 4/8 bytes.

The change makes the `*from` to be written with `sockaddru_len(si->su)` bytes. Is this correct in the way that `*from` has the proper reserved/allocated size? For me it feels incorrect, the `struct sockaddr_in6` is more than 8 bytes:

  - https://man7.org/linux/man-pages/man7/ipv6.7.html

ivanuschak:
@miconda regarding to this change
```
-			*from = &si->su;
+			memcpy(*from, &si->su, sockaddru_len(si->su));
```

This change is made within the function `find_listening_sock_info`. The function declaration looks as follows:
```
inline static int find_listening_sock_info(
		int s, union sockaddr_union **from, int type)
```
The `from` param is a pointer to pointer. It's presumed that at the moment when `find_listening_sock_info` is called the memory for `from` param should be already allocated.
This function is called in only a single place in the source code - [here](https://github.com/kamailio/kamailio/pull/3925/files#diff-aee8567648ae9f54eaf1f7a0feb3302d981966a749b16710e2b457da7c22c412R1373) - and `*from` pointer refers to the `my_name` variable, which is a variable allocated on the stack.
So this `memcpy` function is copying the memory from `si->su` structure object into the `my_name` variable allocated on the stack. I do not see here anything incorrect.
However, maybe the passing of the `from` param as a pointer to pointer (`union sockaddr_union **from`) can be changed to passing as a pointer (`union sockaddr_union *from`) - to exclude a confision.

Vonssgh (approved):
Please fix. I'm not receiving messages

PR #3931, opened 2024-07-24
cdp: fix use after free in transaction call backs

dunst0:
If the traction is set to auto drop,
the memory will be freed and with it the next pointer.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

The use after free caused on a Load tested system crashes.
Now the next pointer is saved before executing the callback and handle the auto drop.
Then the stored next pointer is used to continue the iteration over the callbacks.


vingarzan (approved):
Indeed, good catch.

PR #3934, opened 2024-07-30
msrp: add frame forward

bsantos:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Add a new forward function to the MSRP module that adds the ability to perform MSRP proxying.

The forward function allows frame relaying without multi-path. The frame can be relayed unmodified, or with a custom To/From-Path. If a replacement To-Path is provided, then the From-Path is replace either by the original To-Path or the  replacement From-Path. Either the destination must be set prior or a replacement To-Path must be provided.

### DIFF
--- a/src/modules/msrp/doc/msrp_admin.xml
+++ b/src/modules/msrp/doc/msrp_admin.xml
@@ -248,6 +248,44 @@ end
 
 	<section>
 	<title>Functions</title>
+	<section id="msrp.f.msrp_forward">
+	    <title>
+		<function moreinfo="none">msrp_forward([tpath [, fpath]])</function>
+	    </title>
+	    <para>
+			Forward MSRP frame to target destination. If tpath is provided then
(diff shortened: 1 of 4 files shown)


bsantos:
With the ability to replace the To-Path and From-Path headers and as long you replace both, you can relay between IPv4 and IPv6 with no problem.

As for the SDP, I did it in the routing script (using Lua), I used the sdpops.sdp_get_line_startswith to get the original path and textops.subst_body to replace the path and connection ip/port. I also recorded each session and respective replacement paths in htable vars (INVITE and on-reply callback). I use the Session-Expires header to set the expires for the htable stored data. Then on the MSRP route I fetch the stored data in htable data (using the session as key) to get the replace paths and call forward.

miconda:
Thanks, merging!

PR #3948, opened 2024-08-17
rtpengine: export api

Fr-Soltanzadeh:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Rtpengine module doesn't export api to be used by other modules. Api for four of module functions are added.

### DIFF
--- a/src/modules/rtpengine/api.h
+++ b/src/modules/rtpengine/api.h
@@ -1,0 +1,66 @@
+/*
+ * Rtpengine Module
+ *
+ * Copyright (C) 2003-2008 Sippy Software, Inc., http://www.sippysoft.com
+ * Copyright (C) 2014-2015 Sipwise GmbH, http://www.sipwise.com
+ * 
+ * This file is part of Kamailio, a free SIP server.
+ *
+ * Kamailio is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version
+ *
+ * Kamailio is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+#ifndef _RTPENGINE_API_H_
+#define _RTPENGINE_API_H_
+
+#include "../../core/parser/msg_parser.h"
+
+typedef int (*rtpengine_start_recording_f)(
+		struct sip_msg *msg, char *str1, char *str2);
+typedef int (*rtpengine_answer_f)(struct sip_msg *msg, char *str1, char *str2);
+typedef int (*rtpengine_offer_f)(struct sip_msg *msg, char *str1, char *str2);
+typedef int (*rtpengine_delete_f)(struct sip_msg *msg, char *str1, char *str2);
+
+typedef struct rtpengine_api
+{
+	rtpengine_start_recording_f start_recording;
+	rtpengine_answer_f answer;
+	rtpengine_offer_f offer;
+	rtpengine_delete_f delete;
+} rtpengine_api_t;
+
+typedef int (*bind_rtpengine_f)(rtpengine_api_t *api);
+
+/**
+ * @brief Load the RTPENGINE API
+ */
+static inline int rtpengine_load_api(rtpengine_api_t *api)
+{
+	bind_rtpengine_f bindrtpengine;
+
+	bindrtpengine = (bind_rtpengine_f)find_export("bind_rtpengine", 0, 0);
+	if(bindrtpengine == 0) {
+		LM_ERR("cannot find bind_rtpengine\n");
+		return -1;
+	}
+	if(bindrtpengine(api) < 0) {
+		LM_ERR("cannot bind rtpengine api\n");
+		return -1;
+	}
+
+	return 0;
+}
+
+
+#endif /* _RTPENGINE_API_H_ */
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -89,6 +89,7 @@
 #include "rtpengine_hash.h"
 #include "bencode.h"
 #include "config.h"
+#include "api.h"
 
 MODULE_VERSION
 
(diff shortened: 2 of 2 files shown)


henningw:
Thanks for the PR, good addition. Out of curiosity, which modules do you anticipate will use the rtpengine API?

Regarding the PR, please fix the clang format checks by running clang-format on your files. I have also one question about the rtpengine_load_api(..) function, it should be placed in the module actually using the rtpengine API, and not in the module exporting the API, right?

Fr-Soltanzadeh:
> Thanks for the PR, good addition. Out of curiosity, which modules do you anticipate will use the rtpengine API?

I want to implement ```siprec``` module and I need some of ```rtpengine``` module functionalities. 
> 
> Regarding the PR, please fix the clang format checks by running clang-format on your files. I have also one question about the rtpengine_load_api(..) function, it should be placed in the module actually using the rtpengine API, and not in the module exporting the API, right?

As in:
https://github.com/kamailio/kamailio/blob/f5f7c913d4f008b9937677ae2c2d66baa01dcf0e/src/modules/sanity/api.h#L44
```sanity_load_api``` function is defined in ```sanity``` module and is called in modules using the apis.

mojtabaesfandiari:
As I understood from your message, you are going to implement a new module called siprec for audio recording by rtpengine capabilities in Kamailio. Is it right?
If yes, so your implementation will have a close dependency on rtpengine module.

henningw:
> > Regarding the PR, please fix the clang format checks by running clang-format on your files. I have also one question about the rtpengine_load_api(..) function, it should be placed in the module actually using the rtpengine API, and not in the module exporting the API, right?
> 
> As in:
> 
> https://github.com/kamailio/kamailio/blob/f5f7c913d4f008b9937677ae2c2d66baa01dcf0e/src/modules/sanity/api.h#L44
> 
> `sanity_load_api` function is defined in `sanity` module and is called in modules using the apis.

You are right, there are different patters on how this API is exposed. Thanks for the fixes regarding the format, but I think unfortunately its still failing. Did you executed in the main kamailio directory the clang-format tool? This is usually the easiest way to fix it.

Fr-Soltanzadeh:
> As I understood from your message, you are going to implement a new module called siprec for audio recording by rtpengine capabilities in Kamailio. Is it right? If yes, so your implementation will have a close dependency on rtpengine module.

Yes, the new module will have a close dependency on it.

mojtabaesfandiari:
In my opinion, consider the considerations of strong module dependencies in the design. 
For example, you have used the capabilities of another module in your code and called it recently, while it might be better to design and implement this capability in the new module itself.
Another important point is that this heavy dependency can even reduce the performance of the Rrtpengine module in normal mode. Suppose you have several rtpengine nodes separately from the Kamailio server. In this case, you will not receive the media packets on the Kamailio server itself.  How does the new module want to access the packets on the other server? I think each nodes separately performs the recording operation and saves it to its own file system.
However, this development can be a good contributions for Kamailio.

henningw:
> In my opinion, consider the considerations of strong module dependencies in the design. For example, you have used the capabilities of another module in your code and called it recently, while it might be better to design and implement this capability in the new module itself. Another important point is that this heavy dependency can even reduce the performance of the Rrtpengine module in normal mode. Suppose you have several rtpengine nodes separately from the Kamailio server. In this case, you will not receive the media packets on the Kamailio server itself. How does the new module want to access the packets on the other server? I think each nodes separately performs the recording operation and saves it to its own file system. However, this development can be a good contributions for Kamailio.

Its a bit difficult to comment on the performance differences of different architectural approaches without seeing it implemented and running somewhere. Regarding the rtpengine call recording, this is indeed executed on the individual nodes where the rtpengine is executed.

If the mentioned new module needs to communicate with the rtpengine, its is in my opinion a good design choice to use the existing module and not duplicating the same code again on its own. I don't think this will cause any issue for the performance of the rtpengine module (for its own usage), if the communication path from the Kamailio server to the rtpengine daemons is working fine.

Fr-Soltanzadeh:
> In my opinion, consider the considerations of strong module dependencies in the design. For example, you have used the capabilities of another module in your code and called it recently, while it might be better to design and implement this capability in the new module itself. Another important point is that this heavy dependency can even reduce the performance of the Rrtpengine module in normal mode. Suppose you have several rtpengine nodes separately from the Kamailio server. In this case, you will not receive the media packets on the Kamailio server itself. How does the new module want to access the packets on the other server? I think each nodes separately performs the recording operation and saves it to its own file system. However, this development can be a good contributions for Kamailio.

Thank you for the feedback. I agree with @henningw, to address your concerns:

1) Re-implementing rtpengine functionality in the new module would result in redundant code, which isn't ideal. Leveraging existing modules promotes maintainability.
2) Similar approaches, like SIPREC in OpenSIPS, use existing RTP related modules effectively.
3) Regarding performance concerns, the module assumes recording is handled by each node individually, minimizing any impact on Kamailio.

mojtabaesfandiari:
I see, but SIPREC in Opensip is a general approach and uses the SIPREC protocol. Is your approach the same?

henningw:
Unfortunately there are still clang-format errors in the PR. If you are not able to solve them let us know, it can be also merged manually.

Fr-Soltanzadeh:
> I see, but SIPREC in Opensip is a general approach and uses the SIPREC protocol. Is your approach the same?

Not exactly the same but yes.

Fr-Soltanzadeh:
> Unfortunately there are still clang-format errors in the PR. If you are not able to solve them let us know, it can be also merged manually.

It seems that the problem is that checking clang-format is commit by commit and since my first commit had clang-format error, the problem was still there. Now I squashed all my last commits(include merging with master) and hope to be effective.

linuxmaniac:
sorry but, can you please rebase instead of merge? We don't want any merge commit in PR

Fr-Soltanzadeh:
> sorry but, can you please rebase instead of merge? We don't want any merge commit in PR

Yes, done, thank you.

henningw:
Checks are all fine now, thanks. Lets wait for a bit more time for eventual feedback from @rfuchs, otherwise it will be merged soon.

rfuchs:
No objections from my POV

henningw:
Thanks, it was manually merged (due to the commit message not conforming to the standard) in commit 3d002d561e82b8bb.

miconda:
I think this PR is completely wrong as it was done, because the API exports the functions for kamailio.cfg, which take pseudo-variables in parameters that are fixed-up at startup, thus at runtime the parameters are expected to be pv-string-parsed pointer, not pointer to a string value. I expect crashes on minimal tests as it is done right now. Or the user of the API functions, as they were done, have to call the corresponding fixup functions and provide the result to the functions, afterwards also free the fixed-up result (which are not set right now), otherwise will be memory leaks.

Probably the API should export the corresponding kemi functions, which work directly with `str*` parameters.

henningw:
Yes, you are right, I and also others missed it unfortunately. The author probably did only minimal tests. @Fr-Soltanzadeh can you create a new PR that changes the API to use the KEMI functions as suggested? If there are more questions, let us know.

PR #3951, opened 2024-08-21
kafka: a few additions to the module

stefan-mititelu-idt:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Hello, I recently did some tests with kamailio's kafka module, with focus on what happens when kafka broker goes down.

First thing I noticed is that kamailio won't start initially if kafka broker is down. Second thing I noticed syslog overflowing with err logs, when kafka broker is down. Third, I noticed that when it starts, kamailio queries topic metadata and waited up to 100 seconds for any answer. I've added modparams for all these three to change behavior.  Updated doc for all the new modparams.

I've also updated stats errors counting when exposed cfg functions failed for any reason. And some rework to stats function to accept str* instead of char* topic. Also moved librdkafka's poll() function at the begining of the cfg function logic, so polling happens even when some errors happen with the cfg function.

### DIFF
--- a/src/modules/kafka/doc/kafka_admin.xml
+++ b/src/modules/kafka/doc/kafka_admin.xml
@@ -154,6 +154,70 @@ modparam("kafka", "topic", "name=third_topic")
 		  </programlisting>
 		</example>
 	  </section>
+
+	  <section id="kafka.p.init_without_kafka">
+		<title><varname>init_without_kafka</varname> (string)</title>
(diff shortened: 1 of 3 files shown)


stefan-mititelu-idt:
rebased to latest master

miconda:
Thanks, merging it!

PR #3956, opened 2024-08-26
rtpengine: fix api using ki functions

Fr-Soltanzadeh:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Exported API functions changed to use the KEMI functions.

### DIFF
--- a/src/modules/rtpengine/api.h
+++ b/src/modules/rtpengine/api.h
@@ -26,18 +26,17 @@
 
 #include "../../core/parser/msg_parser.h"
 
-typedef int (*rtpengine_start_recording_f)(
-		struct sip_msg *msg, char *str1, char *str2);
-typedef int (*rtpengine_answer_f)(struct sip_msg *msg, char *str1, char *str2);
-typedef int (*rtpengine_offer_f)(struct sip_msg *msg, char *str1, char *str2);
-typedef int (*rtpengine_delete_f)(struct sip_msg *msg, char *str1, char *str2);
+typedef int (*rtpengine_start_recording_f)(struct sip_msg *msg);
+typedef int (*rtpengine_answer_f)(struct sip_msg *msg, str *str);
+typedef int (*rtpengine_offer_f)(struct sip_msg *msg, str *str);
+typedef int (*rtpengine_delete_f)(struct sip_msg *msg, str *str);
 
 typedef struct rtpengine_api
 {
-	rtpengine_start_recording_f start_recording;
-	rtpengine_answer_f answer;
-	rtpengine_offer_f offer;
-	rtpengine_delete_f delete;
+	rtpengine_start_recording_f rtpengine_start_recording;
+	rtpengine_answer_f rtpengine_answer;
+	rtpengine_offer_f rtpengine_offer;
+	rtpengine_delete_f rtpengine_delete;
 } rtpengine_api_t;
 
 typedef int (*bind_rtpengine_f)(rtpengine_api_t *api);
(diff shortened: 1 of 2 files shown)


src/modules/rtpengine/api.h
```diff
@@ -26,18 +26,23 @@
...
-typedef int (*rtpengine_answer_f)(struct sip_msg *msg, char *str1, char *str2);
-typedef int (*rtpengine_offer_f)(struct sip_msg *msg, char *str1, char *str2);
-typedef int (*rtpengine_delete_f)(struct sip_msg *msg, char *str1, char *str2);
+typedef int (*rtpengine_start_recording_f)(struct sip_msg *msg);
+typedef int (*rtpengine_answer_f)(struct sip_msg *msg, str *str);
+typedef int (*rtpengine_answer0_f)(struct sip_msg *msg);
+typedef int (*rtpengine_offer_f)(struct sip_msg *msg, str *str);
+typedef int (*rtpengine_offer0_f)(struct sip_msg *msg);
+typedef int (*rtpengine_delete_f)(struct sip_msg *msg, str *str);
+typedef int (*rtpengine_delete0_f)(struct sip_msg *msg);
```
rfuchs, line 36: You should only need one version of each function as the functions are able to handle NULL pointers passed to them.

Fr-Soltanzadeh: Done. Thank you.

src/modules/rtpengine/api.h
```diff
@@ -26,18 +26,17 @@
...
+typedef int (*rtpengine_offer_f)(struct sip_msg *msg, str *str);
+typedef int (*rtpengine_delete_f)(struct sip_msg *msg, str *str);
 
 typedef struct rtpengine_api
 {
 	rtpengine_start_recording_f start_recording;
 	rtpengine_answer_f answer;
 	rtpengine_offer_f offer;
-	rtpengine_delete_f delete;
+	rtpengine_delete_f _delete;
```
rfuchs, line 39: Any reason for the underscore? :smile:

J/W, otherwise LGTM

Fr-Soltanzadeh: Thanks for the review. I added the underscore to delete to avoid potential conflicts with the delete keyword in C++. While this code is written in C, there's a possibility that it might be included in a C++ project in the future. Using _delete ensures compatibility and prevents any issues if that happens. If you think it's unnecessary, I'm happy to revert it, but I thought it was a safer choice to avoid potential problems.

rfuchs: > Thanks for the review. I added the underscore to delete to avoid potential conflicts with the delete keyword in C++. While this code is written in C, there's a possibility that it might be included in a C++ project in the future. Using _delete ensures compatibility and prevents any issues if that happens. If you think it's unnecessary, I'm happy to revert it, but I thought it was a safer choice to avoid potential problems.

Ok, fair point, but I do think having the member name with an underscore looks a bit odd. Perhaps use some alternative like `del` or `delet` then, or even just `teardown` or something else.

miconda:
I see that `delete` field was renamed to `_delete`. Any particular reason it was done now?

henningw:
> I see that `delete` field was renamed to `_delete`. Any particular reason it was done now?

Refer to the discussion in the PR, Richard also raised it. We probably should stay with the "delete". We can rename it when necessary due to C++.

Fr-Soltanzadeh:
I changed all function names to rtpengine_xxx . This way is better for readability and fixes the problem.

rfuchs:
> I changed all function names to rtpengine_xxx . This way is better for readability and fixes the problem.

As long as there are no other modules that will ever use this binding ... 😆 (Welcome to one of the only two hard things in computer science)

Anyway, fine with me

Fr-Soltanzadeh:
As this PR is not accepted yet, should I change the functions name? Or is there any other issue with this PR?

Issue #3831, opened 2024-04-26
The tm.cancel function does not generate a “tm:local-request” event.


JuanPascual2:
Hello
The tm.t_uac_start function generates a “tm:local-request” event.
But the tm.cancel function doesn't generete “tm:local-request” event.

Is it possible to activate a parameter in the TM or TMX module to generate the event?

Thanks
Juan

henningw:
Thanks for the report, it was probably not implemented yet for CANCEL generated from tm.

henningw:
@JuanPascual2 a PR was merged that should fix the issue you reported, linked above. Maybe you can test and report back if your problem is solved.

JuanPascual2:
Hello
I tested the correction.
Now ,  the tm.cancel function generete “tm:local-request” event.
But the pseudo variables correspond to the INVITE equest and not to the CANCEL request.
Example: $rm = INVITE
Can you reopen the issue?
BR
Juan

JuanPascual2:
Hello
I tested it. The bug is fixed.
BR
Juan

PR #3962, opened 2024-08-29
tm: Emit request event when tm.cancel is called

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3831

#### Description

The `tm` currently does not trigger the `tm:local-request` event for CANCEL requests that are created by the tm.cancel RPC function. This pull request adds support for triggering the tm:local-request event for CANCEL requests.

According to [docs](https://www.kamailio.org/docs/modules/devel/modules/tm#tm.e.local-request) an event `tm:local-request` should be created for every request that is originates from kamailio and creates a transaction.

### DIFF
--- a/src/modules/tm/t_cancel.c
+++ b/src/modules/tm/t_cancel.c
@@ -29,12 +29,15 @@
 #include "t_funcs.h"
 #include "../../core/dprint.h"
 #include "../../core/ut.h"
+#include "../../core/kemi.h"
 #include "t_reply.h"
 #include "t_cancel.h"
 #include "t_msgbuilder.h"
 #include "t_lookup.h" /* for t_lookup_callid in fifo_uac_cancel */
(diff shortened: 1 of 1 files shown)


miconda:
A couple of issues/suggestions:

  - I would not allocate msg, declaring it as a local sip_msg_t should enough
  - no matter if msg is pointer or sip_msg_t, free_sip_msg() has to be used to clean up allocated memory by parsing
  - I would not run the event route with a faked message, either the cancel is built/parsed fine and run the event route, or throw an error message
  - the condition to execute the event route should be useless, there are some returns before that should ensure that either there is a proper event_route block or a kemi callback event name.

xkaraman:
@miconda All suggestions should be resolved now!

Issue #3823, opened 2024-04-24
Memory usage increases everytime tls.reload is executed


denzs:
### Description

We are using Kamailio 5.7.4 on Debian 12 (from http://deb.kamailio.org/kamailio57) with rtpengine as an Edgeproxy for our clients. The instance terminates SIP/TLS (with Cliencertificates) and forwards the SIP Traffic to internal systems.

After some days we are getting errors like this
`tls_complete_init(): tls: ssl bug #1491 workaround: not enough memory for safe operation: shm=7318616 threshold1=8912896`

First we thought Kamailio just doesnt have enough memory, so we doubled it..

But after some days the Logmessage (and Userissues) occured again.

So we monitored the shmmem statistics and found that used and max_used are constantly growing til it reaches the limit.

As i mentioned we are using client-certificates and so we are also using the CRL feature.
We do have a systemd-timer which fetches the CRL every hour and runs 'kamcmd tls.reload' when finished.

Our tls.cfg looks like this:
```
[server:default]
method = TLSv1.2+
private_key = /etc/letsencrypt/live/hostname.de/privkey.pem
certificate = /etc/letsencrypt/live/hostname.de/fullchain.pem
ca_list = /etc/kamailio/ca_list.pem
ca_path = /etc/kamailio/ca_list.pem
crl = /etc/kamailio/combined.crl.pem
verify_certificate = yes
require_certificate = yes

[client:default]
verify_certificate = yes
require_certificate = yes
```

After testing a bit we found that every time tls.reload is executed Kamailio consumes a bit more memory which eventually leads to all the memory being consumed which leads to issues for our users.

See following example:
```
[0][root@edgar-dev:~]# while true ; do /usr/sbin/kamcmd tls.reload ; /usr/sbin/kamcmd core.shmmem ; sleep 1 ; done
Ok. TLS configuration reloaded.
{
	total: 268435456
	free: 223001520
	used: 41352552
	real_used: 45433936
	max_used: 45445968
	fragments: 73
}
Ok. TLS configuration reloaded.
{
	total: 268435456
	free: 222377960
	used: 41975592
	real_used: 46057496
	max_used: 46069232
	fragments: 78
}
Ok. TLS configuration reloaded.
{
	total: 268435456
	free: 221748664
	used: 42604992
	real_used: 46686792
	max_used: 46698080
	fragments: 77
}
Ok. TLS configuration reloaded.
{
	total: 268435456
	free: 221110832
	used: 43242408
[...]

PR #3972, opened 2024-09-09
tls: Create one shared context and reuse it.

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3823 

#### Description

This PR aims to implement what was discussed in [mailing list](https://lists.kamailio.org/mailman3/hyperkitty/list/sr-dev@lists.kamailio.org/thread/AJMGLWJNQGA6C7SKLVQEXI5RFRRRWBN2/#AJMGLWJNQGA6C7SKLVQEXI5RFRRRWBN2) regarding some `tls.reload` and increasing memory usage.

It adds a new parameter `enable_shared_ctx` in `tls` module that if set to 0, preserves the old behavior and if set to 1 (other than 0 tbh), it creates a single SSL context that is being shared. This have the effect of using way less memory when initialized as well, but also minimizes (can't say it fixes the problem) the `tls.reload` memory increase. 

I have also added a small markdown (comparison.md) file, where some comparisons where made between enabled/disabled shared context and with/without CA file (where the initial problem was occurring by the reporter).

Feedback would be necessary to verify whether this patch, acts as expected and kamailio works as intented.

### DIFF
--- a/comparison.md
+++ b/comparison.md
@@ -1,0 +1,15 @@
+# TLS reload issue
+
+## **Without CA List**
+
+| enable_shared_ctx | At start | After ~1000 calls to tls.reload |
+|--------|--------|-------|
+| 0      | total: 14680064<br> free: 9612064<br> used: 3342048<br> real_used: 5068000<br> max_used: 5068400<br> fragments: 3 | total: 14680064<br> free: 7689136<br> used: 4285648<br> real_used: 6990928<br> max_used: 6994592<br> fragments: 11 |
+| 1      | total: 14680064<br> free: 11115744<br> used: 2769760<br> real_used: 3564320<br> max_used: 3565120<br> fragments: 3 | total: 14680064<br> free: 11026000<br> used: 2811904<br> real_used: 3654064<br> max_used: 3688576<br> fragments: 8 |
+
+## **With CA List**
+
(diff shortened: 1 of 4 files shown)


miconda:
I don't think that the comparison file needs to be added in the git repository, it is specific to current version, and anyhow definitely not in the root folder of the source code. Its content can be added in the comments/description of this PR.

Otherwise, no time yet to review the changes to C code.

xkaraman:
yes i agree! i upload it for ease of comparison in the PR. it will be removed before and if merged

space88man:
@xkaraman : in the unshared case `load_private_key()` goes through each `SSL_CTX` (per proc) to add the private key.

In the shared case is this unnecessary work? IOW, add the private key for `d->ctx[i], i = 0`; no need to call `SSL_CTX_use_PrivateKey` on `d->ctx[i], i > 0` since they are all the same ?

xkaraman:
Hey @space88man,

Yeah, you are probably right and it holds true for other cases like `load_crl`, `load_ca` etc, that also use a for loop to handle contexts, i guess?

Ok i'll have the RSA keys in mind when figuring this out.

PR #3961, opened 2024-08-29
dialog: Remove deprecated dlg_flag

xkaraman:
#### Pre-Submission Checklist

- [x] Commit message has the format required by CONTRIBUTING guide
- [x] Commits are split per component (core, individual modules, libs, utils, ...)
- [x] Each component has a single commit (if not, squash them into one commit)
- [x] No commits to README files for modules (changes must be done to docbook files
in `doc/` subfolder, the README file is autogenerated)

#### Type Of Change
- [ ] Small bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds new functionality)
- [x] Breaking change (fix or feature that would change existing functionality)

#### Checklist:

- [] PR should be backported to stable branches
- [x] Tested changes locally
- [x] Related to issue #2593 

#### Description

This PR removes the `dlg_flag` from the dialog module and recommends `dlg_manage()` instead in the docs.

PR #3975, opened 2024-09-16
dialog: Remove dlg_flag docs

btriller:
Related to #3961

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

### DIFF
--- a/src/modules/dialog/dlg_handlers.c
+++ b/src/modules/dialog/dlg_handlers.c
@@ -101,10 +101,10 @@ int dlg_set_tm_waitack(tm_cell_t *t, dlg_cell_t *dlg);
 /*!
  * \brief Initialize the dialog handlers
  * \param rr_param_p added record-route parameter
- * \param dlg_flag_p dialog flag
  * \param timeout_avp_p AVP for timeout setting
  * \param default_timeout_p default timeout
  * \param seq_match_mode_p matching mode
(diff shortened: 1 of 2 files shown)


xkaraman:
Hey @btriller,

Good catch regarding the function comment. Thanks!

Regarding the docs, I think it's best that the param `dlg_flag` stay but with maybe a small description ie, `deprecated from version 5.8.4. See dlg_manage()`. This would inform the users to look for the alternative and better option of `dlg_manage()`.

What do you think?

btriller:
> Hey @btriller,
> 
> Good catch regarding the function comment. Thanks!
> 
> Regarding the docs, I think it's best that the param `dlg_flag` stay but with maybe a small description ie, `deprecated from version 6.x. See dlg_manage()`. This would inform the users to look for the alternative and better option of `dlg_manage()`.
> 
> What do you think?

I only realized that not removing module parameter in docs was intended after opening this PR. Feel free to close this PR.
Still, IMHO it'd be better to remove this in devel docs and add removal to 6.0 upgrade docs.

xkaraman:
Hey @btriller,

Changes in code were pushed in commit 66fe6eb71e58a02222d1a2fb00f9a0cdb863134c.

I thought it was best to have `dlg_flag` deprecated docs though, so I didn't remove those!

PR #3978, opened 2024-09-18
geoip: module relocated to archive

linuxmaniac:
related #3886


henningw:
As 6.0 has been released, what about merging this (with a possible adaption due to the conflict shown) for git master now?

miconda:
As I commented at https://github.com/kamailio/kamailio/pull/3886#issuecomment-2357669624 , I am fine to skip packaging this module on newer distros, but not to remove it, since it is still used by some older, however still in use, distros.

And I guess the packaging issue was sorted out (https://github.com/kamailio/kamailio/pull/3886#issuecomment-2395456583).

There are other modules that are not (and were not) packaged, so removing is not the way for not packing it.

henningw:
Ok, close it for now then. It can be re-open when its fine to remove it.

PR #3979, opened 2024-09-18
tm: move local-request event route to the right location to access CANCEL

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3831
#### Description

This PR fixes the error introduced in #3962, about not being able to access the generated CANCEL message.

### DIFF
--- a/src/modules/tm/t_cancel.c
+++ b/src/modules/tm/t_cancel.c
@@ -89,11 +89,6 @@ void prepare_to_cancel(
 	int i;
 	int branches_no;
 	branch_bm_t mask;
-	int rt, backup_rt;
-	struct run_act_ctx ctx;
-	sip_msg_t msg;
-	sr_kemi_eng_t *keng = NULL;
-	str evname = str_init("tm:local-request");
 
 	*cancel_bm = 0;
 	branches_no = t->nr_of_outgoings;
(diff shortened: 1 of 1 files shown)


JuanPascual2:
Hello
I tested it. The bug is fixed.
BR
Juan

vsaranov:
Hello! Will this fix be included in the 5.8.4 release?

miconda:
It was not backported beacuse it was not clear for me if it goes more towards an enhancement (feature). The docs say the event route is executed for local requests that creates a transaction, however, CANCEL is associated with an INVITE transaction, it is never with a standalone purpose. It was also mentioned that it was for the case of CANCEL generated via RPC.

Overall I am not against backporting if it is considered more a bug than an enhancement, I just wanted to be sure it doesn't break expecting behaviour in a minor release. It was an email last week when the commits were backported to review if anything was missed.

henningw:
Thanks for the reply. I agree that it could be considered a feature extensions or a bug fix depending on the view point.

There were two commits if I remember correctly, I thought that it might be also not cleanly apply to the stable branches. At least the first commit was about the RPC commands, which will probably not affect a lot of people. Lets have Xenofon a look at the two commits, do a quick summary and decide then.

src/modules/tm/t_cancel.c
```diff
@@ -378,6 +335,46 @@ int cancel_branch(struct cell *t, int branch, sip_msg_t *cancel_msg,
...
+	if(build_sip_msg_from_buf(&msg, crb->buffer, crb->buffer_len, 0) < 0) {
+		LM_ERR("fail to parse msg\n");
+	}
+
+	/* Call event */
+	backup_rt = get_route_type();
+	set_route_type(REQUEST_ROUTE);
+	init_run_actions_ctx(&ctx);
+	if(rt >= 0) {
+		LM_ALERT("tm:local-request found [%d]\n", rt);
```
xkaraman, line 363: This should be changed DBG.

xkaraman:
I would say that the changes fix a bug. According to the documentation, an event should have been emitted, but it was not. 
 
If I remember correctly, an invite was necessary for my tests before issuing the cancel request, and then issuing the RPC was another way of triggering it that did not generate the event as per the reporter. 

Probably other ways of generating a locally `CANCEL` did not produce the event, but I did not verify this before.

xkaraman:
- Backported to 5.8 (701ebefaf630cb7f6039f0030f66f736f384af69)
- Backported to 5.7 (73e860525a52d8c8e07aced4b7aa6a9ee6c95865)

PR #3986, opened 2024-10-01
app_python3: support for Free-Threading Python and legacy clean-up

space88man:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
[Update:

This series of related commits is aimed at cleaning up app_python3 in time for 6.0.0.
I combine the commits in one PR for ease of reference.

*  legacy modules: Router / Core / Ranks  / Logger - already removed in app_python3s; waiting for sr-users feedback
* early imports: skip the legacy modules Router / Core / Ranks  / Logger as they are no longer supported (depends on the previous commit)
* add support for free-threading Python; this commit has no impact on stable distros and Kamailio packaging as the macro Py_GIL_DISABLED is only defined on ft builds of Python; adventurous users that build their own Kamailio may be interested in testing. It is known that Fedora ships a ft-build of Python; also Conda makes ft-Python available
* incorrect use of PyGILState_XXXX functions to ensure thread-correctness; these have been replaced by Py_BLOCK/UNBLOCK_THREADS macros; in addition Python threading behavies correctly now in that Python spawned threads will run when KEMI returns to the Kamailio event loop. Before this fixed Python threads are blocked. ] 

[Previously]
Long overdue clean-up of `app_python3`; remove legacy modules and rely solely on KEMI.
Bring it in sync with `app_python3s`.

This is an effort to build this module with Free-Threading CPython
https://py-free-threading.github.io/
- build with Free-Threading CPython builds `Py_GIL_DISABLED`
- when Python is a free-threading build this feature can be disabled by defining `-DKSR_PYTHON_DISABLE_FREETHREADING`
- when Python is a non-free-threading build this change has no effect


miconda:
It is a useful step to clean up legacy exports, but it has to announced to users mailing lists, many use this module before KEMI and we should get feedback if something from the old exports cannot be replaced by KEMI exports, and in such case, to be implemented before the next major release.

space88man:
Fix clang-format per commit; this PR is ready for review
@linuxmaniac , @miconda

space88man:
Update: this PR has a larger "blast radius" as it removes legacy modules (Logger / Router / Core etc)

No one on the user list has complained - we're assuming they are using KEMI  functions.

I will leave this up until end of October before merging.

Since this is a breaking change it is good that the major number is increased (6.x.x).

Issue #3297, opened 2022-12-08
TM crash when relaying reply, in context of unreachable rtpengine


abalashov:
When relaying this `200 OK` repeatedly with calls to `rtpengine_answer()` failing because all RTPEngines in a set are unreachable...

```
Dec  8 06:24:38 gw kamailio[4823]: ERROR: rtpengine [rtpengine.c:2960]: select_rtpp_node(): rtpengine failed to select new for calllen=36 callid=8279a4ea-151e-48de-a23c-5840104126de
```

```
SIP/2.0 200 OK
v:SIP/2.0/UDP 10.150.20.20;branch=z9hG4bK03e.e4da444e627bb575ef4f46151729a1fa.0,SIP/2.0/UDP 10.151.20.108;received=10.151.20.108;rport=5060;branch=z9hG4bKKQyrBKpp04X0K
Record-Route:<sip:10.150.20.20;lr;ftag=emHm2apy95jFe;rtp_relay=1;rtp_group=0>
f:"14045551212"<sip:14045551212@10.151.20.108>;tag=emHm2apy95jFe
t:<sip:16782001111@internal.evaristesys.com>;tag=tHQt0rDcF7UHK
i:8279a4ea-151e-48de-a23c-5840104126de
CSeq:60686471 INVITE
m:<sip:16782001111@10.160.10.55:5060;transport=udp>
User-Agent:VoiceAppServer
Allow:INVITE,ACK,BYE,CANCEL,OPTIONS,MESSAGE,INFO,UPDATE,REFER,NOTIFY
k:path,replaces
u:talk,hold,conference,refer
c:application/sdp
Content-Disposition:session
l:257

v=0
o=DM 1670428117 1670428118 IN IP4 10.160.10.55
s=DM
c=IN IP4 10.160.10.55
t=0 0
m=audio 52418 RTP/AVP 0 101
a=rtpmap:0 PCMU/8000
a=rtpmap:101 telephone-event/8000
a=fmtp:101 0-16
a=ptime:20
a=rtcp:52419 IN IP4 10.160.10.55
```

Repeat INVITEs eventually result in a crash. This backtrace is taken from a different server but identical scenario:

```
#0  t_should_relay_response (Trans=0x7f4aee56ba80, new_code=200, branch=0, should_store=0x7ffd8b8bbba0, should_relay=0x7ffd8b8bbba4, cancel_data=0x7ffd8b8bbe40,
[...]

PR #3992, opened 2024-10-04
rtpengine: disable aggressive redetection by default

stefan-mititelu-idt:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #3297

#### Description

The setup is a kamailio with no available rtpengine. (e.g. has a list with rtpengine urls but those are not reachable)

Recently had a crash similar to #3297, but in an outdated kamailio version (5.5.7). Tried to reproduce this in 5.5.7 and 5.8.3 but could not crash any of them. However, for both of them, I managed to get the logs similar to ones right before the crash happened:

```
2024-10-04T17:39:58.026206+03:00 kamailio[450237]: ERROR: {1 1 INVITE 203-450290@192.168.100.93} <core> [core/action.c:1595]: run_actions(): alert - action [corefunc (16)] cfg [/home/stefan/kamailio.cfg:625] took too long [29310906 us]
```

...directly related to how much rtpengine_manage() function took to execute. So routing of SIP is delayed by that ammount.

Tracked this in code, down to where this lock is got, when "aggressive_redetection" modparam is enabled: https://github.com/kamailio/kamailio/blob/66fe6eb71e58a02222d1a2fb00f9a0cdb863134c/src/modules/rtpengine/rtpengine.c#L3902 

I will double check that part of the code, since I don't think a lock get is necessary. It only updates a value inside a node, inside the list of nodes, but not changes the list links at all. (this in another PR)

For now I propose to disable this aggressive_redetection mechanism by default. Since it delays the SIP routing logic when no rtpengines available (and in some cases may lead to crashes in transaction module). So rtpengines will be re-pinged by default based on disable timeout modpara.

### DIFF
--- a/src/modules/rtpengine/config.c
+++ b/src/modules/rtpengine/config.c
@@ -32,7 +32,7 @@
 
 struct cfg_group_rtpengine default_rtpengine_cfg = {
 		60,						 /* default disable timeout in seconds */
-		1,						 /* default aggressive_redetection enabled */
+		0,						 /* default aggressive_redetection disabled */
 		1000,					 /* default wait timeout in milliseconds */
 		MAX_RTPP_TRIED_NODES, 5, /* rtprengine retries */
 };
--- a/src/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/src/modules/rtpengine/doc/rtpengine_admin.xml
@@ -219,20 +219,20 @@ modparam("rtpengine", "rtpengine_disable_tout", 20)
 		</para>
 		<para>
 		<emphasis>
-			Default value is <quote>1</quote> (enabled).
+			Default value is <quote>0</quote> (disabled).
 		</emphasis>
 		</para>
 		<para>
 			Can be set at runtime, e.g.:
 			<programlisting>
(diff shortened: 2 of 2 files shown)


stefan-mititelu-idt:
Done a few more tests with a large list of 16 x rtpengines, all disabled. Routing of SIP is still delayed even without the above lock_get(). It depends on the number kamailio processes used and on the "rtpengine_disable_tout" modparam.

So I think this is yet another reason to disable "aggressive_redetection" by default.

mtryfoss:
Seems like a smart change to me.

rfuchs:
> I will double check that part of the code, since I don't think a lock get is necessary. It only updates a value inside a node, inside the list of nodes, but not changes the list links at all. (this in another PR)

I think you're right. However, you would need a memory fence to safely access the values in shared memory without a lock. Also, without the lock, you might get multiple processes doing the health check at the same time, but that's not a big deal. If you want to be fancy, you could make it so that other processes skip the instance while some process is doing busy the health check, and only use it (or continue skipping it) after the health check is done.

Issue #3929, opened 2024-07-23
p-charging-vector bug in siputils with force flag


Micro-Mike:
### Description

I used sip_p_charging_vector() function of siputils with flag f (remove + generate).
This function works fine if call is received without a P-Charging-Vector and only generate needs to be done. 
But if the INVITE already contains a P-Charging-Vector and this vector needs to be removed and replaced by a new generated P-Charging-Vector the output is messed up.
The sip_p_charging_vector() function with flag r (remove) works fine too. P-Charging-Vector is removed if INVITE contains one. So for me this looks like an issue with the combination of remove + generate.

See SIP Traffic attached. The "old" P-Charging-Vector is not deleted and additionaly instead of a new P-Charging-Vector the "old" P-Charging-Vector is added again above SDP. Also some part of the next line "Session-Exp" is added and "TH: dih".

TH: dih
icid-value=0d702d06-5700-103e-00-00-00-10-6b-03-dd-04;icid-generated-at=213.xxx.xxx.xxx
Session-Exp

#### Reproduction

It looks like the issue is triggered always a p-charging-vector is received by kamailio and sip_p_charging_vector() with flag f is called.

#### Log Messages

```
/usr/local/kamailio/sbin/kamailio[486093]: DEBUG: siputils [chargingvector.c:146]: sip_parse_charging_vector(): parsing PCV header [icid-value=0d702d06-5700-103e-00-00-00-10-6b-03-dd-04;icid-generated-at=213.xxx.xxx.xxx]
/usr/local/kamailio/sbin/kamailio[486093]: DEBUG: siputils [chargingvector.c:154]: sip_parse_charging_vector(): parsed P-Charging-Vector icid-value=0d702d06-5700-103e-00-00-00-10-6b-03-dd-04
/usr/local/kamailio/sbin/kamailio[486093]: DEBUG: siputils [chargingvector.c:166]: sip_parse_charging_vector(): parsed P-Charging-Vector icid-generated-at=213.xxx.xxx.xxx
/usr/local/kamailio/sbin/kamailio[486093]: DEBUG: siputils [chargingvector.c:119]: sip_generate_charging_vector(): PCV generate
/usr/local/kamailio/sbin/kamailio[486093]: DEBUG: siputils [chargingvector.c:146]: sip_parse_charging_vector(): parsing PCV header [icid-value=4956537F000101CD6A07001D00000000; icid-generated-at=92.xxx.xxx.xxx#015#012]
[...]

PR #3994, opened 2024-10-08
Fix issue 3929 sip_p_charging_vector destroys SIP message

lachbaer:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #3929

#### Description
The handling of the buffer and pointers in the siputils/chargingvector.c module was inconsistent and led to writing directly in the sip msg buffer, which resulted in the destruction of the SIP message as reported in the issue. During the testing more cases showed up where the PCV handling became ugly and inconsistent, especially in fault scenarios. Therefore more parts have been rewritten to account for those scenarios.

The $pcv(status) pseudo-variable has been added to reflect the state of the P-Charging-Vector in the message, whether it was parsed, had errors (no body or no icid-value), has been deleted or if none is there.

The sip_p_charging_vector() function returns a status value of what it has done, or if it was a no-op call.

A present PCV can only be generated once by the module, either after an explicit deletion or by a forced replacement. After that the call returns with no-op.


linuxmaniac:
@lachbaer you have to fix the format on the commits. The check is commit by commit.

lachbaer:
@linuxmaniac Thank you for the hint. I have updated the commits.

lachbaer:
Can someone start the checks again, please? Thank you.

linuxmaniac:
@lachbaer the problem is that the last commit has ``[skip ci]`` so the checks are skipped. 
Can you, please, rebase and move that documentation commit behind the other one so the last commit message of the PR has no ``[skip ci]`` in it?

linuxmaniac:
please update https://github.com/kamailio/kamailio-wiki/blob/main/docs/features/new-in-devel.md with the new behavior

lachbaer:
new-in-devel.md updated kamailio/kamailio-wiki#60

PR #4016, opened 2024-11-05
Add timer to ping rtpengine

xkaraman:
- Added parameter to control the interval

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR adds a timer that its interval can be controlled with the param `ping_interval`. 

The timer pings all registered rtpengines and disables if any are found not responsive and vice versa.

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -227,6 +227,7 @@ static int rtpengine_delete1_f(struct sip_msg *, char *, char *);
 static int rtpengine_manage1_f(struct sip_msg *, char *, char *);
 static int rtpengine_query1_f(struct sip_msg *, char *, char *);
 static int rtpengine_info1_f(struct sip_msg *, char *, char *);
+static void rtpengine_ping_check_timer(unsigned int ticks, void *);
 
 static int w_rtpengine_query_v(sip_msg_t *msg, char *pfmt, char *pvar);
 static int fixup_rtpengine_query_v(void **param, int param_no);
@@ -365,6 +366,7 @@ static str rtpengine_dtmf_event_sock;
 static int rtpengine_dtmf_event_fd;
 int dtmf_event_rt = -1; /* default disabled */
 static int rtpengine_ping_mode = 1;
+static int rtpengine_ping_interval = 60;
 
 /* clang-format off */
 typedef struct rtpp_set_link {
@@ -489,6 +491,7 @@ static param_export_t params[] = {
 	{"rtpengine_allow_op", PARAM_INT, &rtpengine_allow_op},
 	{"control_cmd_tos", PARAM_INT, &control_cmd_tos},
 	{"ping_mode", PARAM_INT, &rtpengine_ping_mode},
+	{"ping_interval", PARAM_INT, &rtpengine_ping_interval},
 	{"db_url", PARAM_STR, &rtpp_db_url},
 	{"table_name", PARAM_STR, &rtpp_table_name},
 	{"setid_col", PARAM_STR, &rtpp_setid_col},
@@ -1895,7 +1898,6 @@ static int rtpengine_iter_cb_ping(
 		*found_rtpp_disabled = 1;
 		crt_rtpp->rn_disabled = 1;
 	}
-
 	return 0;
 }
 
@@ -2171,6 +2173,11 @@ static int mod_init(void)
 			LM_ERR("Crypto module required in order to have SHA1 hashing!\n");
 			return -1;
 		}
+	}
+
+	/* Enable ping timer if interval is positive */
+	if(rtpengine_ping_interval > 0) {
+		register_timer(rtpengine_ping_check_timer, 0, rtpengine_ping_interval);
 	}
 
 	dtmf_event_rt = route_lookup(&event_rt, "rtpengine:dtmf-event");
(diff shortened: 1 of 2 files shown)


stefan-mititelu-idt:
Hi, we have a similar task for this on our TODO list. There was this idea that the timer ping should take the pinging responsability from routing processes.

Here is some improvement suggestions, if they apply to you too:
1. Register timer callback only if "rtpengine_ping_interval" modparam is > 0 (default can be 0, disabled)
2. In rtpp_test() function check if "force" param is == 0 and if "rtpengine_ping_interval" modparam is > 0, then return imediately => this should take the pinging responsability from SIP routing processes 
3. Please check what happens with build_rtpp_socks() inside the ping timer callback, when there is at least 1 rtpengine in the list down. I think sockets will be re-built each time timer triggers

Thank you,
Stefan

xkaraman:
Hey there @stefan-mititelu-idt,

Thanks for taking time to review this. 

Indeed point 1 and 2 makes sense, and I will implement them as well.

About point for the `build_rtpp_socks()`. What is the concern exactly here? 

I can see that it may rebuild the sockets continuously and I favor finding a way to modify this. 
But even before these changes, it can happen since this function is called in `select_rtpp_node()`, which is called in every `rtpp_function_call`, meaning basically in every usage of `rtp_manage` and similar functions.

Correct me if have a wrong understanding and see some other issue.

Thanks,
Xenofon

stefan-mititelu-idt:
@xkaraman I think you are correct for nr 3. Please ignore it.

Thanks,
Stefan

src/modules/rtpengine/rtpengine.c
```diff
@@ -3549,6 +3606,10 @@ static int rtpp_test(struct rtpp_node *node, int isdisabled, int force)
 		return 1;
 	}
 	if(force == 0) {
+		/* If ping_interval is set, the timer will ping and test
+		the rtps. No need to do something during routing */
+		if(rtpengine_ping_interval > 0)
+			return 1;
```
stefan-mititelu-idt, line 3612: I think this new if condition needs to be after the existing if(isdisabled==0) condition. Otherwise may interfere with the "found:" label in select_rtpp_node_new() and automatically disable a found node

src/modules/rtpengine/rtpengine.c
```diff
@@ -2176,6 +2184,11 @@ static int mod_init(void)
 		}
 	}
 
+	/* Enable ping timer if interval is positive */
+	if(rtpengine_ping_interval > 0) {
+		register_timer(rtpengine_ping_check_timer, 0, rtpengine_ping_interval);
```
stefan-mititelu-idt, line 2189: question: after this timer callback is registered, when will it trigger first time? Is it right away, or after rtpengine_ping_interval seconds?

xkaraman: I can't find documentation about it but it seems to be to be after interval seconds and not right away.

src/modules/rtpengine/rtpengine.c
xkaraman: I think you are right, since this function has a bit different semantics that `rtpp_test_ping()`.

This function should return whether the node `isdisabled` or not (not if it's reachable since other checks are in play, like pings or already enabled).

I have made some modification to the check of `ping_interval` and now we return whatever status we already were if ping timer is enabled.

Does this make more sense?

stefan-mititelu-idt: I like this change, but I think you should still place it inside "if (force==0) {" condition. Otherwise, when timer pinging is enabled, will cancel all the forced pings of rtpp_test(). More specifically will cancel the effect of (at least) these 2 modparams:
1. aggressive_redetection -> maybe should still be ok since the purpose of the timer pinging is to disable pinging done by the routing processes, but I think should be documented
2. ping_mode -> rtpengines won't be able to be pinged first time when kamailio starts, but will be pinged first time when timer triggers (e.g. after the initial rtpengine_ping_interval seconds). Will all be considered available untill that. Please check the child_init()->build_rtpp_socks()->rtpp_test() flow for this.

Please let me know what you think about this.

Thanks,
Stefan

xkaraman: I am currently thinking that if a forced ping is required, regardless of ticks, disabled or not, and so on, the function `rtpp_test_ping` should be used. It does exactly that and only that.

So for 
1. It seems to be ok.
2. We should use the function above instead of `rtpp_test()`

Thoughts on this?

stefan-mititelu-idt: I think rtpp_test_ping() and rtpp_test() are very similar, except rtpp_test() skips sending the actual ping in some cases. Definitely first one can replace the second one when force pinging is required (maybe additionally checking manually disabled rtpengines before calling it -> RTPENGINE_MAX_RECHECK_TICKS)

I think, a much more simpler change would be the previous one suggested. So pings will happen additionally, besides the timer pings, only when they are force pings. I wonder if you see any problems with it, that I am missing?

xkaraman: Well, i don't really see any problems no. Adding the interval check in `force==0` block can work as expected probably. 

What I suggested was just a thought, that might have simplified how all these are handled and more logically structured.

I will update it accordingly.

xkaraman: Updated.

stefan-mititelu-idt: Thank you. Looks good to me.

rfuchs:
Untested but looks OK to me. I can merge if there are no objections.

xkaraman:
@rfuchs Thanks for the quick review. I am merging it!

PR #4017, opened 2024-11-06
lib/srdb1: Allow '/' in database URL passwords

kdanderso:
- Modify parsing of a database URL so that a '/' character is accepted as a valid character in the password part of the URL.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This change modifies the parsing of the DB URL so that '/' characters can be included in the password field. The modified algorithm is similar to that used when a '@' character is contained in the password.

### DIFF
--- a/src/lib/srdb1/db_id.c
+++ b/src/lib/srdb1/db_id.c
@@ -237,14 +237,33 @@ static int parse_db_url(struct db_id *id, const str *url)
 						break;
 
 					case '/':
-						id->host = prev_token;
-						prev_token = 0;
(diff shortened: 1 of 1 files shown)


btriller:
As per RFC1738 
> The user name (and password), if present, are followed by a
>commercial at-sign "@". Within the user and password field, any ":",
>"@", or "/" must be encoded.

[1] https://datatracker.ietf.org/doc/html/rfc1738#section-3.1

Issue #4019, opened 2024-11-07
Kamailio crash on http_async_client module callback invocation


ivanuschak:
### Description
Kamailio service crashed occasionally on `sock_cb` callback function of `http_async_client` module

#### Reproduction
The crash happens occasionally, no specific conditions have been identified for this issue.

#### Debugging Data
```
Program terminated with signal SIGSEGV, Segmentation fault.

(gdb) bt full
#0  0x0000ffff9f4205e4 in event_del_ () from /lib64/libevent-2.1.so.6
No symbol table info available.
#1  0x0000ffff9e7c9b24 in sock_cb (e=0xffffa2027ba8, s=9, what=4, cbp=0xffffa1c0fe58, sockp=0xffffa202d258) at http_multi.c:166
        g = 0xffffa1c0fe58
        cell = 0xffffa202d258
        whatstr = {0xffff9e7eb508 "none", 0xffff9e7eb510 "IN", 0xffff9e7eb518 "OUT", 0xffff9e7eb520 "INOUT", 0xffff9e7eb528 "REMOVE"}
        __func__ = "sock_cb"
#2  0x0000ffff9e73ffa8 in Curl_multi_closed () from /lib64/libcurl.so.4
No symbol table info available.
#3  0x0000ffff9e73d43c in Curl_closesocket () from /lib64/libcurl.so.4
No symbol table info available.
#4  0x0000ffff9e72c8f8 in conn_free.part () from /lib64/libcurl.so.4
No symbol table info available.
#5  0x0000ffff9e72d544 in Curl_disconnect () from /lib64/libcurl.so.4
No symbol table info available.
#6  0x0000ffff9e740948 in multi_done () from /lib64/libcurl.so.4
No symbol table info available.
#7  0x0000ffff9e742b10 in curl_multi_remove_handle () from /lib64/libcurl.so.4
No symbol table info available.
#8  0x0000ffff9e7c7f2c in event_cb (fd=59, kind=1, userp=0xffffa2027ba8) at http_multi.c:123
        error = 0xffff9e7eac70 "TIMEOUT"
        g = 0xffffa1c0fe58
        rc = 1968268110
        easy = 0xffffa2027ba8
        cell = 0xffffa202d258
        __func__ = "event_cb"
        action = 0
#9  0x0000ffff9f423628 in event_process_active_single_queue () from /lib64/libevent-2.1.so.6
No symbol table info available.
[...]

PR #4020, opened 2024-11-07
http_async_client: fixed crash on curl callback

ivanuschak:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4019

#### Description
I assume that the crash #4019 happens, because of an issue in libcurl, namely in the `event_cb` function ([in this line](https://github.com/kamailio/kamailio/blob/master/src/modules/http_async_client/http_multi.c#L110)) kamailio removes the pointer to the `cell`  object (which was attached to this `g->multi` curl descriptor earlier), and the `cell` object memory is released [at this line](https://github.com/kamailio/kamailio/blob/master/src/modules/http_async_client/http_multi.c#L110). However, the crash happens when `event_cb` is called when `curl_multi_remove_handle` is called [at this line](https://github.com/kamailio/kamailio/blob/master/src/modules/http_async_client/http_multi.c#L124).
In this fix `cell` object is retrieved from the corresponding CURL descriptor and then it's compared with the `cell` object passed to this callback.
With this change when this issue happens the `cell` object, which is retrieved from the corresponding CURL descriptor, is NULL pointer, because it was removed earlier [in this line](https://github.com/kamailio/kamailio/blob/master/src/modules/http_async_client/http_multi.c#L110), so the crash should not appear.
However it might be fixed in other way, namely we may not to pass `cell` to the `event_cb` callback and always retrieve it from CURL descriptor. I would appreciate if developers comment that.

### DIFF
--- a/src/modules/http_async_client/http_multi.c
+++ b/src/modules/http_async_client/http_multi.c
@@ -157,9 +157,11 @@ void event_cb(int fd, short kind, void *userp)
 /* CURLMOPT_SOCKETFUNCTION */
 int sock_cb(CURL *e, curl_socket_t s, int what, void *cbp, void *sockp)
 {
+	struct http_m_cell *cell;
 	struct http_m_global *g = (struct http_m_global *)cbp;
-	struct http_m_cell *cell = (struct http_m_cell *)sockp;
 	const char *whatstr[] = {"none", "IN", "OUT", "INOUT", "REMOVE"};
+
+	cell = http_m_cell_lookup(e);
 
 	LM_DBG("socket callback: s=%d e=%p what=%s\n", s, e, whatstr[what]);
 	if(what == CURL_POLL_REMOVE) {


grumvalski:
Thanks for the report and for the fix proposal.
However it is not clear to me how the crash is happening. In case of timeout (like it looks in your case) [this line](https://github.com/kamailio/kamailio/blob/5801ff9612d9eef9a7b06f91ff533a59020181e9/src/modules/http_async_client/http_multi.c#L110) set sockptr to NULL so that, when sock_cb is called, "cell" iS ULL and [thsi condition](https://github.com/kamailio/kamailio/blob/5801ff9612d9eef9a7b06f91ff533a59020181e9/src/modules/http_async_client/http_multi.c#L167) is false. From the backtrace we see that this is not the case in your crash because sockptr still holds the pointer to the cell that has already be unlinked and freed. I'm not sure what's happening in your case but:
- which version of curlare you using?
- could you please clarify/describe the scenario?
On the patch side: you are right that we could, in sock_cb, lookup for the cell by the easy handler, there would be no need of a cell_tmp to compare with in this case.
Anyway this scenario should never occur this way so I would like to understand more.

ivanuschak:
> which version of curl are you using?

The libcurl version is 7.61.1-34

```
[centos@esrp-1b ~]$ rpm -qa | grep curl
libcurl-minimal-7.61.1-34.el8.aarch64
curl-7.61.1-34.el8.aarch64
libcurl-devel-7.61.1-34.el8.aarch64
```


> could you please clarify/describe the scenario?

The crash happens occasionally, I have not yet identified the moment and the context of this crash. The pcap file does not contain HTTP/HTTPs at the moment when it's crashing. I will observe and provide with more details.

> On the patch side: you are right that we could, in sock_cb, lookup for the cell by the easy handler, there would be no need of a cell_tmp to compare with in this case.
Anyway this scenario should never occur this way so I would like to understand more.

I agree that it should be investigated more, and I don't consider this proposed fix as a long term solution. It can be considered as some kind of a temporary workaround until the root cause is identified. At first glance it looks like an issue in libcurl, but more details are needed to understand the context of this crash and to find the root cause. I will gather more details and provide them.

grumvalski:
@ivanuschak: I've not been able to reproduce the crash, but could you try the changes in this branch grumvalski/http_async_cb_fix? It follows your suggestion of not using the *sockp passed to the sock_cb function.

ivanuschak:
@grumvalski yes, I will deploy and test the changes you suggested. It looks more correct than the current pull-request, because it just retrieves the `cell` object from the corresponding CURL descriptor.

grumvalski:
@ivanuschak : have you been able to test the patch?

ivanuschak:
@grumvalski, I tested it on the environment where this issue does not occur and the fix works good there. However I'd like to see how it works on the env, where we experience this crash.

The deployment on the environment, where this issue is observed, is scheduled for tomorrow 05/12/2024. So it's not yet deployed. On average the crash occurs once every two days.
After the fix is deployed I will be tracking the env within a week and will let you know how it works.

grumvalski:
@ivanuschak : thank you, I'll wait for your feedback bedore merging.

ivanuschak:
@grumvalski last week the fix was installed on the environment where the crash occured. After installation the crash no longer occurs.
`http_async_client` module works good with this fix.

henningw:
Thanks for reporting back. So just to conclude this PR, the fix from @grumvalski branches was tested from you? This particular patch should be merged to master and stable branches, and this PR with a different approach is then not needed anymore?

grumvalski:
@ivanuschak : thanks for reporting
@henningw : the patch has been integrated in this PR, I'm going to merge this one and delete my temp branch.

grumvalski (approved):

Issue #3814, opened 2024-04-15
free_fixup functions in cmd_export_t are not called


xkaraman:
### Description
It seems that the fixup_free_* functions are not getting called when used with:

```
static cmd_export_t cmds[] = {
    {"acc_log_request", (cmd_function)w_acc_log_request, 1,
        acc_fixup, free_acc_fixup,
        ANY_ROUTE},
       // ...
{0, 0, 0, 0, 0, 0}
};

static int free_acc_fixup(void **param, int param_no)
{
    LM_CRIT("acc free fixup = %d\n", param_no);
   
    if(*param) {
        pkg_free(*param);
        *param = 0;
    }
    return 0;
}
```

Nothing is getting logged from the function.  Is this expected not to print in free_fixup due to the destruction of kamailio and the logging is not reliable? 

I also used gdb and added some breakpoints in `fixup` and  `free_fixup` functions in some other modules (file_out) and the breakpoint was only found in `fixup` but not in `free_fixup`.

When is the `free_fixup` supposed to be called, at kamailio destruction or after fixing it and before the main loop starts?

#### Reproduction

Use any module that has a custom (maybe also provided from core) `free_fixup` function, and monitor if  `free_fixup` is called using debug mode.

#### Log Messages

```
Apr 15 15:52:03 app01 kamailio[812169]: INFO: <core> [core/mem/q_malloc.c:402]: qm_malloc(): qm_malloc(0x7fcdcab18010, 56) called from file_out: file_out.c: fo_fixup_str_index(287)
Apr 15 15:52:03 app01 kamailio[812169]: INFO: <core> [core/mem/q_malloc.c:449]: qm_malloc(): qm_malloc(0x7fcdcab18010, 64) returns address 0x7fcdcacc4280 frag. 0x7fcdcacc4240 (size=64) on 1 -th hit```

`0x7fcdcacc4240` fragment is not found anywhere in the logs reported that it was freed unlike other malloced resources.
```

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
version: kamailio 5.9.0-dev0 (x86_64/linux) 8bc64a
[...]

PR #4023, opened 2024-11-14
Add fix_actions function parameter in fix_actions to allow call of free_fixup functions and free resources

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3814

#### Description

When the fixup is done in main using fix_rls function (fixing the route lists in cfg file), all the children inherit the memory allocated as a copy that needs separate management. 

The function responsible for the fixup (fix_actions) is also used to call the free_fixup function with an added argument. 

Then on main, we call the free_rls ( as fix_rls) on two separate occasions. Once in cleanup() function for the main process and once in the sig_usr() function for each of the children.

This seems to work fine and after debugging with `file_out` fixups, I can no longer see memory leaks due to the malloced memory in one of the fixups.

- main.c: Call free_rls() in cleanup() function to clean route lists and fixed up parameters in main process.

- main.c: Call free_rls() in sig_usr() function to clean route lists and fixed up parameters in child processes.

My question is, after a parameter is fixed, is it saved somewhere else, therefore we can immediately free the fix-up resources and not on shutdown?

### DIFF
--- a/src/core/route.c
+++ b/src/core/route.c
@@ -261,10 +261,6 @@ int route_lookup(struct route_list *rt, char *name)
 	}
 }
 
-
-int fix_actions(struct action *a); /*fwd declaration*/
-
-
 /** optimize the left side of a struct expr.
  *  @return 1 if optimized, 0 if not and -1 on error
  */
@@ -568,7 +564,7 @@ int fix_expr(struct expr *exp)
 			}
 		}
 		if(exp->l_type == ACTION_O) {
-			ret = fix_actions((struct action *)exp->r.param);
+			ret = fix_actions((struct action *)exp->r.param, 0);
 			if(ret != 0) {
 				LM_CRIT("fix_actions error\n");
 				return ret;
@@ -599,7 +595,7 @@ int fix_expr(struct expr *exp)
 
 /* adds the proxies in the proxy list & resolves the hostnames */
 /* returns 0 if ok, <0 on error */
-int fix_actions(struct action *a)
+int fix_actions(struct action *a, int free)
 {
 	struct action *t;
 	struct proxy_l *p;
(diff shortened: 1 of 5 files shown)


miconda:
First, using `free` as a variable/parameter name is strongly not recommended, many code analyzers started to report such cases, it may result in unexpected behaviour for future changes, by hiding the function.

Them if it is about calling free fixup for `f1("$var(x)....")`, that's by design, to avoid parsing parameter again and again at runtime. If not, can you provided a more specific example when you need it, I looked again over the related reported issue, but I couldn't get where it was left that you needed a fix for.

xkaraman:
@miconda Ahh true about `free`. I will change it to a more sensible one.

I do understand that for performance is best not to parse it again and again, so it must probably not be freed until the end of the shutdown.

The issue was that `free_fixup` function was not called for quoted parameters (ie `func1("$var(x)$var(y)");`, that have been fixed in startup, not even at shutdown. 

Shouldn't the memory be freed as well or am i missing something?

miconda:
The fixup result is in private memory, the clean up on signal/shut down is done only by main process, so it is rather useless, as the main process barely get the chance to execute config file. The fixup-free was mostly intended for dynamic use of the functions (e.g., from embedded interpreters, even before KEMI; or when the parameter is an expression).

Actually chunk-by-chunk cleaning at shut down, even for shared memory, can cause more troubles than benefits (e.g., a module cleans a structure that is referenced in another structure from another module). It is something I mentioned during past devel discussions and it is planned to be addressed also next week at Kamailio Devel Meeting.

xkaraman:
As this was discussed in the Developer Meeting, it can be closed.

PR #4027, opened 2024-11-17
sqlops: new connect mode to limit the amount of connections

jchavanton:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Very often I only need one connection from a worker process, to cache fill provisioning data  to Kamailio.
This new sqlops connect mode will reduce the amount of idle connection to the database server significantly.


henningw:
Thanks for the PR, there are some errors from the format checker that should be addressed.
One question about the motivation for the PR. Usually the DB connections should expire if they are not used, right? What is the advantage of this new mode then, just to avoid the time until the expire or something else?

jchavanton:
@henningw just fixed the formatting ...

From my experience with sqlops and postgres (not sure if all the db drivers are working the exact same way).
Kamailio is creating one connection per process (except for PROC_INIT, PROC_MAIN and PROC_TCP_MAIN) 
The connections remained idle but at not expiring, this is sometimes forcing the DB admin to increase the amount of maximum connections and creating some noise, on the connections.

miconda:
Indeed, the connections stay idle, they are not destroyed, and reused when again needed.

For me is fine to add this new mode, the old mode stays around.

jchavanton:
thanks for your feedbacks !

PR #4030, opened 2024-11-19
cdp: fixed issues discovered during static code analysis

vingarzan:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

This fixes:
- 4-bytes (typically uint32_t) used for time_t storage (Y2K38)
- print of time_t with %d instead of %ld
- reuse of same loop index variable in a sub-loop
- use of random(), when kam_rand() could theoretically be better
- copy-paste mistake in sp2/sp parameter, where sp2 was meant and sp is probably NULL

### DIFF
--- a/src/modules/cdp/diameter_peer.c
+++ b/src/modules/cdp/diameter_peer.c
@@ -233,10 +233,8 @@ int diameter_peer_start(int blocking)
 {
 	int pid;
 	int k = 0;
-	int seed;
 	peer *p;
 
-	seed = random();
 	/* fork workers */
 	for(k = 0; k < config->workers; k++) {
 		pid = fork_process(1001 + k, "cdp_worker", 1);
@@ -245,7 +243,6 @@ int diameter_peer_start(int blocking)
 			return 0;
 		}
 		if(pid == 0) {
-			srandom(seed * k);
 			snprintf(
 					pt[process_no].desc, MAX_PT_DESC, "cdp worker child=%d", k);
 			if(cfg_child_init())
@@ -266,7 +263,6 @@ int diameter_peer_start(int blocking)
 
 
 	/* fork receiver for unknown peers */
-	seed = random();
 	pid = fork_process(1001 + k, "cdp_receiver_peer_unknown", 1);
 
 	if(pid == -1) {
@@ -275,7 +271,6 @@ int diameter_peer_start(int blocking)
 		return 0;
 	}
 	if(pid == 0) {
-		srandom(seed * k);
 		snprintf(pt[process_no].desc, MAX_PT_DESC, "cdp receiver peer unknown");
 		if(cfg_child_init())
 			return 0;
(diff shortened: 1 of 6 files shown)


src/modules/cdp/diameter_peer.c
```diff
@@ -236,7 +236,7 @@ int diameter_peer_start(int blocking)
 	int seed;
 	peer *p;
 
-	seed = random();
+	seed = kam_rand();
```
henningw, line 239: It looks like this seed is later used to seed the glibc random number generator for (sub-) processes? If yes, then you probably want to use cryptorand() for it. This is an internal kamailio function that returns a cryptographically secure random number. The kam_rand() is just a define for rand() from the glibc, which is quite weak security wise.

This applies also to the other places below.

Please also have a look if you are not seeding the main process again, as this is already done in main.c and core/pt.c

henningw: It depends what the child processes actually use then in their code. The code in pt.c seeds all three as the different modules uses a lot of different functions. This is usually not needed in individual module new processes.
My other point was that the change here does not change anything regarding the functionality, besides silence a warning. Of course its still better to use the kam_rand() as a policy, to allow easier auditing or later changes.

```
core/rand/kam_rand.h
31:#define kam_rand(x) rand(x)
```

vingarzan: Right, that function is actually named `fork_process()`, exactly what I needed :rofl:

src/modules/cdp/receiver.c
```diff
@@ -830,7 +830,7 @@ int receive_loop(peer *original_peer)
 #if OPENSSL_VERSION_NUMBER >= 0x10100000L
 							if(enable_tls) {
 								to_ssl(&sp2->tls_ctx, &sp2->tls_conn,
-										sp->tcp_socket, method);
+										sp2->tcp_socket, method);
```
lbalaceanu, line 833: You are right, this seems like a bug.

src/modules/cdp/receiver.c
```diff
@@ -843,7 +843,7 @@ int receive_loop(peer *original_peer)
...
```
lbalaceanu, line 846: This also looks like a typo.

src/modules/cdp/diameter_peer.c
vingarzan: Seems like that function is already called, so I can probably just remove the seeding from here, since it was already seeded (by calling `fork_process()`).

henningw: Great, if you are using fork_process from the core this is indeed not needed and it should be removed.

miconda:
Does this PR still need to be reviewed or is ready to merge?

vingarzan:
> Does this PR still need to be reviewed or is ready to merge?

I have a small concern about session.c:485 if it should be `time_t`... but then that would go a bit deeper, yet I haven't seen issues with that. Hope to check and fix tomorrow.

vingarzan:
Now it should be fine, so if all checks pass, it can be merged.

PR #4043, opened 2024-11-29
Add Initial CMake support 

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR adds initial support for building kamailio with CMake.

See [docs](https://github.com/kamailio/kamailio-wiki/pull/63) for more information on how to build and use CMake.


miconda:
Thanks a lot for your work with the cmake support!

@xkaraman and @adubovikov did also testing of cmake branch during devel meeting in Dusseldorf, this PR should be in good shape to merge rather quick in order to allow time for everyone to test their compile options before freezing. Of course, even after freezing there can be changes to the build system, but it would be good to know by freezing time if we can go with cmake only further.

The old Makefiles can co-exist with the new cmake-based build system for a while, but if cmake one proves to be ready enough before freezing, I would propose to remove the old Makefiles, they got too complex over the past 20 years...

If no comments against till Monday afternoon or soo, I think this PR can be merged.

henningw:
Great work Xenofon, thank you. Indeed we should aim to merge it soon, to also allow different developers to test it more.
Related to the process of merging it, we probably should squash it everything together to one commit, as it contains are over 60 commits, right? Some of the initial commits are using also a different commit message format, no prefix.

miconda:
Fine for me to squash and merge, there are only new files, so it should not interfere with backports.

xkaraman:
Some modules `CMakelists.txt` have `TODO: ` comments, which should initiate discussion regarding deprecated libraries used. A review of those would be also a good addition I think!

xkaraman:
> Fine for me to squash and merge, there are only new files, so it should not interfere with backports.

There are some edited files like `forward.h` and `resolve.c` where I added some variables need to compile or some logic handling macro feature definitions for example. 

i can probably remove those changes and create a separate PR for fixing since they are not cmake-related

PR #4044, opened 2024-11-29
Python exception handling fixes

smurfix:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #4045

#### Description

Python's C interface contract states that C code must *either* raise an exception and return NULL, *or* return a Python object.

Doing both will raise a `SystemError` exception which obscures the original problem. It may also cause a memory leak.

### DIFF
--- a/src/modules/app_python3/apy_kemi.c
+++ b/src/modules/app_python3/apy_kemi.c
@@ -281,7 +281,7 @@ PyObject *sr_apy_kemi_exec_func_ex(
 		if(pobj == NULL) {
 			LM_ERR("null parameter - func: %.*s idx: %d argc: %d\n", fname.len,
 					fname.s, i, (int)alen);
-			return sr_kemi_apy_return_false();
+			return NULL;
 		}
 		if(ket->ptypes[i] == SR_KEMIP_STR) {
 			if(!PyUnicode_Check(pobj)) {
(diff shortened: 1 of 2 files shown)


miconda:
I think the code is better how it is because the second parameter of the function you gave as example in the issue has to be a valid string. In the current form, it indicates there is a problem with the parameter, rather than having it silently not doing anything when it was expected to be a change of the value. As commented in the issue, use sr-users mailing list if you want to discuss further.

smurfix:
You seem to have misunderstood me here. I am **not** trying to suppress the exception. I am trying to make sure that the programmer gets the **correct** exception (`TypeError`) handed to them.

The Python C interface documentation is pretty clear on this point. You MUST return NULL instead of a `PyObject*` when your C code has set an exception. The resulting `SystemError`, on top of the `TypeError` the calling Python code *should* be getting, is a pretty clear and unequivocal demonstration of this, and it even says what's wrong ("SystemError: <built-in function sets> returned a result with an exception set"). See my bug report for the example stack trace.

smurfix:
NB: Two years ago I found another example of this problem which was fixed in b7911b6.

henningw:
Re-opened, as there seems to be still discussions necessary if the python KEMI interface should be adapted to the interface documentation.

miconda:
OK, thanks for explaining further -- as a not much expert in python, it looked like hiding it by not providing the right parameter, so if I got it wrong, then it is fine to merge.

smurfix:
Thank you.

Issue #4041, opened 2024-11-28
The future of internal libraries


miconda:
Considering that it hasn't been much activity around the internal libraries for long time, I was wondering if it still makes sense to keep them like they are or merge them in the core.

Recently, during the devel meeting, I relocated a couple of them to the archive. The `lib/binrpc` (and `print`) can have the same fate, being not actually used.

The srdb1 is very common and I guess it is very rare the case when a deployment is used without a module not depending on it (e.g., even if dispatcher is used with a list file, the code is still linked to this library).
The `trie` library is small and `ims` is actually a set of getter functions for message attribtues/headers.

The `srdb2` is not used much, but I think the right approach for it is to change the modules using it to `libsrdb1` and get rid of it completely sometime in the future. Till then the code is not big and can reside as part of the core.

The benefits I see are in simplifying the build system (current makefiles and hopefully soon-to-be-merged the one based on cmake), reducing also the time to compute compile and link dependencies; getting rid of packaging complexity (same library may be needed by different modules, which, when packaged separately, the library has to be in a 3rd package, or have dependency of the other module's package -- probably now the libs are shipped with the core package).

The code for libs can stay at the same location, so source code for modules doesn't need to be updated, just that they are compiled with the core, removing internal libs from makefiles.

Thinking that we go for 6.0.x, clarifying the plans for this component and cleaning it, if decided so, should fit well in such version number jump.

Opinions?

miconda:
Just pushed the changes to compile/link internal libs to the core for the old-Makefiles build system. I did individual commits just in case it will be needed to revert when some modules get changed/archived.

@xkaraman: do you want to update CMake files for the same purpose: compile/link code from src/lib/{srdb1,srdb2,ims,trie} to the core and remove building/linking of internal libs by modules? I can look into it if you have any constraints.

miconda:
@xkaraman: no, it is about the internal libraries whose code is inside the folders `src/lib/{srdb1,srdb2,ims,trie}`. Instead of building the code in those folders as internal libraries (that were then used by the modules (like acc, msilo, carrierroute, ...)), the code is compiled as part of the core. So, for example, acc no longer needs to be linked to libsrdb1, carrierroute no longer needs to be linked to libsrdb1 and libtrie.

But the external libraries are sill used/linked as they are now (e.g., libssl, libjansson, ...).

[...]

PR #4059, opened 2024-12-06
cmake: Link internal libraries to core and remove them from modules

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4041


miconda:
If I am not misunderstanding the PR and changes in some cmake-files, the internal libs are still compiled as libraries (.so files).

The target is to have them compiled as part of the core, so there will be just kamailio binary and no .so files other than the modules. In other words, the `.c` and `.h` files in the `src/lib/.../` are part of the core and, as an example. should be compiled into `kamailio` in the same way `src/core/mem/*.{c,h}` are compiled.

With the old-Makefiles, once I did the changes for #4041 , the core files are including the internal lib files, see:

- https://github.com/kamailio/kamailio/blob/master/src/Makefile.sources#L13-L21

xkaraman:
Ahh, ok no produced libraries at all and compile everything into the kamailio binary! I will make sure to update the PR accordingly.

xkaraman:
@miconda This should be as expected now. All the libs are compiled in the kamailio binary and the modules are linked against that.

miconda:
Thanks! Fine for me to merge, any tunings that might be needed can be done later once testing

henningw:
Just a note, on a fresh checkout of kamailio there is currently a problem related to the generation of the parser files after this PR. @xkaraman is working on it.

xkaraman:
@henningw This should be fixed in 9ff2bb869dafca8521a0da7f112c3846f86c5f7e.

PR #4066, opened 2024-12-11
pre-commit: cmake-format and cmake-lint hooks [skip ci]

linuxmaniac:
@xkaraman I hope you will agree to add this jobs related to cmake files to keep things tidy.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,4 +11,9 @@ repos:
   - repo: https://github.com/pre-commit/mirrors-clang-format
     rev: v15.0.7
     hooks:
-    - id: clang-format
+      - id: clang-format
+  - repo: https://github.com/cheshirekow/cmake-format-precommit
+    rev: v0.6.13
+    hooks:
+      - id: cmake-format
+      - id: cmake-lint


miconda:
It is fine for me.

xkaraman:
Yes, definitely agree!!

miconda:
If a custom config is going to be used, maybe it would be good to tune it a bit, since we are at the beginning of using cmake files. For example I would suggest a longer `line_width`, 80 was (maybe still is) for code to be easily seen on the very old terminal/console, but considering the current displays as well as terminal apps, it could be more convenient to have it like 120 or so. Otherwise we may end up in splitting lines/strings that are maybe not very long, but they are impacted by indentation.

miconda:
Not to continue the discussion on this closed PR, I created tracker issue for it: # 4075.

xkaraman:
@linuxmaniac @miconda: Question regarding these checks! 

Are they enforced when pushing directly to master, if someone has not setup locally the pre-commit python module and not being warned? 

Ideally, when pushing to master, it must not accept code (either cmake or c), that does not conform to the format requested. 

On pull request we have the checks, so we can know about it. On the direct commit, if the author does not have locally the `pre-commit`, it seems to me that it can push with whatever format.

PR #4069, opened 2024-12-12
ims_diameter_server: fix retrieval ``$diameter_response`` value

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [x] Related to issue https://lists.kamailio.org/mailman3/hyperkitty/list/sr-users@lists.kamailio.org/thread/CZM47Y5ODFPCJCOCX3PQ4FWKHFNNZGV2/

#### Description
``current_msg_id_repl is only`` changed if ``$diameter_reponse`` is changed by via ``pv_set_response()``. The event_route is executed with a ``fake_msg`` so it will not match if is setted outside anyways.

### DIFF
--- a/src/modules/ims_diameter_server/avp_helper.c
+++ b/src/modules/ims_diameter_server/avp_helper.c
@@ -48,7 +48,6 @@
 
 // ID of current message
 static unsigned int current_msg_id = 0;
-static unsigned int current_msg_id_repl = 0;
 
 cJSON *avp2json(AAA_AVP *avp_t)
 {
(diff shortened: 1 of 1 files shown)


vingarzan:
Hey Victor!

sorry, I tried to understand what's going on there, but I don't get it why the id of a sip_msg is used and how that relates to Diameter, so unable to help much. This code was probably contributed by @carstenbock and it seems to be a workaround to use Kamailio routing script for handling Diameter messages (albeit... due to the extensibility nature of Diameter, that might be a never-ending integration job of translating to JSON...). I'm reassigning to him.

IMHO all modules should have an assigned maintainer and we should have a rule to move orphaned ones to `src/modules.attic/` or something like that, which probably would drop them whenever a major Kamailio version is released and there are no upgrades done in time. If anyone misses and older module, they can revive it later, when they upgrade their core. Hope then keeping order would be easier.

Cheers,
-Dragos

miconda:
The current message id is typically used to cache values or detect if values need to be recomputed.

I think this can be merged, based on the diff I could see in the PR.

Issue #4021, opened 2024-11-12
xhttp_prom module with custom tags


ChristianBergerSipgate:
### Description

In Prometheus there is the possibility to tag values with tags. Those tags can then be used provide context to those values, for example to filter them.

### Expected behavior

It would be great if there was a way to provide custom tags to the metrics. These would essentially just be a static string between the name of the metric and its value.

Example:
`system_load15{host="wurst23.dev.sipgate.net",hostname_short="wurst32"} 0.4`

This could probably be achieved by having an `xhttp_prom_ending` parameter similar to the `xhttp_prom_beginning` as we essentially just want to append a static string right after the metric name.

#### Actual observed behavior
`kamailio_sip_lb_de_core_rcv_replies_1xx_invite 1773 1731419851663`

vhernando:
Hello Christian,

I am a bit busy at the moment with health issues, but you could send a patch maintaining backward compatibility if you want this feature added.

henningw:
A PR was submitted for this, linked above.

henningw:
PR was merged last month, close

PR #4073, opened 2024-12-16
xhttp_prom: new xhttp_prom_tags parameter

ChristianBergerSipgate:
- allows custom tags to be added to prometheus metrics
- Addresses GH issue #4021

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #4021

#### Description

This change adds a new parameter `xhttp_prom_tags`. It allows the user to append tags to the prometheus metrics. This makes those metrics a lot more useful when having multiple kamailio instances running on a single host.

This is done by 2 shared memory helper variables which contain the parameter in different formats. This way it is easy to print them out with minimal changes.

### DIFF
--- a/src/modules/xhttp_prom/doc/xhttp_prom_admin.xml
+++ b/src/modules/xhttp_prom/doc/xhttp_prom_admin.xml
@@ -255,6 +255,31 @@ modparam("xhttp_prom", "xhttp_prom_beginning", "");
 		</programlisting>
 	  </example>
 	</section>
+	<section id="xhttp_prom.p.xhttp_prom_tags">
+	  <title><varname>xhttp_prom_tags</varname> (str)</title>
+	  <para>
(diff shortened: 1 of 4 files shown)


xkaraman:
Hey @ChristianBergerSipgate,

The commit itself should be fixed with `clang-format --style=file -i [filename]` iirc the command or use the `pre-commit` python package to not let you push if it fails the `pre-commit` config and it fixes this for you.

The format check is per commit therefore, pushing a new commit with a format fix won't do it.

After fixing the format, rebase and fixup the changes into 1 commit and then force push, so it can pass the checks.

henningw:
@vhernando any comments to this PR from your side?

henningw:
There seems to be no more comments, merging it. Thanks again. There is one corner case where memory is not properly freed during startup errors (don't think it would cause issues) but I will fix it in another commit.

PR #4090, opened 2024-12-31
examples: fix kamailio ims cfg file for basic volte call

herlesupreeth:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/misc/examples/ims/icscf/icscf.cfg.sample
+++ b/misc/examples/ims/icscf/icscf.cfg.sample
@@ -1,20 +1,25 @@
 # SIP / UDP
 listen=udp:11.22.33.44:4060
 # SIP / TCP
-#listen=tcp:11.22.33.44:4060
+listen=tcp:11.22.33.44:4060
 # SIP / TCP/TLS
 #listen=tls:11.22.33.44:4061
 
 alias=ims.mnc001.mcc001.3gppnetwork.org
 
 #!define NETWORKNAME "ims.mnc001.mcc001.3gppnetwork.org"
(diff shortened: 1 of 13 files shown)


miconda:
Thanks, if no other comment, it will be merge soon -- it affect the additional cfg examples/docs.

henningw (commented):
Thanks for the PR. One generic question as there are quite a lot of changes, does it make sense to maybe keep the existing example cfg and create new ones for the VoLTE setup? If the existing cfg is not working anymore, it obviously don't make sense.

Then one specific comment, you are still using the "modules_k" path in the cfg. This is obsolete since many years and you can remove that path component.

herlesupreeth:
Thanks for the review. 

@henningw 

>  One generic question as there are quite a lot of changes, does it make sense to maybe keep the existing example cfg and create new ones for the VoLTE setup? If the existing cfg is not working anymore, it obviously don't make sense.

I dont mind having a separate folder for VoLTE setup. But, last I checked the existing cfg didnt work but maybe others in the community can confirm this.

> Then one specific comment, you are still using the "modules_k" path in the cfg. This is obsolete since many years and you can remove that path component.

Didn't know about this. I will do the necessary changes

henningw:
Thanks for the feedback. If the existing cfgs are not working I am fine in updating them of course, then they don't need to be kept.

henningw:
Thank you, merged. If somebody is still interested in the old examples, they can be found in the existing older releases. If necessary they could be also added as a second directory etc..

PR #4097, opened 2025-01-06
rtpengine: check corner-case, add comments and refactor some functions

Fr-Soltanzadeh:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

Detailed comments for functions related to subscription are added. Bug fixed.


rfuchs:
Can you describe the corner case this is addressing?

src/modules/rtpengine/rtpengine.c
```diff
@@ -3349,7 +3349,10 @@ static bencode_item_t *rtpp_function_call(bencode_buffer_t *bencbuf,
 	} else {
 		ng_flags.dict = extra_dict;
 		ng_flags.flags = bencode_dictionary_get(ng_flags.dict, "flags");
-		bencode_dictionary_get_str(ng_flags.dict, "call-id", &ng_flags.call_id);
```
Fr-Soltanzadeh, line 3352: When ng_flags.dict does not have the "call-id" key, ng_flags.call_id will be NULL. Therefore, I changed the code to set ng_flags.call_id only if ng_flags.dict contains the "call-id" key.

rfuchs: Ok, you mean you don't want to overwrite/reset `ng_flags.call_id` if it was set already, right? I don't think using `shm_str_dup` is correct though. I don't see a corresponding `shm_free`. Shouldn't it just be `ng_flags.call_id = tmp_callid;` ?

Fr-Soltanzadeh: > Ok, you mean you don't want to overwrite/reset `ng_flags.call_id` if it was set already, right?
I mean I want to overwrite /reset `ng_flags.call_id` only if `ng_flags.dict` contains the "call-id" key. Because `ng_flags.call_id` should not be null.

Fr-Soltanzadeh: > Ok, you mean you don't want to overwrite/reset `ng_flags.call_id` if it was set already, right? 

Not exactly. I want to overwrite/reset ng_flags.call_id only if ng_flags.dict contains the "call-id" key, ensuring it's never null.

> I don't think using `shm_str_dup` is correct though. I don't see a corresponding `shm_free`. Shouldn't it just be `ng_flags.call_id = tmp_callid;` ?

You're right—shm_str_dup isn't necessary here. I updated it to directly assign tmp_callid to ng_flags.call_id. Thanks for pointing that out!

rfuchs:
OK LGTM

Issue #4100, opened 2025-01-10
ims_dialog: missing srdb1 schema definition


linuxmaniac:
there's no ``kamailio-ims_dialog.xml`` at ``src/lib/srdb1/schema`` but there's a manual file created at ``utils/kamctl/mysql/ims_dialog-create.sql``

Same thing for:
- ``utils/kamctl/mysql/ims_charging-create.sql``
- ``utils/kamctl/mysql/ims_usrloc_pcscf-create.sql``
- ``utils/kamctl/mysql/ims_usrloc_scscf-create.sql``

linuxmaniac:
ims_usrloc_* changes are still missing

PR #4101, opened 2025-01-10
lib/srdb1/schema: ims_dialog definitions

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [x] Related to issue #4100

#### Description

db_schema for ims_dialog module


src/lib/srdb1/schema/dialog_in.xml
```diff
@@ -0,0 +1,180 @@
...
+		<name>from_tag</name>
+		<type>string</type>
+		<size>&from_tag_len;</size>
+		<description>The tag parameter serves as a general mechanism to
+		identify a dialog, which is the combination of the Call-ID along
+		with two tags, one from participant in the dialog.</description>
+	</column>
+
+	<column>
+		<name>caller_cseq</name>
```
henningw, line 83: This column was called "caller_original_cseq" previously, and its also like this in the code. We can change it, but then we also need to adapt the module code I think.

src/lib/srdb1/schema/dialog_in.xml
```diff
@@ -0,0 +1,180 @@
...
+
+	<column>
+		<name>caller_sock</name>
+		<type>string</type>
+		<size>&domain_len;</size>
+		<description>Local socket used to communicate with caller</description>
+	</column>
+
+	<column>
+		<name>callee_sock</name>
```
henningw, line 119: This is not needed for the dialog_in table, it was also not in the previous scheme.

henningw (commented):
Thanks for the PR. I did a quick review for the differences. The general standardisation regarding the column sizes is great, the re-ordering of some colums is also fine of course. I have added two remarks that in my opinion should be addressed.

src/lib/srdb1/schema/dialog_in.xml
linuxmaniac: yes, correct. Fixed.

src/lib/srdb1/schema/dialog_in.xml
linuxmaniac: indeed, fixed

henningw:
Thanks, I think its ready for merging then. We might want to merge it after 6.0, to not cause eventual other issues for the packaging.

linuxmaniac:
Packaging? I think this is something missing and 6.0 will be better with the fix

henningw:
Sure, nothing against merging it.

PR #4099, opened 2025-01-10
 Fix Makefile.defs and influxdbc for build on FreeBSD

drTr0jan:
#### Pre-Submission Checklist
- [x] Commit message has the format required by CONTRIBUTING guide
- [x] Commits are split per component (core, individual modules, libs, utils, ...)
- [x] Each component has a single commit (if not, squash them into one commit)
- [ ] No commits to README files for modules (changes must be done to docbook files
in `doc/` subfolder, the README file is autogenerated)

#### Type Of Change
- [x] Small bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would change existing functionality)

#### Checklist:

- [x] PR should be backported to stable branches
- [x] Tested changes locally
- [ ] Not related to issues

#### Description
`Makefile.defs` is incorrect (`man_dir` should contain `share`) and `influxdbc` is not built (`sockaddr_in` is undefined) on FreeBSD environment.
The proposed patch fixes it and makes the module is consistent with the rest of modules.

PR #4104, opened 2025-01-14
cmake: add FreeBSD support

drTr0jan:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
I've tested cmake-build in the FreeBSD environment, as requested by @miconda in #4099.
And founded some troubles.
First of all are: specific flags for FreeBSD and arch-name for FreeBSD cmake-environment. I've fixed it in the PR.

But there're some unresolved problems I want to discuss...

A `LOCK_METHOD` problem:
There's a `LOCK_METHOD` variable in defs.cmake:
https://github.com/kamailio/kamailio/blob/6b0b8cb84b7d0e965d50bdf0dfa5423004879f55/cmake/defs.cmake#L171-L179
And the explicit usage `LOCK_METHOD` as `USE_FUTEX` in a compile_definitions:
https://github.com/kamailio/kamailio/blob/6b0b8cb84b7d0e965d50bdf0dfa5423004879f55/cmake/defs.cmake#L304-L327
I suppose it's incorrect while some OSs (FreeBSD, ex.) doesn't tested with futexlock.h earlier.

A resolv-link problem:
There're an explicit links of the resolv library:
https://github.com/kamailio/kamailio/blob/6b0b8cb84b7d0e965d50bdf0dfa5423004879f55/src/CMakeLists.txt#L116-L124
https://github.com/kamailio/kamailio/blob/6b0b8cb84b7d0e965d50bdf0dfa5423004879f55/utils/kamcmd/CMakeLists.txt#L32
I think it's incorrect while some OSs (FreeBSD, ex.) uses builtin resolv (libc, etc).

Initially сmake is built successful after the above fixes.

### DIFF
--- a/cmake/compiler-specific.cmake
+++ b/cmake/compiler-specific.cmake
@@ -8,7 +8,7 @@ option(PROFILE "Enable profiling" OFF)
 add_library(common_compiler_flags INTERFACE)
 
 # Define the flags for the C compiler
-if(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|amd64")
 
   if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
     target_compile_definitions(common_compiler_flags INTERFACE CC_GCC_LIKE_ASM)
--- a/cmake/defs.cmake
+++ b/cmake/defs.cmake
@@ -14,6 +14,8 @@ message(STATUS "OS version: ${OSREL}")
 
 if(CMAKE_SYSTEM_PROCESSOR MATCHES "i386|i486|i586|i686")
   set(TARGET_ARCH "i386")
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|amd64")
+  set(TARGET_ARCH "x86_64")
 else()
   set(TARGET_ARCH "${CMAKE_SYSTEM_PROCESSOR}")
 endif()
@@ -139,7 +141,7 @@ option(USE_FAST_LOCK "Use fast locking if available" ON)
 # Check the system processor type and set USE_FAST_LOCK accordingly
 if(USE_FAST_LOCK)
   if(CMAKE_SYSTEM_PROCESSOR MATCHES
-     "i386|i486|i586|i686|x86_64|sparc64|sparc|ppc|ppc64|alpha|mips2|mips64")
+     "i386|i486|i586|i686|x86_64|amd64|sparc64|sparc|ppc|ppc64|alpha|mips2|mips64")
     set(USE_FAST_LOCK YES)
   elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "arm64")
     set(USE_FAST_LOCK NO)
--- a/cmake/os-specific.cmake
+++ b/cmake/os-specific.cmake
@@ -12,6 +12,9 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "DragonFly")
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
   # Mac OS X specific flags
   include(${OS_SPECIFIC_DIR}/darwin.cmake)
+elseif(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
+  # FreeBSD specific flags
+  include(${OS_SPECIFIC_DIR}/freebsd.cmake)
 elseif()
   message(FATAL_ERROR "Unsupported system: ${CMAKE_SYSTEM_NAME}")
 endif()
--- a/cmake/os-specific/freebsd.cmake
+++ b/cmake/os-specific/freebsd.cmake
@@ -1,0 +1,37 @@
+message(
+  STATUS
+    "Configuring for FreeBSD"
+)
+
+target_compile_definitions(
+  common
+  INTERFACE HAVE_SOCKADDR_SA_LEN
+            HAVE_GETHOSTBYNAME2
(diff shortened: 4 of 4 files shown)


henningw:
Thanks for testing and the PR. @xkaraman will review it and merge or commit in an adapted version.

xkaraman:
> A `LOCK_METHOD` problem: There's a `LOCK_METHOD` variable in defs.cmake:
> 
> https://github.com/kamailio/kamailio/blob/6b0b8cb84b7d0e965d50bdf0dfa5423004879f55/cmake/defs.cmake#L171-L179
> 
> 
> And the explicit usage `LOCK_METHOD` as `USE_FUTEX` in a compile_definitions:
> https://github.com/kamailio/kamailio/blob/6b0b8cb84b7d0e965d50bdf0dfa5423004879f55/cmake/defs.cmake#L304-L327
> 
> 
I suppose it's incorrect while some OSs (FreeBSD, ex.) doesn't tested with futexlock.h earlier. 

That is the default one used if none is provided. 
If FreeBSD requires some other we can overwrite the value in the `freebsd.cmake` to have a more reasonable one. Or use an empty string and maybe force the user to choose the appropriate one, which seems to break the common build case. 
There should also be a way to offer only supported ones per OS, but that might need some more changes but might be worth it in the end.
@henningw @miconda Thoughts on this?

> A resolv-link problem: There're an explicit links of the resolv library:
> 
> https://github.com/kamailio/kamailio/blob/6b0b8cb84b7d0e965d50bdf0dfa5423004879f55/src/CMakeLists.txt#L116-L124
> 
> 
> https://github.com/kamailio/kamailio/blob/6b0b8cb84b7d0e965d50bdf0dfa5423004879f55/utils/kamcmd/CMakeLists.txt#L32
> 
> 
> I think it's incorrect while some OSs (FreeBSD, ex.) uses builtin resolv (libc, etc).
> Initially сmake is built successful after the above fixes.

So, for you to build you have removed the link against `resolv`? Or you mean did it compile with unnessecary linkage

drTr0jan:
> That is the default one used if none is provided.

I think, the default value should be a `USE_SYSV_SEM` as in Makefiles.defs:
https://github.com/kamailio/kamailio/blob/6b0b8cb84b7d0e965d50bdf0dfa5423004879f55/src/Makefile.defs#L2156-L2160

> If FreeBSD requires some other we can overwrite the value in the `freebsd.cmake` to have a more reasonable one.

Do you think it's need to explicit set `LOCK_METHOD` to `USE_PTHREAD_MUTEX` (example) in the `freebsd.cmake` and also duplicate it as `target_compile_definitions` in os-specific, isn't it?
https://github.com/kamailio/kamailio/blob/6b0b8cb84b7d0e965d50bdf0dfa5423004879f55/cmake/os-specific/darwin.cmake#L25-L27
For example for freebsd.cmake:
```
message(STATUS "USE_FAST_LOCK = ${USE_FAST_LOCK}")
if(NOT ${USE_FAST_LOCK})
  target_compile_definitions(common INTERFACE USE_PTHREAD_MUTEX)
endif()
set(LOCK_METHOD USE_PTHREAD_MUTEX)
```
I think it seems ugly.
Also MacOS doesn't tested with futexlock.h earlier too.

> I think freebsd just uses fastlock if executed on a standard architecture

Yes. But I said, the explicit `USE_FUTEX` as a default isn't correct, in my mind.

> So, for you to build you have removed the link against `resolv`

Yep, I've built with removed link only.

xkaraman:
The PR itself looks ok to be merged. 

Indeed we may have to alter some of the link stuff of `kamailio` and `kamcmd` as noted by @drTr0jan for proper `FreeBSD` support. 

Regarding the https://github.com/kamailio/kamailio/blob/6b0b8cb84b7d0e965d50bdf0dfa5423004879f55/cmake/defs.cmake#L321, it will probably go away and just define it per OS like it was in the makefiles.

cmake/compiler-specific.cmake
```diff
@@ -8,7 +8,7 @@ option(PROFILE "Enable profiling" OFF)
 add_library(common_compiler_flags INTERFACE)
 
 # Define the flags for the C compiler
-if(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
```
xkaraman, line 11: is `amd64` reported by cmake due to `FreeBSD`?

cmake/compiler-specific.cmake
```diff
@@ -8,7 +8,7 @@ option(PROFILE "Enable profiling" OFF)
...
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|amd64")
```
xkaraman, line 11: is amd64 reported by cmake due to FreeBSD?

xkaraman (approved):

cmake/compiler-specific.cmake
drTr0jan: Yes, of course.
```sh
boris@boris:~/sources/kamailio% cmake --system-information | grep CMAKE_SYSTEM_PROCESSOR
CMAKE_SYSTEM_PROCESSOR "amd64"
```

drTr0jan:
> Indeed we may have to alter some of the link stuff of `kamailio` and `kamcmd` as noted by @drTr0jan for proper `FreeBSD` support.
> 
> Regarding the
> 
> https://github.com/kamailio/kamailio/blob/6b0b8cb84b7d0e965d50bdf0dfa5423004879f55/cmake/defs.cmake#L321
> 
> , it will probably go away and just define it per OS like it was in the makefiles.

Sound like a good idea. Should I create a new issue?

PR #4105, opened 2025-01-15
outbound: encode bind address into flow token as destination

alexbakker:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This is an attempt to fix the combination of the outbound/rr modules and ``tcp_accept_haproxy=yes``. This combination is currently broken, as previously reported on the mailing list: https://www.mail-archive.com/sr-users@lists.kamailio.org/msg21854.html

The issue is that the destination address/port combo of the haproxy side is encoded into the flow token. When the rr module decodes the flow token and tries to look up the listening socket using ``find_si``, it cannot be found, because it is given the destination address/port combo of the connection of the haproxy side, not the address/port combo that Kamailio is actually listening on.

This patch fixes the issue by encoding the ``bind_address`` of the socket the connection was received on into the flow token as the destination address. The source address remains unchanged, as it *can* be used to look up the TCP connection (with a minor patch to ``_tcpconn_find`` in ``tcp_main.c``)

While I've done quite a bit of testing with this patchset, I find it difficult to oversee the full impact of this change in Kamailio's large codebase. Any feedback is welcome.

### DIFF
--- a/src/core/tcp_main.c
+++ b/src/core/tcp_main.c
@@ -1739,7 +1739,8 @@ struct tcp_connection *_tcpconn_find(int id, struct ip_addr *ip, int port,
 			print_ip("ip=", &a->parent->rcv.src_ip, "\n");
 #endif
 			if((a->parent->state != S_CONN_BAD) && (port == a->port)
-					&& ((l_port == 0) || (l_port == a->parent->rcv.dst_port))
+					&& ((l_port == 0) || (l_port == a->parent->rcv.dst_port)
+							|| (l_port == a->parent->cinfo.dst_port))
 					&& (ip_addr_cmp(ip, &a->parent->rcv.src_ip))
 					&& (is_local_ip_any
 							|| ip_addr_cmp(l_ip, &a->parent->rcv.dst_ip)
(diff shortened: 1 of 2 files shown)


miconda:
Thanks, merging!

henningw:
Just to add a bit of context, the TCP connection match was changed globally to include the local destination port, its not conditional to the tcp_accept_haproxy=yes.

miconda:
Indeed, but the tcp struct field `cinfo.dst_port` is only set in such cases, otherwise, for normal tcp, is 0.

I let this PR in background for a while to think if other option would be better, but ultimately adding more conditions would add complexity without benefits, because for normal TCP connections `cinfo.dst_port` is 0 and the `(l_port == 0)` was already there and it will match before comparing ` (l_port == a->parent->cinfo.dst_port)`.

```
					&& ((l_port == 0) || (l_port == a->parent->rcv.dst_port)
							|| (l_port == a->parent->cinfo.dst_port))
```

henningw:
I agree that adding more conditions does not make this code path easier to understand. I don't expect any regressions, just wanted to add a bit more info if something is discovered later anyway.

PR #4113, opened 2025-01-16
cmake: Add Windows support

### LINKED ISSUE #4109: cmake: Add Windows support
I suggest port Kamailio to windows, if you agree with me, I can try to port that

- [ ] Make cmake happy to generation build files in Windows #4113 
- [ ] Remove hard dependency to strings.h #4114 
- [ ] Make dprint.h cross platform with `OutputDebugString`
- [ ] Reinvention core threading system with C17 `threads.h`
- [ ] Make core networking system cross platform with `winsock2`

soroshsabz:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [x] Related to issue #4109

#### Description
This PR fixes #4109

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,8 +63,10 @@ add_subdirectory(src)
 
 # Add utils
 # add_subdirectory(utils)
-add_subdirectory(utils/kamctl)
-add_subdirectory(utils/kamcmd)
+if(NOT WIN32)
+  add_subdirectory(utils/kamctl)
+  add_subdirectory(utils/kamcmd)
+endif()
 
 # TODO: Packaging stuuf. These should be on different file propably
 set(CPACK_PACKAGE_VERSION_MAJOR ${PROJECT_VERSION_MAJOR})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,13 @@ cmake_minimum_required(VERSION 3.10)
 
 # Use cmake -DCMAKE_SYSTEM_NAME .. for cross-compiling (inside build directory)
 
+# cmake_policy must be before project for working with toolchain together
+if(WIN32)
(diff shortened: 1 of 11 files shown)


soroshsabz:
blocks by #4114

soroshsabz:
In this step cmake make happy and green to generate build files in Windows

henningw:
Hello, before going further this way you should start a discussion e.g. on the developer list or during one of the next online developer meetings. There will be many changes necessary to really support Windows, and right now I am not 100% sure if this is really a feasible thing to do that will work in the end. In the old Makefile structure there was some support for Cygwin, but I have never really seen this working or used from somebody.

linuxmaniac:
I don't see the value of porting kamailio to windows. But if you want to do it... no one will stop you from doing it, in your fork. Pushing changes just because, makes no sense to me. This will just make things harder to maintain.

soroshsabz:
@linuxmaniac thanks for your reply, many company use Windows environment for IT infrastructure, so if I can port kamailio to Windows, and merge into main repo, windows has come to second citizenship class of kamailio, and many companies can use better kamailio with default their IT environment, and reducing IT staff cost.

I think we have only two big changes, one for networking and another one for threading. in threading changes, I plan to use just more standard approach, so I think not only does it not reduce maintainability, but it also increases it. You are absolutely right about the changes needed for the network, But I think we can limit these changes as much as possible and reduce their scope by abstracting them.

lazedo:
isn't `docker` or `wsl` a valid alternative ? I think we don't need these `windows` specific changes.

soroshsabz:
@lazedo many Windows first environment does not have WSL on Windows Server and does not use Linux container and they use [Windows Container](https://learn.microsoft.com/en-us/virtualization/windowscontainers/about/) instead.

PR #4120, opened 2025-01-23
tcp: fix for local TCP/TLS socket matching with 'tcp_reuse_port' enabled

pinacolada1610:
- Changed the logic for matching a listening TCP/TLS-socket in tcp_reuse_port scenario, now it considers local port as well. With 'tcp_reuse_port' option enabled, the local port is meaningful and helps to differentiate between sockets on the same IP but different ports.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
In our installation Kamailio runs multiple SIP-trunks, including TCP/TLS, which may be bound to the same local IP address but different local ports, and it's crucial for us to know exactly which trunk, i.e. local socket/port, was used to receive a SIP message.
Additionally, the option "tcp_reuse_port" is also enabled in our environment to start outbound connections from the same TCP/TLS ports that we are listening at.

The problem we hit is that identifying a trunk receiving a SIP message by $Rn(Received socket name) $Rp(Received port), which works perfectly fine for UDP trunks, not working for TCP/TLS with "tcp_reuse_port": wrong port and sockname are reported in preudovars.
Investigation showed that it always reports the first listening socket (in the order defined in config file) which matches IP and transport protocol. Port is being ignored here

e.g. if kamailio is listening on:
listen=tcp:192.168.0.11:5060 name "int_tcp_5060"
listen=tcp:192.168.0.11:5062 name "int_tcp_5062"
listen=tcp:192.168.0.11:5063 name "int_tcp_5063"
it will always report $Rn: int_tcp_5060 and $Rp: 5060 for incoming SIP-messages regardless of the actual socket/port being used, even if it's 5062 or 5063
This applies for both TCP and TLS.

Additionally we noticed that the issue happens only in case if an incoming SIP message from remote SBC is coming via the tcp-connection which was initially opened as outbound with well-known ports (rather than via a new incoming tcp-connection with ephemeral src/remote port)

This pointed us to the function core/tcp_main.c:tcp_do_connect() which handles outbound connections.
While opening a new outbound connection to remote SBC, it ignores local/src port while matching it to a listening socket and saves that incorrect info for future use.
This behavior makes sense if tcp_reuse_port is disabled as matching ephemeral port would not give any result.
But with tcp_reuse_port enabled, we need to take the port into consideration as well.

With suggested changes applied we get correct values in $Rn(Received socket name) $Rp(Received port) for inbound messages via all sockets.
Tests with tcp_reuse_port disabled show no change in behavior, as expected.

### DIFF
--- a/src/core/tcp_main.c
+++ b/src/core/tcp_main.c
@@ -1332,6 +1332,7 @@ inline static int tcp_do_connect(union sockaddr_union *server,
 	union sockaddr_union my_name;
 	socklen_t my_name_len;
 	struct ip_addr ip;
+	unsigned short port;
 #ifdef TCP_ASYNC
 	int n;
 #endif /* TCP_ASYNC */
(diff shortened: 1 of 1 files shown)


miconda:
Thanks!

PR #4128, opened 2025-01-29
db_oracle: fix db_oracle cmake build for oracle 18.3.0.0.0

ovidiusas:
This pull request enables db_oracle to be built against oracle 18.3.0.0.0 dev libraries using cmake.

It requires the following env variables to be set:
```
declare -x ORACLE_HOME="/usr/lib/oracle/18.3/client64/"
declare -x ORACLE_INCLUDE_DIR="/usr/include/oracle/18.3/client64"
```


xkaraman:
I think it can be merged from my side! Looks good!

Issue #2923, opened 2021-11-08
Kamailio 5.4.x dialog profiles functionality dead-lock on certain high-load scenarios


henningw:
### Description

The Kamailio 5.4.x dialog profiles functionality can lead to dead-lock on certain high-load scenarios.

The Kamailio dialog profiles are used to track parallel channels for about 200 outgoing PSTN carrier interconnections. During high traffic times (like several thousands parallel calls) the Kamailio server will frequently (e.g. hourly) goes into an end-less loop while executing get_profile_size in the configuration script. This causes the locking for the dialog profiles never be released and Kamailio will stop serving traffic. Internal monitoring tools and RPC commands stay working, as long as they do not touch the dialog functionality.

A similar (dedicated) Kamailio setup is used for tracking parallel channels for customers. Here the dead-lock is not observed that frequently, but aparentely also some crashes happens in a much longer time interval.

### Troubleshooting

After analysis of the back-traces with GDB the get_profile_size() function was removed from the configuration script. After this change the crash did not happened anymore for several days.

#### Reproduction

Issue could not be reproduced so far.

##### bt 1 (some data removed)

(gdb) bt
\# 0  0x00007f57cf3b00da in get_profile_size (profile=0x7f50ccbc7e80, value=0x7ffd9928f300) at dlg_profile.c:859
        n = 364
        i = 12
        ph = 0x7f50d3e4b7d0
\# 1  0x00007f57cf419c67 in w_get_profile_size_helper (msg=0x7f57d699d418, profile=0x7f50ccbc7e80, value=0x7ffd9928f300, spd=0x7f57d6916960) at dialog.c:941
\# 2  0x00007f57cf41a459 in w_get_profile_size3 (msg=0x7f57d699d418, profile=0x7f50ccbc7e80, value=0x7f57d6935118, result=0x7f57d6916960) at dialog.c:982
\# 3  0x0000000000463fea in do_action (h=0x7ffd99293610, a=0x7f57d6936488, msg=0x7f57d699d418) at core/action.c:1094
\# 4  0x00000000004711ee in run_actions (h=0x7ffd99293610, a=0x7f57d6936488, msg=0x7f57d699d418) at core/action.c:1581
\# 5  0x000000000046058b in do_action (h=0x7ffd99293610, a=0x7f57d690fda8, msg=0x7f57d699d418) at core/action.c:700

[...]

PR #4151, opened 2025-02-19
SEGV in dialog module

Den4t:
Fixes for SEGV in dialog module.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
Hi !

The problem - periodically, 2-3 times in а month, kamailio dumps core with SEGV.

Different OS versions (Ubuntu 18,20,22), different kamailio versions from 5.7.1 up to 5.8.4,
high call load, python kemi, actively used dialog module.

Analyzing core files shows more or less the same picture - SEGV happends in dialog mododule
around the dialog vars, can be in rpc call (dlg.list), in destroy dlg, in accessing dlg variable.

Backtrace can look like:

```
Core was generated by `/usr/sbin/kamailio -P /run/kamailio/kamailio.pid -f /etc/kamailio/kamailio.cfg'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  get_dlg_variable_unsafe (dlg=dlg@entry=0x7fa2818849b8, key=key@entry=0x7ffe67990d68) at dlg_var.c:236
236     dlg_var.c: No such file or directory.
(gdb) bt full
#0  get_dlg_variable_unsafe (dlg=dlg@entry=0x7fa2818849b8, key=key@entry=0x7ffe67990d68) at dlg_var.c:236
        var = 0x8
        var_list = <optimized out>
#1  0x00007fa2a2ce0397 in get_dlg_varref (dlg=dlg@entry=0x7fa2818849b8, key=key@entry=0x7ffe67990d68) at dlg_var.c:297
        var = 0x0
        __func__ = "get_dlg_varref"
#2  0x00007fa2a2ccd429 in ki_dlg_var_get_mode (msg=<optimized out>, name=0x7ffe67990d68, rmode=0) at dialog.c:2462
        dlg = 0x7fa2818849b8
        pval = <optimized out>
#3  0x000055e2d5b87cbb in sr_kemi_exec_func (ket=ket@entry=0x7fa2a2d0c0c0 <sr_kemi_dialog_exports+1728>, msg=msg@entry=0x7fa2a34577c0, pno=pno@entry=1, vps=vps@entry=0x7ffe67990d60) at core/kemiexec.c:82
        ret = <optimized out>
        __func__ = "sr_kemi_exec_func"
#4  0x00007fa2a2930376 in sr_apy_kemi_exec_func_ex (ket=ket@entry=0x7fa2a2d0c0c0 <sr_kemi_dialog_exports+1728>, self=self@entry=0x7fa280ce2b80, args=args@entry=0x7fa2804ba3a0, idx=idx@entry=485) at apy_kemi.c:327
        fname = <optimized out>
        i = <optimized out>
        ret = <optimized out>
        vps = {{vtype = 2, v = {n = -2142122656, l = 140335914274144, s = {s = 0x7fa28051cd60 "rtpe_setid", len = 10}, dict = 0x7fa28051cd60}}, {vtype = 0, v = {n = 0, l = 0, s = {s = 0x0, len = 0}, dict = 0x0}}, {vtype = 0, v = {
              n = 0, l = 0, s = {s = 0x0, len = 0}, dict = 0x0}}, {vtype = 0, v = {n = 0, l = 0, s = {s = 0x0, len = 0}, dict = 0x0}}, {vtype = 0, v = {n = 0, l = 0, s = {s = 0x0, len = 0}, dict = 0x0}}, {vtype = 0, v = {n = 0, l = 0,
              s = {s = 0x0, len = 0}, dict = 0x0}}}
        env_P = <optimized out>
        lmsg = 0x7fa2a34577c0
        xret = <optimized out>
        slen = 10
        alen = 1
        pobj = <optimized out>
        __func__ = "sr_apy_kemi_exec_func_ex"
#5  0x00007fa2a2931f45 in sr_apy_kemi_exec_func (self=0x7fa280ce2b80, args=0x7fa2804ba3a0, idx=485) at apy_kemi.c:356
        ket = 0x7fa2a2d0c0c0 <sr_kemi_dialog_exports+1728>
        ret = 0x0
        pstate = 0x0
        pframe = 0x0
        tvb = {tv_sec = 0, tv_usec = 0}
        tve = {tv_sec = 0, tv_usec = 0}
        tz = {tz_minuteswest = 2, tz_dsttime = 0}
        tdiff = <optimized out>
        __func__ = "sr_apy_kemi_exec_func"
#6  0x00007fa2a266f697 in ?? () from /lib/x86_64-linux-gnu/libpython3.8.so.1.0

```

Inspecting dlg vars, the chain look like this:

```
(gdb) frame 0
#0  get_dlg_variable_unsafe (dlg=dlg@entry=0x7fa2818849b8, key=key@entry=0x7ffe67990d68) at dlg_var.c:236
236     in dlg_var.c
(gdb) p *dlg.vars
$1 = {key = {s = 0x7fa2833ea040 "mos", len = 3}, value = {s = 0x7fa2833ea0b0 "4.3", len = 3}, vflags = 1, next = 0x7fa2825fa3d0}
(gdb) p *$.next
$2 = {key = {s = 0x7fa200332e34 <error: Cannot access memory at address 0x7fa200332e34>, len = -1061109568}, value = {s = 0xabcdefed <error: Cannot access memory at address 0xabcdefed>, len = 49}, vflags = 1, next = 0x8}
(gdb)
```

Additionaly, inspecting logs i found that this is happends in SIP race conditions, f.e. when caller and calle sends BYE at the same time and
kamailio processing two different transactions of same dialog in two different workers, so present some competition from identical scenarios.
Long time i can't reproduce this in lab, until insert an artificial delay into the scenario which process BYE request to increase competition.
This code quickly causes SEGV in race conditions with BYE:

```
#kemi get/set dlg var
while time.time()-start < 0.5:
    for i in ('bbb','ccc'):
        dlgvar=KSR.dialog.var_get(i)
        KSR.dialog.var_sets(i,str(random.randint(1,500000)))
```

SEGV not fire if i change dialog var_get/set to kamscript variant:

```
while time.time()-start < 0.5:
    for i in ('bbb','ccc'):
        dlgvar=KSR.pv.get(f"$dlg_var({i})")
        KSR.pv.sets(f"$dlg_var({i})", str(random.randint(1,500000)))

```

By analyzing source i found some places which are not properly covered by locks, and some dangerous places
where direct pointer to dialog var structure used, which can change in concurrent process,
so in first case i insert the locks, in second - it is better to use the clone of variable in private PV buffer.
Also found some mistakes with dialog flags, which can potentionaly affect DMQ and DB dialog behavior.

### DIFF
--- a/src/modules/dialog/dialog.c
+++ b/src/modules/dialog/dialog.c
@@ -1857,27 +1857,29 @@ static str *ki_dlg_get_var_helper(
 {
 	dlg_cell_t *dlg = NULL;
 	unsigned int dir = 0;
-	str *val = NULL;
+	static str val;
 
 	if(sc == NULL || sc->s == NULL || sc->len == 0) {
 		LM_ERR("invalid Call-ID parameter\n");
-		return val;
+		return NULL;
 	}
 	if(sf == NULL || sf->s == NULL || sf->len == 0) {
 		LM_ERR("invalid From tag parameter\n");
-		return val;
+		return NULL;
 	}
 	if(st == NULL) {
 		LM_ERR("invalid To tag parameter\n");
-		return val;
+		return NULL;
 	}
 
 	dlg = get_dlg(sc, sf, st, &dir);
 	if(dlg == NULL)
-		return val;
-	val = get_dlg_varref(dlg, key);
+		return NULL;
+	if(get_dlg_varval(dlg, key, &val) != 0) {
+		return NULL;
+	}
 	dlg_release(dlg);
-	return val;
+	return &val;
 }
 
 /**
@@ -2504,8 +2506,13 @@ static int ki_dlg_var_sets(sip_msg_t *msg, str *name, str *val)
 	int ret;
 
 	dlg = dlg_get_msg_dialog(msg);
+	if(dlg) {
+		dlg_lock(d_table, &(d_table->entries[dlg->h_entry]));
+	}
+
 	ret = set_dlg_variable_unsafe(dlg, name, val);
 	if(dlg) {
+		dlg_unlock(d_table, &(d_table->entries[dlg->h_entry]));
 		dlg_release(dlg);
 	}
 
@@ -2518,7 +2525,7 @@ static int ki_dlg_var_sets(sip_msg_t *msg, str *name, str *val)
 static sr_kemi_xval_t *ki_dlg_var_get_mode(sip_msg_t *msg, str *name, int rmode)
 {
 	dlg_cell_t *dlg;
-	str *pval;
+	str pval;
 
 	memset(&_sr_kemi_dialog_xval, 0, sizeof(sr_kemi_xval_t));
 
@@ -2527,14 +2534,14 @@ static sr_kemi_xval_t *ki_dlg_var_get_mode(sip_msg_t *msg, str *name, int rmode)
 		sr_kemi_xval_null(&_sr_kemi_dialog_xval, rmode);
 		return &_sr_kemi_dialog_xval;
 	}
-	pval = get_dlg_varref(dlg, name);
-	if(pval == NULL || pval->s == NULL) {
+
+	if(get_dlg_varval(dlg, name, &pval) != 0) {
 		sr_kemi_xval_null(&_sr_kemi_dialog_xval, rmode);
 		goto done;
 	}
 
 	_sr_kemi_dialog_xval.vtype = SR_KEMIP_STR;
-	_sr_kemi_dialog_xval.v.s = *pval;
+	_sr_kemi_dialog_xval.v.s = pval;
 
 done:
 	dlg_release(dlg);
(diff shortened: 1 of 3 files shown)


Den4t:
> What was the reasoning by changing the string memory to static and also passing str by value instead of reference? Is this also related to the crash fix?

Hi !
I think you mean this in ki_dlg_get_var_helper:
static str val;

This is because of natute of function get_dlg_varval, whiсh i reuse it to clone dlg var in private pv buffer, we can't use here stack variable (without static), because no garanty the stack is changed after return from ki_dlg_get_var_helper (if I understood the question correctly), using get_dlg_varref as before is unsafe, because of data under the pointer can be changed in concurrent process.

Den4t:
> Thanks for the PR, good point with the additional locks and also the error related to the DMQ flags. There have been similar reports (e.g. #2923) for crashes related to the dialog profiles.

I do not analyze dialog profile functions, but think it is possible that they can have similar problems.

miconda:
Thanks! I split manually the PR in 2, doing differently the part for dlg variables to use directly the static kemi global var, the second being about fixes to the flags.

If you find any related problems, open a bug on the tracker.

Den4t:
> Thanks! I split manually the PR in 2, doing differently the part for dlg variables to use directly the static kemi global var, the second being about fixes to the flags.
> 
> If you find any related problems, open a bug on the tracker.

Thanks !

Should we expect backports?

PR #4158, opened 2025-02-26
dispatcher: weight documentation and workaround for limitations

ChristianBergerSipgate:
- describe limitation of weight parameter
- describe a potential workaround against those limitations

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

We had problems understanding the weight parameter of the dispatcher module and wrote a little workaround against the problems we have found. This workaround in the configuration file improves behavior for setups with many dispatcher targets at the cost of more CPU load.

### DIFF
--- a/src/modules/dispatcher/doc/dispatcher_admin.xml
+++ b/src/modules/dispatcher/doc/dispatcher_admin.xml
@@ -2558,6 +2558,34 @@ setid(int) destination(sip uri) flags(int,opt) priority(int,opt) attrs(str,opt)
 	</section>
 	</section>
 
+	<section id="dispatcher.weight.info">
+		<title>
+		<function moreinfo="none">Limitation on the weight parameter</function>
+		</title>
+		<para>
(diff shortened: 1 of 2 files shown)


miconda:
I do not really get the meaning of:

```
This means that approaching 100 entries the distribution will become more and more uneven.
```

After:

```
The weight is implemented as an internal table with 100 entries. Each entry points to one dispatcher target.
```

Can you elaborate more. Distributing the target to a 100 array is by design, not really a limitation, same that the weights sum has to be 100. A new algorithm can be implemented if the existing ones do not suit some use cases.

From my point of view, this could be better done as a full config (starting from the one in dispatcher docs) that is placed in `misc/examples/`, having at the top in comments the description you added in the PR as docs. The readme for modules try to be reference documentation for what the module implementation offers.

henningw:
@ChristianBergerSipgate Any comments from your side regarding the previous comment?

PR #4166, opened 2025-03-07
ims_registrar_scscf: regpv fix

lyatanski:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/modules/ims_registrar_scscf/regpv.c
+++ b/src/modules/ims_registrar_scscf/regpv.c
@@ -45,6 +45,7 @@ typedef struct _regpv_profile
 	str pname;
 	str domain;
 	str aor;
+	str impi;
 	int flags;
 	unsigned int aorhash;
 	int nrc;
@@ -120,6 +121,11 @@ static void regpv_free_profile(regpv_profile_t *rpp)
 		pkg_free(rpp->aor.s);
 		rpp->aor.s = 0;
 		rpp->aor.len = 0;
+	}
+	if(rpp->impi.s != NULL) {
+		pkg_free(rpp->impi.s);
+		rpp->impi.s = 0;
+		rpp->impi.len = 0;
 	}
 
 	rpp->flags = 0;
@@ -245,6 +251,9 @@ int pv_get_ulc(struct sip_msg *msg, pv_param_t *param, pv_value_t *res)
 		case 17: /* count */
 			return pv_get_sintval(msg, param, res, rpp->nrc);
 			break;
+		case 18: /* impi */
+			return pv_get_strval(msg, param, res, &rpp->impi);
+			break;
 	}
 
 	return pv_get_null(msg, param, res);
@@ -323,6 +332,8 @@ int pv_parse_ulc_name(pv_spec_p sp, str *in)
 				rp->attr = 4;
 			else if(strncmp(pa.s, "cseq", 4) == 0)
 				rp->attr = 9;
+			else if(strncmp(pa.s, "impi", 4) == 0)
+				rp->attr = 18;
 			else
 				goto error;
 			break;
@@ -437,10 +448,18 @@ int pv_fetch_contacts(
 	if(res != 0) {
 		LM_DBG("'%.*s' Not found in usrloc\n", aor.len, ZSW(aor.s));
 		ul.unlock_udomain((udomain_t *)table, &aor);
-		return -1;
+		goto error;
 	}
 
-	ptr = 0; //r->contacts;TODO
+	rpp->impi.s = (char *)pkg_malloc(r->private_identity.len * sizeof(char));
+	if(rpp->impi.s == NULL) {
+		LM_ERR("no mem for impi\n");
+		goto error;
+	}
+	memcpy(rpp->impi.s, r->private_identity.s, r->private_identity.len);
+	rpp->impi.len = r->private_identity.len;
+
+	ptr = r->linked_contacts.head ? r->linked_contacts.head->contact : NULL;
 	ptr0 = NULL;
 	n = 0;
 	while(ptr) {
@@ -455,8 +474,8 @@ int pv_fetch_contacts(
 			goto error;
 		}
 		memcpy(c0, ptr, ilen);
-		//c0->domain = {0,0};//NULL;TODO
-		//c0->aor = {0,0};//NULL;
+		c0->domain = (str){NULL, 0};
+		c0->aor = (str){NULL, 0};
 		c0->next = NULL;
 		c0->prev = NULL;
 
--- a/src/modules/ims_registrar_scscf/doc/ims_registrar_scscf_admin.xml
+++ b/src/modules/ims_registrar_scscf/doc/ims_registrar_scscf_admin.xml
@@ -744,7 +744,7 @@ switch ($retcode) {
       profile)</function></title>
 
       <para>The function fetches the contacts for 'uri' from table 'domain' to
-      pseudo-variable $imssulc(profile) [imssulc = ims scscf ulc].</para>
+      pseudo-variable $ulc(profile) [ulc = ims scscf ulc].</para>
 
       <para>Meaning of the parameters is as follows:</para>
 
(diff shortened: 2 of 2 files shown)


lyatanski:
Hi, any updates or recommendations for improvement?

miconda:
I am not using the module, if no other comments soon, it can be merged.

henningw:
Thanks for the PR. Interesting that the previous name of the PV in the docs was "imsulc" - even if the code defines it otherwise. But as the module defines also other existing functions (save, lookup..) from main registration modules, it can be not used together with them. Could you maybe fix the commit message by adding the module name as prefix? Then it can be merged.

lyatanski:
Hi and thank you for the feedback.

I believe the commit message alredy has the module name as prefix. The merge request name seems to have replaced undersvores with spaces and if this is what you mean - I'll try to update it

src/modules/ims_registrar_scscf/regpv.c
```diff
@@ -440,7 +451,15 @@ int pv_fetch_contacts(
 		return -1;
 	}
 
-	ptr = 0; //r->contacts;TODO
+	rpp->impi.s = (char *)pkg_malloc(r->private_identity.len * sizeof(char));
+	if(rpp->impi.s == NULL) {
+		LM_ERR("no mem for impi\n");
+		return -1;
```
henningw, line 457: Do we need to free the earlier allocated memory here? Maybe also just jump to error?

henningw: I think a similar issue is probably in line 439?

src/modules/ims_registrar_scscf/regpv.c
```diff
@@ -455,8 +474,8 @@ int pv_fetch_contacts(
 			goto error;
 		}
 		memcpy(c0, ptr, ilen);
-		//c0->domain = {0,0};//NULL;TODO
-		//c0->aor = {0,0};//NULL;
+		c0->domain = (str){NULL, 0};
```
henningw, line 477: I don't think you need the casts here, it would be probably more consistent to use the {0, 0} as elsewhere in this function.

henningw (commented):
Thanks for the feedback. Regarding the commit message, I misread it, sorry. I looked to the code, and add a few remarks. Please have a look, thank you.

src/modules/ims_registrar_scscf/regpv.c
lyatanski: yes, I have done it using AOR as example. I'm not sure if it is going to be a problem as in the config script we should have `reg_free_contacts("caller");` and even in failure case it should free the memory.

Anyway, if you recommend to use `goto error;` instead, I have no objections and I'll change it in both places. Please confirm.

src/modules/ims_registrar_scscf/regpv.c
lyatanski: there is compilation error without it. I assume this is the reason it was commented out and although it is not being used, it would probably be better to be initialized/zerroed.

I doubt it is related to particular compiler version, but with `gcc (Alpine 14.2.0) 14.2.0` I get:

```
Building C object src/modules/ims_registrar_scscf/CMakeFiles/ims_registrar_scscf.dir/regpv.c.o
FAILED: src/modules/ims_registrar_scscf/CMakeFiles/ims_registrar_scscf.dir/regpv.c.o 
/usr/bin/cc -DADAPTIVE_WAIT -DADAPTIVE_WAIT_LOOPS=1024 -DARCH=\"x86_64\" -DCC_GCC_LIKE_ASM -DCFG_DIR=\"/usr/local/etc/kamailio/\" -DCOMPILER="\"gcc 14.2.0\"" -DDBG_SR_MEMORY -DDISABLE_NAGLE -DDNS_IP_HACK -DFAST_LOCK -DFMSTATS -DF_MALLOC -DHAVE_ALLOCA_H -DHAVE_EPOLL -DHAVE_GETHOSTBYNAME2 -DHAVE_IP_MREQN -DHAVE_MSGHDR_MSG_CONTROL -DHAVE_MSG_NOSIGNAL -DHAVE_RESOLV_RES -DHAVE_SCHED_SETSCHEDULER -DHAVE_SCHED_YIELD -DHAVE_SELECT -DHAVE_SIGIO_RT -DHAVE_TIMEGM -DHAVE_UNION_SEMUN -DKMSTATS -DKSR_PTHREAD_MUTEX_SHARED -DMALLOC_STATS -DMOD_NAME=\"ims_registrar_scscf\" -DMOD_NAMEID=ims_registrar_scscf -DNAME=\"kamailio\" -DOS=Linux -DOS_QUOTED=\"Linux\" -DPKG_MALLOC -DQ_MALLOC -DRAW_SOCKS -DRUN_DIR=\"/run/kamailio\" -DSHARE_DIR=\"/usr/local/share/kamailio/\" -DSHM_MMAP -DSIGINFO64_WORKAROUND -DSTATISTICS -DTLSF_MALLOC -DTLS_HOOKS -DUSE_DNS_CACHE -DUSE_DNS_FAILOVER -DUSE_DST_BLOCKLIST -DUSE_MCAST -DUSE_NAPTR -DUSE_RAW_SOCKS -DUSE_SCTP -DUSE_TCP -DUSE_TLS -DVERSION=\"6.1.0-dev0\" -DVERSIONVAL=6001000 -D__CPU_x86_64 -D__OS_linux -Dims_registrar_scscf_EXPORTS -isystem /usr/include/libxml2 -O3 -DNDEBUG -std=gnu11 -fPIC -ffile-prefix-map=/code/src/modules/ims_registrar_scscf/= -fPIC -Wall -funroll-loops -Wcast-align -Werror=implicit-function-declaration -Werror=implicit-int -m64 -minline-all-stringops -falign-loops -ftree-vectorize -fno-strict-overflow -mtune=generic -MD -MT src/modules/ims_registrar_scscf/CMakeFiles/ims_registrar_scscf.dir/regpv.c.o -MF src/modules/ims_registrar_scscf/CMakeFiles/ims_registrar_scscf.dir/regpv.c.o.d -o src/modules/ims_registrar_scscf/CMakeFiles/ims_registrar_scscf.dir/regpv.c.o -c /code/src/modules/ims_registrar_scscf/regpv.c
/code/src/modules/ims_registrar_scscf/regpv.c: In function 'pv_fetch_contacts':
/code/src/modules/ims_registrar_scscf/regpv.c:477:30: error: expected expression before '{' token
  477 |                 c0->domain = {NULL, 0};
      |                              ^
ninja: build stopped: subcommand failed.
```

lyatanski:
Hi @henningw,

did you have the chance to check out my comments and confirm the direction we should go for error handling

src/modules/ims_registrar_scscf/regpv.c
henningw: My expection would be that if a script function fails, there is no need to cleanup necessary in the kamailio cfg. So we should do a proper cleanup here and also in line 439. If you compare to the standard registrar module function, it also just uses the "goto error" logic.

henningw:
> Hi @henningw,
> 
> did you have the chance to check out my comments and confirm the direction we should go for error handling

Thanks for the reminder, I have added a comment.

src/modules/ims_registrar_scscf/regpv.c
```diff
@@ -423,7 +434,7 @@ int pv_fetch_contacts(
 	rpp->aor.s = (char *)pkg_malloc(aor.len * sizeof(char));
 	if(rpp->aor.s == NULL) {
 		LM_ERR("no more pkg\n");
-		return -1;
+		goto error;
```
henningw, line 437: Sorry to be annoying, but I would prefer to keep the -1 here, as we are not having any memory allocated yet, and to also stay in sync with the main registrar module implementation

lyatanski: done, returned it back to just exiting, but changed 451 line to `goto error;` as there we have allocated memory for the AOR and have to clean it up before exiting

lyatanski:
the changes have been pushed and the PR branch have been rebased on the master as there were failing checks, now they seem to be passing

Issue #4164, opened 2025-03-05
Topoh module does not hide 180 Ringing and 200 Ok to the caller UE in IMS Lab


Fr-Soltanzadeh:
### Description
In a IMS lab using Kamailio PCSCF, ICSCF and SCSCF default configs, I have used TOPOH module in PCSCF for topology hiding. Topoh config is as follows:
```
modparam("topoh", "mask_key", "my-key")
modparam("topoh", "mask_ip", "127.0.0.2")
modparam("topoh", "sanity_checks", 1)
```
As shown in attached traffic, SCSCF IP is not hided in 180 ringing and 200ok messages to the caller. In invite, ack and bye messages hiding is performed well.

#### Debugging Data

```
(paste your debugging data here)
```

#### Log Messages

```
(paste your log messages here)
```

#### SIP Traffic

```

[topoh.zip](https://github.com/user-attachments/files/19086518/topoh.zip)

```

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
kamailio 5.8.3
```

* **Operating System**:

```
ubuntu:22.04
```

kamailio-sync:
Hi Farzaneh,
You could use topos module instead of topoh.
As alternative way, SEMS works right just for topology hiding.
With best regards.
Mojtaba

Fr-Soltanzadeh:
It seems that `Topoh` module considers that a proxy is only one time in the path of signaling . So, In IMS which two record routes are present for PCSCF, It can not hide IPs properly.

PR #4170, opened 2025-03-09
topoh: fix bug in hiding record routes in reply messages

Fr-Soltanzadeh:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4164

#### Description

This PR fixes https://github.com/kamailio/kamailio/issues/4164.  
It seems that the Topoh module assumes a proxy appears only once in the signaling path. As a result, when it encounters a Record-Route with its IP, it returns immediately. However, in an IMS setup with multiple Record-Route headers for the PCSCF, the module fails to hide the IP addresses properly in the Record-Route headers of the 180 Ringing and 200 OK messages. I suggest that all Record-Route headers should be hidden, except for the last one.

### DIFF
--- a/src/modules/topoh/th_msg.c
+++ b/src/modules/topoh/th_msg.c
@@ -148,12 +148,12 @@ int th_get_uri_type(str *uri, int *mode, str *ip, str *value)
 		if(ret < 0)
 			return -1;
 		if(ret == 1)  /* not found */
-			return 0; /* skip */
+			return 3; /*encode if it is not the last rr with my IP, else skip*/
 		LM_DBG("VALUE [%.*s]\n", value->len, value->s);
 		if(value->len == 2 && strncasecmp(value->s, "on", 2) == 0)
 			*mode = 1;
 		memset(value, 0, sizeof(str));
-		return 0; /* skip */
+		return 3; /*encode if it is not the last rr with my IP, else skip*/
 	}
 	/* not myself & not mask ip */
 	return 1; /* encode */
@@ -554,7 +554,7 @@ int th_flip_record_route(sip_msg_t *msg, str *uri_prefix, str *ip, int mode)
 {
 	hdr_field_t *hdr;
 	struct lump *l;
-	int i;
+	int i, j, last_myself_rr;
 	rr_t *rr;
 	str out;
 	int utype;
@@ -566,6 +566,28 @@ int th_flip_record_route(sip_msg_t *msg, str *uri_prefix, str *ip, int mode)
 		LM_DBG("no record route header\n");
 		return 0;
 	}
+	hdr = msg->record_route;
+	j = 0;
+	while(hdr != NULL) {
+		if(parse_rr(hdr) < 0) {
+			LM_ERR("failed to parse RR\n");
+			return -1;
+		}
+
+		rr = (rr_t *)hdr->parsed;
+		while(rr) {
+			j++;
+			sip_uri_t puri;
+			if(parse_uri(rr->nameaddr.uri.s, rr->nameaddr.uri.len, &puri) < 0)
+				return -1;
+			if(check_self(&puri.host, puri.port_no, 0) == 1) {
+				last_myself_rr = j;
+			}
+			rr = rr->next;
+		}
+		hdr = next_sibling_hdr(hdr);
+	}
+
 	hdr = msg->record_route;
 	i = 0;
 	act = 0;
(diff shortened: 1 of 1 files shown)


Fr-Soltanzadeh:
@miconda 
Thank you for the feedback. I agree that my initial approach—only hiding up to the last Record-Route (RR) header—was incorrect, as it doesn’t account for scenarios where the Topoh proxy (e.g., P-CSCF) is not the last entity in the RR chain.

However, I believe there’s a misunderstanding about the specific issue in the IMS topology I’m working with. Let me clarify the problem more precisely.

Scenario
In an IMS setup, due to spiraling behavior (e.g., S-CSCF -> I-CSCF -> S-CSCF), the same proxy (like P-CSCF) may appear multiple times in the Record-Route list. For example, in the 180 Ringing or 200 OK responses sent from the core network back to the UE, the RR headers might look like:

Record-Route: PCSCF
Record-Route: SCSCF
Record-Route: SCSCF
Record-Route: PCSCF

Here, P-CSCF (which is running Topoh) appears twice—once at the beginning and once at the end. However, the current Topoh implementation works by scanning RRs top-down and stops at the first occurrence of its own IP. That means in this example, none of the RR headers are hidden, because the first occurrence of its IP is at the top of the list.
In this case, Topoh fails to hide any RR headers, which results in exposing internal IPs (like those of S-CSCF). This breaks the topology hiding requirement expected in IMS.

Suggested Behavior
Instead of stopping at the first match of its own IP in the RR headers, Topoh should:
* First, scan all Record-Route headers to find the last occurrence of its own IP.
*Then, hide all Record-Route headers up to that position.

This ensures that:
All internal hop headers before and including the last instance of the Topoh proxy are hidden.
Any legitimate Record-Route headers after the Topoh proxy (e.g., from a downstream proxy or UE) remain intact.

I'd appreciate your thoughts on whether this approach aligns with the original design goals of Topoh, especially with respect to spiraling.

Fr-Soltanzadeh:
@miconda 
Hi again. I’d appreciate any feedback when you get a chance, as I’m pending your response to move forward.

Fr-Soltanzadeh:
@miconda 
Hi again. I've implemented the solution I explained above.  
Could you please take another look and let me know whether the new solution properly implements the record‑route hiding logic in spiral behavior? Thanks!

miconda:
Travelling lately because of the Easter holidays, I haven't had the chance to look much at it yet. Is it about un/masking the IPs in the SIP response, right?

Fr-Soltanzadeh:
Yes. In Record-Route headers.

miconda:
The masking/unmasking cannot be done using the last one corresponding to Topoh-Server, because it has to be done in the way that the intermediary nodes see the same Record-Route as for the INVITE.

Lets take the example:

```
UA1 <=> P1 <=> P2(TH) <=> P3 <=> P4 <=> P3 <=> P2(TH) <=> P1 <=> UA2
```

When INVITE goes to UA2, the Record-Route headers leaving a proxy are:

  - at UA1
  - at P1: P1
  - at P2(TH): P2, X1
  - at P3: P3, P2, X1
  - at P4: P4, P3, P2, X1
  - at P3: P3, P4, P3, P2, X1
  - at P2(TH): P2, X3, X4, X3, X2, Y1
  - at P1: P1, P2, X3, X4, X3, X2, Y1
  - at UA2: P1, P2, X3, X4, X3, X2, Y1

The P is clear address, X is encoded address, Y is double encoded address.

When the response goes the opposite direction, it has to be:

  - at UA2: P1, P2, X3, X4, X3, X2, Y1
  - at P1: P1, P2, X3, X4, X3, X2, Y1
  - at P2(TH): X1, P2, P3, P4, P3, P2, X1
  - at P3: X1, P2, P3, P4, P3, P2, X1
  - at P4: X1, P2, P3, P4, P3, P2, X1
  - at P3: X1, P2, P3, P4, P3, P2, X1
  - at P2(TH): Y1, X2, X3, X4, X3, P2, P1
  - at P1: Y1, X2, X3, X4, X3, P2, P1
  - at UA1: Y1, X2, X3, X4, X3, P2, P1

Using the last route matching the TH-Server is not correct, I assume that you set mask ip to be P2, so all the X and Y record-routes have P2 address. If the mask ip is different that P2 (it can be any address, say 127.0.0.2), then it should work as it is now. If mask ip is P2 address, then when response comes first to it, it has to encode the record-routes towards UA2 via P1 and decode those towards UA1 via P3. When it comes second time, it has to encode those towards UA2 via P3 and decode those via UA1 via P1.

PR #4182, opened 2025-03-21
DMQ improvements related to node status

stefan-mititelu-idt:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

Some improvements related to DMQ node status:
1. Allow other modules that use DMQ to detect NOT_ACTIVE nodes via default response callback. Note that only internal dmq ping mechanism can set them back to DMQ_NODE_ACTIVE status.
2. Simplify node status and merge DMQ_NODE_PENDING + DMQ_NODE_TIMEOUT into DMQ_NODE_NOT_ACTIVE since those specific states were never used to be compared against.
3. Add new mod parameters which will control after how many fails the node should be moved to DMQ_NODE_NOT_ACTIVE state.

Under testing currently. Will update if any bugs found.


stefan-mititelu-idt:
Testing finished and looks ok with this last commit updates.

henningw:
Thanks for the PR. I don't fully understand the commits. They are all prefixed with "dmq" but also changing htable and presence module. If this was done intentionally, please format the commit messages with the respective prefixes. There is also a format check error, but this seems to be related to the script, we can investigate later.

stefan-mititelu-idt:
Yes, this is done intentionally.

I've added them under "dmq: " because: those changes in htable, presence, dialog, dmq_usrloc (e.g. remove locally defined dmq reponse callback and use NULL instead) are strictly coupled with the change in dmq (add default response callback to be used when those modules use NULL). Shall I still reformat the commit in this case?

linuxmaniac:
> Shall I still reformat the commit in this case?

Yes, we use prefix to track changes in modules easily from commit messages.

And while at it, can you please rebase?

PR #4192, opened 2025-03-27
cmake_format all files

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,8 @@ project(
   kamailio
   VERSION 6.1.0
   DESCRIPTION "Kamailio SIP Server"
-  HOMEPAGE_URL "https://www.kamailio.org")
+  HOMEPAGE_URL "https://www.kamailio.org"
+)
 
 # Set the version number
 set(EXTRAVERSION "-dev0")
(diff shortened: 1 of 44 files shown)


linuxmaniac:
Can you please rebase first. I would like to check the recent changes in cmake-format checks in GitHub PR action

xkaraman:
@linuxmaniac rebased

xkaraman:
i am not sure but i think the PR action check was not honored before. i can see in the config file that `dangle_parens = True` was set already but the PR checks didn't fail on older PRs. (I am quite sure there were a cmake PRs before this change with the wrong format)

linuxmaniac:
>  Hope your new way with script it does take into account the config file.

What do you mean?

xkaraman:
As i said before:

> i am not sure but i think the PR action check was not honored before. i can see in the config file that `dangle_parens = True` was set already but the PR checks didn't fail on older PRs. (I am quite sure there were a cmake PRs before this change with the wrong format).
> 
> EDIT: I just saw you made some changes last week regarding the `cmake-format` with some script. i was not referring to this but to the older way of checking, just to be clear. Hope your new way with script it does take into account the config file.

I think this is the first PR regarding CMake after you have introduced the scripts. Before it was done directly (I guess?) but it seemed that it was not honoring the config file (realised this due to config already have the `dangle_parens = True` but the format in files was not like that) . 

The script check now seems to pass, but let me push a wrongly formatted file to verify it.

Cheers,
Xenofon

xkaraman:
As you can see pushing a wrongly formatted file like `kamcmd/CMakeLists.txt` did NOT trigger a failure regarding its format. 

`cmake-format.py` in master has the check enabled (even before this PR) that it should trigger fail the check.

linuxmaniac:
Yes, something is missing. I will fix it ASAP. Thanks

linuxmaniac:
> Yes, something is missing. I will fix it ASAP. Thanks

fixed at e9852eff76

henningw:
More necessary changes or can it be merged?

PR #4195, opened 2025-03-31
pua_json: add support for as-feature-event

ThomasSevestre:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description

This PR allow usage of the new presence_dfks module with pua_json

### DIFF
--- a/src/modules/pua_json/defs.h
+++ b/src/modules/pua_json/defs.h
@@ -68,12 +68,20 @@
 #define MWI_JSON_FROM "From"
 #define MWI_JSON_TO "To"
 
+#define AS_FEATURE_JSON_CALLID "Call-ID"
+#define AS_FEATURE_JSON_FROM "From"
+#define AS_FEATURE_JSON_FROM_USER "From-User"
+#define AS_FEATURE_JSON_FROM_REALM "From-Realm"
+#define AS_FEATURE_JSON_DND_STATUS "DND-Status"
+#define AS_FEATURE_JSON_EXPIRES "Expires"
+
 #define TO_TAG_BUFFER_SIZE 128
 #define FROM_TAG_BUFFER_SIZE 128
 #define SENDER_BUFFER_SIZE 1024
 #define DIALOGINFO_BODY_BUFFER_SIZE 8192
 #define MWI_BODY_BUFFER_SIZE 2048
 #define PRESENCE_BODY_BUFFER_SIZE 4096
+#define AS_FEATURE_BODY_BUFFER_SIZE 4096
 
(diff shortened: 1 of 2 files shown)


src/modules/pua_json/defs.h
```diff
@@ -139,4 +147,12 @@
 </dialog>\
 </dialog-info>"
 
+#define AS_FEATURE_BODY \
+	"<?xml version=\"1.0\" encoding=\"ISO-8859-1\"?>\
+<DoNotDisturbEvent xmlns=\"http://www.ecma-international.org/standards/ecma-323/csta/ed3\">\
```
henningw, line 152: This URL returns a 404, seems that they changed their page structure

src/modules/pua_json/pua_json_publish.c
```diff
@@ -359,6 +360,57 @@ int pua_json_publish_dialoginfo_to_presentity(struct json_object *json_obj)
...
+
+	json_api.extract_field(json_obj, AS_FEATURE_JSON_DND_STATUS, &dnd_status);
+
+	struct json_object *ExpiresObj =
+			json_api.get_object(json_obj, AS_FEATURE_JSON_EXPIRES);
+	if(ExpiresObj != NULL) {
+		expires = json_object_get_int(ExpiresObj);
+	}
+
+	len = snprintf(body, AS_FEATURE_BODY_BUFFER_SIZE, AS_FEATURE_BODY,
```
henningw, line 398: We probably want to check if you've got a from and dnd value from the json before doing the snprintf. Compare to line 294 in this file, for example. If there are other mandatory fields we need to get from the json, we should also check for them. If they are not present, just go to the error case.

henningw (commented):
Thanks for the PR. I have added a few remarks to the changes, please have a look.

src/modules/pua_json/defs.h
ThomasSevestre: I've used the same link as in presence_dfks module. I'm not sure what link should be used

src/modules/pua_json/pua_json_publish.c
ThomasSevestre: I've added check on `from_user` and `dnd_status`

src/modules/pua_json/defs.h
henningw: Ok, if its the same link - then lets keep it in sync.

henningw:
Thanks for the additions. Lets wait a bit for more feedback, otherwise it will be merged in a few days.

PR #4221, opened 2025-04-23
app_jstd: Added dynamic buffer sizing to load bigger .js files and Added ability to load all JavaScript files from a specified directory

nickgashcc:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

- The code that loaded a JavaScript file used a fix buffer length of 128K on the stack.  This has been changed so that the file size is determined and an attempt to allocate an appropriate buffer size temporarily is made.  The file contents are then loaded into that buffer before being passed to the duktape engine.
- In addition a new module param 'loaddir' has been added that allows you to specify a directory containing .js files rather than specifying a single .js file to load with the existing 'load' param.  If loaddir is set it will take a higher priority than load.  All .js files in the directory are loaded into a temporary buffer and combined before passing to the duktape engine.   This allows you to split logic/routes into separate .js files but load them all into the JavaScript engine.
- Updated documentation


nickgashcc:
Force pushed to resolve some clang formatting issues and commit typo

miconda:
Thanks for the PR!

It would be good to keep coherence for local variable naming style, e.g., using lowercase -- now there are `Result` and `len`, or `FullBuffer` and `totallen`, or `sizeToAllocate`.

Also, declare the variables at the beginning of the block (e.g., `size_t name_len = strlen(dir->d_name);` is in the middle).

nickgashcc:
thanks for the feedback - I'll make those changes

PR #4228, opened 2025-04-29
core: docs: Update type identifiers that can be used in the RPC API

xkaraman:
- format as well

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [] PR should be backported to stable branches
- [ ] Tested changes locally
- [x] Related to issue #4227

#### Description

There was missing information of type identifiers that can be used in [rpc tutorial docs](https://www.kamailio.org/docs/docbooks/5.7.x/rpc_api/rpc_api.html#rpc.data_types). This is an update to those docs. The most recent I could find was for 5.7. 

Maybe there is a need for manual updating the page?

Also during the relevant PR #4227 , we found the `t` identifier is never used in the code. since it can overflow on 64bit systems due to being a `long` and not `int` type when saved/load, maybe we can remove it completely?. One can just cast the `time_t` to the relevant type and use the correct identifier as done in https://github.com/kamailio/kamailio/blob/2081b3aac394223b68b849fbac2a9e5ac4392334/src/modules/dispatcher/dispatcher.c#L1978-L1990

### DIFF
--- a/doc/tutorials/rpc/kamailio_rpc.xml
+++ b/doc/tutorials/rpc/kamailio_rpc.xml
@@ -16,7 +16,7 @@
         <pubdate role="cvs">$Date$</pubdate>
     </sectioninfo>
     -->
- 
+
     <title>
 	The Kamailio RPC Control Interface
     </title>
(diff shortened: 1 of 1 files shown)


miconda:
Thanks! Maybe we should migrate these tutorials to markdown and keep them on a more visible place on git repo.

Issue #4232, opened 2025-05-07
rtpengine: improve compatibility of per-leg statistics parsing


tsearle:
### Description

For getting A and B leg QoS stats, Kamailio expects there to be a SSRC key inside the Stream object.

However some versions of rtpengine (tested with mr13.3.1.1) do not send this key.  Looking at the master version of rtpengine, I see this key has been added to correspond to ingress SSRCs[0].

This patch adds a checking ingress SSRCs[0] as a fallback if SSRC
#### Reproduction

use kamailio 6.0.1 and rtpengine mr13.3.1.1 using baresip as the calling and called party

the following in the kamailio.cfg

```
modparam("rtpengine", "mos_A_label_pv", "$avp(mos_A_label)")
#The name of a pseudovariable to hold the minimum encountered MOS value for the call. The value typically has a range of 1.0 through 5.0.
modparam("rtpengine", "mos_min_A_pv", "$avp(mos_min_mos_A)")
#The name of a pseudovariable to hold the amount of packetloss in percent at the time the minimum MOS value was encountered;
modparam("rtpengine", "mos_min_packetloss_A_pv", "$avp(mos_min_packetloss_A)")
#The name of a pseudovariable to hold the amount of jitter in milliseconds at the time the minimum MOS value was encountered;
modparam("rtpengine", "mos_min_jitter_A_pv", "$avp(mos_min_jitter_A)")
#The name of a pseudovariable to hold the packet round-trip time in microseconds at the time the minimum MOS value was encountered;
modparam("rtpengine", "mos_min_roundtrip_A_pv", "$avp(mos_min_roundtrip_A)")

#The name of a pseudovariable to hold the maximum encountered MOS value for the call. The value typically has a range of 1.0 through 5.0.
modparam("rtpengine", "mos_max_A_pv", "$avp(mos_max_mos_A)")
#The name of a pseudovariable to hold the amount of packetloss in percent at the time the maximum MOS value was encountered;
modparam("rtpengine", "mos_max_packetloss_A_pv", "$avp(mos_max_packetloss_A)")
#The name of a pseudovariable to hold the amount of jitter in milliseconds at the time the maximum MOS value was encountered;
modparam("rtpengine", "mos_max_jitter_A_pv", "$avp(mos_max_jitter_A)")
#The name of a pseudovariable to hold the packet round-trip time in microseconds at the time the maximum MOS value was encountered;
[...]

PR #4233, opened 2025-05-07
rtpengine: improve compatibility of rtpengine per call leg stats parsing

tsearle:
#### Pre-Submission Checklist

#### Type Of Change
- [*] Small bug fix (non-breaking change which fixes an issue)

#### Checklist:

- [*] PR should be backported to stable branches
- [*] Tested changes locally
- [*] Related to issue #4232

#### Description
not all versions of rtpengine sends a key SSRC per stream.  For those that do not the same information can be found in ingress SSRCs.  Add logic to check for the SSRC value there if the SSRC key is not present.


src/modules/rtpengine/rtpengine.c
```diff
@@ -4545,9 +4546,21 @@ static void parse_call_stats_1(struct minmax_mos_label_stats *mmls,
...
-			if((ssrc = bencode_dictionary_get_integer(stream, "SSRC", -1))
-					== -1)
-				continue;
+			ssrc = bencode_dictionary_get_integer(stream, "SSRC", -1);
+			if(ssrc == -1) {
+				ingress_ssrcs = bencode_dictionary_get_expect(
+						stream, "ingress SSRCs", BENCODE_LIST);
+				if(!ingress_ssrcs || !ingress_ssrcs->child)
+					continue;
+				LM_DBG("rtpengine: XXX got ingress SSRCs\n");
```
rfuchs, line 4555: Is this here intentionally or as a leftover?

tsearle: I'm not sure I fully understand the question... but this is here intentionally to make sure that ingress_ssrcs exists and has at least 1 element.

and the log message to indicate the affirmative

tsearle: so in the case there is no SSRC but an "ingress SSRC" the log looks as follows
```
rtpengine: XXX got ingress SSRCs
rtpengine: found SSRC '1778154774' for label 'Bleg_label'
```

henningw: I also noticed that. Generally we should try to have log messages that are understandable and have some meaning. I don't get the meaning of this XXX in this case. ;-)

tsearle: all the log messages in the method have XXX in them... I was simply matching the existing code

Issue #4187, opened 2025-03-24
tcp_send_timeout does not apply for outbound connections


ovedeneev:
### Description

[tcp_send_timeout](https://www.kamailio.org/wikidocs/cookbooks/5.8.x/core/#tcp_send_timeout) option does not apply for outbound connections. 

#### Reproduction

1. Establish TCP OPTIONS between Kamailio and target node.
2. On target node block incoming traffic from Kamailio node using iptables.
3. Notice on Kamailio node that TCP retransmission packets don't follow tcp_send_timeout settings, instead they fall back to kernel's net.ipv4.tcp_retries2 settings.

### Additional Information

  * **Kamailio Version**
 
```
# kamailio -V
version: kamailio 5.8.2 (aarch64/linux) 446039
flags: USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, MEM_JOIN_FREE, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT-NOSMP, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLOCKLIST, HAVE_RESOLV_RES, TLS_PTHREAD_MUTEX_SHARED
ADAPTIVE_WAIT_LOOPS 1024, MAX_RECV_BUFFER_SIZE 262144, MAX_SEND_BUFFER_SIZE 262144, MAX_URI_SIZE 1024, BUF_SIZE 65535, DEFAULT PKG_SIZE 8MB
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
id: 446039
compiled on 14:32:09 Nov 14 2024 with gcc 8.5.0
```

* **Operating System**:

```
Linux xxx 4.18.0-553.6.1.el8.aarch64 #1 SMP Thu May 30 04:10:32 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux
```

sergey-safarov:
To reproduce this issue, you need to start Kamailio with config file
```
listen=eth0

tcp_send_timeout=3

loadmodule "xlog.so"
loadmodule "pv.so"
loadmodule "ctl.so"
loadmodule "jsonrpcs.so"
loadmodule "tm.so"

[...]

PR #4234, opened 2025-05-09
core: tcp_send_timeout applied for outbound connections

sergey-safarov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #4187

#### Description
core: tcp_send_timeout applied for outbound connections


sergey-safarov:
Question.
Is required output next message using `LM_DBG`
```
LM_INFO("Set TCP_USER_TIMEOUT=%d ms\n", optval);
```

henningw:
Thanks for the PR. Regarding the question about the "Set TCP_USER_TIMEOUT" log, in my opinion its ok to keep it with INFO for now, also to inform users about it, as in the past it was not applied. If its too noisy we can always set to DBG later one.

miconda:
Thanks!

PR #4245, opened 2025-05-20
Adding kemi exports play_dtmf event_callback and dlg_req_within

ggontariu:

- [x] Tested changes locally

#### Description
Added missing kemi exports for RTP Engine module play_dtmf event_callback functions and Dialog module dlg_req_within function


src/modules/dialog/dialog.c
```diff
@@ -2602,6 +2602,26 @@ static int ki_dlg_var_is_null(sip_msg_t *msg, str *name)
 	return -1;
 }
 
+/**
+ *
+ */
+static int ki_dlg_req_within4(struct sip_msg *msg, char *side, char *method,
+		char *content_type, char *content)
```
miconda, line 2609: The `char*` params have to be changed to `str*`, that's the prototype expected by KEMI engine. FTR, the `str->s` values are NULL terminated, so they can be used for functions that expect `char*`.

src/modules/dialog/dialog.c
```diff
@@ -2602,6 +2602,26 @@ static int ki_dlg_var_is_null(sip_msg_t *msg, str *name)
...
+{
+	dlg_cell_t *dlg;
+	int ret;
+
+	dlg = dlg_get_msg_dialog(msg);
+	if(dlg == NULL) {
+		return 1;
+	}
+	ret = w_dlg_req_with_content(dlg, side, method, content_type, content);
```
miconda, line 2618: Calling this function is not the correct way, because it is the function exported to kamailio.cfg, and its parameters are expected to be `fixed-up` at startup, being evaluated at runtime for cfg vars.

I split the function w_dlg_req_with_headers_and_content() in commit 0cb4dd3282ec12cdaefca491f9ed43acd0f21bdd, you need to update this functuin to use the new one ki_dlg_req_with_headers_and_content() directly.

miconda (commented):
Thanks for submissions, there are some comments made inline within the patches.

src/modules/dialog/dialog.c
ggontariu: Fixed

src/modules/dialog/dialog.c
ggontariu: fixed

PR #4246, opened 2025-05-20
topos: Allow contact a/b host like user part 

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

In this PR we introduce a new mode and two new parameters for accomplishing the same thing for user part also in domain part.

Docs incoming if no significant changes are needed.


miconda:
Haven't gone much into it, but I think it is ok, still needs docs.

Would it work if the values are also with more than the host part (e.g., `host:port;some=param`)?

xkaraman:
Hey @miconda,

As of now this will copy whatever was in the modparam string to the URI including port and params. 

Bearing in mind the append_host function tries to append the port and transport if available as well. So, it could end up with two ports and params not at the end for example, so it depends on what's available, i think

`user@host:port;some=param:port_if_available;transport=if_found`

xadhoom:
Just jumping in, without searching for open PRs I implemented a similar thing, the only difference is that I don't have a specific flag (contact_mode) for it, but just work on the host part like the xavu contact does.

Your seems more complete, I have only a question: if I'm not mistaken, with this PR you cannot enable at same time the xavu contact_user and contact host. What if a user needs to manage at same time the user part and the host part?

xkaraman:
Yeah i think we have discussed this when implementing it. I think right now it's not possible to manage both user and host for both sides. 

This might require a seperate mode or maybe combine all this into one mode if it makes sense?

xadhoom:
In my approach I've just avoided the flag for contact_mode, and relied on the presence of both a/b params for host, like contact_host xavu is handled.  

A quick "works for me" diff is here https://github.com/kamailio/kamailio/compare/master...VoiSmart:kamailio:add_xavu_field_a-b_contact_host?expand=1

In that way you can have manage both user part and host part, and be compatible with existing contact_host.

miconda:
Merging. 

@xadhoom : further enhancements can be done with another PR.

thanhloi2603:
Hi, I just wanted to input a memoir that applying this PR may introduce a regression if the contact mode was configured to 0 by default. i.e: there were port and transport before but there were none after.

This commit [9d14693cb41a838b210222d625ac2754139b5d41](https://github.com/kamailio/kamailio/pull/4246/changes/9d14693cb41a838b210222d625ac2754139b5d41) in particular in deleting the duplicated code block

hoilchoi:
> Hi, I just wanted to input a memoir that applying this PR may introduce a regression if the contact mode was configured to 0 by default. i.e: there were port and transport before but there were none after.
> 
> This commit [9d14693cb41a838b210222d625ac2754139b5d41](https://github.com/kamailio/kamailio/pull/4246/changes/9d14693cb41a838b210222d625ac2754139b5d41) in particular in deleting the duplicated code block

+1 using default contact_mode 0 caused removal of port and transport in contact header.

henningw:
@thanhloi2603 @hoilchoi Thanks for bringing it up. Does it happen all the time, or e.g. only when you are using a protocol different then UDP and another port as 5060?

hoilchoi:
@henningw  Yea, it was noticed when using TLS with different port than 5060.

thanhloi2603:
Hi @henningw, got a PR for restoring the port and transport in #4699.
For other people to migrate from 6.0.x to hopefully 6.1.2 do not encounter the regression.

Issue #4252, opened 2025-05-22
Duplicated presentities (and notifies) when usrloc and presence are replicated via dmq and pua_reginfo used


xadhoom:
### Description

A kamailio cluster which replicates presentities and usrloc via dmq may incur into duplicated presentities and notifications depending on config.

The problematic setup, where each kamailio node has the following setup:
- presence enabled with dmq
- usrloc distributed with dmq
- pua_reginfo used to publish reginfo to self

Really is multiplicated by the numbers of nodes, so 4 nodes -> 4 times notifications and presentities.

### Troubleshooting
The problem arise because pua_reginfo hooks to usrloc and every time a "save()" is done a publish is sent, which is by design. Then the handled publish generates a presentity which is replicated via dmq from presence module.

Now, if also usrloc is replicated via DMQ, the handling of such message still trigger the hook, causing another publish with generates a different presentiy, since is new and the e-tag is different.

This causes subscribers to receive N (number of nodes) notifications on every status change.

### Possible Solutions

The approach can be to optionally disable that hook (basically making pua_reginfo skipping publish) when handling a message via DMQ, so the usrloc replica is just saved.

In other modules, like pua_dialoginfo or pua_usrloc this can be already done, since publishing can be controlled from script.

### Additional Information

  * **Kamailio Version**

```
6.0.1
```

xadhoom:
I think this has been merged in #4253 , so it can be closed.

miconda:
Thanks for the heads up!

PR #4253, opened 2025-05-22
pua_reginfo: add helper to disable publishing for current message

xadhoom:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4252

#### Description
This PR adds a new helper `reginfo_disable_publish` to `pua_reginfo` module which disabled the publishing for the current message, which is useful to control when to actually send it or not (eg in DMQ replication scenarios).

The helper is "negative" (it *disables* not enables) on purpose in order to mantain backward compatibility to current behaviour.

### DIFF
--- a/src/modules/pua_reginfo/doc/pua_reginfo_admin.xml
+++ b/src/modules/pua_reginfo/doc/pua_reginfo_admin.xml
@@ -238,6 +238,33 @@ reply_route[1] {
 				</programlisting>
 			</example>
 		</section>
+		<section id="pua_reginfo.f.reginfo_disable_publish">
+			<title>
+				<function moreinfo="none">reginfo_disable_publish()</function>
+			</title>
+			<para>
+				This function disables sending the PUBLISH request in the registrar
+				callback invocation, for the current message only.
+			</para>
+
+			<example>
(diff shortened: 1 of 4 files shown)


henningw:
> Also, the functions are split to be ready to be added to kemi, only later I found out that pua_reginfo is not kemi aware yet. I can simplify the two functions in one, leave as is or add all kemi stuff, too

Not sure if I fully understood your question. If you also want to add KEMI support for pua_reginfo module, its better to do it as a dedicated PR and commit, as this makes reviewing and merging easier. The KEMI could be also backported, which is not possible for new functionality.

xadhoom:
> Not sure if I fully understood your question. If you also want to add KEMI support for pua_reginfo module, its better to do it as a dedicated PR and commit, as this makes reviewing and merging easier. The KEMI could be also backported, which is not possible for new functionality.

No I don't want to add kemi here. I was just commenting of my own PR which has 2 functions, one of which is "ready" for kemi. I can just leave as is, or collapse into one function (because without kemi is not really needed), that's all.

xadhoom:
Updated with docs.

miconda:
Thanks!

Issue #4273, opened 2025-06-06
expose subscripe request/answer commands in rtpengine module


tsearle:
### Description

For forking media and sip rec flows the following rtpengine commands need to be exposed
* subscribe reqeust
* subscribe answer
* unsubscribe

currently these commands are only exposed for internal api use.

This patch exposes these commands in both kemi and the kamailio scripting language
### Possible Solutions

merge the corresponding pr

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
version: kamailio 6.0.1 (aarch64/linux) 
flags: USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, MEM_JOIN_FREE, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT-NOSMP, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLOCKLIST, HAVE_RESOLV_RES, TLS_PTHREAD_MUTEX_SHARED
ADAPTIVE_WAIT_LOOPS 1024, MAX_RECV_BUFFER_SIZE 262144, MAX_SEND_BUFFER_SIZE 262144, MAX_URI_SIZE 1024, BUF_SIZE 65535, DEFAULT PKG_SIZE 8MB
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
id: unknown 
compiled with gcc 12.2.0```

* **Operating System**:

```
Linux dev-ew1-blue-app-208 6.1.0-37-cloud-arm64 #1 SMP Debian 6.1.140-1 (2025-05-22) aarch64 GNU/Linux
```

PR #4274, opened 2025-06-06
rtpengine: add support for commands needed for siprec

tsearle:
add rtpengine_subscribe_request rtpengine_subscribe_answer and rtpengine_unsubscribe methods to both th cfg and kemi interfaces

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4273 

#### Description
new kemi and cfg script commands added.  existing internal api modifed in a backwards compatible way (though no means of testing this)

### DIFF
--- a/src/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/src/modules/rtpengine/doc/rtpengine_admin.xml
@@ -3194,8 +3194,291 @@ rtpengine_manage();
 </programlisting>
 		</example>
         </section>
+		<section id="rtpengine.f.rtpengine_subscribe_request">
+			<title>
+				<function moreinfo="none">rtpengine_subscribe_request(flags,sdp_avp,to_tag_avp,stream_xavp[,via-branch])</function>
+			</title>
+			<para>
+				Generates &sdp; body for forking the current &rtp; stream and stores it in the
+				xavp specified <quote>sdp_avp</quote> AVP. The new call leg will be identified
+				by the generated (or specified) to tag that will be put in the
+				<quote>to_tag_avp</quote> AVP.
+				on INVITE for the cases the &sdp; bodies are in INVITE and 200 OK and on 200 OK
+				when &sdp; bodies are in 200 OK and ACK.
+			</para>
+			<para>
+				The function will return true on success and false (-1) on various failures,
+				like using rtpengine_subscribe_request() before an rtpengine session was established.
+			</para>
+			<para>Meaning of the parameters is as follows:</para>
+			<itemizedlist>
+				<listitem>
+					<para>
+						<emphasis>flags</emphasis> - flags to turn on some features.
+					</para>
+					<para>The <quote>flags</quote> string is a list of space-separated items. Each item
+						is either an individual token, or a token in <quote>key=value</quote> format. The
+						possible tokens are described below. This offer SDP can be manipulated with the same flags as
+						used in <function>rtpengine_offer</function> additionally, the following extra flags can be used
+					</para>
+					<itemizedlist>
+						<listitem><para>
+							<emphasis>from-tags=...</emphasis> - Specifies which participants in the call should have
+							their media forked. This value can be repeated multiple times to specify multiple
+							participants. To specify all participants, use <quote>all</quote> as an alternative.
+						</para></listitem>
+						<listitem><para>
+							<emphasis>all</emphasis> - forks the media stream for all participants in the call. Use
+							instead of <quote>from-tags=...</quote>
+						</para></listitem>
+						<listitem><para>
+							<emphasis>siprec</emphasis> - indicates that the &sdp; should be in SIPREC format. The flag
+							<quote>all</quote> should also be used in this case.
+						</para></listitem>
+					</itemizedlist>
+					<para>
+						Check also the documentation of RTPEngine, these flags are documented there as well:
+						<ulink url="https://github.com/sipwise/rtpengine">https://github.com/sipwise/rtpengine</ulink>.
+					</para>
+				</listitem>
+				<listitem>
+					<para>
+						<emphasis>sdp_avp</emphasis> - the name of the AVP where the generated &sdp; will be stored.
+					</para>
+					<para>The <quote>sdp_avp</quote> parameter is mandatory and must be a writable AVP.
+						The AVP will be set to the generated &sdp; body, which can then be used in a subsequent
+						<quote>t_uac_send()</quote> call to send the offer to the new participant.
+					</para>
+				</listitem>
+				<listitem>
+					<para>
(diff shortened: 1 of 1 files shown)


src/modules/rtpengine/rtpengine.c
```diff
@@ -3376,10 +3426,12 @@ static int parse_flags(struct ng_flags_parse *ng_flags, struct sip_msg *msg,
...
 					ng_flags->transport |= 0x100;
 					ng_flags->transport &= ~0x002;
-				} else if(str_eq(&key, "TOS") && val.s)
+				} else if(str_eq(&key, "TOS") && val.s) {
 					bencode_dictionary_add_integer(
 							ng_flags->dict, "TOS", atoi(val.s));
-				else
+				} else if(str_eq(&key, "all")) {
+					ng_flags->all_legs = 1;
+				} else
```
rfuchs: I think this is redundant? Just putting the word "all" in the argument string would already lead to "all" being put in the flags list without the need for special handling, no?

src/modules/rtpengine/rtpengine.c
```diff
@@ -3441,7 +3493,17 @@ static int parse_flags(struct ng_flags_parse *ng_flags, struct sip_msg *msg,
...
+					err = "missing value";
+					if(!val.s)
+						goto error;
+
+					if(!ng_flags->from_tags) {
+						ng_flags->from_tags =
+								bencode_list(ng_flags->dict->buffer);
+					}
+					bencode_list_add_str(ng_flags->from_tags, &val);
+				} else
```
rfuchs, line 3541: Ok if this is how you want to do it, but some suggestions for alternatives: `subscribe request` can take either a single `from-tag` or multiple `from-tags` as a list, and would treat them the same. So here you could also treat "from-tag" and "from-tags" the same, and making building the list conditional on the opmode.

Or: omit handling altogether and require a modern version of rtpengine which parses the flags internally (`parsed_by_module` false). In this case you can use syntax like `rtpengine_blah("flag flag ... from-tags=[tag1 tag2 tag3] ...")`

Up to you, just a hint.

src/modules/rtpengine/rtpengine.c
```diff
@@ -500,6 +519,18 @@ static cmd_export_t cmds[] = {
...
+	{"rtpengine_subscribe_offer", (cmd_function)rtpengine_subscribe_request_wrap_f, 5,
+	fixup_rtpengine_subscribe_request_v, fixup_free_rtpengine_subscribe_request_v, ANY_ROUTE},
+	{"rtpengine_subscribe_answer", (cmd_function)rtpengine_subscribe_answer_wrap_f, 1,
+	fixup_spve_null, fixup_free_spve_null, ANY_ROUTE},
+	{"rtpengine_subscribe_answer", (cmd_function)rtpengine_subscribe_answer_wrap_f, 2,
+	fixup_spve_spve, fixup_free_spve_spve, ANY_ROUTE},
+	{"rtpengine_unsubscribe", (cmd_function)rtpengine_unsubscribe_wrap_f, 1,
+	fixup_spve_null, fixup_free_spve_null, ANY_ROUTE},
+	{"rtpengine_unsubscribe", (cmd_function)rtpengine_unsubscribe_wrap_f, 2,
+	fixup_spve_spve, fixup_free_spve_spve, ANY_ROUTE},
```
rfuchs, line 535: These would have to be added to the documentation

rfuchs (commented):
An example of the intended usage would be helpful

src/modules/rtpengine/rtpengine.c
tsearle: I'm adding the documentation as a 2nd commit onto this PR

src/modules/rtpengine/rtpengine.c
tsearle: from_tags and from_tag are different keys in rtpengine.

from_tag is used to identify the call, from_tags is used as an alternative to all to specify which participants to for streams for

src/modules/rtpengine/rtpengine.c
tsearle: I was hoping so too,
unknown flags get put into ng_flags->flags

however
https://github.com/kamailio/kamailio/blob/7de2b9b7625d9cb856944627a3a5d7c167006dc6/src/modules/rtpengine/rtpengine.c#L3594

ignores ng_flags->flags when doing the encoding.  I was not brave enough to change this behaviour as the effect would be quite global!

rfuchs: I don't quite follow. The flags are constructed within `rtpp_function_call` and then put into the dictionary: https://github.com/kamailio/kamailio/blob/7de2b9b7625d9cb856944627a3a5d7c167006dc6/src/modules/rtpengine/rtpengine.c#L3730

src/modules/rtpengine/rtpengine.c
rfuchs: Yes they're different keys, but for a subscribe request putting `from-tag=foobar` is identical to `from-tags=[foobar]`

But doesn't really matter, anyway.

tsearle:
Example usage included in the documentation

src/modules/rtpengine/doc/rtpengine_admin.xml
```diff
@@ -3169,8 +3169,291 @@ rtpengine_manage();
...
+					<itemizedlist>
+						<listitem><para>
+							<emphasis>to-tag=...</emphasis> - value of the to tag returned by
+							<quote>rtpengine_subscribe_request</quote>
+						</para></listitem>
+						<listitem><para>
+							<emphasis>from-tag=...</emphasis> - value of the from tag of the original call.
+						</para></listitem>
+						<listitem><para>
+							<emphasis>callid=...</emphasis> - value of the call id of the original call.
```
rfuchs: `call-id` with a dash (also below)

tsearle: fixed

src/modules/rtpengine/rtpengine.c
tsearle: I've re-tested with the "all" flag removed and it still gets put in flags.  So the only thing lost is enforcing that either "all" or "from_tags" is sent, but not both. updating the patch to remove this code

src/modules/rtpengine/rtpengine.c
tsearle: I looked some more about merging the logic.... from-tag is a string in flags, so it doesn't support specifying it multiple times...

also there is som logic for checking that from-tag is set when no sip_msg is present,  so I feel more comfortable keeping it as is

PR #4275, opened 2025-06-06
presence: preserve etag uniqueness when marking presentity for removal

Ozzyboshi:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [ ] Related to issue

#### Description
When the `presence` module attempts to clean up records in the `presentity` table, it replaces the `etag` with a static, hardcoded string:
```
 "*#-OFFLINE-#*" 
```
However, the `presentity` table enforces a uniqueness constraint on the following combination of fields:

```
MariaDB [kamailio]> show create table presentity;
| presentity | CREATE TABLE `presentity` (
...
  UNIQUE KEY `presentity_idx` (`username`,`domain`,`event`,`etag`),
```
This approach breaks the uniqueness of etag, which is intended to differentiate records , especially in cases such as parallel calls, or when using kamailio as a forking proxy. 

Instead of overwriting the etag with the static offline marker, this patch appends the marker to the existing etag value. This ensures the offline flag is retained while preserving the uniqueness of the etag.

The deletion query was also updated accordingly: it now identifies offline records using a LIKE '%*#-OFFLINE-#*' clause.

Based on testing, the duplicate key errors no longer occur, and the cleanup logic now behaves as expected. This patch improves the reliability of presentity cleanup without requiring schema or index changes.

Please consider merging into master.

Thanks!

### DIFF
--- a/src/modules/presence/presentity.c
+++ b/src/modules/presence/presentity.c
@@ -2273,6 +2273,8 @@ int mark_presentity_for_delete(presentity_t *pres, str *ruid)
 	str *cur_body = NULL, *new_body = NULL;
 	db_query_f query_fn = pa_dbf.query_lock ? pa_dbf.query_lock : pa_dbf.query;
 
+	LM_DBG("mark_presentity_for_delete called\n");
+
 	if(pres->event->agg_nbody == NULL) {
 		/* Nothing clever to do here... just delete */
 		if(delete_presentity(pres, NULL) < 0) {
@@ -2359,6 +2361,8 @@ int mark_presentity_for_delete(presentity_t *pres, str *ruid)
 		goto done;
 	}
 
+	LM_DBG("Found 1 presentity as expected\n");
+
 	row = RES_ROWS(result);
 	value = ROW_VALUES(row);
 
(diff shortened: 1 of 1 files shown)


miconda:
I expect that introducing the use of raw_query breaks the usage of the module with non-sql backends (e.g., db_text), so another solution has to be found.

Maybe leveraging another column (e.g., priority set to -1 for cleaning up records) or adding a new column.

linuxmaniac:
but this problem is about unique keys. So it's a relational database problem no?

miconda:
Well, if I understood correctly, the code replaces the unique ETags of the presentity records with a predefined value (ie., `#-OFFLINE-#'`) in order to delete the records later, but that can lead to infringement of unique constraints. So the current behaviour seems to be an uninspired `solution` based on this report, so we have to look for alternatives that are compatible with the internal DB API, without relying on raw queries.

Ozzyboshi:
I’m proposing an alternative solution, which I had previously developed and appears to work well.

This approach uses the "event" field to mark records with #-OFFLINE-#, allowing us to retain the etag, which is the key element that differentiates records.

One major advantage of this solution is that it eliminates the need for raw SQL queries. In the current implementation, I had to rely on raw SQL because I needed to retrieve all records ending with #-OFFLINE-#, and I couldn’t find a way to achieve that without raw queries.

linuxmaniac:
Maybe it's time to introduce a new field into the schema. This approach doesn't feels right. If we don't want to relay on raw queries. I would introduce a ``offline`` (boolean) field to mark the rows so it can be filtered with a simple query

miconda:
If a new column is introduced, I would name it a bit more generic (e.g., `status`), so it can be eventually reused if new processing modes are needed in the future, for now it can be 0 for active, 1 for delete.

Ozzyboshi:
ok but... someone is willing to create the PR? Should I do it? Changing db schema is required I need to go over the procedure in this case because I have no clue how to do it.

linuxmaniac:
@Ozzyboshi I will create a PR with that approach

Ozzyboshi:
ok  in this case i will just close this PR

linuxmaniac:
@Ozzyboshi I forgot this branch. Can you please test the solution from your side?

Ozzyboshi:
sorry victor just saw your comment now, i think this branch is part of our solution, i can check it, why are you interested? do you have the same problem?

linuxmaniac:
> @Ozzyboshi I will create a PR with that approach

Well I said I was going to do it

> Maybe it's time to introduce a new field into the schema. This approach doesn't feels right. If we don't want to relay on raw queries. I would introduce a ``offline`` (boolean) field to mark the rows so it can be filtered with a simple query

Ozzyboshi:
ok perfect if upstream is able to provide a solution i can remove mine from my codebase

linuxmaniac:
@Ozzyboshi did you manage to check it?

Ozzyboshi:
not yet but we are having some issues with presentity and I will take this into consideration in this period

PR #4276, opened 2025-06-10
pv_headers: Add remove all headers function 

fprdkhti:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

This function removes all the headers. it is useful when calling it and setting `skip_headers` modparam. `pvh_apply_headers()` function remove all the headers except those specified in the `skip_headers` .

### DIFF
--- a/src/modules/pv_headers/doc/functions.xml
+++ b/src/modules/pv_headers/doc/functions.xml
@@ -128,6 +128,18 @@
 		pvh_collect_headers()</link> or with <quote>auto_msg</quote> parameter enabled.
 		</para>
 	</section>
+	<section id="pv_headers.f.pvh_remove_all_headers">
+		<title>
+		<function moreinfo="none">pvh_remove_all_headers()</function>
+		</title>
+		<para>
+		Removes all sip message headers. 
+		</para>
+		<para>
+		This function can be used from ANY_ROUTE but only after <link linked="pv_headers.f.pvh_collect_headers">
+		pvh_collect_headers()</link> or with <quote>auto_msg</quote> parameter enabled.
+		</para>
+	</section>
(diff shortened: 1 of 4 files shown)


miconda:
Shouldn't the function be better named `pvh_remove_all_headers` instead of `pvh_remove_all_header`?

@linuxmaniac: any other comments on this PR?

fprdkhti:
> Shouldn't the function be better named `pvh_remove_all_headers` instead of `pvh_remove_all_header`?
> 
> @linuxmaniac: any other comments on this PR?
Yes. I fixed it. 
Thanks.

src/modules/pv_headers/doc/functions.xml
```diff
@@ -128,6 +128,18 @@
 		pvh_collect_headers()</link> or with <quote>auto_msg</quote> parameter enabled.
 		</para>
 	</section>
+	<section id="pv_headers.f.pvh_remove_all_headers">
+		<title>
+		<function moreinfo="none">pvh_remove_all_headers()</function>
+		</title>
+		<para>
+		Removes all headers from the XAVP.
```
linuxmaniac: Uhm... this it's not clear to me. It seems this will only remove the headers that are in the message, not the headers that could be added via ``$x_hdr``, for instance. 

If that's what you want... please explain it here. Because I would expect all the headers to be removed not only the ones in the message.

linuxmaniac: please rebase before sending the next version

miconda:
@linuxmaniac: is this ok to merge?

linuxmaniac:
@fprdkhti please document this new feature on https://github.com/kamailio/kamailio-wiki at docs/features/new-in-devel.md

PR #4280, opened 2025-06-11
rr: loose_route_mode: add vmode bit 2 to skip outbound

furmur:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

`loose_route()` in `rr` checks request src ip/port against ip/port in flow tokens to detect direction and decide whether to apply flow tokens routing:
```
static inline int process_outbound(struct sip_msg *_m, str flow_token)
{
...
	} else if(!ip_addr_cmp(&rcv->src_ip, &_m->rcv.src_ip)
			  || rcv->src_port != _m->rcv.src_port) {
		LM_DBG("\"incoming\" request found. Using flow-token for "
			   "routing\n");
...
```

there are cases when remote UA changes ip/port for in-dialog messages and this approach does not work.
incoming message has route set with flow tokens and src ip/port different from ones in the flow tokens.
as a result, `loose_route()` routes message back to the client using flow tokens instead of using the normal part of the Route headers as expected.

new `force_lr_no_outbound_flag` allows to control flow tokens usage in the `loose_route()` when we know direction based on the some high level config logic (for example: by the result of `ds_is_from_list()` call)
UPD:
changed to use new _vmode_ value in `loose_route_mode(vmode)` instead of the `force_lr_no_outbound_flag` param

### DIFF
--- a/src/modules/rr/doc/rr_admin.xml
+++ b/src/modules/rr/doc/rr_admin.xml
@@ -446,11 +446,16 @@ if(!loose_route_preloaded()) {
 	<section id="rr.f.loose_route_mode">
       <title><function moreinfo="none">loose_route_mode(vmode)</function></title>
 
-	  <para>The function is similar to `loose_route()`, but it does only
-	  loose routing processing if vmode==1, skipping the testing of r-uri==myself
-	  for performing strict routing. If vmode==0, it behaves like loose_route().
+      <para>The function is similar to `loose_route()`,
+      but it accepts route mode bitmask as <emphasis>vmode</emphasis> parameter.
       </para>
-      <para>It is a convenient function to use with application servers that
+      <para>
+      If bit one is set, then do only loose routing processing, skipping the testing of r-uri==myself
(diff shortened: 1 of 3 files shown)


miconda:
I haven't looked at the code or tried to understand properly the use case, but I would suggest that the new parameter has a more generic name, like `mode` (or `lr_mode` if it is more specific to loose-routing), with the value being set based on flags (bits), so later it can be reused if a new matching (or processing) mode needs to be added.

In general is not recommended to have long names for parameters, what they supposed to do should be in the docs, not necessarily reflected by their names.

furmur:
@miconda, as an option, i can also rewrite it as an additional mode id for `loose_route_mode(vmode)`. what do you think?

furmur:
changed to use new _vmode_ value in `loose_route_mode(vmode)` instead of the `force_lr_no_outbound_flag` param

miconda:
I think it is better so, more flexible via function param.

PR #4289, opened 2025-06-20
dialplan: improve performance by reusing PCRE structure

lbalaceanu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

By reusing the pcre_md structure needed by pcre2_match function we see improvements of 10x in the time of dp_translate() function call (via benchmark module).

### DIFF
--- a/src/modules/dialplan/dp_repl.c
+++ b/src/modules/dialplan/dp_repl.c
@@ -409,7 +409,7 @@ int rule_translate(sip_msg_t *msg, str *instr, dpl_node_t *rule,
 	int repl_nb, offset, match_nb, rc, cap_cnt;
 	struct replace_with token;
 	struct subst_expr *repl_comp;
-	pcre2_match_data *pcre_md = NULL;
+	static pcre2_match_data *pcre_md = NULL;
 	str match;
 	pv_value_t sv;
 	str *uri;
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the PR - nice catch. I don't use this particular module that often, lets wait a bit for other feedback.

src/modules/dialplan/dp_repl.c
```diff
@@ -427,6 +427,14 @@ int rule_translate(sip_msg_t *msg, str *instr, dpl_node_t *rule,
 		return 0;
 	}
 
+	if(pcre_md == NULL) {
+		pcre_md = pcre2_match_data_create(MAX_REPLACE_WITH, NULL);
```
linuxmaniac, line 431: if we don't set the second parameter we are no longer using ``pcre2_compile_context *dpl_ctx`` that we were using implicitly when using  ``pcre2_match_data_create_from_pattern`` so we no longer using ``pcre2_malloc/pcre2_free`` are you aware of that change?

Since you are using ``static`` now I'm assuming we don't really need to use shm here. But, nevertheless, we will end up using plain ``malloc/free`` and not ``pkg_malloc/pkg_free``

lbalaceanu: Hi @linuxmaniac - thank you, I was not aware of that compile context usage you mentioned. However, I also think that because we are using a static variable we can get away with allocating via malloc once. Please tell me if you want it changed however.

linuxmaniac:
fine for me

lbalaceanu:
Can I merge then?

PR #4297, opened 2025-06-24
permissions: Perform LPM to find the longest matching subnet

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR allows the functions `allow_address_group` and `allow_source_address_group` to search for Longest Prefix Match (LPM) when searching for an IP is in a subnet instead of the first found. 

Since this is a stricter check, i don't think we require an extra param for it, but feel free to suggest otherwise if there are any use-cases that required the first matched.

### DIFF
--- a/src/modules/permissions/doc/permissions_admin.xml
+++ b/src/modules/permissions/doc/permissions_admin.xml
@@ -221,6 +221,25 @@
 		(see tag_col module parameter) is added as value to
 		peer_tag AVP if peer_tag_avp module parameter has been defined.
 		</para>
+		<note>
+		<para>
+			Starting with Kamailio version 6.1.x, the <function>allow_address()</function>
+			function and its related functions use the Longest Prefix Match (LPM) method to
+			find matching entries.
+		</para>
+		<para>
+			This means the <function>_group</function> variants will now return the most specific
+			(longest) subnet match, instead of the first match (which was previously the entry with
+			the lowest group ID).
+			This LPM behavior is now consistent across the following functions:
+		</para>
+		<itemizedlist>
+			<listitem><para><function>allow_address()</function></para></listitem>
+			<listitem><para><function>allow_source_address()</function></para></listitem>
+			<listitem><para><function>allow_source_address_group()</function></para></listitem>
+			<listitem><para><function>allow_address_group()</function></para></listitem>
+		</itemizedlist>
+		</note>
 	</section>
 	<section id="sec-trusted-requests">
 		<title>Trusted Requests</title>
@@ -1259,6 +1278,9 @@ if (allow_uri("basename", "$avp(i:705)") {  // Check URI stored in $avp(i:705)
 		matches any port. The <quote>group_id</quote> argument can be an integer
 		string or a pseudo variable.
 		</para>
+		<note>
+		See <link linkend="sec-address-permissions"> Address permissions </link> for more details.
+		</note>
 		<para>
 		This function can be used from REQUEST_ROUTE, FAILURE_ROUTE.
 		</para>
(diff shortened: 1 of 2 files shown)


miconda:
It is a stricter check, but still changing the existing behaviour, so I am not sure what would be the best. Personally I don't think that I am affected, I don't remember having need to have records of overlapping subnets, but maybe there are others that are affected, so they have to comment.

My concern comes from another perspective, whether now the behaviour is consistent with the matching a subnet record with `allow_[source_]address()`. If one has overlapping subnets in the same group, will the same record as `allow_[source_]address_group()` be matched? Even I don't have it so far, one reason I could think of right now to have overlapping subnets is to return different values from the tag column.

So whatever mode it ends up with, new parameter or not, matching should be coherent across the module functions.

xkaraman:
@miconda Thanks for the review!

Indeed, the `allow_source_address` could return a different tag value with this approach! So with the new commit a similar LPM is performed to verify whether is in that group for overlapping domains! 

We confirmed that now, both functions return the same tag!

miconda:
I am fine to merge it, but a note has to be added in the docs for the affected functions to be clear how the matching is done with this commit.

xkaraman:
@miconda 

Added notes regarding the search in the 4 related functions.

miconda:
The note seems to be for `allow_[source_]address_group()` because it says `which was previously the entry with the lowest group ID`. For `allow_[source_]address()`, the matching is within same group, but still can be different network prefixes. Maybe that parenthesis should be removed or clarified is for the get-the-group functions.

I also see you mentioned the version 6.0.3, I am not sure about backporting, is it considered a bug fix? I guess this has to be discussed in a larger group. Or add a mod param, defaulting to old behaviour and backport in 6.0 and then switch to the new behaviour in 6.1, so nobody has unexpected surprises after upgrading within 6.0.x series in the future.

Also, instead of having the note for each function, maybe is better added to the section at the top about address permissions:

  - https://www.kamailio.org/docs/modules/stable/modules/permissions.html#sec-address-permissions

miconda:
Thinking more about it, a modparam would be good to have anyhow, because the PR has also impact on performance. 

Instead of returning on first match, now all the records are tried always, so, for example, even when the match would be the 1st or the 10th within 100 records, all the 100 records are tried to see if there is a longer match.

xkaraman:
> The note seems to be for `allow_[source_]address_group()` because it says `which was previously the entry with the lowest group ID`. For `allow_[source_]address()`, the matching is within same group, but still can be different network prefixes. Maybe that parenthesis should be removed or clarified is for the get-the-group functions.
> 

Ah good catch, yes i should clarify that about the group variants.

> I also see you mentioned the version 6.0.3, I am not sure about backporting, is it considered a bug fix? I guess this has to be discussed in a larger group. Or add a mod param, defaulting to old behaviour and backport in 6.0 and then switch to the new behaviour in 6.1, so nobody has unexpected surprises after upgrading within 6.0.x series in the future.
> 

No, not a bug fix. Just confused and wrote it for the next release, but  didn't consider if its a bug. I will change to 6.1.x or maybe remove it completerly.

> Also, instead of having the note for each function, maybe is better added to the section at the top about address permissions:
> 
> * https://www.kamailio.org/docs/modules/stable/modules/permissions.html#sec-address-permissions

Hmm, yeah could do it there as well, but will it be read if someone goes directly to the function doc? Maybe if i add link to it as the note in functions.



> Thinking more about it, a modparam would be good to have anyhow, because the PR has also impact on performance.
> 
> Instead of returning on first match, now all the records are tried always, so, for example, even when the match would be the 1st or the 10th within 100 records, all the 100 records are tried to see if there is a longer match.

Yeah that's true. Not sure if it's noticable in that range of records, but for a lot more it might be a problem.

xkaraman:
@miconda Docs updated. are these better?

henningw:
Merged, thanks. Further adaptions can be done in git master if necessary.

miconda:
For the records, I added a new module parameter `subnet_match_mode` to control if the subnet matching is on first prefix or longest prefix, it feels unnecessary to match always all the records when usually is a single match. The default is first prefix match to ensure backward compatibility and don't affect deployments on upgrade.

PR #4302, opened 2025-06-27
core: Add null and empty check for AVP name

Tomohare:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Ensure that the pointer 'name' is not null and its size not negative.

- Returns 0 in case of invalid name values: null or empty or name length have a negative value.


Tomohare:
@linuxmaniac Please, could you have a look a this.

miconda:
Have you faced any issue that requires such check? As I can see in the code, `match_by_name()` is used only once inside `search_next_avp()`, which expects to have the name set, because one avp was found already.

Tomohare:
Yes, we had a core dump in that region, for out of an abundance of caution this check was added.

linuxmaniac:
Can you show an example of such coredump?

Tomohare:
Here is the coredump: [coredump_kamailio.txt](https://github.com/user-attachments/files/20941539/coredump_kamailio.txt)

Tomohare:
PS: this occur on Debian Trixie

linuxmaniac:
``s`` is clearly not empty. That value is comming from:
```
#4  0x00007ff1b164d7bc in pv_get_avp (msg=0x7ff1b5cade00, param=0x7ff1b5fdca08, res=0x7fff555caca8) at ./src/modules/pv/pv_core.c:1849
[...]
state = {flags = 497, id = 64, name = {n = 140676117154143, s = {s = 0x7ff1b5fdc95f "eventapi_lock)", len = 13}, re = 0x7ff1b5fdc95f}, avp = 0x6874656d5f687365}
```

I would assume the issue here is our friend ``transaction is gone`` instead. The check will not help.

miconda:
I guess you still have the coredump, can you get the output with gdb for the next commands?

```
frame 4
print *param
```

The avp name seems to be set down in the trace (`eventapi_lock`), is it used like `$avp(eventapi_lock)` or via a variable as name (e.g., `$avp($var(name))`)?

Tomohare:
The coredump was deleted by mistake. We will be re-running tests to try to generate this error again. Also, we are using AVP as this `$avp(s:eventapi_lock)`.

Tomohare:
Here is the result of those commands with a new coredump:

```
(gdb) frame 4
#4  0x00007f87918217bc in pv_get_avp (msg=0x7f8795ec8100, param=0x7f87961f6d08, res=0x7ffdbb3a7b38) at ./src/modules/pv/pv_core.c:1849
1849		if((avp = search_first_avp(name_type, avp_name, &avp_value, &state)) == 0)
(gdb) print *param
$1 = {pvn = {type = 0, nfree = 0x0, u = {isname = {type = 1, name = {n = 140220315954271, s = {s = 0x7f87961f6c5f "eventapi_lock)", len = 13}, re = 0x7f87961f6c5f}}, dname = 0x1}}, pvi = {type = 4, u = {ival = 0, dval = 0x0}}}
```

miconda:
It looks good, so likely something happened with the avp.

The backtrace shows that params/vars are optimized, but let's try to see them anyhow, in gdb do:

```
frame 0
p avp_name
p *avp_name
p avp
p *avp
p name
p *name
p id
```

Tomohare:
```
(gdb) p *avp_name
value has been optimized out
(gdb) p avp
$1 = (avp_t *) 0x205049532e797469
(gdb) p *avp
Cannot access memory at address 0x205049532e797469
(gdb) p name
$2 = (str *) 0x7ffdbb3a77a8
(gdb) p *name
$3 = {s = 0x7f87961f6c5f "eventapi_lock)", len = 13}
(gdb) p id
$4 = 64
```

miconda:
The `name` seems to be ok, so the patch of the PR won't save it. The `avp` seems to be the issue.

Tomohare:
Yes, will be closing it since it doesn't do what I expected to do (and the problem seems to be in the avp as miconda said).

Issue #4266, opened 2025-05-29
branch_failure_route does not handle reply while wt_timer period


IgorrG:
### Description

While implementing mid-register inspired by pull request (https://github.com/kamailio/kamailio/pull/3360) we found the case when it goes impossible to handle responses for registrar servers. In case some of registrars busy and respond with 401 after 200ok received - it is not possible using any routes and configuration options to handle responses received in branch transactions. 

 According to documentation I think that setting wt_timer should allow to absorb and handle responses on failure-route while transaction kept in memory. In fact wt_timer does not affect handling responses in any way.

![Image](https://github.com/user-attachments/assets/14feb472-91ea-4028-8780-1fee1d4ee0e8)

### Troubleshooting

To reproduce this case we have configured sipp to response instant with 200ok on registration request.

#### Reproduction

To reproduce issue SIPP could be setup as one of the dispatcher nodes with instant 200ok reply to register. Kamailio should use following configuration snippet: https://github.com/ovoshlook/kamailio-mid-registrar-config-snippets/blob/master/shared_creds_at_scale/kamailio.cfg

#### Log Messages

REGISTER|401tm [t_lookup.c:912]: t_reply_matching(): t_reply_matching: hash 12337 label 0 branch 1
REGISTER|401tm [t_lookup.c:986]: t_reply_matching(): reply (0x7f1fc67f83a8) matched an active transaction (T=0x7f1fc0f0f740)!
REGISTER|401tm [t_lookup.c:1122]: t_check_msg(): msg (0x7f1fc67f83a8) id=13/1915035 global id=13/1915035 T end=0x7f1fc0f0f740
REGISTER|401tm [t_reply.c:2363]: reply_received(): transaction found - T:0x7f1fc0f0f740 branch:1
REGISTER|401tm [t_reply.c:2376]: reply_received(): original status uas=200, uac[1]=0 local=0 is_invite=0)
REGISTER|401tm [t_reply.c:1363]: t_should_relay_response(): ->>>>>>>>> T_code=200, new_code=401
REGISTER|401tm [t_reply.c:1374]: t_should_relay_response(): final reply already sent
REGISTER|401tm [t_reply.c:1625]: t_should_relay_response(): finished with rps discarded - uas status: 200

### Additional Information

  * **Kamailio Version**

```
version: kamailio 5.6.4 (x86_64/linux) 
[...]

PR #4304, opened 2025-07-01
tm: allow pending forked REGISTER auths being handled the same way as…

ovoshlook:
… forked INVITEs auths

#### Pre-Submission Checklist

#### Type Of Change
- [*] New feature (non-breaking change which adds new functionality)

#### Checklist:

- [ *] Tested changes locally
- [ ]* Related to issue #[4266](https://github.com/kamailio/kamailio/issues/4266)

#### Description
If a Negative reply for forked REGISTER comes after the first 200 response, it won't be ignored by Kamailio and will be processed the same way as replies processed for the INVITEs

### DIFF
--- a/src/modules/tm/t_reply.c
+++ b/src/modules/tm/t_reply.c
@@ -1365,7 +1365,7 @@ static enum rps t_should_relay_response( struct cell *Trans , int new_code,
 	*/
 	LM_DBG("->>>>>>>>> T_code=%d, new_code=%d\n", Trans->uas.status,new_code);
 	inv_through=new_code>=200 && new_code<300 && is_invite(Trans);
-	/* if final response sent out, allow only INVITE 2xx  */
+	/* if final response sent out, allow only INVITE 2xx && REGISTER > 299 */
 	if ( Trans->uas.status >= 200 ) {
 		if (inv_through) {
 			LM_DBG("200 INV after final sent\n");
@@ -1373,11 +1373,18 @@ static enum rps t_should_relay_response( struct cell *Trans , int new_code,
 			Trans->uac[branch].last_received=new_code;
 			*should_relay=branch;
 			return RPS_PUSHED_AFTER_COMPLETION;
-		} else {
+		} 		
+		/* Do not discard negative replies for REGSITER messages 
+		 * after 200 for the initial trasaction already been sent.
+		 * In case of forking REGISGER requests to multiple Registrars
+		 * some of registrars may reply later than others being authenticated.
+		 * This lests late forked transactions being finalised correctly
+		*/
+		if (!(strncmp(Trans->method.s, "REGISTER", 8) == 0 && new_code >= 300)) {
+			/*Except the exceptions above, too late  messages will be discarded */
 			LM_DBG("final reply already sent\n");
+			goto discard;
 		}
-		/* except the exception above, too late  messages will be discarded */
-		goto discard;
 	}
 
 	/* if final response received at this branch, allow only INVITE 2xx */


ovoshlook:
Having 4 commits with no lack to resolve formatting for commit and code (profiling failed), is it better to create a new PR with proper changes in one commit, or is it minor? ( BTW, don't get why commit message formatting failed)

linuxmaniac:
> Having 4 commits with no lack to resolve formatting for commit and code (profiling failed), is it better to create a new PR with proper changes in one commit, or is it minor? ( BTW, don't get why commit message formatting failed)

You just need to rebase and push force your branch

miconda:
I am not sure I understand exactly the use case. Is it about a REGISTER request getting to a SIP proxy that does parallel forking to many downstream SIP registrars? If yes, what do you want to happen with replies having the code >= 300? To be forwarded back to the initial sender of the REGISTER request?

If it is about authentication, why not only 401/407? Shall a 400 be forwarded as well?

Also, the description does not seem right, for the INVITE, if the transaction uas status is `>= 200`, then only 2xx responses are forwarded (in the code it does not seem that >=300 are still forwarded for the INVITEs). Or I am missing something?

Looking at the overall diff at this moment:

  - https://patch-diff.githubusercontent.com/raw/kamailio/kamailio/pull/4304.diff

The first changed line to the LM_DBG() does not seem useful, just breaks the code formatting. Also, a non-necessary empty line is added.

ovoshlook:
The case is a little bit different (it is made in response to this issue https://github.com/kamailio/kamailio/issues/4266):

The case: 

When Kamailio  receives a REGISTER request, it makes a parallel forking to registrars ( mid-registrar case, let's say )
The Kamailio catches the Auth response (401/407) from Registrars at the branch-failure route.
It does the authentication procedure and sends a REGISTER with an authentication header to the Registrar.

However, there is a case where one of the Registrars makes authentication successfully and sends a 200 OK response earlier than other Registrars responses to Kamailio with 401/407 responses.

existing logic doesn't give those responses being handled anyhow and discards them, as the transaction status already changed to 200, and logic hits this point: https://github.com/kamailio/kamailio/blob/master/src/modules/tm/t_reply.c#L1399
> goto: discard; 

Added code allows handling such responses within failure routes; however, it won't propagate failure replies  to UAC, because the new code is >=300
https://github.com/kamailio/kamailio/blob/master/src/modules/tm/t_reply.c#L1430
and the status of the transaction is already  200
https://github.com/kamailio/kamailio/blob/master/src/modules/tm/t_reply.c#L1553
200 responses to late parallel registrations also won't be propagated to UAC because the transaction status 200 already exists.

The description of the added code might indeed be messy - I'll try to improve the clarity of the comment

ovoshlook:
Hi,
May I request some help to figure out what is wrong with the formatting checks? It appears that the code and commit messages are now meeting the requirements, but the parsers are still failing.

henningw:
I have re-run the checks. The check-commit is failing at an unrelated commit, and can be probably ignored. The check-format is failing at your commit it seems, some whitespace change maybe not correct? Just execute clang-format and force-push again.

gaaf:
Why only REGISTER requests? Shouldn't this be more general? At least SUBSCRIBE requests would be similar.

ovoshlook:
Thanks for the comment. That's a good point.
Initially, I was thinking of covering all initial uncovered requests. 
However:
    -  `t_should_relay_response` is quite a sensitive function for large changes.
    - The patch covers a particular edge case related to a particular issue.
    - Spreading the context may disrupt the clarity of the edge case and code.
Moving forward with atomic features was a better approach for me.

According to the questions, it seems the feature's purpose is much clearer now.
If no objections, then I'll cover the rest within this feature branch.

henningw:
I think its a good idea to do smaller incremental changes in sensitive modules, like tm. There seems to be no more comments regarding this patch, lets wait a few more days before merging it. There are some spelling issues in the code comment, but they can be also fixed later.

miconda:
I want to look more at this PR because it is not clear yet for me and touches important components, but I got into a series of trips (e.g., ClueCon, ...), so it can still take some time.

henningw:
@miconda You mentioned that you want to review the PR more closely, just asking if there is maybe an update. Thank you

miconda:
Thanks for the reminder, I am still not sure if is safe/worth to do it without a config parameter, but given that the patch is small, we can add that later if side effects are  discovered. For now it can be merged.

PR #4306, opened 2025-07-02
db_redis: add Redis auto expiry support by changing sets to hashes

lbalaceanu:
Kudos go to Pawel Kuzak.

Since Redis v7.4, hashes support auto-expiry for individual keys. By transforming sets into hashes we can support auto-expiry.

Adding module parameters use_redis_hashes and expires to control the described feature.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/modules/db_redis/db_redis_mod.c
+++ b/src/modules/db_redis/db_redis_mod.c
@@ -42,6 +42,12 @@ MODULE_VERSION
 str redis_keys = str_init("");
 str redis_schema_path = str_init(SHARE_DIR "db_redis/kamailio");
 int db_redis_verbosity = 1;
+int mapping_struct_type = 0;
+str db_redis_hash_value = str_init("DUMMY");
+
+int db_redis_hash_expires = 0;
+str db_redis_hash_expires_str = str_init("");
+char db_redis_hash_expires_buf[20] = {0};
 
 static int db_redis_bind_api(db_func_t *dbb);
 static int mod_init(void);
@@ -60,6 +66,9 @@ static param_export_t params[] = {
 		{"keys", PARAM_STRING | PARAM_USE_FUNC, (void *)keys_param},
 		{"schema_path", PARAM_STR, &redis_schema_path},
 		{"verbosity", PARAM_INT, &db_redis_verbosity},
+		{"mapping_struct_type", PARAM_INT, &mapping_struct_type},
+		{"hash_value", PARAM_STRING, &db_redis_hash_value},
+		{"hash_expires", PARAM_INT, &db_redis_hash_expires},
 #ifdef WITH_SSL
 		{"opt_tls", PARAM_INT, &db_redis_opt_tls},
 		{"ca_path", PARAM_STRING, &db_redis_ca_path},
@@ -120,6 +129,18 @@ int mod_register(char *path, int *dlflags, void *p1, void *p2)
 static int mod_init(void)
 {
 	LM_DBG("module initializing\n");
+
+	if(db_redis_hash_expires && (mapping_struct_type != MS_HASH)) {
+		LM_ERR("expires parameter is only supported with mapping_struct_type "
+			   "set to 1 (hash)\n");
+		return -1;
+	}
+
+	if(db_redis_hash_expires) {
+		db_redis_hash_expires_str.s = db_redis_hash_expires_buf;
+		db_redis_hash_expires_str.len = snprintf(
+				db_redis_hash_expires_str.s, 20, "%d", db_redis_hash_expires);
+	}
 	return 0;
 }
 
--- a/src/modules/db_redis/doc/db_redis_admin.xml
+++ b/src/modules/db_redis/doc/db_redis_admin.xml
@@ -48,11 +48,11 @@ username/string,domain/string,contact/string,received/string,path/string,expires
 		<para>
 			For instance, usrloc relies on a key of "username@domain", but in order to store
 			multiple contacts per AoR, it cannot be constrained to uniqueness. To
-			work around this, db_redis supports mapping sets in such a way as to, in the case of
-			the usrloc module, have a set with a key of "username@domain" and its entries being
-			unique keys per contact based on the ruid of a contact. Thus, one contact in usrloc
-			consists of a unique key "location:entry::example-ruid-1" being a hash with the columns like
-			username, domain, contact, path etc. In addition, this unique key is stored
+			work around this, db_redis supports mapping structures (either sets or hashes).
+			If sets are chosen, in the case of the usrloc module for example, one would have a set with a key
+			of "username@domain" and its entries being unique keys per contact based on the ruid of a contact.
+			Thus, one contact in usrloc consists of a unique key "location:entry::example-ruid-1" being a hash
+			with the columns like username, domain, contact, path etc. In addition, this unique key is stored
 			in a set "location:usrdom::exampleuser:exampledomain.org". When usrloc does
 			a lookup based on "username@domain", db_redis figures out via the keys/values
 			the query constructed by usrloc to look for the final entry key in the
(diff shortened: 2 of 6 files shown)


src/modules/db_redis/db_redis_mod.c
```diff
@@ -42,6 +42,11 @@ MODULE_VERSION
 str redis_keys = str_init("");
 str redis_schema_path = str_init(SHARE_DIR "db_redis/kamailio");
 int db_redis_verbosity = 1;
+int db_redis_use_hashes = 0;
+int db_redis_expires = 0;
+
+str expires = str_init("");
+char expires_buf[20] = {0};
```
miconda, line 51: These should be also prefixed with the module name, `expires` is quite common name in SIP and other libs,  being declared with `extern` in other files, the symbol can conflict and be resolved to other globals. Like:

```
str db_redis_expires_str = str_init("");
char db_redis_expires_buf[20] = {0};
```

src/modules/db_redis/redis_connection.c
```diff
@@ -315,6 +328,38 @@ int db_redis_connect(km_redis_con_t *con)
...
 	reply = NULL;
+
+	if(expires.len) {
+		reply = redisCommand(con->con, "INFO server");
+		if(!reply) {
+			LM_ERR("failed to get INFO from Redis server\n");
+			goto err;
+		}
+
+		char *version_str = strstr(reply->str, "redis_version:");
```
miconda: It is recommended that variables are declared at the beginning of the function or of the block to work fine with older C versions.

src/modules/db_redis/redis_connection.c
```diff
@@ -315,6 +328,38 @@ int db_redis_connect(km_redis_con_t *con)
...
+		if(!version_str) {
+			LM_ERR("Redis version not found in INFO reply\n");
+			goto err;
+		}
+
+		version_str += strlen("redis_version:"); // Skip past the field name
+		int major = 0, minor = 0, patch = 0;
```
miconda: It is recommended that variables are declared at the beginning of the function or of the block to work fine with older C versions.

src/modules/db_redis/doc/db_redis_admin.xml
```diff
@@ -243,6 +243,52 @@ modparam("db_redis", "opt_tls", 1)
...
+			<title><varname>use_redis_hashes</varname> (int)</title>
+			<para>
+				Controls whether to use Redis hashes for storing the records in the database.
+				If 0, sets are used (which is the default).
+			</para>
+			<para>
+				Motivation of hashes is the implementation of HEXPIRE command in Redis,
+				available since Redis v 7.4.0 onwards, which allows expiring individual
+				keys inside. Hashes do bring the disadvantage of storing DUMMY values, 
+				since only the keys are useful for us.
```
miconda: Is it a reason to store exactly `DUMMY`, or it was just a random pick? Maybe this value should be made a mod param (e.g., `hash_value`)?

src/modules/db_redis/doc/db_redis_admin.xml
```diff
@@ -243,6 +243,52 @@ modparam("db_redis", "opt_tls", 1)
...
+				<programlisting format="linespecific">
+...
+modparam("db_redis", "use_redis_hashes", 1)
+...
+				</programlisting>
+			</example>
+		</section>
+
+		<section id="db_redis.p.expires">
+			<title><varname>expires</varname> (int)</title>
```
miconda: To be more clear that is related to the hash table, I would propose to name this modparam `hash_expires`.

src/modules/db_redis/doc/db_redis_admin.xml
```diff
@@ -243,6 +243,52 @@ modparam("db_redis", "opt_tls", 1)
 			</example>
 		</section>
 
+		<section id="db_redis.p.use_redis_hashes">
+			<title><varname>use_redis_hashes</varname> (int)</title>
+			<para>
```
miconda, line 271: I am not sure if in the future would be another need for a different value, so it is just a suggestion to name this modparam `hash_mode`, so another value (e.g., `2`) can be considered instead of adding a new parameter in the future.

src/modules/db_redis/doc/db_redis_admin.xml
lbalaceanu: Hi @miconda, this "DUMMY" name was just a random pick, the value is currently not used. Sure, we will make it configurable.

src/modules/db_redis/doc/db_redis_admin.xml
lbalaceanu: Sure.

src/modules/db_redis/doc/db_redis_admin.xml
lbalaceanu: Hi, these hash data structures refer to the mapping structures that db_redis uses to implement uniqueness. use_redis_hashes means in fact use_redis_hashes_for_mapping_structures, maybe renaming the parameter to `mapping_struct_type` would be clearer ?!

lbalaceanu:
Hi @miconda,

Thank you for the comments, I will implement them.

src/modules/db_redis/doc/db_redis_admin.xml
miconda: Fine for me for the suggested name, sounds better for future extensibility that the previous one.

miconda:
A few questions to understand the design for the new struct storage model: did I understand correctly that a DB record (ie., row values) are now keys in a dedicated hash table (since the value is `DUMMY`)? If yes, they are prefixed with the column name, or how is the mapping column-name to hash-key (the value) is done? Afaik, uasualy the order of the items/keys in a hash table is not guaranteed, or does Redis guarantees that for the same list of keys, it is the same order of the items?

Where per-item expiration helps (e.g., for usrloc), isn't the entire hash table that has to be deleted on expire? Or is not like each DB record is a dedicated htable?

Sorry for the laziness of not going through the code to understand from there some of the questions ...

lbalaceanu:
Sorry for not making this clearer.

Until now, when using usrloc + redis, when 497xxxxxxxxx @ domain would register, then the following data structs would be created:

HASH "location:entry::uloc-686296ea-5773d-1" -> contact info
SET "location:usrdom::497xxxxxxxxx:domain" -> contains "location:entry::uloc-686296ea-5773d-1" , but might have more entries
SET "location::index::usrdom" -> contains "location:usrdom::497xxxxxxxxx:domain" etc
SET "location:timer::2025-06-30 17:54:03" "location:entry::uloc-686296ea-5773d-1"
SET "location::index::timer" "location:timer::2025-06-30 17:54:03"

This commit changes all the sets to hashes, the keys being the previous entries in the hashes.
We do this because from v7.4 on, Redis allows auto-expiry on key level.
With this patch we set expires on location:entry::uloc-686296ea-5773d-1 itself and on its corresponding keys in "location:usrdom::497xxxxxxxxx:domain",  "location::index::usrdom", "location:timer::2025-06-30 17:54:03" , "location::index::timer".

We saw this also as a relatively low hanging fruit since there was a todo comment in the db_redis code stating:  
// TODO: utilize auto-expiry? on insert/update, also update expire value
// of mappings

miconda:
Thanks for the details!

> This commit changes all the sets to hashes, the keys being the previous entries in the hashes.

I guess that a typo was made and the above should be: `... the keys being the previous entries in the sets`

lbalaceanu:
You are right about the typo.

lbalaceanu:
I am still doing some changes so please don't close this PR.

lbalaceanu:
Hi, I removed the work-in-progress label, I have nothing more to add to these changes. If anyone wants to suggest anything, please do.

miconda:
Fine for me to merge, if there are no other comments.

PR #4319, opened 2025-07-13
Remove some references to CVS and SVN

dilyanpalauzov:


dilyanpalauzov:
I do not know what prefix to use here.

PR #4320, opened 2025-07-13
dialog: use capital letter at start of sentence

dilyanpalauzov:


henningw:
Thanks for the PR. Please note that the README files should not be changed, as they are generated automatically from the XML sources. Can you remove this file from the commit and force-push? Alternatively you can also close this one and create a new PR.

PR #4336, opened 2025-07-23
Parse urn fix

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue https://kamailio.org/mailman3/hyperkitty/list/sr-users@lists.kamailio.org/thread/DV3AFOJUP3V5YNEIERHCO5FLHG6Q4E7F/
#### Description

This PR fixes the bug when failing to parse a URN scheme with a '@' character in the host name.

- Format of file 
- Add some comments on magic numbers

### DIFF
--- a/src/core/parser/parse_uri.c
+++ b/src/core/parser/parse_uri.c
@@ -198,6 +198,10 @@ int parse_uri(char *buf, int len, struct sip_uri *uri)
 	str comp_val; /* not returned for now */
 #endif
 
+/* derived from the ASCII values of the protocol scheme strings
+ ("sip:", "sips", "tel:", "urn:") interpreted as 4-byte
+ little-endian integers.
+ */
 #define SIP_SCH 0x3a706973
 #define SIPS_SCH 0x73706973
 #define TEL_SCH 0x3a6c6574
@@ -467,6 +471,8 @@ int parse_uri(char *buf, int len, struct sip_uri *uri)
 		goto error_too_short;
 	scheme = ((uint32_t)buf[0]) + (((uint32_t)buf[1]) << 8)
 			 + (((uint32_t)buf[2]) << 16) + (((uint32_t)buf[3]) << 24);
+	/* make it case insensitive by converting to lowercase
+	OR with difference between 'A' and 'a' is 32 (0x20) */
 	scheme |= 0x20202020;
 	if(scheme == SIP_SCH) {
 		uri->type = SIP_URI_T;


sergey-safarov:
Do I understand properly that here should be "if equal urn schema go to error" 
```	case '@':                      \
		if(scheme != URN_SCH) {    \
			goto error_bad_char;   \
		}  
```

xkaraman:
> Do I understand properly that here should be "if equal urn schema go to error"
> 
> ```
> 		if(scheme != URN_SCH) {    \
> 			goto error_bad_char;   \
> 		}  
> ```

No. 

If it's NOT a URN scheme, then '@' is an invalid charachter in host part and should error out.

If it is a URN scheme, '@' is just another character in the host name and it's valid.

miconda:
> If it's NOT a URN scheme, then '@' is an invalid charachter in host part and should error out.
> 
> If it is a URN scheme, '@' is just another character in the host name and it's valid.

I haven't looked at the code, but from your comments above I guess that the URN value is stored in the host field of the URI structure? Would it be a case where the part after the `@` should be actually used as a host/domain (ie., the urn with the role of the username, and the value after `@` as domain/ip address)?

henningw:
> > If it's NOT a URN scheme, then '@' is an invalid charachter in host part and should error out.
> > If it is a URN scheme, '@' is just another character in the host name and it's valid.
> 
> I haven't looked at the code, but from your comments above I guess that the URN value is stored in the host field of the URI structure? Would it be a case where the part after the `@` should be actually used as a host/domain (ie., the urn with the role of the username, and the value after `@` as domain/ip address)?

I don't work daily with URNs, but I think according to their definition[1] there is no concept of a host name in it. Its basically a character for the purposes of the URN. There are other standards which are using URNs (e.g.  Canonical Text Services) which uses it for adressing of sub-elements inside a specific content. The parser is taking some shortcuts, it would be probably better if it would be separate from the URI/URL part. Its also probably does not fully support the mentioned RFC.

[1] https://datatracker.ietf.org/doc/html/rfc8141#section-2

sergey-safarov:
As I know "urn" cannot contain `@` char.
Here is [RFC link](https://www.rfc-editor.org/rfc/rfc5031.html#page-5)
```
     service-URN  = "URN:service:" service
     service      = top-level *("." sub-service)
     top-level    = let-dig [ *25let-dig-hyp let-dig ]
     sub-service  = let-dig [ *let-dig-hyp let-dig ]
     let-dig-hyp  = let-dig / "-"
     let-dig      = ALPHA / DIGIT
     ALPHA        = %x41-5A / %x61-7A   ; A-Z / a-z
     DIGIT        = %x30-39 ; 0-9
```
As I understanding allowed letters, digits and "-" char only.

xkaraman:
Hmm interesting
Following https://datatracker.ietf.org/doc/html/rfc8141#section-2 which is a generic URN RFC i imagine, the `@` character is allowed as part of `pchar` defined in [RFC 3986 ](https://datatracker.ietf.org/doc/html/rfc3986).

Now about service URNs seems to not follow this RFC, or i might misread something, or something else that i don't fully understand.

--EDIT:

After some more reading, `NSS` from https://datatracker.ietf.org/doc/html/rfc8141#page-10 reads

> The NSS is a string, unique within a URN namespace, that is assigned
   and managed in a consistent way and that conforms to the definition
   of the relevant URN namespace.  

Therefore i think we need to decide whether we need to interpited as 

`URN service` namespace definition where it seems that service can NOT contain a `@` to be compliant 

OR

a generic defintion where `@` is allowed in the `NSS`.

I believe due to https://datatracker.ietf.org/doc/html/rfc8141#section-4.2, we probably need to parse it as a generic URN, and if we need a more compliant versions, we need to have some other mechanism.

henningw:
> As I know "urn" cannot contain `@` char. Here is [RFC link](https://www.rfc-editor.org/rfc/rfc5031.html#page-5)
> 
> ```
>      service-URN  = "URN:service:" service
>      service      = top-level *("." sub-service)
>      top-level    = let-dig [ *25let-dig-hyp let-dig ]
>      sub-service  = let-dig [ *let-dig-hyp let-dig ]
>      let-dig-hyp  = let-dig / "-"
>      let-dig      = ALPHA / DIGIT
>      ALPHA        = %x41-5A / %x61-7A   ; A-Z / a-z
>      DIGIT        = %x30-39 ; 0-9
> ```
> 
> As I understanding allowed letters, digits and "-" char only.

Yes, but this is the standard for the URN service namespace, not the generic URN standard.

sergey-safarov:
@xkaraman @henningw thanks for the clarification.

miconda:
OK, if `@` is not indicating the start of a fqdn/ip, but it is part of the urn itself, then I am fine to merge.

PR #4339, opened 2025-07-24
tls_tracker: added module for tls encriptions key export

sergey-safarov:
#### Pre-Submission Checklist
- [x] No commits to README files for modules (changes must be done to docbook files in `doc/` subfolder, the README file is autogenerated)

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description
Added TLS encryption keys export into the database.
Later keys can be searched and saved to a file. Key search can be done by timestamp, connection IP addresses, and port numbers. For key search, used kamcmd utility.


sergey-safarov:
@xkaraman could you help me compile this module using cmake.
I have added `tls_tracker` module to the module list for compilation and get an error
```
-- Found libphonenumber include directory: /usr/include
-- Found libphonenumber library: /usr/lib64/libphonenumber.so;/usr/lib64/libgeocoding.so
-- Module pua_rpc has already been added. Skipping...
-- PCRE2 library not found... looking with pkg-config
-- Checking for module 'libpcre2-8'
--   Found libpcre2-8, version 10.45
-- Found NETSNMP: /usr/bin/net-snmp-config
-- Found ZLIB: /usr/lib64/libz.so (found version "1.3.1")
CMake Error at src/modules/CMakeLists.txt:129 (message):
  Module directory
  /root/rpmbuild/BUILD/kamailio-6.1.0-build/kamailio-6.1.0-dev1/src/modules/tls_tracker
  does not exist.
Call Stack (most recent call first):
  src/modules/CMakeLists.txt:170 (add_module_group)
```

What is the correct fix to compile a new module?

miconda:
Thanks for the PR!

However, I think it needs some redesign, because now it seems that it adds mutual dependency and quite some code duplication in the modules that are updated to use the new tls_tracker module (in this PR, the tls and http_client).

IMO, the better approach is to have API functions in the tls and http_client module that allow setting callbacks on some events and then the tls_tracker will leverage them to have its own functions executed.

Also, the additional app data stored in SSL structure has to be done more generic. Adding fields directly to tls_extra_data that are not tls module specific is not scalable in long term. As another note here, the values in tls_extra_data have to be in share memory, not in private/pkg, tls connection and context structures can be accessed from many processes.

The associated tcp connection can be added to tls_extra_data structure, being from the core and could be useful as a common resource, but db_session_id and session_key are very specific to tls_tracker, so they have to be stored in a more generic way, that other modules can leverage a similar mechanism without changing tls code every time.

I will think about it and try to come with some proposal.

miconda:
Also, the database table structures have to be defined in xml format inside `src/lib/srdb1/schema/`, the sql files are generated from them (with `make dbschema`). The names of the tables should be prefixed with the name of the module (or a prefix that suggests the module name), like `tls_tracker_connections` and `tls_tracker_keys` (or `tlstrack_connections` `tlstrack_keys`).

xkaraman:
Hey @sergey-safarov,

You seem to be missing the `CMakeLists.txt` file, see https://github.com/kamailio/kamailio/blob/master/src/modules/tls/CMakeLists.txt for how to write one for it.

PR #4343, opened 2025-07-29
TM Event local-request for local ACKs

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR introduces a new mode on the modparam `evlreq_mode` to allow for choosing what should emit the local-request event. It enables the ACK that are generated from kamailio to also emit the event if the user chose to do so.

With this PR the modes are
- 0 for None
- 1 for Cancel
- 2 for Acks
- 3 for Both


henningw:
Thanks for the PR - can you fix the one wrong commit message?

xkaraman:
Fixed, it can also me be merged with the first commit after some review and testing, if need.

henningw:
Thanks, lets wait a bit more for feedback before merging.

miconda:
I guess this can be merged.

src/modules/tm/config.h
```diff
@@ -88,8 +88,14 @@
 #define CANCEL_DONE "ok -- no more pending branches"
 #define CANCELED "Request canceled"
 
-/* execute event_route[tm:local-request] for hop-by-hop CANCEL */
-#define TM_EVLREQ_CANCEL_HBH 1
+/* execute event_route[tm:local-request] for hop-by-hop messages
+1 = execute for Cancel
+2 = execute for ACK
+4 = execute for both
```
miconda: Actually I just spotted that this comment should list the value 3 for both, right?

xkaraman: Ahh yeah good catch! Fixed!

henningw:
Thank you, merged.

Issue #4345, opened 2025-08-04
segfault due to expired timer broken "next" list


stefan-mititelu-idt:
### Description
Crash in kamailio 5.8.5 due to broken "next" pointer of circular double linked list of expired timers.

#### Reproduction
Not easily reproducible

#### Debugging Data

```
(gdb) bt full
#0  timer_list_expire (slow_mark=<optimized out>, slow_l=0x7ff891b59d60, h=0x7ff891b56ff0, t=<optimized out>) at core/timer.c:850
        tl = 0x7ff89d214810
        ret = <optimized out>
        tl = <optimized out>
        ret = <optimized out>
#1  timer_handler () at core/timer.c:925
        saved_ticks = <optimized out>
        run_slow_timer = 0
        i = <optimized out>
        saved_ticks = <optimized out>
        run_slow_timer = <optimized out>
        i = <optimized out>
        __func__ = "timer_handler"
        __llevel = <optimized out>
        __kld = {v_facility = <optimized out>, v_level = <optimized out>, v_lname = <optimized out>, v_fname = <optimized out>, v_fline = <optimized out>, v_mname = <optimized out>,
          v_func = <optimized out>, v_locinfo = <optimized out>, v_pid = <optimized out>, v_pidx = <optimized out>}
        __llevel = <optimized out>
        __kld = {v_facility = <optimized out>, v_level = <optimized out>, v_lname = <optimized out>, v_fname = <optimized out>, v_fline = <optimized out>, v_mname = <optimized out>,
          v_func = <optimized out>, v_locinfo = <optimized out>, v_pid = <optimized out>, v_pidx = <optimized out>}
#2  timer_main () at core/timer.c:963
No locals.
#3  0x000055e671830461 in main_loop () at main.c:1933
        i = <optimized out>
        pid = <optimized out>
        si = 0x0
        si_desc = "udp receiver child=15 sock=x.x.x.x:5080 (y.y.y.y:5080)", '\000' <repeats 16 times>, "\002\000\000\000\004\000\000\000\n_\022:\372\177\000\000\003\000\000\000\000\000\000\000\000kt\234s8\b\rThu Jul \a\000\000\000\000\000\000"
        nrprocs = <optimized out>
        woneinit = 1
        __func__ = "main_loop"
        error = <optimized out>
#4  0x000055e671824ff2 in main (argc=<optimized out>, argv=<optimized out>) at main.c:3257
        cfg_stream = <optimized out>
        c = <optimized out>
[...]

PR #4346, opened 2025-08-04
core: check and recover if broken circular list

stefan-mititelu-idt:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

Possible fix for #4345 .

### DIFF
--- a/src/core/coreparam.c
+++ b/src/core/coreparam.c
@@ -25,12 +25,14 @@
 #include "coreparam.h"
 
 int ksr_coreparam_store_nval(str *pname, ksr_cpval_t *pval, void *eparam);
+long ksr_timer_sanity_check = 0;
 
 /* clang-format off */
 static ksr_cpexport_t _ksr_cpexports[] = {
 	{ str_init("random_engine"), KSR_CPTYPE_STR,
 		ksr_xrand_cp, NULL },
-
+	{ str_init("timer_sanity_check"), KSR_CPTYPE_NUM,
+		ksr_coreparam_store_nval, &ksr_timer_sanity_check },
 	{ {0, 0}, 0, NULL, NULL }
 };
 /* clang-format on */
--- a/src/core/timer.c
+++ b/src/core/timer.c
@@ -800,6 +800,76 @@ void timer_allow_del(void)
 }
 
 
+#define timer_sanity_check(                                                  \
+		h, tl1, tl2, last_tl1, last_tl2, dir1, dir2, dir1str, dir2str)       \
+	(last_tl1) = (void *)h;                                                  \
+	(tl1) = h->dir1;                                                         \
+	do {                                                                     \
+		if((tl1) == NULL) {                                                  \
+			int fixed = 0;                                                   \
+			LM_WARN("timer %s list is broken at cell %p with %s=%p "         \
+					"%s=%p; trying to fix it...",                            \
+					dir1str, (last_tl1), dir1str, (last_tl1)->dir1, dir2str, \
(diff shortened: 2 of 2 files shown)


henningw:
Thanks for the PR. There is also one change included for the tm module timer infrastructure, which might be introduced by mistake?

stefan-mititelu-idt:
Thank you, will PR that one separately, since was different issue.

miconda:
Since the macro is called twice, I would add another parameter to provide the first direction parameter also as a string to make the log messages easier to show the direction of the check, like:

```
#define timer_sanity_check(h, tl1, tl2, last_tl1, last_tl2, dir1, dir2, dir1name) \
...
			LM_WARN("dir1=%s: timer dir1 list is broken at cell %p with dir1=%p "     \
					"dir2=%p; trying to fix it...",  dir1name,                        \
					(last_tl1), (last_tl1)->dir1, (last_tl1)->dir2);         \
...

	timer_sanity_check(h, tl1, tl2, last_tl1, last_tl2, next, prev, "next");
	timer_sanity_check(h, tl1, tl2, last_tl1, last_tl2, prev, next, "prev");
``` 

The other comment I would have is related to the processing impact: whether it makes sense to add a core parameter to enable/disable this check. It might be time-costly if it walks the list every time twice, when there are a lot of items.

If it sounds like be possible concern, a coreparam can be added easily for it inside `src/core/coreparam.c`, like:

```
extern long ksr_timer_sanity_check;

static ksr_cpexport_t _ksr_cpexports[] = {
	{ str_init("random_engine"), KSR_CPTYPE_STR,
		ksr_xrand_cp, NULL },
	{ str_init("timer_sanity_check"), KSR_CPTYPE_NUM,
		ksr_coreparam_store_nval, &ksr_timer_sanity_check },

	{ {0, 0}, 0, NULL, NULL }
};
```

Inside src/core/timer.c:

```
long ksr_timer_sanity_check = 0;

. . .
	if(ksr_timer_sanity_check != 0 && !timer_list_sanity_check(h)) {
		abort();
	}
```

Then in kamailio.cfg can be enabled with:

```
coreparam[timer_sanity_check] = 1
```

stefan-mititelu-idt:
Thanks for feedback! All makes sense. Did the following:
1. amend the initial commit with improvements for "next" and "prev" logging; diff in 1'st force-push, timer.c file
2. add 1 more core commit for timer_sanity_check coreparam, after rebase to latest master; diff in 2'nd force push, coreparam.c and cpreparam.h files

Basically tested point nr 2. and value is set correctly from cfg, using "coreparam[timer_sanity_check]=1"

henningw:
A related question about this core extension. The motivation for introducing it was to help a debugging aid for potential issues on the timer list? So it would be not permanently enabled on a production environment?

stefan-mititelu-idt:
Hi, we definitely intend to enable it in the production environment. And constantly check for the new WARN logs in the new macro. However, we could not reproduce the issue during QA, so I can imagine someone maybe not wanting to enable it in production.

Motivation for this was that coredump at the linked issue could have been avoided using this patch. Because the "prev" list was ok, only "next" list was broken.

miconda:
I am fine to merge it.

One more comment, just to plug in an idea here, since this is good for safety/troubleshooting, it might be the case the the item with the prev/next broken was supposed to be already consumed, so maybe as a new `timer_sanity_check` value/mode that can be added in the future, the fix removes the item. But first let's see what you observe watching the logs.

PR #4355, opened 2025-08-07
pua_reginfo: use expires from contact for publish request

xadhoom:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
Current implementation of pua_reginfo, when `publish_reginfo` is used, sets a fixed `Expires` value of `3600` in the PUBLISH message, which may lead to many entries in pua db for the same reginfo data if the client registers sin an expiration < 1hr.

For example: a device registers every 2 minutes, in one hour pua db will have 30 entries for the same reg, because they're set to expire after 1hr.

In addition if this info is subscribed by another entity, the entity will receive notifies with a lot of entries for the same reg.

Instead, like is done with the pua_dialoginfo, the Expires of the publish should follow the expire of the registration.

This patch does exactly that, it sets the PUBLISH Expire to the reg expire value.

There's only one caveat: that each usrloc record can have many contacts with different expires, so the approach followed here is to set the PUBLISH Expire value to the max reg expire value of all contacts of the same aor.

### DIFF
--- a/src/modules/pua_reginfo/usrloc_cb.c
+++ b/src/modules/pua_reginfo/usrloc_cb.c
@@ -257,6 +257,33 @@ error:
 	return NULL;
 }
 
+int reginfo_get_expires(urecord_t *record)
+{
+	// Loop through all record contacts and set the publish Expire value
+	// to the max expire of the contacts
+	ucontact_t *ptr;
+	time_t cur_time = time(0);
+	int expires = -1;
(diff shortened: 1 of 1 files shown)


miconda:
Fine for me to merge.

henningw:
Thanks for the PR, merged

Issue #4331, opened 2025-07-18
dispatcher: EWMA formula


furmur:
have a question about EWMA formula in the `dispatcher`  module.

[documentation](https://www.kamailio.org/docs/modules/devel/modules/dispatcher.html#dispatcher.p.ds_latency_estimator_alpha) for `ds_latency_estimator_alpha` points to this [explanation](https://www.itl.nist.gov/div898/handbook/pmc/section3/pmc324.htm)
which specifies EWMA formula as: $EWMA_t = \lambda*Y_t + (1 - \lambda) * EWMA_{t-1}$

related code at [`src/modules/dispatcher/dispatch.c`](https://github.com/kamailio/kamailio/blob/master/src/modules/dispatcher/dispatch.c#L3191):
```
static inline void latency_stats_update(
		ds_latency_stats_t *latency_stats, int latency)
{
...
    /* exponentially weighted moving average */
    if(latency_stats->count < 10) {
        latency_stats->estimate = latency_stats->average;
    } else {
        latency_stats->estimate =
                latency_stats->estimate * ds_latency_estimator_alpha
                + latency * (1 - ds_latency_estimator_alpha);
    }
```

so:
* `latency_stats->estimate` is $EWMA$,
* `ds_latency_estimator_alpha` is $\lambda\$.
* `latency` is $Y_t$

looks like `latency_stats->estimate` and `latency` are swapped regarding to the formula and as a result higher `ds_latency_estimator_alpha` will rate old average over the actual latency value and that contradicts both `dispatcher` module documentation and meaning in the referenced formula.

i believe that code should be:
```
        latency_stats->estimate =
                latency * ds_latency_estimator_alpha
                + latency_stats->estimate * (1 - ds_latency_estimator_alpha);
```

could someone verify those observations? will make PR if i am not mistaken.

miconda:
Maybe @jchavanton can comment, he was the one that added latency stats to dispatcher, iirc.

henningw:
The linked PR was merged, but apparently there are some documentation changes missing.

jchavanton:
Sorry, I did not notice the email ...

It is true that ALPHA formula was inverted, compared to the referenced documentation.

With the default it was working as intended, now the default value is not making much sense.
I will do an MR to update the doc and the default value, as commented on : 

https://github.com/kamailio/kamailio/pull/4358#issuecomment-3271202397

henningw:
Documentation PR was merged, close this issue.

[...]

PR #4358, opened 2025-08-11
dispatcher: fix ds_latency_estimator_alpha meaning

furmur:
swap 'latency' and 'latency_stats->estimate' in the EWMA formula to ensure ds_latency_estimator_alpha rates new values over the old average estimate

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4331

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -3214,8 +3214,8 @@ static inline void latency_stats_update(
 		latency_stats->estimate = latency_stats->average;
 	} else {
 		latency_stats->estimate =
-				latency_stats->estimate * ds_latency_estimator_alpha
-				+ latency * (1 - ds_latency_estimator_alpha);
+				latency * ds_latency_estimator_alpha
+				+ latency_stats->estimate * (1 - ds_latency_estimator_alpha);
 	}
 }
 


henningw:
Thanks for the PR, also read the linked issue. Looks like a straightforward fix to me. Lets wait a bit for other feedback before merging.

miconda:
Not using latency, I can't assert it properly, but given the long time being without comments and being a small change, I am going to merge it, thanks! If someone has amendments, can push new commits.

jchavanton:
@furmur sorry I did not see that MR before.

I can see that your modification is aligned with the general formula of the EWMA, I am in favor of making the swap,  but we also need to update the documentation and the default value. 

We have the default value of .90 (900), that is use as the weight to the estimator (not the new sample).
`new-estimate = alpha * previous-estimate + (1-apha) * new sample`



```
3.28. ds_latency_estimator_alpha (int)

The value to be used to control the memory of the estimator EWMA "exponential weighted moving average" or "the speed at which the older samples are dampened" a good explanation can be found here : http://www.itl.nist.gov/div898/handbook/pmc/section3/pmc324.htm Because Kamailio doesn't support float parameter types, the value in the parameter is divided by 1000 and stored as float. For example, if you want to set the alpha to be 0.75, use value 750 here. 
```
Should become
```
3.28. ds_latency_estimator_alpha (int)

The parameter decides how important the current observation is in the calculation of the EWMA. The higher the value of alpha, the more closely the EWMA tracks the original time series.
Because Kamailio doesn't support float parameter types, the value in the parameter is divided by 1000 and stored as float. For example, if you want to set the alpha to be 0.25, use value 250 here. 
```

Default value should now be `0.10`

furmur:
@jchavanton,
thank you for clarifying it and for the updated documentation

PR #4363, opened 2025-08-12
CMake Auto detect lock method according to priority

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

This PR implements an option for locking methods.

It detects what options are available and applys the nessecary definitions.

It introduces the AUTO which, automatically selects one of the available on the platform according to the priority found in https://github.com/kamailio/kamailio/blob/2ca2ba1b2cbd73924f5b09396ff194402196258c/src/core/lock_ops.h#L80-L91

Also fixes some other missing definitions

### DIFF
--- a/cmake/defs.cmake
+++ b/cmake/defs.cmake
@@ -87,14 +87,12 @@ option(Q_MALLOC "Use q_malloc" ON)
 # Same goes for fmalloc and tlsf malloc
 # cmake_dependent_option(DBG_QM_MALLOC "Enable debugging info for q_malloc" OFF "Q_MALLOC" OFF)
 option(TLSF_MALLOC "Use tlsf_malloc" ON)
-option(MALLOC_STATS "Use malloc stats" ON)
 
 option(USE_DNS_FAILOVER "Use DNS failover" ON)
 option(USE_DST_BLOCKLIST "Use destination blacklist" ON)
 option(HAVE_RESOLV_RES "Have resolv_res" ON)
 
 option(KSR_PTHREAD_MUTEX_SHARED "Use shared mutex for TLS" ON)
-option(STATISTICS "Statistics" ON)
 
 # if(${MEMPKG})
 #   target_compile_definitions(common INTERFACE PKG_MALLOC)
@@ -241,10 +239,6 @@ if(TLSF_MALLOC)
   target_compile_definitions(common INTERFACE TLSF_MALLOC)
 endif()
 
-if(MALLOC_STATS)
-  target_compile_definitions(common INTERFACE MALLOC_STATS)
-endif()
-
 if(MEMDBG)
   target_compile_definitions(common INTERFACE DBG_SR_MEMORY)
   if(MEMDBGSYS)
@@ -313,11 +307,11 @@ if(KSR_PTHREAD_MUTEX_SHARED)
 endif()
 
 if(FMSTATS)
-  target_compile_definitions(common INTERFACE FMSTATS)
+  target_compile_definitions(common INTERFACE MALLOC_STATS)
 endif()
 
 if(KMSTATS)
-  target_compile_definitions(common INTERFACE KMSTATS)
+  target_compile_definitions(common INTERFACE USE_CORE_STATS STATISTICS)
 endif()
 
 include(${CMAKE_SOURCE_DIR}/cmake/compiler-specific.cmake)
@@ -369,7 +363,6 @@ target_compile_definitions(
             # TODO: We can use the generator expression to define extra flags
             # instead of checking the options each time
             $<$<BOOL:${USE_SCTP}>:USE_SCTP>
-            $<$<BOOL:${STATISTICS}>:STATISTICS>
 )
 target_link_libraries(common INTERFACE common_compiler_flags)
 
--- a/cmake/defs.cmake
+++ b/cmake/defs.cmake
@@ -298,10 +298,6 @@ if(NO_DEV_POLL)
   target_compile_definitions(common INTERFACE NO_DEV_POLL)
 endif()
 
-if(RAW_SOCKS)
-  target_compile_definitions(common INTERFACE RAW_SOCKS)
-endif()
-
 if(KSR_PTHREAD_MUTEX_SHARED)
   target_compile_definitions(common INTERFACE KSR_PTHREAD_MUTEX_SHARED)
 endif()
--- a/cmake/defs.cmake
+++ b/cmake/defs.cmake
@@ -246,6 +246,10 @@ if(MEMDBG)
   endif()
 endif()
 
+if(MEM_JOIN_FREE)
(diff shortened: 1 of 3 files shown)


xkaraman:
We are now supplying only one method for locking out of five available: `FUTEX`, `FAST_LOCK` and so on.

Previsouly both `USE_FUTEX` and `FAST_LOCK` were provided to the build, but according to https://github.com/kamailio/kamailio/blob/2ca2ba1b2cbd73924f5b09396ff194402196258c/src/core/lock_ops.h#L80-L91 only one is of course used.

But it seems 
https://github.com/kamailio/kamailio/blob/2ca2ba1b2cbd73924f5b09396ff194402196258c/src/core/lock_alloc.h#L59 and  https://github.com/kamailio/kamailio/blob/2ca2ba1b2cbd73924f5b09396ff194402196258c/src/core/lock_alloc.h#L81 do not include the `USE_FUTEX` one. 

Therefore since now we only apply the USE_FUTEX and not the FAST_LOCK one, it fails to build. 

Is this expected and both are required? or a code adoption to include `USE_FUTEX` in the `ifdef` statement is nessecary?

miconda:
I haven't looked at the code, but I guess USE_FUTEX has to be added in the ifdef.

xkaraman:
Ok will update as well.

My main question is that the intention is to provide only one of the five available lock methods: `FUTEX`, `FAST_LOCK`, `PTHREAD`, `POSIX_SEM` and `SYSV_SEM`. That's what i was thinking it should behave as lock_ops.h will ever use one.

Previous builds provided both `USE_FUTEX` AND `FAST_LOCK` as seen by

```
sbin/kamailio -I
Print out of kamailio internals
  Version: kamailio 5.8.1 (x86_64/linux) ba7a49
  Default config: /home/xenofon/kamailio-builds/kam-5.8.1/etc/kamailio/kamailio.cfg
  Default paths to modules: /home/xenofon/kamailio-builds/kam-5.8.1/lib64/kamailio/modules
  Compile flags: ... **USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT**, ...```.

Is this intentional or was it something that went unnoticed?

miconda:
Not being the one that developed the locks APIs, I can't say for sure. I based my opinion on your snippets showing the `#if ... #elif`. The change should go into devel branch and tested to be safe.

xkaraman:
For `aarch64` it fails to build but does build for make. Why this happens.

For old `Makefiles`, `futexlock.h` seems to not have support for it. This happens beacuse it needs the `HAVE_ASM_INLINE_ATOMIC_OPS` which is only defined for `arm6` or `arm7` in `atomic_arm.h`. 

In `futexlock.h` then it undefines the USE_FUTEX, which is provided (either by user or auto), and due to https://github.com/kamailio/kamailio/blob/3f315d0f583b253919572498f47371fddca4371f/src/Makefile.defs#L920-L923, it falls back to `fast_lock`. If we also disable `FAST_LOCK`, it will fallback even more to `USE_PTHREAD_MUTEX` due to https://github.com/kamailio/kamailio/blob/3f315d0f583b253919572498f47371fddca4371f/src/Makefile.defs#L1853-L1858.

So, in summary, what i am seeing is inconsisteny that i it's not even visible. 
On `x86_64`, `kamailio` use the `futex` lock method, but `kamailio -I` will report `USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT` and `aarch64` (and `arm64` probably didn't tested it yet), will use `FAST_LOCK` lock method but report `USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT-NOSMP` (tested on an aarch64 server).

In CMake and this PR, i check for futex availability through https://github.com/xkaraman/kamailio/blob/2f1c3d2447dcf860365314e6345b67dde68c0f94/cmake/lock_methods.cmake#L35-L41, and since it's the first in priority i passed that one to the build.
Since in this PR we only provide a single method `USE_FUTEX` and no fallbacks, the build fails due to `HAVE_ASM_INLINE_ATOMIC_OPS` which is only defined for `arm6` or `arm7` in `atomic_arm.h`, therefore it's `USE_FUTEX` gets undefined and does not pick any other lock mechanism.

I know this gets a bit convoluted and bit lower level than i was expecting, so up to you what should be the best way to go.

From my side i would expect consistency across platforms and easily identifiable lock mechanism.
I would also expect respecting the user choice and either error out if not available and not fallback without warning or information.

miconda:
In general it used to be a fallback to the best option available, like for sockets with epoll, poll, kpoll, select ... so as long as it is some indication in the defined flags (e.g., via `kamailio -v` or `kamailio -I`), I see no problem to have different locks on different systems. If falling back is not possible with cmake, maybe the priority can be set per CPU/OS.

If the user wants to enforce an unavailable option, with some explicit configuration, then building can fail, but that should not be the behavior with the default configs for cmake.

xkaraman:
> In general it used to be a fallback to the best option available, like for sockets with epoll, poll, kpoll, select ... so as long as it is some indication in the defined flags (e.g., via `kamailio -v` or `kamailio -I`), 

Of course, fallback to the best available that's good. but what i meant consistency is about the reporting. One is using `futex` lock but still reports both and the `aarch64` is using the `fast_lock` but reports both. So you can't know which is the one being used actually. 

Also this decision, is a compile time decision, so we can know from start which ones are available and the decision should there. We don't need to define multiple ones ie (both `USE_FUTEX` and `FAST_LOCK` and `PTHREAD_MUTEX`) and then let code pick up the first. That's my belief at least.

Also, it might that my checks about the availability of each locks are not enough. For example, is using `futex` lock method, available by only checking if `linux/futex.h` only, or is something else needed as well?

> I see no problem to have different locks on different systems. 
>

Absolutely agree. but we need the checks for all the supported systems right? ie `aarch64` is not there in `make`, which i believe supports `futex`. correct me if i am wrong.

> If falling back is not possible with cmake, maybe the priority can be set per CPU/OS.

This [lock_methods.cmake](https://github.com/xkaraman/kamailio/blob/2f1c3d2447dcf860365314e6345b67dde68c0f94/cmake/lock_methods.cmake) file proposed, actually does exactly that in first point. It detects in configuration (even before compilation), which methods are currently available for the platform, and then using the priority we defined, pick the best one available.

> If the user wants to enforce an unavailable option, with some explicit configuration, then building can fail, but that should not be the behavior with the default configs for cmake.

Agree, i think cmake already fails in configuration currently, noting the available methods.

henningw:
As said, its a compile time definition. So the "kamailio -I" etc.. should report the method that was actually used at compilation time and not report two different methods with no clear way for the user to know which one is used, in my opinion. As we observed some locking method specific bugs some recently, its important for the user to know which locking method is used on the installation.

miconda:
Thanks, it can be merged!

PR #4367, opened 2025-08-13
presence: segfault on presentity replace

n2p-litvinov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Use after free issue that leads to crash in ps_ptable_replace function.

### DIFF
--- a/src/modules/presence/hash.c
+++ b/src/modules/presence/hash.c
@@ -1058,6 +1058,7 @@ int ps_ptable_replace(ps_presentity_t *ptm, ps_presentity_t *pt)
 	ps_presentity_t ptc;
 	ps_presentity_t ptv;
 	ps_presentity_t *ptn = NULL;
+	ps_presentity_t *ptnx = NULL;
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the PR. Just to understand the context better, does it happen in a specific situation or configuration?

n2p-litvinov:
Hello!
This bug was introduced by me in https://github.com/kamailio/kamailio/commit/c9f120f707ccc1692c5b142a51548d0c502fec43
Break the loop was replaced by presentity_free to delete all matched presentity instead of only first one, because on highly loaded system we see presentity duplicates in case if few publishes are processed in parallel.
We were not able to reproduce it on stage, but got few segfaults on production. Should I create a separate issue with attached backtraces?

henningw:
Thanks for the context. If its a bugfix for an earlier patch from your side, its fine for me. Lets wait a bit more for eventual other feedback, before merging it.

PR #4368, opened 2025-08-14
test: Add contention benchmark tests

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

We added a small tool that tries to benchmark the locking behaviour of different implementations like fast_lock, kamailio's implementation of futex with adapative and not adaptive behaviour and system futex.

It tracks the total locks, how many where contented and wait counter until it acquires it.


miconda:
I guess that the commit message prefix should be `misc:` instead of `test:`.

PR #4373, opened 2025-08-18
db_redis: add sentinel support

lbalaceanu:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This change adds redis sentinels support to db_redis module. The commits do the following:

- add **_with_sentinels, sentinels_config, redis_master_name, use_replicas_** kamailio config params
- sentinels_config example:
    password=password;sentinels=host1:port1|host2:port2|host3:port3
- when **_with_sentinels_** param is set, sentinels configuration in db_redis module is used to retrieve host and port of redis master, obsoleting the host and port part in usrloc's db_url 
- in sentinel scenarios with master/replica setup, if use_replicas is enabled and for whatever reason no replicas are available,   the module will fall back to connecting to the master (this behavior cannot be toggled off)
- introduced **_recheck_replicas_interval_** module param: it sets the interval in seconds to recheck for available replicas so that we don't remain connected to the master. Because of implementation details, the first check can happen after half of this interval. Value 0 means no recheck will be done.

### DIFF
--- a/src/modules/db_redis/db_redis_mod.c
+++ b/src/modules/db_redis/db_redis_mod.c
@@ -52,12 +52,18 @@ int db_redis_hash_expires = 0;
 str db_redis_hash_expires_str = str_init("");
 char db_redis_hash_expires_buf[20] = {0};
 
+time_t *db_redis_shared_time = NULL;
+time_t last_seen_time = 0;
+int using_master_read_only = 0;
+
 static int db_redis_bind_api(db_func_t *dbb);
 static int mod_init(void);
 static void mod_destroy(void);
 int use_replicas = 0; // 0 master/RW, 1 replica/RO
 int keys_param(modparam_t type, void *val);
 int sentinels_param(modparam_t type, void *val);
+int recheck_replicas_interval = 120; // seconds
+int min_recheck_interval = 60;		 // seconds
 
 static cmd_export_t cmds[] = {
 		{"db_bind_api", (cmd_function)db_redis_bind_api, 0, 0, 0, 0},
@@ -79,6 +85,7 @@ static param_export_t params[] = {
 				(void *)sentinels_param},
 		{"use_replicas", PARAM_INT, &use_replicas}, // 0 master/RW, 1 replica/RO
 		{"redis_master_name", PARAM_STR, &db_redis_master_name},
+		{"recheck_replicas_interval", PARAM_INT, &recheck_replicas_interval},
 #ifdef WITH_SSL
 		{"opt_tls", PARAM_INT, &db_redis_opt_tls},
 		{"ca_path", PARAM_STRING, &db_redis_ca_path},
@@ -141,6 +148,13 @@ int mod_register(char *path, int *dlflags, void *p1, void *p2)
 	return 0;
 }
 
+static void db_redis_timer(unsigned int ticks, void *param)
+{
+	if(db_redis_shared_time) {
+		*db_redis_shared_time = time(NULL);
+	}
+}
+
 static int mod_init(void)
 {
 	LM_DBG("module initializing\n");
@@ -157,9 +171,33 @@ static int mod_init(void)
 				db_redis_hash_expires_str.s, 20, "%d", db_redis_hash_expires);
 	}
 
-	if(db_redis_with_sentinels && sc.sentinel_list == NULL) {
-		LM_ERR("sentinels_config parameter is required when using sentinels\n");
+	db_redis_shared_time = shm_malloc(sizeof(time_t));
+	if(db_redis_shared_time == NULL) {
+		SHM_MEM_ERROR;
 		return -1;
+	}
+
+	*db_redis_shared_time = 0;
+
+	if(db_redis_with_sentinels) {
+		if(db_redis_sc.sentinel_list == NULL) {
+			LM_ERR("sentinels_config parameter is required when using "
+				   "sentinels\n");
+			return -1;
+		}
+
+		if(recheck_replicas_interval <= 0) {
+			LM_WARN("Considering replica recheck is not desired.\n");
+			return 0;
+		} else {
+			min_recheck_interval = recheck_replicas_interval / 2;
+			if(register_timer(db_redis_timer, 0, recheck_replicas_interval)
+					< 0) {
+				LM_ERR("Failed to register timer for rechecking replica "
+					   "availability\n");
+				return -1;
+			}
+		}
 	}
 	return 0;
 }
--- a/src/modules/db_redis/db_redis_mod.c
+++ b/src/modules/db_redis/db_redis_mod.c
@@ -30,6 +30,7 @@
 #include "db_redis_mod.h"
 #include "redis_dbase.h"
 #include "redis_table.h"
+#include "redis_sentinels.h"
 
 #ifdef WITH_SSL
 int db_redis_opt_tls = 0;
@@ -44,6 +45,8 @@ str redis_schema_path = str_init(SHARE_DIR "db_redis/kamailio");
 int db_redis_verbosity = 1;
 int mapping_struct_type = 0;
 str db_redis_hash_value = str_init("DUMMY");
+int db_redis_with_sentinels = 0;
+char *db_redis_master_name = "mymaster";
 
 int db_redis_hash_expires = 0;
 str db_redis_hash_expires_str = str_init("");
@@ -52,7 +55,9 @@ char db_redis_hash_expires_buf[20] = {0};
 static int db_redis_bind_api(db_func_t *dbb);
 static int mod_init(void);
 static void mod_destroy(void);
+int use_replicas = 0; // 0 master/RW, 1 replica/RO
 int keys_param(modparam_t type, void *val);
+int sentinels_param(modparam_t type, void *val);
 
 static cmd_export_t cmds[] = {
 		{"db_bind_api", (cmd_function)db_redis_bind_api, 0, 0, 0, 0},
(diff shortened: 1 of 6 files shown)


lbalaceanu:
Hi,

Does anyone have any comments about this?

Thank you

miconda:
I guess it can be merged, nobody having comments.

src/modules/db_redis/redis_connection.c
```diff
@@ -214,6 +249,69 @@ int db_redis_connect(km_redis_con_t *con)
...
+		reply = redisCommand(sentinel_con,
+				"SENTINEL get-master-addr-by-name %s", db_redis_master_name);
+		if(!reply) {
+			LM_ERR("Failed to run SENTINEL get-master-addr-by-name.\n");
+			goto err;
+		}
+		// The reply should be an array with two elements: the master IP and port.
+		if(reply->type == REDIS_REPLY_ARRAY && reply->elements == 2) {
+			if(con->id->host)
+				pkg_free(con->id->host);
```
henningw: Is this actually allocated as as dedicated segment?

src/modules/db_redis/redis_connection.c
```diff
@@ -374,6 +456,8 @@ int db_redis_connect(km_redis_con_t *con)
 		redisFree(con->con);
 		con->con = NULL;
 	}
+	if(sentinel_con)
```
henningw: If its correct to allocate new memory for con->id->host, we should probably also release it here in error cases.

henningw (commented):
I was not in the office for some time, so just had a quick look. Could you quickly check the memory allocation noted in the diff, just to be sure that its correct? Thank you.

lbalaceanu:
Thank you for the comments. Adding the label work-in-progress on this as we are actually extending this feature atm.

lbalaceanu:
Hi everyone, I finally updated the code for this redis sentinels PR. It works for us. Please review and comment.

lbalaceanu:
I have one question though. In tests, on Kamailio startup with fork=no I see the db_do_init() being called 3 times for a process that on the long run becomes the stand-alone receiver (kamctl ps). I see these are: 

1. fix_rls -> save_fixup -> domain_fixup 
2. main_loop -> init_child(PROC_SIPINIT) -> db_redis_init
3. main_loop -> init_child(PROC_POSTCHILDINIT) -> db_redis_init

What is this PROC_POSTCHILDINIT? How many connections to db_redis should be created in this case for the said process? 1 or 3? I am asking because at some point in set_con_details() I am actually changing con->id->host/ con->id->port details.

Thank you

henningw:
The PROC_POSTCHILDINIT will get you called after forking of all child processes. There is a special registration function (e.g. compare to ndb_redis) for that. If you don't set it, I think you should not see it. Maybe there is an issue in the core? You can find the introduction of this function in usrloc module here, for example  4780d46976a7

Usually you want to skip database initialization for non worker processes, this is what some of the other DB modules do.

lbalaceanu:
Hi @henningw,

I appreciate you pointing me to the commit. While I don't understand its rationale, given that Daniel specifically added that, I think it is safe to consider that the connection is really needed there.

Thank you

lbalaceanu:
Hi,

Does anybody have any comments?

Thanks

src/modules/db_redis/redis_sentinels.c
```diff
@@ -0,0 +1,529 @@
...
+			sc.sentinel_list_tail = elem;
+		} else {
+			sc.sentinel_list_tail->next = elem;
+			sc.sentinel_list_tail = elem;
+		}
+
+		sentinel = strtok(NULL, "|");
+	}
+
+	LM_INFO("Adding usrloc dburl sentinels:\n");
```
henningw: This log seems a bit odd, not sure what the relationship to usrloc is. The actual log output is also LM_DBG, maybe it make sense to also make this one LM_DBG?

src/modules/db_redis/db_redis_mod.c
```diff
@@ -141,6 +170,30 @@ static int mod_init(void)
 		db_redis_hash_expires_str.len = snprintf(
 				db_redis_hash_expires_str.s, 20, "%d", db_redis_hash_expires);
 	}
+
+	shared_time = shm_malloc(sizeof(time_t));
```
henningw: The malloc can fail, consider adding a check for NULL and SHM_MEM_ERROR or similar. Its also not freed in error cases, but arguably it don't matter much as the server will not start anyway. But it would be cleaner.

Actually, do we need a shm_malloc for this? Its an integer in the end.

src/modules/db_redis/db_redis_mod.c
```diff
@@ -119,13 +136,25 @@ int keys_param(modparam_t type, void *val)
...
 int mod_register(char *path, int *dlflags, void *p1, void *p2)
 {
 	if(db_api_init() < 0)
 		return -1;
 	return 0;
 }
 
+static void db_redis_timer(unsigned int ticks, void *param)
+{
+	if(shared_time) {
```
henningw: I am wondering if we need to protect the access to this shared variable with a Kamailio lock. The timer will run indepently and then the individual process might access when its changed.

src/modules/db_redis/redis_connection.c
```diff
@@ -537,11 +604,22 @@ void *db_redis_command_argv(km_redis_con_t *con, redis_key_t *query)
 	}
 	LM_DBG("query has %d args\n", argc);
 
-	redisReply *reply =
-			redisCommandArgv(con->con, argc, (const char **)argv, NULL);
-	if(con->con->err != REDIS_OK) {
-		LM_DBG("redis connection is gone, try reconnect. (%d:%s)\n",
-				con->con->err, con->con->errstr);
+	current_time = *shared_time;
```
henningw: See previous comment about the locking

src/modules/db_redis/db_redis_mod.c
```diff
@@ -44,15 +45,25 @@ str redis_schema_path = str_init(SHARE_DIR "db_redis/kamailio");
...
 int mapping_struct_type = 0;
 str db_redis_hash_value = str_init("DUMMY");
+int db_redis_with_sentinels = 0;
+char *db_redis_master_name = "mymaster";
 
 int db_redis_hash_expires = 0;
 str db_redis_hash_expires_str = str_init("");
 char db_redis_hash_expires_buf[20] = {0};
 
+time_t *shared_time = NULL;
```
henningw: I think we should add a db_redis prefix to this variables, as there are pretty generic and there is a risk that they overlapping with another module.

src/modules/db_redis/redis_sentinels.c
```diff
@@ -0,0 +1,529 @@
...
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+#include <stdlib.h>
+
+#include "redis_sentinels.h"
+#include "db_redis_mod.h"
+
+sentinel_config_t sc;
```
henningw: db_redis prefix - as above

src/modules/db_redis/redis_connection.c
```diff
@@ -607,17 +685,31 @@ int db_redis_get_reply(km_redis_con_t *con, void **reply)
...
 		LM_ERR("Internal error passing null connection\n");
 		return -1;
 	}
 
-	*reply = NULL;
-	ret = redisGetReply(con->con, reply);
-	if(con->con->err != REDIS_OK) {
-		LM_DBG("redis connection is gone, try reconnect. (%d:%s)\n",
-				con->con->err, con->con->errstr);
+	current_time = *shared_time;
```
henningw: locking as above

src/modules/db_redis/redis_sentinels.h
```diff
@@ -0,0 +1,107 @@
...
+} redis_role_t;
+
+typedef struct redis_sentinel
+{
+	char *host;
+	unsigned int port;
+	struct redis_sentinel *next;
+} redis_sentinel_t;
+
+typedef struct
```
henningw, line 59: Maybe I missed it, but I failed to find the actual usage of the "user" and "attr" (besides helping with the parsing) fields? It is of course ok to keep it for future use, if this was the intention.

henningw (commented):
Thank you for the update! I had a look to the usual topics, and added some remarks about smaller things I noticed.

src/modules/db_redis/redis_sentinels.c
lbalaceanu: Hi, you are right about this log mentioning usrloc and needing to be LM_DBG.

src/modules/db_redis/db_redis_mod.c
lbalaceanu: I consider we should use a shared mem value there because this is a value used by all Kamailio processes. Here is the background:

The timer is there for the "failover" cases when _use_replicas=1_ : when no Redis replica is reachable for some reason, for some process, Kamailio should try to connect, even for reading, to the Redis master (maybe this one is working). After some time, Kamailio should try to recheck the replica availability.

The thing is that we have multiple Redis connections and to different Redis replicas (each Kamailio receiver child holds a Redis connection of its own).

I implemented the recheck for replicas by using a "central timer" for ALL Kamailio processes and then only the processes interested (_using_master_read_only=1_) would look into this value. Possibly I should have implemented with a timer per proc - please advise or point me to some code example.

src/modules/db_redis/db_redis_mod.c
lbalaceanu: Since the shared_time is read by many processes and written to by only one, possibly we can get away with no lock ?!

src/modules/db_redis/redis_sentinels.h
lbalaceanu: Nice catch. You are right, attr is for parsing and user authentication is somehow for future use. At least as far as I know, user authentication is not yet supported for Redis.

lbalaceanu:
Hi,

I just pushed a change for LM_INFO -> LM_DBG, usrloc mention removal and db_redis prefix add.

Thank you

lbalaceanu:
Hi @miconda,

Since there are no more comments on this issue, can I merge the PR?

Thank you

src/modules/db_redis/db_redis_mod.c
henningw: It is an integer value, and you are also doing an assigment and not an increment. So its probably fine without a lock. Even if some process access it during the assignment, it should just receive a stale value.

src/modules/db_redis/db_redis_mod.c
henningw: A shared timer is probably fine, in the end you will be not execute it multiple times per second.

henningw:
Thanks for addressing the discussed topics, merged it.

lbalaceanu:
Thank you!

xkaraman:
@lbalaceanu 

Hey there, i was taking a look at #4773 and it seems that this won't compile when `hiredis_cluster` is also present in the OS. 

Did you had the same issue at all? or you only tested for `hiredis` installation?

Thanks!

lbalaceanu:
Hi @xkaraman,

Sorry, I didn't see this earlier, indeed I just tested only for hiredis installation. Also looking on the hiredis_cluster issue now.

Issue #3001, opened 2022-01-12
xHTTP_PROM does not include metric_type


mneirynck:
### Description
We have a couple of Kamailio containers running on AWS ECS Fargate. Using a Cloudwatch Agent sidecar container I should be able to collect metrics from the Kamailio containers and push them to Cloudwatch for monitoring.

```
xHTTP_PROM exposes the metrics like this:
# Kamailio whole internal statistics
kamailio_app_python3_traced_replies 0 1640265350616
kamailio_app_python3_traced_requests 0 1640265350616
kamailio_core_bad_URIs_rcvd 0 1640265350616
kamailio_core_bad_msg_hdr 0 1640265350616
```

While other prometheus metrics I have seen are throwing metrics like this:
```
# HELP kamailio_shmmem_free_size Free shared memory.
# TYPE kamailio_shmmem_free_size gauge
kamailio_shmem_free_size {value} {timestamp}
```
### Troubleshooting
Cloudwatch error:
```
2022-01-12T21:16:38Z E! metric_type ERROR: kamailio_shmem_free_size|cwagent-ecs-file-sd-config|10.102.99.61:8080|  
```

### Possible Solutions
Have a look at the exporter Callr:
https://github.com/florentchauveau/kamailio_exporter

### Additional Information

Latest Kamailio built from source on a Debian base image

vhernando:
Hello mneirynck,

in Prometheus metrics when there is no # TYPE metadata it defaults to UNKNOWN metric type. This is suitable for xHTTP_PROM because it blindly takes Kamailio registered counters and export them. We do not know if they are Prometheus counters or if they are Prometheus gauges. 

Currently OpenMetrics recommend not to use UNKNOWN metric types 
https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#unknown 
and Cloudwatch do not support unknown Prometheus metrics 
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/ContainerInsights-Prometheus-metrics-conversion.html

On the other hand kamailio_exporter exports a set of metrics and defines clearly if they are counters, gauges, etc. So at this moment I recommend you to use kamailio_exporter to interface with Cloudwatch
https://github.com/florentchauveau/kamailio_exporter/blob/master/collector.go

mneirynck:
Hi,
Sad we have to resort to a third party for this.
[...]

PR #4392, opened 2025-09-10
xhttp_prom: add support for `TYPE` and `HELP` metadata output

a12n:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #3001

#### Description

This adds support for conditionally generating `HELP` and `TYPE` Prometheus metadata strings into the output.

It may be a waste of bytes most of the time, but sometimes it may be useful (e.g., see #3001).

I noticed that most of the module parameters are prefixed with module name. Is there a reason it can't be just `metadata_flags` (`uptime_stat`, etc) instead of `xhttp_prom_metadata_flags` (`xhttp_prom_uptime_stat`, etc)? 

While working on this I've also tried to reduce code duplication in functions to create and free metrics. The related changes (in their current state) are in [a separate branch](/a12n/kamailio/tree/xhttp_prom_code_dedup). It's significant refactoring of code that seems to be working fine, so I'm not sure about it. Let me know if I should open another PR for this.

### DIFF
--- a/src/modules/xhttp_prom/doc/xhttp_prom_admin.xml
+++ b/src/modules/xhttp_prom/doc/xhttp_prom_admin.xml
@@ -255,6 +255,91 @@ modparam("xhttp_prom", "xhttp_prom_beginning", "");
 		</programlisting>
 	  </example>
 	</section>
+	<section id="xhttp_prom.p.xhttp_prom_metadata_flags">
+	  <title><varname>xhttp_prom_metadata_flags</varname> (integer)</title>
+	  <para>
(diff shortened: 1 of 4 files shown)


miconda:
I guess it can be merged, if no other comments.

henningw:
Thanks for the PR, merged. Regarding the code duplication refactoring, this is usually also a good idea and should be integrated. Only rarely we decide to not add it, if its for example much more complicated for only small savings. Just create a PR for the relevant commits (the linked tree from above contained several commits, some of them seems to be unrelated extensions).

PR #4394, opened 2025-09-12
CMakeLists.txt: drop EXTRAVERSION

mslehto:
I think stable branch does not need EXTRAVERSION string after MAJOR.MINOR.0 is released.

This patch sets EXTRAVERSION to empty value.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ project(
   HOMEPAGE_URL "https://www.kamailio.org")
 
 # Set the version number
-set(EXTRAVERSION "-rc0")
+set(EXTRAVERSION "")
 set(RELEASE "${PROJECT_VERSION}${EXTRAVERSION}")
 
 message(STATUS "PROJECT_VERSION: ${PROJECT_VERSION}")


mslehto:
Without this minor tweak stable Kamailio 6.0.3 CMake build reports version as 6.0.3-rc0`

```
git checkout -b testv603 5095b7b21eaca218d696df2ee0da8f53e5fec585

mkdir buildv603
cd buildv603

cmake -S .. -B .
cmake --build . --target kamailio
```
```
src/kamailio -V|head -1
version: kamailio 6.0.3-rc0 (x86_64/FreeBSD) 5095b7
```

( Maybe some other `CMakeLists.txt` approach is better to distinguish dev / stable branch. )

xkaraman:
I believe this should be merged as it relates to `6.0` branch. `Makefile` for 6.0 branch already set that to empty.

@miconda can this be changed even after the release of `6.0.3`?

henningw:
Daniel might missed the question. We usually don't change the source code of a released Kamailio version. But its not a big issue, more a reporting thing. Lets merge it to branch 6.0, main branch sets dev (as appropriate).

PR #4395, opened 2025-09-13
core: add `xavp_add_value_reserve()`, don't copy `SR_XTYPE_STR` data if it's `NULL`

a12n:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

This change should allow a small optimization/convenience improvement when building a combined string for XAVP value.

Consider the `xavp_example-1.c`. It builds a semicolon-separated string of parameters out of some data and makes XAVP from this combined string. It has to build the string entirely before passing it to `xavp_add_value()`. It either needs to use some fixed-length pad buffer or has to allocate/deallocate a temporary PKG buffer to build the string.

This change allows `SR_XTYPE_STR` values with `NULL` data (but with non-zero length) into `xavp_add_value()`. In that case `xavp_add_value()` will create a shared memory buffer for the string, but wouldn't try to copy string data into the buffer, leaving it uninitialized. The string content should be filled in-place later through the returned `sr_xavp_t *`. This allows simpler `xavp_example-2.c`.

<details>
<summary>xavp_example-1.c</summary>

```c
struct some_third_party_item
{
	const char *key;
	const char *value;
	size_t key_sz, value_sz;
};

int add_items_xavp1()
{
	struct some_third_party_item **list = some_third_party_func();

	str key = STR_STATIC_INIT("items");
	sr_xval_t value = {.type = SR_XTYPE_STR, .v = {.s = STR_NULL}};

	/* Compute total size of the combined string */
	for(int i = 0; list[i]; ++i) {
		value.v.s.len += (i != 0); /* ; */
		value.v.s.len += (int)list[i]->key_sz;
		value.v.s.len += 1; /* = */
		value.v.s.len += (int)list[i]->value_sz;
	}

	/* Allocate temporary buffer */
	if((value.v.s.s = pkg_malloc(value.v.s.len)) == NULL) {
		LM_ERR("couldn't allocate string value");
		return -1;
	}

	/* Build the string in the temporary buffer */
	char *dest = value.v.s.s;
	for(int i = 0; list[i]; ++i) {
		if(i != 0) {
			*dest++ = ';';
		}
		memcpy(dest, list[i]->key, list[i]->key_sz);
		dest += list[i]->key_sz;
		*dest++ = '=';
		memcpy(dest, list[i]->value, list[i]->value_sz);
		dest += list[i]->value_sz;
	}

	/* Allocate XAVP and its SHM for string, copy into SHM */
	if(!xavp_add_value(&key, &value, NULL)) {
		LM_ERR("couldn't allocate XAVP\n");
		pkg_free(value.v.s.s);
		return -1;
	}

	/* Deallocate the temporary buffer */
	pkg_free(value.v.s.s);

	return 0;
}
```
</details>

<details>
<summary>xavp_example-2.c</summary>

```c
struct some_third_party_item
{
	const char *key;
	const char *value;
	size_t key_sz, value_sz;
};

int add_items_xavp2()
{
	struct some_third_party_item **list = some_third_party_func();

	str key = STR_STATIC_INIT("items");
	sr_xval_t value = {.type = SR_XTYPE_STR, .v = {.s = STR_NULL}};

	/* Compute total size of the combined string */
	for(int i = 0; list[i]; ++i) {
		value.v.s.len += (i != 0); /* ; */
		value.v.s.len += (int)list[i]->key_sz;
		value.v.s.len += 1; /* = */
		value.v.s.len += (int)list[i]->value_sz;
	}

	sr_xavp_t *xavp;

	/* Allocate XAVP and its SHM for string, leave its content uninitialized */
	if(!(xavp = xavp_add_value(&key, &value, NULL))) {
		LM_ERR("couldn't allocate XAVP\n");
		return -1;
	}

	/* Build the string in the XAVP shared memory directly */
	char *dest = xavp->val.v.s.s;
	for(int i = 0; list[i]; ++i) {
		if(i != 0) {
			*dest++ = ';';
		}
		memcpy(dest, list[i]->key, list[i]->key_sz);
		dest += list[i]->key_sz;
		*dest++ = '=';
		memcpy(dest, list[i]->value, list[i]->value_sz);
		dest += list[i]->value_sz;
	}

	return 0;
}
```
</details>

### DIFF
--- a/src/core/xavp.c
+++ b/src/core/xavp.c
@@ -101,7 +101,10 @@ void xavp_free_unsafe(sr_xavp_t *xa)
 	shm_free_unsafe(xa);
 }
 
-static sr_xavp_t *xavp_new_value(str *name, sr_xval_t *val)
+/* Like xavp_new_value(), but allows NULL data in SR_XTYPE_STR
+ * value. In that case, it only reserves memory for the string and
+ * doesn't try to initialize string data. */
+static sr_xavp_t *xavp_new_value_reserve(str *name, sr_xval_t *val)
 {
 	sr_xavp_t *avp;
 	int size;
@@ -128,12 +131,24 @@ static sr_xavp_t *xavp_new_value(str *name, sr_xval_t *val)
 	memcpy(&avp->val, val, sizeof(sr_xval_t));
 	if(val->type == SR_XTYPE_STR) {
 		avp->val.v.s.s = avp->name.s + avp->name.len + 1;
-		memcpy(avp->val.v.s.s, val->v.s.s, val->v.s.len);
+		/* Leave the string uninitialized if no data is provided, it
+		 * should be filled in-place later. */
+		if(val->v.s.s) {
+			memcpy(avp->val.v.s.s, val->v.s.s, val->v.s.len);
+		}
 		avp->val.v.s.s[val->v.s.len] = '\0';
 		avp->val.v.s.len = val->v.s.len;
 	}
 
 	return avp;
+}
+
+/* Like xavp_new_value_reserve(), but requires non-NULL data pointer
+ * in SR_XTYPE_STR value. */
+static sr_xavp_t *xavp_new_value(str *name, sr_xval_t *val)
+{
+	assert(val->type != SR_XTYPE_STR || val->v.s.s);
+	return xavp_new_value_reserve(name, val);
 }
 
 int xavp_add(sr_xavp_t *xavp, sr_xavp_t **list)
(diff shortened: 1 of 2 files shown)


miconda:
I think this makes it hard to follow code and be safe of unintended mistakes. Rather than making the existing functions accept null pointers with greater-than-zero length, add new functions like xavp_reserve_value() so it is explicitly known what is supposed to happen.

a12n:
> I think this makes it hard to follow code and be safe of unintended mistakes. Rather than making the existing functions accept null pointers with greater-than-zero length, add new functions like xavp_reserve_value() so it is explicitly known what is supposed to happen.

Makes sense. I've added `xavp_add_value_reserve()`, please see the updates.

miconda:
As I thought more about it, I pushed a commit (e7f7e2dd73e45d82279ab5a9b1f8b4de891740c1) with a different implementation, adding a function that gets a new parameter `mode` to control how the xavp value is built. This should be easier to extend if other needs related to creating the xavp value come up in the future. The mode 1 should give you the behaviour you wanted.

PR #4398, opened 2025-09-19
Update Debian copyright file [skip ci]

mslehto:
Update debian/copyright file collection in pkg/kamailio/deb/.../copyright as suggested by Debian Lintian warnings.
Mostly file path adjustments, one license name clarification.


linuxmaniac:
Thanks!

PR #4405, opened 2025-09-23
core: try fix timer list only in case of issues

stefan-mititelu-idt:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #4346

#### Description

We have noticed constant PKG mem leak in "slow timer", while running #4346 fix in prod-like environment.
The fix is to run the sanity checks only when needed, not all the time; no PKG mem leak now.

### DIFF
--- a/src/core/timer.c
+++ b/src/core/timer.c
@@ -892,18 +892,36 @@ inline static void timer_list_expire(ticks_t t, struct timer_head *h
 	first = h->next;
 #endif
 
-	// check if timer circular double linked list has broken links
-	// try to recover if so, otherwise abort execution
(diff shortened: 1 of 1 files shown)


miconda:
Fine to merge from my point of view, if no other comments.

henningw:
Good improvement, should reduce the overhead considerable. Perhaps it make sense to even activate it by default, after some testing. The timer corruption is clearly a bug somehow, which also should be fixed. Maybe you can report somewhere if you find out some details what might cause it.

Issue #4418, opened 2025-10-06
VRF (Virtual Routing and Forwarding) support


linuxmaniac:
VRF (Virtual Routing and Forwarding) is a technology that allows multiple instances of a routing table to coexist within the same router, enabling traffic segmentation. Each VRF instance is isolated, meaning that packets from different VRFs do not interfere with each other, which is essential for managing traffic in complex networks. This capability is particularly useful in environments requiring distinct routing paths for different customers or services, enhancing security and efficiency.

Things that applications have to do in order to support [VRF](https://docs.kernel.org/networking/vrf.html#applications)

> Applications that are to work within a VRF need to bind their socket to the VRF device:
```
setsockopt(sd, SOL_SOCKET, SO_BINDTODEVICE, dev, strlen(dev)+1);
```
> or to specify the output device using cmsg and IP_PKTINFO.

sergey-safarov:
Here, it is not only binding. Here is the user identification and `Record-Route` and `Route` headers management.
Example.
Two users registered via different VRF:
1. user1 registered via vrf-1 and uses IP address 192.168.1.2, Kamailio IP 192.168.1.1;
2. user2 registered via vrf-2 and uses IP address 192.168.1.2; Kamailio IP 192.168.1.1.

In this case, we need to separate users and the Kamailio interface not only by IP. In the provided example, both are the same. We need add VRF ID for Kamailio interfaces and users.

Also, add 5 cents more. In the current IT, containerization technology is present everywhere. For containers separation used `namespace` technology. This technology, like VRF.  The difference is the loopback interface. When you create namespace, it creates a new loopback interface. In VRF loopback interface is not created.

Here ChatGPT example of how to open a socket in the namespace
```
#include <sched.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/socket.h>
#include <netinet/in.h>
#include <arpa/inet.h>

int main() {
    int fd = open("/var/run/netns/nsB", O_RDONLY);  // target namespace
    setns(fd, CLONE_NEWNET);                        // switch to nsB
    close(fd);

    int sock = socket(AF_INET, SOCK_STREAM, 0);
    struct sockaddr_in addr = {
        .sin_family = AF_INET,
        .sin_addr.s_addr = inet_addr("10.0.0.1"),
        .sin_port = htons(8080)
    };

    bind(sock, (struct sockaddr *)&addr, sizeof(addr));

    // Optionally, switch back to original namespace
    // (if you saved it before switching)

[...]

PR #4419, opened 2025-10-06
core: VRF support

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4418

#### Description

This will allow to define the VRF device associated to the listen interface
```
listen=10.10.10.1:5060 vrf "blue"
```

related #4418

### DIFF
--- a/src/core/cfg.lex
+++ b/src/core/cfg.lex
@@ -313,6 +313,7 @@ ADVERTISE	advertise|ADVERTISE
 VIRTUAL		virtual
 STRNAME		name|NAME
 AGNAME		agname|AGNAME
+VRF		vrf|VRF
 ALIAS		alias
 DOMAIN		domain
 SR_AUTO_ALIASES	auto_aliases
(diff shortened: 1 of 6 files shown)


henningw:
Thanks for the PR. Quick question, shouldn't we add the #if defined(__OS_linux) also to the UDP case? And maybe it makes also sense to add a #else case to output a LM_WARN if this feature is activated on a non-Linux OS?

linuxmaniac:
> Thanks for the PR. Quick question, shouldn't we add the #if defined(__OS_linux) also to the UDP case? 

It's already inside:
https://github.com/kamailio/kamailio/blob/ae13e4114bb38e7b62cab9a9b2a2c19dcdfee518/src/core/udp_server.c#L499

miconda:
Does this have to work only for `listen=proto:addr:port vrf "id"` variant? Or it should be also supported for the other cases, like when `advertise` or socket `name` are provided?

If the later, then it might be better to add it to the `socket` definition:

- https://www.kamailio.org/wikidocs/cookbooks/6.0.x/core/#socket

The grammar to define all the combinations possible for `listen` has to be extended and it is getting too complex, comparing with the one for `socket`.  You just have to add `vrf` to `socket_lattr:` group inside the `src/core/cfg.lex`, extend the `socket_attrs_t` structure to hold its value, ... But then, all combinations with or without advertise, socket name, ... should be possible.

linuxmaniac:
> Does this have to work only for `listen=proto:addr:port vrf "id"` variant? Or it should be also supported for the other cases, like when `advertise` or socket `name` are provided?

It would need to be supported for the other cases too.

> If the later, then it might be better to add it to the `socket` definition:
> 
> * https://www.kamailio.org/wikidocs/cookbooks/6.0.x/core/#socket
> 
> The grammar to define all the combinations possible for `listen` has to be extended and it is getting too complex, comparing with the one for `socket`. You just have to add `vrf` to `socket_lattr:` group inside the `src/core/cfg.lex`, extend the `socket_attrs_t` structure to hold its value, ... But then, all combinations with or without advertise, socket name, ... should be possible.

But should I remove the listen variant and just go for the socket version?

miconda:
I added `agname` only to `socket` for the same reason that extending the grammar for `listen` would have made it too complex.

So I would say to remove it from listen unless you want to extend it to cover all variants, then have it only for `socket`, which should be seen as a more advanced/flexible alternative to listen.

miconda:
Ahh, I see the forced-push does that, so I am fine with it.

PR #4423, opened 2025-10-06
cdp: support vrf on acceptor and peer definitions

linuxmaniac:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4418

#### Description

```
<Peer FQDN="hss.open-ims.test" Realm="open-ims.test" port="3868" src_addr="192.168.2.1" vrf="blue" />
<Acceptor port="3870" bind="192.168.2.1" vrf="blue"/>
```

### DIFF
--- a/src/modules/cdp/peer.c
+++ b/src/modules/cdp/peer.c
@@ -56,7 +56,7 @@
  * @param port - port of the peer to connect to
  * @returns the new peer* if ok, NULL on error
  */
-peer *new_peer(str fqdn, str realm, int port, str src_addr, str proto)
+peer *new_peer(str fqdn, str realm, int port, str src_addr, str proto, str vrf)
 {
 	peer *x;
 	x = shm_malloc(sizeof(peer));
@@ -73,6 +73,9 @@ peer *new_peer(str fqdn, str realm, int port, str src_addr, str proto)
 		goto error;
 	shm_str_dup_macro(x->src_addr, src_addr);
 	if(!x->src_addr.s)
+		goto error;
+	shm_str_dup_macro(x->vrf, vrf);
+	if(!x->vrf.s)
 		goto error;
 	shm_str_dup_macro(x->proto, proto);
 	x->port = port;
@@ -111,6 +114,8 @@ void free_peer(peer *x, int locked)
 		shm_free(x->realm.s);
 	if(x->src_addr.s)
 		shm_free(x->src_addr.s);
+	if(x->vrf.s)
+		shm_free(x->vrf.s);
 	lock_destroy(x->lock);
 	lock_dealloc((void *)x->lock);
 	shm_free(x);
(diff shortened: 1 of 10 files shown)


linuxmaniac:
@vingarzan I see no issue adding this to the acceptor and peer definitions. Do you think otherwise?

src/modules/cdp/peer.c
```diff
@@ -74,6 +74,9 @@ peer *new_peer(str fqdn, str realm, int port, str src_addr, str proto)
 	shm_str_dup_macro(x->src_addr, src_addr);
 	if(!x->src_addr.s)
 		goto error;
+	shm_str_dup_macro(x->vrf, vrf);
+	if(!x->vrf.s)
+		goto error;
```
vingarzan, line 79: I don't remember how this macro works... because if the there is no parameter, in my mind it could skip uselessly allocating an empty pointer. So I would've wrapped this in an `if`, since the `vrf` parameter is optional, right?

linuxmaniac: Well, I'm just following the same logic as, for instance, optional ``src_addr``.

linuxmaniac:
> Looks like a clean new (optional) feature, so 👍 But do please check my comment below, to ensure it won't crash without the vrf parameter.

It doesn't crash without ``vrf``. That was tested.

If you think the optional parameters should not reserve empty pointers. I would say I could do that in a different PR.

src/modules/cdp/peer.c
vingarzan: https://github.com/kamailio/kamailio/blob/d2c3db281e08f53f4b4e892978ebf43df6be58ce/src/modules/cdp/utils.h#L67

So it seems it is wasteful, but safe (including zero-padding). I was more of a fan of a variant where it would not allocate anything on empty source and the check if "empty" was on str->len, not str->s. Hence all is good.

vingarzan:
> > Looks like a clean new (optional) feature, so 👍 But do please check my comment below, to ensure it won't crash without the vrf parameter.
> 
> It doesn't crash without `vrf`. That was tested.
> 
> If you think the optional parameters should not reserve empty pointers. I would say I could do that in a different PR.

Nah, that's a huge change :stuck_out_tongue_winking_eye: . I was just biased by a parallel/similar platform where I used cdp, but str was there a purely binary blob. 

The macro here is wasteful but arguably safer, e.g. if someone uses str->s like a char* (albeit that's still not a good thing, since str was defined to compensate for the limitations of char* when storing binary data, so that's like going back and un-compensating :upside_down_face: ).

henningw:
Nice addition, thanks. Could you also document the new parameter "vrf" briefly in the module docs and example cfg? Right now its only listed, but not explanation what it actually does.

Issue #4415, opened 2025-10-03
Topos_htable Not sending SIP notify talk event


ServanVcare:
### Description

In Kamailio 6.0.3, we enabled the topos and topos_htable modules. We initiate a call from an Asterisk and then send a SIP notify with the talk event header. If topos is enabled, all messages are properly forwarded to the phone, except for the SIP notify with the talk event header. If we pick up the call via the phone and then send a SIP notify with the hold event header, it is properly forwarded to the phone and the phone set the call on hold.

If we only use the topos module via SQL, the SIP notify with the Talk event header is properly forwarded to the device.

#### Reproduction

Use the topos_htable and send a SIP notify with event header Talk to the phone.

We used the following module parameters for topos and topos_htable:
```
modparam("topos", "storage", "htable")
modparam("topos", "branch_expire", 180)    
modparam("topos", "dialog_expire", 10800)  
modparam("topos", "clean_interval", 3600)
modparam("topos", "sanity_checks", 0)
modparam("topos", "contact_host", "<ip-adres>")
modparam("topos", "rr_update", 1)
modparam("topos", "contact_mode", 1)
```
#### SIP Traffic

PCAP of this happening: [sip_notify_examples.zip](https://github.com/user-attachments/files/22680836/sip_notify_examples.zip)

### Possible Solutions

Yes, we have modified topos_htable_storage.c with the following function modified in the code:
```
// we change the arguments in the functions
// static int tps_htable_load_initial_method_branch(tps_data_t *md, tps_data_t *sd)
static int tps_htable_load_initial_method_branch(sip_msg_t* msg, tps_data_t *md, tps_data_t *sd)
{
	str hkey;
	ht_cell_t *hval;
	char *ptr;
	int ret = 0;
	int i = 0;
	str xuuid = str_init("");
	str xtag = str_init("");
	str smethod = str_init("INVITE");

	// checks
	if(md == NULL || sd == NULL) {
		LM_DBG("NULL pointers");
		return -1;
	}

	if(md->a_callid.len <= 0 || md->b_tag.len <= 0) {
		LM_DBG("no call-id or to-tag for this message\n");
		return -1;
	}

	LM_DBG("HERE\n");

	// build key
	if(md->direction == TPS_DIR_DOWNSTREAM) {
		if(md->b_tag.s != NULL) {
			xtag.s = md->b_tag.s;
			xtag.len = md->b_tag.len;
		}
	} else {
		if(md->a_tag.s != NULL) {
[...]

PR #4424, opened 2025-10-06
topos_htable: Check on sip notify talk

ServanVcare:
- Check if an SIP Notify in an dialog with event-header value is talk, then it is not an subcribe

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #4415 

#### Description

In Kamailio 6.0.3, we enabled the topos and topos_htable modules. We initiate a call from an Asterisk and then send a SIP notify with the talk event header. If topos is enabled, all messages are properly forwarded to the phone, except for the SIP notify with the talk event header. If we pick up the call via the phone and then send a SIP notify with the hold event header, it is properly forwarded to the phone and the phone set the call on hold.

If we only use the topos module via SQL, the SIP notify with the Talk event header is properly forwarded to the device.

### DIFF
--- a/src/modules/topos_htable/topos_htable_storage.c
+++ b/src/modules/topos_htable/topos_htable_storage.c
@@ -213,7 +213,7 @@ static int tps_htable_insert_initial_method_branch(
 	return 0;
 }
 
-static int tps_htable_load_initial_method_branch(tps_data_t *md, tps_data_t *sd)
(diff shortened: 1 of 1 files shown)


ServanVcare:
Close because of an better PR #4425

PR #3627, opened 2023-11-02
topos: add compatiblity for early in-dialog NOTIFY

mauritsderuiter95:
#### Pre-Submission Checklist

- [x] Commit message has the format required by CONTRIBUTING guide
- [x] Commits are split per component (core, individual modules, libs, utils, ...)
- [x] Each component has a single commit (if not, squash them into one commit)
- [x] No commits to README files for modules (changes must be done to docbook files
in `doc/` subfolder, the README file is autogenerated)

#### Type Of Change
- [x] Small bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would change existing functionality)

#### Checklist:

- [x] PR should be backported to stable branches
- [x] Tested changes locally
- [ ] Related to issue #XXXX (replace XXXX with an open issue number)

#### Description
After enabling the Topos module, we ran into a bit of trouble with early in-dialog notifies. We use the `event: talk` package from Cisco Broadsoft to be able to do call control, as described in chapter 6 here: https://pubhub.devnetcloud.com/media/broadsoft-docs/docs/pdf/BW-SIPAccessSideExtensionsInterfaceSpec.pdf

This will send a NOTIFY with a header `Event: talk` before the first `200 OK`. Currently this doesn't work with the topos module, since the NOTIFY method without a B-side contact will not check the INVITE record. When added to those methods in `tps_msg.c`, it tries to check the SUBSCRIBE for all NOTIFY's in `tps_storage.c`. The NOTIFY with  `Event: talk` will follow on an INVITE, so there was a small change needed there.

I've limited it to just the NOTIFY's with the `Event: talk` header, to minimize any impact.

PR #4425, opened 2025-10-07
topos_htable: remove smethod from transaction key

stefan-mititelu-idt:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #4415

### DIFF
--- a/src/modules/topos_htable/topos_htable_storage.c
+++ b/src/modules/topos_htable/topos_htable_storage.c
@@ -143,23 +143,20 @@ static int tps_htable_insert_initial_method_branch(
 
 	// base64 encode key values
 	if(_tps_base64) {
-		base64url_enc(md->s_method.s, md->s_method.len, _tps_base64_buf[0],
+		base64url_enc(md->a_callid.s, md->a_callid.len, _tps_base64_buf[0],
 				TPS_BASE64_SIZE - 1);
-		base64url_enc(md->a_callid.s, md->a_callid.len, _tps_base64_buf[1],
-				TPS_BASE64_SIZE - 1);
-		base64url_enc(md->b_tag.s, md->b_tag.len, _tps_base64_buf[2],
+		base64url_enc(md->b_tag.s, md->b_tag.len, _tps_base64_buf[1],
 				TPS_BASE64_SIZE - 1);
 		base64url_enc(
-				xuuid.s, xuuid.len, _tps_base64_buf[3], TPS_BASE64_SIZE - 1);
+				xuuid.s, xuuid.len, _tps_base64_buf[2], TPS_BASE64_SIZE - 1);
 
-		ret = snprintf(ptr, TPS_HTABLE_SIZE_KEY, "%s|%s|%s|x%s",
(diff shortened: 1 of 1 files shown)


stefan-mititelu-idt:
To be merged after some more internal testing with the new format.

henningw:
I looked to the PR, and wondered if this is the right fix. If you compare e.g. to the fixes from #3627 , in particular commit 511964b, I think it just ignores the method name for the specific NOTIFY and not for all different SIP messages. Don't you need the method name for other messages?

stefan-mititelu-idt:
@henningw I had double checked the code flow for this and still think method name is not needed for the reasons below.

Short answer:
No, callid should be enough for the other methods too, because that key format will always identify either INVITE or SUBSCRIBE transactions only, in the first place.

Long answer:
1. When storing transaction or dialog data, topos_htable storage must build both key and value and store them as key -> value
2. When storing transaction, topos_htable storage builds the key like this:
    - for all transactions, use via-branch as key
    - for "initial_method_branch" transactions in particular (INVITE/SUBSCRIBE), one more metadata-like entry is stored in addition, which used the "method|callid|tag|uuid" as key and via-branch as value. This is just to have a reference to those INVITE/SUBSCRIBE transactions. But since both methods are dialog-forming, **callid is different**, so no need to differentiate them by method.
3. When loading transaction, the "nosql" storages builds the key to be looked up like this, depending on the "mode" function param:
    - if mode = 0 => looks up using via-branch directly and gets the transaction values
    - if mode = 1 => looks up using "method|callid|tag|uuid", gets via-branch value. Then looks it up again using that via-branch as key, and gets the transaction values. This is the case for that NOTIFY, and can only use its own callid to figure out to which "initial_method_branch" it belongs (INVITE or SUBSCRIBE). Generally speaking, is the same for any "METHOD_BYE | METHOD_INFO | METHOD_PRACK | METHOD_UPDATE | METHOD_NOTIFY", no need for method.

The above is also applicable to topos_redis (since it was the dev start point of topos_htable).
We will do some more testing for this, but just looking at the code seems like no problem.

Thanks you,
Stefan

ServanVcare:
@stefan-mititelu-idt Hi, we have tested you're changes in a load test with more then 1000 simultanious calls and we don't see issues.
When will this be merged?

Kind regards,
Servan

stefan-mititelu-idt:
From our side, can be merged too. No further issues observed, related to this.

henningw:
Thank you both for the additional testing and feedback. Merged.

ServanVcare:
Thanks @stefan-mititelu-idt and @henningw !

ServanVcare:
Hi, 
Is it true that this wasn't merged into 6.0.5? Can it be included in the next release?
Kind regards,
Servan Kroes

henningw:
> Hi, Is it true that this wasn't merged into 6.0.5? Can it be included in the next release? Kind regards, Servan Kroes

Hello, it was not cherry-picked, but just did it. It will then be in the next release and the daily packages.

PR #4427, opened 2025-10-09
core: lower log level for qm_free

stefan-mititelu-idt:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Hello, can this log be lowered to at least WARNING? Or maybe even lower?

### DIFF
--- a/src/core/mem/q_malloc.c
+++ b/src/core/mem/q_malloc.c
@@ -552,7 +552,7 @@ void qm_free(void *qmp, void *p)
 					p, file, func, line, f->file, f->func, f->line);
 			abort();
 		} else {
-			LM_CRIT("BUG: freeing already freed pointer (%p),"
+			LM_WARN("BUG: freeing already freed pointer (%p),"
 					" called from %s: %s(%d), first free %s: %s(%ld) - "
 					"ignoring\n",
 					p, file, func, line, f->file, f->func, f->line);


miconda:
That log is about a situation that should not happen, when do you get it that becomes annoying?

stefan-mititelu-idt:
```/usr/sbin/kamailio[524249]: CRITICAL: <core> [core/mem/q_malloc.c:519]: qm_free(): BUG: freeing already freed pointer (0x7fc45fbbed40), called from tls: tls_init.c: ser_free(323), first free tls: tls_init.c: ser_free(323) - ignoring```

It's on an older 5.5.7 kamailio on a debian 11, and probably is related to the libssl library version:
 ```
 ii  libssl1.1:amd64                   1.1.1w-0+deb11u4                   amd64        Secure Sockets Layer toolkit - shared libraries
 ```
 
It does not happen too often; the annoying part is that can trigger CRITICAL alerts even though one could probably live with that log message for once in a while.

henningw:
I also think its a condition which should not happen in a crucial area of the core. So I don't think its a good idea to reduce it to WARN level.

stefan-mititelu-idt:
Ok. Thanks for the feedback! Closing this PR.

PR #4435, opened 2025-10-16
core/parser: minor bugfix for URI_ENCLOSED

groissen:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
Two small bugfixes improving parser stability and PCSCF reliability.

- **core/parser**: added missing `URI_ENCLOSED` case in `parse_addr_spec()`  
  → prevents `LM_ERR("unexpected char ...")` when parsing `<...>` segments.
- **ims_registrar_pcscf**: fixed incorrect `memset()` target in `pcscf_unregister()`  
  → ensures `search_ci` is properly zero-initialized, preventing undefined behavior.

### DIFF
--- a/src/core/parser/parse_addr_spec.c
+++ b/src/core/parser/parse_addr_spec.c
@@ -784,6 +784,8 @@ char *parse_addr_spec(char *const buffer, const char *const end,
 						to_b->display.s = tmp;
 						status = DISPLAY_QUOTED;
 						break;
+					case URI_ENCLOSED:
+						break;
 					case DISPLAY_QUOTED:
 						status = E_DISPLAY_QUOTED;
 						to_b->display.len = tmp - to_b->display.s + 1;


linuxmaniac:
I've merged the ``ims_registrar_pcscf`` path manually. Please rebase and push force for the core part.

henningw:
Thanks for the PR - regarding the core fix, can you give a bit more details (like valid Header that is not correctly parsed right now) to properly evalute it? Its a sensitive part of the server in the end, thanks.

groissen:
The Issue appears when a header has a quoted display-name and address-spec enclosed in <...>.
For Example:

```
To: "Max Example" <sip:max@example.com>;tag=abc
```

While parsing, the state machine might still encounter a " token after it already transitioned to URI_ENCLOSED.
Without the "case URI_ENCLOSED; break;" guard, the parser hits the default branch and log.

The URI_ENCLOSED treats the case as a no-operation, making sure the parser ignores the quote character when already inside the <...>.

henningw:
> The Issue appears when a header has a quoted display-name and address-spec enclosed in <...>. For Example: To: "Max Example" sip:max@example.com;tag=abc
> 
> While parsing, the state machine might still encounter a " token after it already transitioned to URI_ENCLOSED. Without the "case URI_ENCLOSED; break;" guard, the parser hits the default branch and log.
> 
> The URI_ENCLOSED treats the case as a no-operation, making sure the parser ignores the quote character when already inside the <...>.

Sorry, there is no "<..>" in the Example above. Maybe you need to add it as code inside the comment to prevent display problems.

groissen:
Yeah, my bad - GitHub seems to have stripped the `<...>` part. 
Here is the corrected example:
`To: "Max Example" <sip:max@example.com>;tag=abc`

While parsing, the state machine might still encounter a '"' token after it already transitioned to URI_ENCLOSED.
Without the `case URI_ENCLOSED; break; ` guard, the parser hits the default branch and log.

The URI_ENCLOSED treats the case as a no-operation, making sure the parser ignores the quote character when already inside the `<...>`.

henningw:
Your example don't contain an additional  "**"**" char. 
You are saying that the parser should accept the following format?

`To: "Max Example" <sip:"max"@example.com>;tag=abc`

I have not checked the RFC in full detail, but it does not look valid on a first view.

miconda:
To-header with URI enclosed in angle brackets `<...>` is so common, can you give the entire SIP message that throws such error?

I did a quick test with sipexer and kamailio devel version and no error was printed. I used:

```
$ sipexer -fv tname:"Max Example" -fv tuser:max -fv tdomain:example.com -fv totag:abc
```

Log from sipexer:

```
[info] [sipexer.go:1591] main.SIPExerDialogLoop(): local socket address: 127.0.0.1:58019 (udp)
[info] [sipexer.go:1592] main.SIPExerDialogLoop(): local via address: 127.0.0.1:58019
[info] [sipexer.go:1593] main.SIPExerDialogLoop(): sending to udp 127.0.0.1:5060: [[---
OPTIONS sip:127.0.0.1:5060 SIP/2.0
Via: SIP/2.0/UDP 127.0.0.1:58019;rport;branch=z9hG4bKSG.075f344e-081a-4101-b649-7e5d726649df
From: <sip:alice@localhost>;tag=d0f30d9e-6704-4e99-8944-90d3f5d04a1c
To: "Max Example" <sip:max@example.com>;tag=abc
Call-ID: 44595a09-7905-472b-8140-37e084e0c3a3
CSeq: 336064 OPTIONS
Date: Thu, 16 Oct 2025 13:38:18 CEST
User-Agent: SIPExer v1.1.0
Max-Forwards: 10
Content-Length: 0

```

The To header is the same as in comments above, but Kamailio printed no error message.

miconda:
Somehow I overlooked Henning's comment, but if that's the case when quotes are in the user part of the enclosed URI, I also think it is invalid. I tested with the example from the comment https://github.com/kamailio/kamailio/pull/4435#issuecomment-3409648751

groissen:
Thanks everyone for the feedback!

I discussed this internally with a colleague who has more background on the original change, and we realized that this patch was addressing the wrong layer.

The real issue was related to an asterisk (*) not being allowed in the URI host part, which should be handled during URI parsing rather than in `parse_addr_spec()`.

I'll close this Pull-Request and later on open a new one with the correct fix at the URI-parsing level.

Thanks again for reviewing and the discussion - and sorry for the confusion.

PR #4438, opened 2025-10-18
Multiple Modules: integer support for extra DB fields

devopsec:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
This feature addition adds support for loading integer-type DB fields into AVPs, for the "extra" fields that are loaded when matching the authentication / route.  
The following modules are effected by this change:
- dispatcher: support integer fields for ds_db_extra_attrs
- permissions: support integer fields for peer_tag
- htable: support integer fields for htable param (cols attribute key_name)

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -1408,6 +1408,7 @@ int ds_load_db(void)
 
 		attrs.s = 0;
 		attrs.len = 0;
+		memset(ds_attrs_buf, 0, DS_ATTRS_MAXSIZE);
 		if(nrcols >= 5) {
 			if(!VAL_NULL(values + 4)) {
 				attrs.s = VAL_STR(values + 4).s;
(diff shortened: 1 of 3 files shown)


devopsec:
IMO, the [output format](https://github.com/kamailio/kamailio/actions/runs/18610861848/job/53068752816?pr=4438) from check-format is much harder to read..  
I always have to reformat sections to understand the code when it does hard wraps so often.  
LMK your thoughts, I can run it back through clang-format if needed.

devopsec:
seems like the format hook is still failing.  
ran formatter like so (should be using rules in repo):

```
clang-format -i src/modules/{permissions,htable,dispatcher}/*.{c,h}
```

henningw:
The format checker runs over each individual commits, so it fails on the first commit already. You probably need to squash the commits somehow, or maybe just close this PR and create a new one with the correct commits. This should also fix the commit check error.

devopsec:
fixed formatting.. awaiting final checks but i think this is formatted correctly

devopsec:
@henningw any more updates needed for this PR?

henningw:
Thanks for the reminder. I have not reviewed the logic in full detail, but it looks like a good addition to the modules to me. Just to double check, does it need a database structure change for the respective modules as well?

As it touches several important modules, it would be good to have also feedback from another developer on this PR.

devopsec:
No changes in DB structure.  
We have a few customers using integer fields for the "extra" fields loaded by those modules and for a few hash tables so this is essentially adding support for those types to get loaded instead of hacking around it with generated varchar columns

devopsec:
@henningw checking in on this one

devopsec:
@henningw I found an edge case where the kextra field bufs for those ints were not reset to 0 prior to the next loop iteration using them.  
fixed that and rebased on current master..

devopsec:
hey team, any movement on merging this one?  
if there are concerns or changes needed please let me know but we have had this in prod for one of our customers for a month already..

miconda:
I am merging this one, thanks!

Issue #4315, opened 2025-07-11
outbound + loose_route: new exit code when flow token does not exist


sergey-safarov:
### Description
Use case WebRTC client connected to Kamilio Edge proxy.

For network connectivity reasons, the flow token may no longer be valid. In this case, behavior should be described in the [RFC5626, 5.3.1 Processing Incoming Requests](https://datatracker.ietf.org/doc/html/rfc5626#section-5.3.1)
> If the flow no longer exists, the proxy SHOULD send a 430 (Flow Failed) response to the request.

Also [9.3  Incoming Call and Proxy Crash](https://datatracker.ietf.org/doc/html/rfc5626#section-9.3)
> Bob's authoritative proxy first tries the flow to EP1, but EP1 no longer has a flow to Bob, so it responds with a 430 (Flow Failed) response.

To get it working, it will be fine to use construction like
```
if (!loose_route()) {
	switch($rc) {
	case -4:
		sl_send_reply("430", "Flow Failed");
		exit;
	case -2:
		sl_send_reply("403", "Forbidden");
		exit;
	default:
		xlog("L_ERR", "in request_route\n");
		sl_reply_error();
		exit;
	}
}
```
Where `loose_route` returns a new error code (`-4` for example). This error code is generated when the flow token points to not existing TCP/TLS/WSS connection.

### Expected behavior
Kamailio `loose_route` allows the check of flow token validity before messages relayed via non-existent TCP/TLS/WSS connections.

#### Actual observed behavior
For now, Kamailio tries to relay messages to non-existent client connections and generate an error message like
```
477 Unfortunately error on sending to next hop occurred (477/TM)
500 Message processing error (2/TM)
```

#### How to reproduce

Required to start Kamailio with a config like
```
listen=udp:127.0.0.1:5060
listen=tcp:127.0.0.1:5060

loadmodule "pv.so"
loadmodule "xlog.so"
loadmodule "sl.so"
loadmodule "tm.so"
loadmodule "siputils.so"
loadmodule "rr.so"

loadmodule "stun.so"
loadmodule "outbound.so"
modparam("outbound", "force_outbound_flag", 1)
modparam("outbound", "flow_token_secret", "kamailio")

loadmodule "debugger.so"
modparam("debugger", "cfgtrace", 1)
modparam("debugger", "log_level_name", "exec")

request_route {
	loose_route();
	switch($rc) {
	case -3:
		xlog("L_ERR", "loose_route code -3\n");
		break;
	case -2:
[...]

PR #4449, opened 2025-10-26
outbound: added check_flow_token function

sergey-safarov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4315

#### Description
Added a function to check the flow token.
Useful to make sure the client TCP connection is alive.

### DIFF
--- a/src/modules/outbound/doc/outbound_admin.xml
+++ b/src/modules/outbound/doc/outbound_admin.xml
@@ -488,7 +488,85 @@ event_route[tm:branch-failure:FAIL-BRANCH] {
 	</section>
 	</section>
 
-
+	<section>
+		<title>Functions</title>
+		<section id="outbound.f.check_flow_token">
+			<title><function moreinfo="none">check_flow_token()</function></title>
+			<para>This function search flow token for the current
+				message and make checks:</para>
+			<orderedlist>
+				<listitem><para>headers can be parsed;</para></listitem>
+				<listitem><para><emphasis>Route</emphasis> header can be parsed;</para></listitem>
+				<listitem><para>top route header is myself;</para></listitem>
+				<listitem><para>flow token present in the <emphasis>Route</emphasis> header;</para></listitem>
+				<listitem><para>flow token can be decoded;</para></listitem>
+				<listitem><para>TCP/TLS or WS/WSS connection present;</para></listitem>
+				<listitem><para>flow token is not expired;</para></listitem>
+			</orderedlist>
+			<para>Return codes:</para>
+			<orderedlist>
+				<listitem><para><emphasis>1</emphasis> - flow token is
+					valid and can be used to route message to the next hop</para></listitem>
+				<listitem><para><emphasis>-1</emphasis> - no <emphasis>Route</emphasis> header present</para></listitem>
+				<listitem><para><emphasis>-2</emphasis> - host in the first <emphasis>Route</emphasis> value is not current host</para></listitem>
+				<listitem><para><emphasis>-3</emphasis> - no flow token present</para></listitem>
+				<listitem><para><emphasis>-4</emphasis> - cannot decode flow token</para></listitem>
+				<listitem><para><emphasis>-5</emphasis> - connection associated with this token does not exist </para></listitem>
+				<listitem><para><emphasis>-6</emphasis> - flow token expired</para></listitem>
+				<listitem><para><emphasis>-7</emphasis> - cannot parse <emphasis>Route</emphasis> header</para></listitem>
+				<listitem><para><emphasis>-8</emphasis> - cannot parse headers</para></listitem>
+			</orderedlist>
+			<example>
+				<title><function>check_flow_token</function> usage</title>
+				<programlisting>
+...
+	check_flow_token();
+	switch $rc {
+		case -8:
+			xlog("cannot parse headers\n");
+			sl_send_reply("403", "Forbidden");
+			break;
+		case -7:
+			xlog("cannot parse Route header\n");
+			sl_send_reply("403", "Forbidden");
+			break;
+		case -6:
+			xlog("expired flow token\n");
+			sl_send_reply("430", "Flow Failed");
+			break;
+		case -5:
+			xlog("no TCP conection\n");
+			sl_send_reply("430", "Flow Failed");
+			break;
+		case -4:
+			xlog("cannot decode flowtoken\n");
+			sl_send_reply("403", "Forbidden");
+			break;
+		case -3:
+			xlog("no flow token\n");
+			loose_route();
+			break;
+		case -2:
+			xlog("Route is not my self\n");
+			sl_send_reply("403", "Forbidden");
+			break;
+		case -1:
+			xlog("no Route header\n");
+			break;
+		case 1:
+			xlog("correct flow token\n");
+			loose_route();
+			break;
+		default:
+			xlog("not supported return code: $rc\n");
+			sl_send_reply("403", "Forbidden");
+			break;
+	}
+...
+				</programlisting>
+			</example>
+		</section>
+	</section>
 	<section>
 	<title>Parameters</title>
 	<section>


src/modules/outbound/doc/outbound_admin.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding='ISO-8859-1'?>
-<!DOCTYPE book PUBLIC "-//OASIS//DTD DocBook XML V4.4//EN"
+<!DOCTYPE chapter PUBLIC "-//OASIS//DTD DocBook XML V4.4//EN"
```
henningw, line 2: Why you changed from book to chapter type here? Type book is used in 99% of all places.

src/modules/outbound/doc/outbound.xml
```diff
@@ -6,6 +6,9 @@
 <!ENTITY % docentities SYSTEM "../../../../doc/docbook/entities.xml">
 %docentities;
 
+<!ENTITY % xinclude SYSTEM "http://www.docbook.org/xml/4.4/xinclude.mod" >
```
henningw, line 9: Why you added this? Its not used elsewhere.

src/modules/outbound/outbound_mod.c
```diff
@@ -503,3 +518,137 @@ int bind_ob(struct ob_binds *pxb)
...
 	return 0;
 }
+
+/*!
+ * \brief Check if URI is myself
+ * \param _host host
+ * \param _port port
+ * \return 0 if the URI is not myself, 1 otherwise
+ */
+static inline int is_myself(sip_uri_t *_puri)
```
henningw: More a note - it might make sense to move this to the core, or to expose it with the RR module API, as its duplicated in some modules already. Could be also done later if necessary.

src/modules/outbound/outbound_mod.c
```diff
@@ -503,3 +518,137 @@ int bind_ob(struct ob_binds *pxb)
...
+
+	return ret;
+}
+
+/*!
+ * \brief Parse the message and find first occurrence of Route header field.
+ * \param _m SIP message
+ * \return -1 or -2 on a parser error, 0 if there is a Route HF and 1 if there is no Route HF
+ */
+static inline int find_first_route(struct sip_msg *_m)
```
henningw: More a note - it might make sense to move this to the core, or to expose it with the RR module API, as its duplicated in some modules already. Could be also done later if necessary.

src/modules/outbound/outbound_mod.c
```diff
@@ -503,3 +518,137 @@ int bind_ob(struct ob_binds *pxb)
...
+
+	if(ret == -2) {
+		LM_DBG("no flow token found\n");
+		return CHECK_FLOW_ERROR_NO_FLOW_TOKEN;
+	} else if(ret == -1) {
+		LM_DBG("failed to decode flow token\n");
+		return CHECK_FLOW_ERROR_DECODE;
+	} else if(rcv->proto == PROTO_TCP || rcv->proto == PROTO_TLS
+			  || rcv->proto == PROTO_WS || rcv->proto == PROTO_WSS) {
+		tcp_connection_t *con =
```
henningw: Please move variable initialization to the beginning of the function.

src/modules/outbound/doc/outbound_admin.xml
sergey-safarov: I use `Oxygen XML Editor` for XML validation and see an error
> E [Xerces] Document root element "chapter", must match DOCTYPE root "book".

Please check the example
https://tdg.docbook.org/tdg/5.0/ch02
Here you can find.
image
When used in this form, the XML validator does not claim.

src/modules/outbound/doc/outbound.xml
sergey-safarov: XML validator reporting error
> E [Xerces] Attribute "xmlns:xi" must be declared for element type "book".

I think this is well documented at
https://www.oxygenxml.com/doc/versions/27.1/ug-author/topics/including-document-parts-with-XInclude.html

src/modules/outbound/outbound_mod.c
sergey-safarov: I'm fine with using the core function.

src/modules/outbound/outbound_mod.c
sergey-safarov: I'm fine with using the core function.
Could core developers create these core functions and then we will use in this PR.

src/modules/outbound/outbound_mod.c
sergey-safarov: done

src/modules/outbound/outbound_mod.c
```diff
@@ -503,3 +518,137 @@ int bind_ob(struct ob_binds *pxb)
...
+	str uri;
+	int ret;
+	struct receive_info *rcv = NULL;
+	tcp_connection_t *con = NULL;
+
+	switch(find_first_route(msg)) {
+		case -1:
+			LM_DBG("there is no Route HF\n");
+			return CHECK_FLOW_NO_ROUTE_HEADER;
+			break;
```
ivanuschak: please remove this line, this `break` is not needed as well as the `break` on 601 line

src/modules/outbound/outbound_mod.c
```diff
@@ -503,3 +518,137 @@ int bind_ob(struct ob_binds *pxb)
...
+{
+	struct hdr_field *hdr;
+	struct sip_uri puri;
+	rr_t *rt;
+	str uri;
+	int ret;
+	struct receive_info *rcv = NULL;
+	tcp_connection_t *con = NULL;
+
+	switch(find_first_route(msg)) {
```
ivanuschak: please check `find_first_route` function implementation. It returns 4 values: -2, -1, 0, 1.
The value `1` is returned when no route header found.
Service crashes when execution goes over this function and `sip_msg` does not contain Route header. Need to add `1` in the switch.

henningw: Good catch

src/modules/outbound/outbound_mod.c
sergey-safarov: done

src/modules/outbound/outbound_mod.c
sergey-safarov: done

miconda:
I added the functions (with different names) to core with the commits 521c974a200569b39f48b397853f50299d06ae7e and d11a33dc80431b69a3e7f29e1b00071ca24b5f04 . I will update rr module to use them.

src/modules/outbound/outbound_mod.c
sergey-safarov: done

src/modules/outbound/outbound_mod.c
sergey-safarov: done

miconda:
The description of the commit lists return 0, which should not be done by a module config function because it stops config execution -- this as a note, because it seems the code is returning negative and positive int values. But I would suggest to add explicit values to the enum fields, not only the -7 and 1, it should be safer in the future and easier to understand when looking at the code, avoiding shifting values when a new field might be added:

```
typedef enum check_flow_retcode
{
	CHECK_FLOW_ERROR_HEADERS_PARSE = -7,
	CHECK_FLOW_ERROR_ROUTE_PARSE,
	CHECK_FLOW_EXPIRED,
	CHECK_FLOW_NO_TCP_CONNECTION,
	CHECK_FLOW_ERROR_DECODE,
	CHECK_FLOW_ERROR_NO_FLOW_TOKEN,
	CHECK_FLOW_ERROR_URI_NOT_MYSELF,
	CHECK_FLOW_NO_ROUTE_HEADER = 1,
	CHECK_FLOW_SUCCESS
} check_flow_retcode_t;
```

Then, in my opinion CHECK_FLOW_NO_ROUTE_HEADER belongs better to the negative/failure range (false in native scripting), and have only CHECK_FLOW_SUCCESS as positive/success (true).

sergey-safarov:
@miconda I have updated the ENUM definition as proposed in the previous message, and updated description the XML documentation.
Also fixed the commit description.

sergey-safarov:
Maybe the `check_flowtoken` function name is more preferred than `check_flow_token`?

miconda:
Merging it, thanks!

PR #4450, opened 2025-10-27
pv: Always convert int to string in {s.int} transformation

### LINKED ISSUE #4444: PV's inconsistent in numeric context
### Description

Trying to access a PV in numeric context is highly inconsistent in case the PV contains a string as well as a number. Examples of such PV's are $dP, $rP, $nh(P). These set the protocol as string and as number.

#### Reproduction

The following script fragment logs the various interactions with such a PV.

```
xinfo("Target proto 1: $rP");    # "tls"

xinfo("Target proto 2: $(rP{s.int})");    # "tls"   WRONG!

$vn(rP) = $(rP{s.int});
xinfo("Target proto 3: $vn(rP)");    # 3

$vn(rP) = $rP;
xinfo("Target proto 4: $vn(rP)");    # "tls"

$vn(rP) = $rP + 0;
xinfo("Target proto 5: $vn(rP)");    # "tls0"  WRONG!

$vn(rP) = 0 + $rP;
xinfo("Target proto 6: $vn(rP)");    # 3

$vn(rP) = $rP + "";
xinfo("Target proto 7: $vn(rP)");    # "tls"

$vn(rP) = "" + $rP;
xinfo("Target proto 8: $vn(rP)");    # "tls"

$vn(isTLS) = $rP == "tls";
xinfo("Target is TLS 9: $vn(isTLS)");    # 1

$vn(isTLS) = $rP == 3;
xinfo("Target is TLS 10: $vn(isTLS)");    # 0   WRONG!

$vn(isTLS) = $(rP{s.int}) == 3;
xinfo("Target is TLS 11: $vn(isTLS)");    # 1

$vn(isTLS) = 3 == $rP;
xinfo("Target is TLS 12: $vn(isTLS)");    # 1

```

Resulting in the following log entries:
```
INFO: Target proto 1: tls
INFO: Target proto 2: tls
INFO: Target proto 3: 3
INFO: Target proto 4: tls
INFO: Target proto 5: tls0
INFO: Target proto 6: 3
INFO: Target proto 7: tls
INFO: Target proto 8: tls
INFO: Target is TLS 9: 1
INFO: Target is TLS 10: 0
INFO: Target is TLS 11: 1
INFO: Target is TLS 12: 1
```

Scenario's 2, 3, 5, 6, 10, 11, 12 all access the PV in numeric context, yet only 3, 6, 11 and 12 yield the correct result.

Especially 2 is worrying as the clear intention of the author is to fetch a numerical value, not a string.

For 4 and 10 I can understand that the type of the first variable item used, but the script parser should throw an error as the concatenated value is not a string. The script parser should not convert a number to a string automatically.

And then there is also the fact that the whole numeric part gets lost on assignment to another PV:
```
$vn(rP) = $rP;
xinfo("Target proto 13: $(vn(rP){s.int})");

$vn(rP) = $(vn(rP){s.int});
xinfo("Target proto 14: $vn(rP)");
```

Log:
```
INFO: Target proto 13: 
ERROR: <core> [core/lvalue.c:346]: lval_pvar_assign(): non existing right pvar
ERROR: <core> [core/lvalue.c:404]: lval_assign(): assignment failed at pos: (734,13-734,28)
INFO: Target proto 14: <null>
```

gaaf:
PV's may hold a string as well as an integer, and the s.int transformation always assumes that an existing string value always represents the integer. This is not the case for some PV's such as $rP where the string value is not equal to the integer (ie, string "tls" and integer 3).

By ignoring the existing string value and always creating a new string from the integer value, it is guaranteed that s.int always returns an numerical value.

Fixes #4444

N.B. I first tried removing the `PV_VAL_STR` flag and removing the int2str altogether so conversion would be "on demand", but other code seem to contain a bug and blindly use the string value even if `PV_VAL_STR` is not set.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4444

### DIFF
--- a/src/modules/pv/pv_trans.c
+++ b/src/modules/pv/pv_trans.c
@@ -223,8 +223,8 @@ int tr_eval_string(
 				if(str2slong(&val->rs, &val->ri) != 0)
 					return -1;
 			} else {
-				if(!(val->flags & PV_VAL_STR))
-					val->rs.s = int2str(val->ri, &val->rs.len);
+				// Replace string as an existing string value may not be equal to the int
+				val->rs.s = int2str(val->ri, &val->rs.len);
 			}
 
 			val->flags = PV_TYPE_INT | PV_VAL_INT | PV_VAL_STR;


miconda:
It doesn't feel like the right approach here, overwriting the existing string value, it can break existing config conditions.

Maybe it would be good to add new PV with has only the numeric value, also as string, like $rPn (received-protocol-numeric) and eventually $rPs (received-protocol-string), with no numeric value... I think there are no many proto-related variables where this case happens now and it would be cleaner after all.

gaaf:
> It doesn't feel like the right approach here, overwriting the existing string value, it can break existing config conditions.

Unless I oversee something, those existing configs must be very wrong then. Do you mean the case where someone is comparing something that is deliberately converted to an int with a string? Like `$(rP{s.int}) == "tls"`? I don't think that's very likely.

And, is there really an overwriting of the value? AFAIK, transformations do not touch the input, they produce _new_ output. Otherwise, doesn't `{s.md5}` also overwrite the "existing string value"?
```
compute_md5(_tr_buffer, val->rs.s, val->rs.len);
val->rs.s = _tr_buffer;
```
If needed I can rewrite the in2str conversion to use a _tr_buffer (which should probably be done anyway to prevent overwriting in a cascade/nesting of transformations).

 
> Maybe it would be good to add new PV with has only the numeric value

Well, that doesn't fix the bug with the `{s.int}` transformation.

NB. It also adds to the already overwhelming amount of PV's with somewhat duplicated/similar contents. I'd rather remove all those separate string and int PV's and replace them with the `strintval` variant.

miconda:
What I meant by being wrong is to replace the string value when one was set for it already, in other words, if there is a string value, that should be kept/propagated, don't replace it with the string representation of variable's int value.

The {s.int} transformation is to return the integer conversion of a variable that holds only a string value. If the variable has an integer value, that is returned. What bug you are referring here to be for {s.int}?

Can you list some of "the already overwhelming amount of PVs with somewhat duplicated/similar contents"?

We are open for improvements of course, but to be clearly aware of what is impacted and how, what are the duplicates, ...

gaaf:
> What I meant by being wrong is to replace the string value when one was set for it already, in other words, if there is a string value, that should be kept/propagated, don't replace it with the string representation of variable's int value.

You've not disputed my analysis, so I still stand by it: It's not replaced in the original value. Only the output of the transformation returns the updated string value.

 
> The {s.int} transformation is to return the integer conversion of a variable that holds only a string value. If the variable has an integer value, that is returned.

Not when the result is used in a string context, like in xlog().


> What bug you are referring here to be for {s.int}?

That `{s.int}` on a PV and used in a string context does not represent an integer.

`xlog("$(rP{s.int})");` prints "tls", not "3". The output of `{s.int}` should never contain letters, only digits.

See the referenced issue #4444.

miconda:
It is replacing the original string value that is in the expression: it was never about a variable value what I mean, but that a string value was set previously for that expression.

The `{s.int}` is supposed to be used for arithmetic operations. For protocol variable $rP the equivalent of protocol ID 3 to string translates to `tls`, it is what I want to have printed in such case. The $pR type is string, the integer value for it is adjacent/complementary to get the corresponding protocol id.

Many things may not make "clear" sense if one wants to analyze the behaviour of the config interpreter, including the evaluation of negative to false and positive to true, exit for 0. Similar is with the config variables, that's why they are called also "pseudo-variables", they have specific behaviour per implementation, what the developer needed for them.

You also didn't list any of the "overwhelming amount of PVs ...".

miconda:
For a similar case, there is $pr and $prid, where the second gives only the numeric value, while the first gives the string with complementary protocol id.

miconda:
I added `{s.intv}` which returns the printed int value in string context, even an associated string exists:

- https://www.kamailio.org/wikidocs/cookbooks/devel/transformations/#sintv

PR #4456, opened 2025-10-30
Adjust Docbook tags in many modules

mslehto:
Some module documentation entries have capitalised "Kamailio.cfg".
This patch uses `&kamailioconfig;` and few other minor Docbook related updates.

#### Type Of Change

### DIFF
--- a/doc/docbook/entities.xml
+++ b/doc/docbook/entities.xml
@@ -93,7 +93,6 @@
 <!ENTITY xmpp "<acronym>XMPP</acronym>">
 <!ENTITY gnu "<ulink url='https://www.gnu.org/'>GNU</ulink>">
 <!ENTITY gpl "<ulink url='https://www.gnu.org/licenses/licenses.html#GPL'>GPL</ulink>">
-<!ENTITY openssl "<ulink url='https://www.openssl.org'>OpenSSL</ulink>">
 
 <!-- Often used RFCs -->
 


henningw:
Thanks for the PR - minor remark, it seems that the OpenSSL entity is two times present, if I read it correctly? Probably make sense to also cleanup this.

mslehto:
Sure, good catch, updated!

src/modules/tls/doc/certs_howto.xml
```diff
@@ -109,9 +109,9 @@ Setting &kamailio; to use the TLS certificate
 	intended machine (preferably in your &kamailio; configuration directory,
 	 this is the default place &kamailio; searches for).
 
-3. Set up &kamailio;.cfg to use the certificate
+3. Set up &kamailio; to use the certificate
```
oej: I think this is wrong, see line 146 below. Should possibly be kamailioconfig

mslehto: Tack, many styles to choose from, good to be consistent. Pushing updated branch next.

mslehto:
New commit address oej note, thanks.
Another new commit fixes more of the same (TLS certs_howto.xml).

We have DocBook entity &kamailioconfdir; but I decided to not use it as I think its content is less common seen in hard coded documentation files.

I will squash-push before merge to main when acceptable.

src/modules/tls/doc/certs_howto.xml
mslehto: Updated (f0f849443667b069ff63d189dcc263498f09edba)

miconda:
When PR is ready, it has to be merged manually, some commit messages don't follow the guidelines, like `tls: Fix and add DocBook tags MR4456 WIP2 msl` (reference to PR in first line and I guess some user id in there), ideally with a commit per component.

mslehto:
I think manual merge is not needed anymore, patches are now splitted per component and commit messages cleaned.

henningw:
Thank you, merged

PR #4458, opened 2025-11-04
db_redis: fix coverity scan defect

lbalaceanu:
* fixes "Null pointer dereferences"

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

### DIFF
--- a/src/modules/db_redis/redis_sentinels.c
+++ b/src/modules/db_redis/redis_sentinels.c
@@ -150,12 +150,14 @@ int replica_list_free(struct reply_list *list)
 
 static redisReply *replica_list_pop(struct reply_list *list, size_t idx)
 {
-	struct reply_node *current = list->head;
+	struct reply_node *current = NULL;
 	struct reply_node *prev = NULL;
 	size_t current_idx = 0;
 	if(!list || idx >= list->count) {
 		return NULL;
 	}
+
+	current = list->head;
 
 	//traverse the list to the idx-th node and remove it from the list also return it
 	while(current && current_idx < idx) {


miconda:
The commit message first line should indicate what was fixed, not the tool used to detect issues. Like:

```
db_redis: fix null pointer dereferences
```

If you want to make a reference to the tool in the body of the message, is fine, but not necessary. Many warnings/errors are reported by compilers/tools, we do not need to know how discovery was done, but what has been fixed/added to the code. There is a note about it at:

  - https://github.com/kamailio/kamailio/blob/master/.github/CONTRIBUTING.md#commit-message-content

lbalaceanu:
Thank you @miconda,

I will change the commit message.

Can you maybe also suggest something for this:
```
 CID 638428:         Security best practices violations  (DC.WEAK_CRYPTO)
 "rand" should not be used for security-related applications, because linear congruential algorithms are too easy to break.
486     		idx = rand() % replica_list.count;
```


What should I use instead of rand()?

miconda:
You can ignore that one, some are more recommendations, just to be aware if it is some crypto/security critical case.

henningw:
> ```
>  CID 638428:         Security best practices violations  (DC.WEAK_CRYPTO)
>  "rand" should not be used for security-related applications, because linear congruential algorithms are too easy to break.
> 486     		idx = rand() % replica_list.count;
> ```
> 
> What should I use instead of rand()?

If its not security related, its fine to use rand() as Daniel mentioned. If you want to get a strong pseudo random number, which is not used for cryptographic purposes, you can use fastrand() from core. For cryptographic purposes there is also cryptorand().

lbalaceanu:
Then I will let the rand() and just rebase and merge this PR. 
Thanks

Issue #4426, opened 2025-10-08
KEMI: Add htable count/match API equivalent to $shtcn() (to avoid pv_cache growth with dynamic PV strings)


wedhornsipgate:
**Description:**

We're missing a KEMI function in the htable module that provides the same functionality as the PV function `$shtcn(table=>pattern)` to count the number of keys in an htable that match a given pattern (prefix/suffix/regex). The goal is to let KEMI scripts perform pattern counts without constructing dynamic PV expressions, which currently causes unbounded pv_cache growth in PKG memory.

**Use case / motivation:**

We implemented rolling-window rate limiting by writing short-lived, per-call entries into an htable and then counting keys that match a number-derived pattern. Using $shtcn() works well for the logic, but in KEMI (JS/TypeScript) we must call it via pv.gete("$shtcn(...)"). Because the PV string varies per INVITE (different pattern/key each time), these dynamic PV expressions accumulate in the PV cache, growing PKG/private memory over time.

We’d like a native KEMI function to perform the same count/match so we can avoid the PV layer entirely.

**Environment**

Kamailio: 5.8.6 on Debian 12

Scripting: KEMI (JavaScript/TypeScript)

Symptom: steady PKG/private memory increase over ~24h under load when using pv.gete("$shtcn(...)"). Disabling the $shtcn() call stops the growth.

Related log line: the one discussed in [#3440](https://github.com/kamailio/kamailio/issues/3440) (“pv cache … pkg memory may get filled with pv declarations”).

Current line in our script:

```if (ksr.pv.gete(`$shtcn(${htable_count}=>%~${number})`) > threshold) { ... }```

**Current workaround**

Restarting the Kamailio every day or disabling rate limit on instances with KEMI.

**Idea for API**

Minimal, PV-parity:

```
// KEMI export
int shtcn_match(str table, str pattern, int mode /* 0=exact, 1=prefix, 2=suffix, 3=regex */);
```

JS example:

`const n = KSR.htable.shtcn_match("from_number_count", number, 3); // regex`

More explicit helpers (avoids integer modes):

```
KSR.htable.shtcn_exact(table, key)      -> int
KSR.htable.shtcn_prefix(table, prefix)  -> int
KSR.htable.shtcn_suffix(table, suffix)  -> int
KSR.htable.shtcn_regex(table, pattern)  -> int
```

**Behavioral notes / expectations**

[...]

PR #4492, opened 2025-11-17
Add more htable KEMI functions to htable module

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [ ] Related to issue #4426

#### Description

This PR add the functionaility discussed in the referenced issue for KEMI!

- PV shtcn equivalent
- match with mode
- seperate specific functions for exact, prefix, suffix, regex

### DIFF
--- a/src/modules/htable/htable.c
+++ b/src/modules/htable/htable.c
@@ -1489,6 +1489,51 @@ static int ki_ht_dec(sip_msg_t *msg, str *htname, str *itname)
 	return ki_ht_add_op(msg, htname, itname, -1);
 }
 
+static int ki_sht_cn(sip_msg_t *msg, str *table, str *op, str *pattern)
+{
+	ht_t *ht;
+	char *pat_buf;
+	str pat;
+	int cnt;
+
+	ht = ht_get_table(table);
+	if(!ht) {
+		LM_ERR("cannot get hash table [%.*s]\n", table->len, table->s);
+		return -1;
+	}
+
+	if(op->len != 0 && op->len != 2) {
+		LM_ERR("invalid operator [%.*s]\n", op->len, op->s);
+		return -1;
+	}
+
+	/* Build pattern with operator prefix */
+	pat_buf = pkg_malloc(pattern->len + 3);
+	if(!pat_buf) {
+		LM_ERR("no pkg memory\n");
+		return -1;
+	}
+
+	pat.s = pat_buf;
+	pat.len = pattern->len + 2;
+	pat.s[pat.len] = '\0';
+
+	/* Prepend operator prefix to pattern */
+	if(op->len == 0) {
+		pat.s[0] = '~';
+		pat.s[1] = '~';
+	} else {
+		pat.s[0] = op->s[0];
+		pat.s[1] = op->s[1];
+	}
+
+	memcpy(pat.s + 2, pattern->s, pattern->len);
+	cnt = ht_count_cells_re(&pat, ht, 0);
+
+	pkg_free(pat_buf);
+	return cnt;
+}
+
 #define RPC_DATE_BUF_LEN 21
 
 /* clang-format off */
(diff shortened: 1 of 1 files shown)


xkaraman:
@miconda While this is a new feature it only involves KEMI new functions, would this made the changes eligible for backport to 6.0 or even 5.8?

miconda:
Usually it was about exporting existing kamailio.cfg functions to the KEMI framework that can be backported. Here seems to be about adding a function that offers equivalent of some cfg variables, which can be already used in KEMI via `KSR.pv.get(...)`. If other developers feel ok about backporting, I am neutral.

Regarding the PR itself, I think it would be better to have a single KEMI function with the operator as parameter having same options like in the cfg variables. I see the function is with a integer parameter which is then converted inside the code to the strings. Something like:

```
ki_sht_cn(sip_msg_t *msg, str *table, str *op, str *pattern)
```

Then have something like `KSR.htable.sht_cn("x", "re", "^a.+")`

xkaraman:
Hey @miconda,

Thanks for the review. 

Got some question to clarify some things:

> KEMI function with the operator as parameter having same options like in the cfg variables
Is there somewhere documented what would these look like? The only docs i could find is this https://www.kamailio.org/wikidocs/cookbooks/6.0.x/pseudovariables/#shtcvhtableexp. But this do not include any operation paramater. The operator is included in the string given to search for like `shtcn(htable=>~~regexp`. 

So maybe the new function just accept only pattern as parameter that already is described here? with no mode or operator paramater at all?

I prefer the operator to be explicit to be honest just like you suggested but it is not how cfg, pv variables are used. Correct me if i am wrong here.

> I think it would be better to have a single KEMI function

So no implementation for the specific operation at all? They are all using the common function with selected operation as a convenience. I don't have a strong preference here, if one can use the base for everything.

miconda:
It does not have to be like in cfg, because there is a variable and in KEMI is a function, it is not really 1-to-1 equivalent. So having explicit operator parameter is fine. The cfg variables have some constrains for their names, which don't have to be propagated to functions.

xkaraman:
OK understood.

We probably want to all options found available in here https://www.kamailio.org/wikidocs/cookbooks/6.0.x/pseudovariables/#shtcvhtableexp supported and not the ones i included already (regex, left prefix, right prefix, exact), like `==` for int comparison and `**` for all?

xkaraman:
Hey @miconda,

Adopted with the operator as parameter with available values as `regex` , `exact`, `prefix` `suffix` `match_all`.

Convenience functions can be removed of course as discussed.

henningw:
Any more comments or remarks, or can it be merged?

src/modules/htable/htable.c
```diff
@@ -1489,6 +1489,83 @@ static int ki_ht_dec(sip_msg_t *msg, str *htname, str *itname)
 	return ki_ht_add_op(msg, htname, itname, -1);
 }
 
+static int ki_sht_cn(sip_msg_t *msg, str *table, str *op, str *pattern)
```
miconda, line 1492: Wouldn't be enough to export only this function as `KSR.htable.sht_cn(...)`?

The others are just wrappers, the native cfg does not have a PV per each type of match.

src/modules/htable/htable.c
```diff
@@ -1489,6 +1489,83 @@ static int ki_ht_dec(sip_msg_t *msg, str *htname, str *itname)
...
+	}
+
+	pat.s = pat_buf;
+	pat.len = pattern->len + 2;
+	pat.s[pat.len] = '\0';
+
+	/* Prepend operator prefix to pattern */
+	if(op->len == 5 && strncmp(op->s, "regex", 5) == 0) {
+		pat.s[0] = '~';
+		pat.s[1] = '~';
```
miconda, line 1524: Why is not the `op` value directly what is expected internally (and documented for `$shtcn()`), like `~~`, `==` ... It only does an extra translation in this style.

src/modules/htable/htable.c
xkaraman: Yep, removed wrappers

src/modules/htable/htable.c
xkaraman: Got it. Modified so that it fowrards whatever the user provided. Add a small input check for length either to be 0 (act like a regex) or 2 for the rest of the operators, and fail if something else was given.

miconda:
Thanks, merging this one, then doing a small update to the name of the kemi function export name.

PR #4493, opened 2025-11-17
rtpengine: add connect() command

vsaranov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
Support for the `connect` NG method has been added. This method is available in rtpengine starting from release mr13.2.1.1. 
https://rtpengine.readthedocs.io/en/latest/ng_control_protocol.html#connect-message
Method can be used to directly connect the media flow of two call parties without the need for a full offer/answer exchange. Call parties from two different calls (different call IDs) can also be connected.


henningw:
Thanks for the PR, looks straightforward. I think it would be benefitial to add the minimum version requirement for this new command to the docs, maybe to chapter 2 "recommended versions".

rfuchs:
LGTM.

Perhaps even a generic method that is able to issue any command to rtpengine would be useful, to have future-proof support for new features.

vsaranov:
> I think it would be benefitial to add the minimum version requirement for this new command to the docs, maybe to chapter 2 "recommended versions".

It’s possible, but I think the information about the required rtpengine version will soon become outdated and will need to be removed. I believe that if a user uses the NG "connect" command, it means they have already read the rtpengine documentation and know which version supports this command.

> Perhaps even a generic method that is able to issue any command to rtpengine would be useful, to have future-proof support for new features.

It is possible that all existing functions of the rtpengine module can be removed, or rather consolidated into a single universal function. I suggest considering this in the future, but for now, apply the proposed changes.

henningw:
No further comments, merged.

PR #4500, opened 2025-11-26
dispatcher: selected hash more accessible in failover

devopsec:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
This PR moves the selected hash up the call stack so it is more accessible in failover, and by the higher order functions.  

It is a little hard to see the benefit until I get the next few patchsets submitted, but I will link them below this description.  
The biggest benefit for other dispatcher algorithm developers is that the hash selected is now available on failover and can be used to lookup attrs (or any other info) needed for implementing more complex failure scenarios.  

These are the examples I am working on that depend on this change (as an example of how it can be used):
1. [rpc commands][1] - an example where a user can test selection algorithms with a dataset, without stepping into gdb
2. [failover selection][2] - an example algorithm that makes use of this change for re-balancing the distribution on failover (in mode 1)

[1]: https://github.com/devopsec/kamailio/tree/dispatcher_rpc_updates
[2]: https://github.com/devopsec/kamailio/tree/dispatcher_alg_updates

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -2276,6 +2276,16 @@ int ds_add_xavp_record(
 		return -1;
 	}
 
+	/* add dstidx field */
+	memset(&nxval, 0, sizeof(sr_xval_t));
+	nxval.type = SR_XTYPE_LONG;
+	nxval.v.l = pos;
+	if(xavp_add_value(&ds_xavp_dst_dstidx, &nxval, &nxavp) == NULL) {
+		xavp_destroy_list(&nxavp);
+		LM_ERR("failed to add destination dstidx xavp field\n");
+		return -1;
+	}
+
 	if(((ds_xavp_dst_mode & DS_XAVP_DST_SKIP_ATTRS) == 0)
 			&& (dsidx->dlist[pos].attrs.body.len > 0)) {
 		memset(&nxval, 0, sizeof(sr_xval_t));
@@ -2356,7 +2366,7 @@ int ds_add_xavp_record(
  */
 int ds_select_dst(struct sip_msg *msg, int set, int alg, int mode)
 {
-	return ds_select_dst_limit(msg, set, alg, 0, mode);
+	return ds_select_dst_limit(msg, set, alg, 0, mode).ret;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


devopsec:
fixed formatting with clang-format

miconda:
I would prefer not to return structures from functions, they can be inefficient if their size is big, which tends to happen over the time. It is also not the common practice we have in the code so far.

If you need the value of the hash for failover, wouldn't it be better to keep it in the context xavp?

  - https://www.kamailio.org/docs/modules/stable/modules/dispatcher.html#dispatcher.p.xavp_ctx

devopsec:
Hmmm, ds_xavp_dst works accessing it on failover but the other use case (the one the ds_hres_t was created for) is accessing it on the first selected destination, even when failover is not turned on.  
Is there a better way to grab the original ds_dest_t selected in that case?

The example would be the 1st branch i linked above, showing the user the order of destinations when using a specific algorithm with their current data, via an rpc command.  
I did not see another simple way to pull that selected hash up the call stack.  
Here is an example (WIP) feature where I used the returned strcture: https://github.com/kamailio/kamailio/compare/master...devopsec:kamailio:dispatcher_rpc_updates#diff-016b0bd9b0a333124f034e636fb1b8a75d8200bbac06149fcfeb8737c98f3f96R2668

I'll highlight that use case at a higher level for more context.  
Lets say that I have a customer call in and say they were expected their traffic to be distributed in a different manner than currently is (lets say they have some indication they are not getting the level of service or features they expect).  
Lets say I have a NOC staffed with technicians that needs to review that customers routing configuration or need to check if changing their routing configuration (dispatcher data / algorithm) would/is giving them the intended distribution.  
They could test the proposed changes (or current route configuration) using the rpc command in that branch I linked to check where the requests will be distributed without needing to try replicating the live traffic that led to the customer complaint.  

The struct made that selected destination available, even when failover was not enabled or the algorithm did not use failover.  
I am of course willing to refactor to meet that requirement but I am unsure how I could accomplish that use case without returning the dlist[] index selected as well.

miconda:
I am going to merge this one, but then change to return again int from functions and have the res structure (ds_hres_t) as output parameter, to match the common coding practice we have so far.

miconda:
Check to see if the updated varian suits your needs and works as expected -- see the last commit referenced above. If there are issues then, open a bug in the tracker.

devopsec:
Understood, looks like that will still accomplish the same task, allowing hash selected to be used when calling ds_select_dst_limit() if needed.  
Thanks for looking through this @miconda

PR #4513, opened 2025-12-08
Grumvalski/db async refactor

grumvalski:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR https://github.com/kamailio/kamailio/pull/4027 introduced a new connection mode for sqlops.
Unfortunately this new connection mode doesn't work with asynchronous operations because:
- the connectiono will be created by the kamailio worker handling the sip message and when the async worker takes the relay it won't have any connection with the database so it'll create one
- after executing the query the connection's refcount will be decreased and, having reached '0' (it had just been created) it wil lbe destroyed, meaning that for each query a new connection will be open/closed.
Thsi pull request tries to adress this issue with a refactoring of the database backends async operation refactoring.


miconda:
Thanks, merging!

PR #4515, opened 2025-12-09
core: pthread shared mutexes (default for ARM64) and stdatomic for ARM64

n2p-litvinov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
For pthread mutex added proc share attribute support.
This locking method is enabled by default for ARM64.
stdatomic is add and used as default for ARM64.

### DIFF
--- a/src/core/atomic/atomic_native.h
+++ b/src/core/atomic/atomic_native.h
@@ -65,6 +65,10 @@
 
 #include "atomic_arm.h"
 
+#elif defined __CPU_aarch64
+
+#include "atomic_stdatomic.h"
+
 #elif defined __CPU_alpha
 
 #include "atomic_alpha.h"
--- a/src/core/atomic/atomic_stdatomic.h
+++ b/src/core/atomic/atomic_stdatomic.h
@@ -1,0 +1,359 @@
+/*
+ * Copyright (C) 2025 Viktor Litvinov
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+/**
+ * @file
+ * @brief Atomic operations and memory barriers using C11 stdatomic.h
+ *
+ * Atomic operations and memory barriers implemented using C11 standard
+ * atomic library (stdatomic.h). This provides portable atomic operations
+ * for platforms supporting C11 or newer.
+ *
+ * \warning atomic ops do not include memory barriers unless using mb_* variants
+ * @ingroup atomic
+ */
+
+#ifndef _atomic_stdatomic_h
+#define _atomic_stdatomic_h
+
+#include <stdatomic.h>
+
+#ifndef HAVE_ASM_INLINE_MEMBAR
+
+#ifdef NOSMP
+#define membar() \
+	do {         \
+	} while(0)
+#else /* SMP */
+
+/* memory barriers using C11 atomic thread fence */
+#define membar() atomic_thread_fence(memory_order_seq_cst)
+#define membar_write() atomic_thread_fence(memory_order_release)
+#define membar_read() atomic_thread_fence(memory_order_acquire)
+
+#ifndef __CPU_alpha
+#define membar_depends() \
+	do {                 \
+	} while(0) /* empty on non-Alpha */
+#else
+#define membar_depends() membar_read()
+#endif
+
+#endif /* NOSMP */
+
+#define membar_enter_lock() atomic_thread_fence(memory_order_acquire)
+#define membar_leave_lock() atomic_thread_fence(memory_order_release)
+
+/* membars after or before atomic_ops */
+#define membar_atomic_op() membar()
+#define membar_atomic_setget() membar()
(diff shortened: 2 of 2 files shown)


miconda:
Thanks, looks good for me, let's wait a few days to see if anyone else has comments.

More context: the coding/PR has been made by @n2p-litvinov based on discussions during Kamailio Devel Meeting 2025, being decided to use stdatomic by default for ARM64, not having a native/asm variant, then later maybe switch the other archs to it. Also, the fast-locks (busy-waiting) are performing poorly on ARM, pthread process-shared mutexes should be more optimal. Maybe we can review the use of them for other archs as well, probably their implementation is a lot more optimized now than last time when they were evaluated, many years ago.

henningw:
Thanks for the PR. I think its a good addition and we should add it as default for ARM64. As said, we should then also probably make it the default in the next release cycle (6.2.x) for the other architectures after some benchmarks have been done.

Can you rebase the PR to solve the conflict, which probably got introduced from some other changes in the make/cmake.

xkaraman:
Hey @n2p-litvinov,

I have some comments on this after some testing with it.

TLDR:

So, 
1. for Make, `CC_GCC_LIKE_ASM` def for `aarch64` should be provided for the header to be included at all.
2. def `HAVE_ASM_INLINE_ATOMIC_OPS` and `HAVE_ASM_INLINE_MEMBAR` so atomic_unknown is not included if not nessecary

------
As it currently stands, i don't think the `atomic_stdatomic.h` is included anywhere as of now (at least on linux aarch64 sever i am testing with).

There is this definition `CC_GCC_LIKE_ASM` missing in the current Makefiles for aarch64, so that `atomic_native.h` include anything at all. Therefore the build might succeed in the current form but it's not using `atomic_stdatomic` but `atomic_unknown.h`.

Even after i modify the makefiles to define the nessecary defintion then the build fails with:

```
core/atomic/atomic_unknown.h:260:19: error: redefinition of ‘mb_atomic_get_int’
  260 | inline static int mb_atomic_get_int(volatile int *v)
      |                   ^~~~~~~~~~~~~~~~~
core/parser/../mem/../atomic/atomic_stdatomic.h:263:19: note: previous definition of ‘mb_atomic_get_int’ with type ‘int(volatile int *)’
  263 | inline static int mb_atomic_get_int(volatile int *v)
      |                   ^~~~~~~~~~~~~~~~~
core/atomic/atomic_unknown.h:273:20: error: redefinition of ‘mb_atomic_get_long’
  273 | inline static long mb_atomic_get_long(volatile long *v)
      |                    ^~~~~~~~~~~~~~~~~~
core/parser/../mem/../atomic/atomic_stdatomic.h:332:20: note: previous definition of ‘mb_atomic_get_long’ with type ‘long int(volatile long int *)’
  332 | inline static long mb_atomic_get_long(volatile long *v)
      |                    ^~~~~~~~~~~~~~~~~~
```

because both stdatomic and unknown are included. this is due to 
 
```
/*! \brief if no native operations, emulate them using locks */
#if !defined HAVE_ASM_INLINE_ATOMIC_OPS || !defined HAVE_ASM_INLINE_MEMBAR
#warning "no native atomic ops, atomic_unknown.h will be used"
#include "atomic/atomic_unknown.h"
 
#endif /* if HAVE_ASM_INLINE_ATOMIC_OPS */
 ```
which requires `HAVE_ASM_INLINE_ATOMIC_OPS`  and `HAVE_ASM_INLINE_MEMBAR` to be defined or otherwise it will include the unknown variant.

These two definition are each defined in their respective `atomic_[arch].h` if applicable, so i guess `atomic_stdatomic` should too.
 
If you try to compile with cmake you get these errors from the beginning because we already provide the `CC_GCC_LIKE_ASM` for aarch64. 

At first try, i tried using `FUTEX` lock method which heavily used atomic, but these errors and some other probably related to this, compilation did not complete.

Then i tried compiling with `PTHREAD_MUTEX` lock method, with the same errors as makefile, redefinition of multiple functions.

miconda:
I think the conflict happens with cmake files after merging the PR related to selection/priority of the locking type. Since the problems are not in the code, I am going to push the commits individually, so it can be worked on directly on the public Kamailio repo, rather than PR from an external repo.

xkaraman:
Yep the conflict is a minor due to missing block moved to `lock_methods.cmake` and can be removed completely. 

There is no need to `set(USE_FAST_LOCK NO)`, this is only for assessing availability for said platform. Actual method is selected based on the priority and the available methods, or if the user sets the `-DLOCK_METHOD=` (and of course is available hence it should stay as such) when configuring the build.

miconda:
I pushed two of the Viktor's patches manually ([7600dc5](https://github.com/kamailio/kamailio/pull/4515/commits/7600dc5a4495722afff561b9bfff00f73f1af73f) and [abc6537](https://github.com/kamailio/kamailio/pull/4515/commits/abc653738622986dc8344f2034fd2f4c89ea3cdc)).

Two new patches were pushed by me to get the stdatomic enabled for aarch64, in the header file and Makefile.defs. And it seems it got compiled with stdatomic, but I haven't done further tests yet.

@xkaraman: can you also add to changes needed for the CMake files?

n2p-litvinov:
Hello guys. Sorry for delay with replay.

Let me shortly force push rebased and fixed branch.

Also want to mention that one strange memory leak in tm module was found during testing of this PR. Can not say is it stdatomic or pthread related. I'll back with investigation results as soon as will have them.

xkaraman:
@miconda thanks for the commits! I can confirm that could use `CMake` to build kamailio with both `futex` and `pthread` lock methods using the `stdatomic.h` header.

There is no changes in `CMake` as i was already providing `CC_GCC_xxx` defintion for aarch64. 

The only change i can provide currently is allowing `aarch64` to use FUTEX lock method, as the build was with a local change.

n2p-litvinov:
Thanks guys, I see @miconda you are merged everything already (and fixed flags definitions).
Should we change default locking from fast_lock to pthread for aarch64?

miconda:
@n2p-litvinov: your commits, besides the one for c/make files were merged manually, which is no longer needed. This PR will be closed.

@xkaraman: to change so FUTEX can be also used for aarch64 if wanted at compile time, or to be the default?

miconda:
@n2p-litvinov: I am fine to switch to pthread mutex as default for aarch64.

xkaraman:
@miconda With the change i had in mind, is to just enable futex building for aarch64. but by doing so it will pick it up as a default due to the priority list. 

I ll see if i can make an exception for aarch64 to pick up pthread instead.

xkaraman:
@miconda 

I pushed 18119539680183d5815df84c67d64a94357afbfd for using different priority list for each arch, or back up to a default priority. 

Right now only aarch64 is implementing it, prioritizing pthread_mutex then fast_lock then futex, but of course will respect the user choice if given.

I think this issue can be closed then.

miconda:
@xkaraman: thanks, closing it!

Anything new problem that comes on this topic should be tracked via a new PR or issue.

PR #4518, opened 2025-12-10
rtpengine: changed the hash_algo parameter to be by default 2 (CRC32) 

AlexisHadj:
rtpengine: Changed the hash_algo parameter to be by default 2 (CRC32)  instead of 0 (legacy algorithm)

### DIFF
--- a/src/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/src/modules/rtpengine/doc/rtpengine_admin.xml
@@ -2343,7 +2343,7 @@ modparam("rtpengine", "control_cmd_tos", 144)
 			CRC32 - 2 (calculate crc32 sum over the callid).
 		</para>
 		<para>
-			Default value is 0, legacy algorithm.
+			Default value is 2, CRC32.
 		</para>
 		<para>
 			The values not falling into the range <quote>0-2</quote> are ignored.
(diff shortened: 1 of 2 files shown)


miconda:
Not sure if should be merged, others can comment, there is a modparam to set the value as desired, but if majority thinks changing default value should be done, I would be fine.

My comment is actually to say that you added spaces to a line in the code and that the commit message is not formatted according to the contributing guidelines (you used that for PR title, not for the commit first line). You have to fix them and do a force push (or close this one and make a new PR).

henningw:
About the change of the default algorithm, there were multiple reports in the past that the legacy algorithm is not working well, and provide an uneven distribution, e.g. [here](https://kamailio.org/mailman3/hyperkitty/list/sr-users@lists.kamailio.org/message/LUJJMNMKXNZ2COB6Q6KXUBM72SXLKOFE/). From a performance point of view the CRC32 should be almost identical to the legacy algorithm. I think the default should be an algorithm that work in common scenarios with multiple rtpengine servers.

rfuchs:
FWIW I would agree that the default for new users should not be the legacy hash and CRC32 is probably a good choice. The legacy hash could even be retired at some point.

henningw:
Thank you for the comments, merged. Regarding the proposed retirement of the legacy hash, we will create another PR. It can be also merged after the 6.1 branch was created, there is no need to rush it.

PR #4536, opened 2025-12-23
dmq: Fix DNS resolution corruption when search list is used

xadhoom:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

When `dns_use_search_list=1` (the default), `get_record()` may prepend a "fake CNAME" record to map short hostnames to their expanded form. The DMQ module's `get_dmq_host_list()` function was iterating through all returned records and casting `rdata` to `struct a_rdata *` without checking the record type. This caused CNAME record data (hostname bytes) to be interpreted as IP address bytes, resulting in bogus IP addresses.

For example, a hostname containing `.ord` would produce the IP `43.111.114.100` (the ASCII bytes `+ord` = 0x2B 0x6F 0x72 0x64).

This has been tested and fixed with Kamailio 6.0.x, backported also to master.

#### Fix

Add a type check to skip non-A records in the iteration loop:

```c
if(prec->type != T_A) {
    continue;
}
```

#### Affected Environments
This bug affects Kamailio deployments using the following setup:
* DMQ module with dmq_notification_address containing partial hostnames
* DNS search lists (common in Kubernetes environments)
* Default dns_use_search_list=1 setting

#### Workarounds
Before this fix, users could work around the issue by:
* Using fully qualified domain names in dmq_notification_address
* Setting dns_use_search_list=0 in kamailio.cfg

### DIFF
--- a/src/modules/dmq/notification_peer.c
+++ b/src/modules/dmq/notification_peer.c
@@ -228,6 +228,7 @@ int get_dmq_host_list(
 	for(prec = phead; prec; prec = prec->next) {
 		/**********
 		* o check max
+		* o skip non-A records (e.g., CNAME)
 		* o create URI
 		**********/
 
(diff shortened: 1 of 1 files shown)


henningw:
Thanks, merged and also cherry-picked to 5.8 and 6.0 branches

PR #4537, opened 2025-12-24
peerstate: Add peer state tracking module

Serdar51:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR introduces a new Kamailio module named `peerstate`.

The peerstate module provides real-time tracking of SIP peer state by
correlating dialog lifecycle events with registration status. It
maintains a shared-memory cache of peer states and supports multiple
simultaneous dialogs and contacts per peer.

The design and behavior of the module are inspired by existing presence
and PUA-related modules such as `pua_reginfo` and `pua_dialoginfo`.
Some concepts and mechanisms used in those modules (dialog and
registration event handling, some parameters) are reused and adapted for peer state tracking without requiring
full SIP presence functionality. 

Peer state changes are exposed through:
- Kamailio event routes (event_route[peerstate:state_changed])
- A callback-based API for other modules
- RPC commands for runtime inspection and monitoring

The module integrates with the `dialog` and `usrloc` modules to receive
dialog and registration events. It supports both URI-derived and
AVP-based peer identification and handles multi-contact registration
scenarios conservatively.

Documentation is provided in DocBook XML format following the standard
Kamailio Admin Guide / Developer Guide structure.

### DIFF
--- a/src/modules/peerstate/docs/peerstate_admin.xml
+++ b/src/modules/peerstate/docs/peerstate_admin.xml
@@ -1,0 +1,596 @@
+<?xml version="1.0" encoding='UTF-8'?>
+<!DOCTYPE book PUBLIC "-//OASIS//DTD DocBook XML V4.4//EN"
+"http://www.oasis-open.org/docbook/xml/4.4/docbookx.dtd" [
+
+<!ENTITY % docentities SYSTEM "../../../../doc/docbook/entities.xml">
+%docentities;
+
+]>
+
+<chapter>
+    
+    <title>&adminguide;</title>
+    
+    <section>
+	<title>Overview</title>
+	<para>
+	    The peerstate module provides peer state tracking for the 
+	    Kamailio SIP proxy. It monitors both dialog events (call states) 
+	    and registration events (usrloc) to maintain an accurate view of 
+	    each peer's current status.
+	</para>
+	<para>
+	    For example, a common need in call center applications is to track 
+	    agent availability in real-time. In order to monitor agent states, 
+	    it is necessary for the proxy to be aware of both call activity and 
+	    registration status. This is just one common application; there are 
+	    many others such as presence services, load balancing, and 
+	    concurrent call limiting.
+	</para>
+	<para>
+	    The module exports RPC commands for monitoring and management, 
+	    provides an event route mechanism for custom logic execution, and 
+	    offers a callback API for other Kamailio modules.
+	</para>
+    </section>
+    
+    <section>
+	<title>How it works</title>
+	<para>
+	    The peerstate module registers callbacks with the dialog and usrloc 
+	    modules. When dialog state changes (EARLY, CONFIRMED, TERMINATED) 
+	    or registration events (REGISTER/UNREGISTER) occur, the module 
+	    updates its internal cache and determines if a peer's state has 
+	    changed.
+	</para>
+	<para>
+	    Peer state is maintained in a shared memory hash table. Each peer 
+	    entry contains current state, active call count, registration flag, 
+	    and timestamp.
+	</para>
+	<para>
+	    State transitions:
+	    <itemizedlist>
+		<listitem>
+		    <para>EARLY events increment call count and set peer to RINGING (or 
+		    keep INUSE if already in call)</para>
+		</listitem>
+		<listitem>
+		    <para>CONFIRMED events set peer to INUSE</para>
+		</listitem>
+		<listitem>
+		    <para>TERMINATED events decrement call count; peer returns to 
+		    NOT_INUSE (if registered) or UNAVAILABLE (if not registered) 
+		    when call count reaches zero</para>
+		</listitem>
+		<listitem>
+		    <para>REGISTER events set registration flag and may transition from 
+		    UNAVAILABLE to NOT_INUSE</para>
+		</listitem>
+		<listitem>
+		    <para>UNREGISTER events clear registration flag and may transition 
+		    from NOT_INUSE to UNAVAILABLE (if no active calls)</para>
+		</listitem>
+	    </itemizedlist>
+	</para>
+	<para>
+	    When a peer's state changes, the module triggers notifications via 
+	    event routes and callback functions.
+	</para>
+    </section>
+    
+    <section>
+	<title>Peer states</title>
+	<para>
+	    Peer states tracked by the module:
+	    <itemizedlist>
+		<listitem>
+		    <para>0 : NOT_INUSE - Peer is registered and idle (no active calls)</para>
+		</listitem>
+		<listitem>
+		    <para>1 : INUSE - Peer is in an active call (dialog confirmed)</para>
+		</listitem>
+		<listitem>
+		    <para>2 : RINGING - Peer has incoming or outgoing call ringing (early dialog)</para>
+		</listitem>
+		<listitem>
+		    <para>3 : UNAVAILABLE - Peer is not registered</para>
+		</listitem>
+		<listitem>
+		    <para>4 : NA - Not applicable or unknown state</para>
+		</listitem>
+	    </itemizedlist>
+	</para>
+	<para>
+	    State transitions are designed to accurately reflect the peer's availability:
+	    <itemizedlist>
+		<listitem>
+		    <para>A peer transitions to RINGING when receiving or making a call</para>
+		</listitem>
+		<listitem>
+		    <para>A peer transitions to INUSE when the call is answered</para>
+		</listitem>
+		<listitem>
+		    <para>A peer returns to NOT_INUSE when all calls end and the peer is still registered</para>
+		</listitem>
+		<listitem>
+		    <para>A peer transitions to UNAVAILABLE when unregistering with no active calls</para>
+		</listitem>
+		<listitem>
+		    <para>Active calls are preserved during registration changes</para>
+		</listitem>
+	    </itemizedlist>
+	</para>
+    </section>
+    
+    <section>
+	<title>Dependencies</title>
+	<section>
+	    <title>&kamailio; Modules</title>
+	    <para>
+		The following modules must be loaded before this module:
+		<itemizedlist>
+		    <listitem>
+			<para><emphasis>Dialog</emphasis> - Dialog tracking module</para>
+		    </listitem>
+		    <listitem>
+			<para><emphasis>Usrloc</emphasis> - User location module</para>
+		    </listitem>
+		</itemizedlist>
+	    </para>
+	</section>
+	<section>
+	    <title>External Libraries or Applications</title>
+	    <para>
+		The following libraries or applications must be installed before 
+		running &kamailio; with this module loaded:
+		<itemizedlist>
+		    <listitem>
+			<para><emphasis>None</emphasis>.</para>
+		    </listitem>
+		</itemizedlist>
+	    </para>
+	</section>
+    </section>
+    
+    <section>
+	<title>Parameters</title>
+	<section id="peerstate.p.cache_hash_size">
+	    <title><varname>cache_hash_size</varname> (integer)</title>
+	    <para>
+		The size of the hash table internally used to keep the peer state 
+		information. A larger table is much faster but consumes more 
+		memory. The hash size must be a power of two.
+	    </para>
+	    <para>
+		<emphasis>
+		    Default value is <quote>4096</quote>.
+		</emphasis>
+	    </para>
+	    <example>
+		<title>Set <varname>cache_hash_size</varname> parameter</title>
+		<programlisting format="linespecific">
+...
+modparam("peerstate", "cache_hash_size", 8192)
+...
+</programlisting>
+	    </example>
+	</section>
+	
+	<section id="peerstate.p.cache_expire">
+	    <title><varname>cache_expire</varname> (integer)</title>
+	    <para>
+		The time in seconds after which inactive peer entries are removed 
+		from the cache. Set to 0 to disable automatic expiration. This 
+		helps conserve memory by removing stale peer information.
+		Minimum allowed value is 60 seconds (when enabled).
+	    </para>
+	    <para>
+		<emphasis>
+		    Default value is <quote>3600</quote> (1 hour).
+		</emphasis>
+	    </para>
+	    <example>
+		<title>Set <varname>cache_expire</varname> parameter</title>
+		<programlisting format="linespecific">
+...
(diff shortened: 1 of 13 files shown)


Serdar51:
@henningw Hi,

Could you please review this PR when you have time? All changes have been completed.

miconda:
I haven't really looked at the code, I see that all checks passed, so it is a good sign for merging soon.

What I noticed, which can be done also after merge, is that the code files do not have the license(GPL or BSD)/copyright (your own) at the top.

Serdar51:
Hi Daniel,

It is more ready now, tnx.

henningw:
> Hi Daniel,
> 
> It is more ready now, tnx.

Thanks, will also have a look to it today and give you feedback. As mentioned on the list, as the PR was submitted before, we would be able to merge it for 6.1 even after the freeze period starts in some days.

oej:
The terminology "real-time peer" seems to come from Asterisk and is not a generic terminology used in the SIP world. We need another way of documenting and naming this. I strongly object to this terminology. 

We also need to clearly document if this is the state of a single UA or the state of an account - where multiple UAs has registred for the same account.

henningw:
> The terminology "real-time peer" seems to come from Asterisk and is not a generic terminology used in the SIP world. We need another way of documenting and naming this. I strongly object to this terminology.
> 
> We also need to clearly document if this is the state of a single UA or the state of an account - where multiple UAs has registred for the same account.

Do you have some suggestions for a better name in your opinion?

miconda:
I think in this case the `realtime` attribute applies to the `state` of the `peer`, not to the `peer` itself. With all these words being commonly used in realtime communications, I don't see it confusing, but fine for me if people want it differently.

oej:
I don't see that we have used "peer" anywhere in Kamailio before. aorstate would be more in line with the docs of USRLOC, or even userstate.

- The README file is autogenerated and should not be part of the PR
- Why does this module implement its own hash table? Using the htable module would enable replication and much more. There may be good reasons, but I would like to see them :-) Having this module without being able to run multiple servers will be a problem.

miconda:
The `peer` as a term is used a lot in presence modules, dmq, tls, ... there is even a module named `peering` (which is about trusting peers). I haven't looked at the code, but I understood it is also about dialog states and calls can be made by gateways or interconnect partners, which are not local subscribers ... so aor or user can be restrictive/not reflecting all kind of entities that can be tracked by this module.

The README must be in the initial commit of a new module. Further updates have to be done only in the xml files.

The htable module has its hash table designed for (key, value) items, where the value can be integer or string. If that is enough, it can be reused, otherwise lot of other modules have their own specific designed hash tables, including tm, usrloc, dialog, ...

oej:
Just do not use "realtime" and "peer" in combination - that will surely lead to problems for those that have used the asterisk realtime peers. It will be confusing.

miconda:
Fine for me whatever is decided/preferred, my comments were just to present how I perceived it and what is already in kamailio.

Serdar51:
Firstly, thanks for the reviews.

> I don't see that we have used "peer" anywhere in Kamailio before. aorstate would be more in line with the docs of USRLOC, or even userstate.

Actually, I initially considered the terms `aor` and `user`, but wanted to use a more generic concept. These terms, however, are primarily associated with `usrloc`.

```
modparam("peerstate", "use_avps", 1) 
modparam("peerstate", "caller_avp", "$avp(caller_xxx)")
modparam("peerstate", "callee_avp", "$avp(callee_xxx)")

$avp(caller_xxx) = $_s($fU@$fd);
$avp(callee_xxx) = $_s($rU@$rd);
```

This approach enables tracking and modifying state for multiple custom `peers` on both caller and callee sides at the `dialog` level, even for non-local entities.

> * Why does this module implement its own hash table? Using the htable module would enable replication and much more. There may be good reasons, but I would like to see them :-) Having this module without being able to run multiple servers will be a problem.

I considered using the `htable` api, but I would have needed to structure the data to be stored for peer as json etc. I thought this could create extra load for update and build processes. I wanted to provide peer access with O(1) complexity this way.

Finally, while the content differs, using "realtime peer state" as a title could cause confusion with Asterisk's realtime concept. Can be changed, it's not a must-have. As @miconda mentioned, what's realtime is the "state of peer", not the "peer" itself.

src/modules/peerstate/peerstate.h
```diff
@@ -0,0 +1,63 @@
...
+
+#ifndef _PEERSTATE_H
+#define _PEERSTATE_H
+
+#include "../../core/parser/parse_uri.h"
+struct str_list;
+
+#define PEERSTATE_EVENT_ROUTE "dialog:state_changed"
+
+typedef enum pstate
```
henningw: Please add the module name as prefix, as there is already a pstate variable in the codebase.

src/modules/peerstate/peerstate.h
```diff
@@ -0,0 +1,63 @@
...
+
+static const char *pstate_strs[] __attribute__((unused)) = {
+		"NOT_INUSE", "INUSE", "RINGING", "UNAVAILABLE", "N/A"};
+
+#define PSTATE_TO_STR(s)                                                   \
+	((s) >= 0 && (s) < (int)(sizeof(pstate_strs) / sizeof(pstate_strs[0])) \
+					? pstate_strs[s]                                       \
+					: "UNKNOWN")
+
+struct dlginfo_cell
```
henningw: Please add the module name as prefix, as there is already a definition with this name

src/modules/peerstate/peerstate_cache.c
```diff
@@ -0,0 +1,688 @@
...
+
+#include "../../core/mem/shm_mem.h"
+#include "../../core/locking.h"
+#include "../../core/str_hash.h"
+#include "../../core/dprint.h"
+#include "../../core/timer.h"
+#include "../../core/timer_proc.h"
+#include "peerstate_cache.h"
+
+struct peer_state
```
henningw: If this is used only here, please make it static. There is also already a peer_state.

oej:
Thank you for the explanations! Let's agree on not using "realtime peer" in any texts to avoid any confusion. Otherwise I'm ok.

src/modules/peerstate/peerstate_cache.c
```diff
@@ -0,0 +1,688 @@
...
+	old_count = p->call_count;
+
+	// Process based on candidate state
+	if(candidate_state == RINGING) {
+		// EARLY event occurred
+		// Always increment counter for new dialog
+		p->call_count++;
+		new_count = p->call_count;
+
+		// Update state based on priority: INUSE > RINGING > NOT_INUSE
```
henningw, line 230: Just a comment, maybe it would be easier to use a switch statement

src/modules/peerstate/peerstate_cache.c
```diff
@@ -0,0 +1,688 @@
...
+
+		LM_DBG("REGISTERED: peer='%.*s', old_state=%s, new_state=%s, count=%d, "
+			   "is_registered=%d\n",
+				peer->len, peer->s, PSTATE_TO_STR(old_state),
+				PSTATE_TO_STR(new_state), call_count, p->is_registered);
+	} else if(candidate_state == UNAVAILABLE) {
+		// UNREGISTERED event occurred
+		// Update registration flag
+		p->is_registered = 0;
+		if(old_state == NOT_INUSE)
```
henningw: Just a comment, maybe it would be easier to use a switch statement

src/modules/peerstate/peerstate_cache.c
```diff
@@ -0,0 +1,688 @@
...
+	for(i = 0; i < peer_state_hash_size; i++) {
+		p = peer_state_hash_table[i];
+		while(p) {
+			if(p->state == target_state) {
+				if(found >= capacity) {
+					capacity *= 2;
+					str *new_result =
+							(str *)pkg_realloc(result, capacity * sizeof(str));
+					if(!new_result) {
+						lock_release(peer_state_lock);
```
henningw, line 573: There should be probably also some error logging here, e.g. PKG_MEM_ERROR

src/modules/peerstate/peerstate_cache.c
```diff
@@ -0,0 +1,688 @@
...
+						pkg_free(result);
+						return -1;
+					}
+					result = new_result;
+				}
+
+				result[found].s = (char *)pkg_malloc(p->peer.len + 1);
+				if(!result[found].s) {
+					lock_release(peer_state_lock);
```
henningw, line 583: There should be probably also some error logging here, e.g. PKG_MEM_ERROR

src/modules/peerstate/peerstate_cache.c
```diff
@@ -0,0 +1,688 @@
...
+
+	for(i = 0; i < peer_state_hash_size; i++) {
+		p = peer_state_hash_table[i];
+		while(p) {
+			if(found >= capacity) {
+				capacity *= 2;
+				str *new_result =
+						(str *)pkg_realloc(result, capacity * sizeof(str));
+				if(!new_result) {
+					lock_release(peer_state_lock);
```
henningw, line 636: There should be probably also some error logging here, e.g. PKG_MEM_ERROR

src/modules/peerstate/peerstate_cache.c
```diff
@@ -0,0 +1,688 @@
...
+					pkg_free(result);
+					return -1;
+				}
+				result = new_result;
+			}
+
+			result[found].s = (char *)pkg_malloc(p->peer.len + 1);
+			if(!result[found].s) {
+				lock_release(peer_state_lock);
```
henningw, line 646: There should be probably also some error logging here, e.g. PKG_MEM_ERROR

src/modules/peerstate/peerstate_mod.c
```diff
@@ -0,0 +1,1329 @@
...
+#define DEF_ENABLE_NOTIFY_ON_TRYING 0
+#define DEF_DISABLE_CALLER_NOTIFY_FLAG -1
+#define DEF_DISABLE_CALLEE_NOTIFY_FLAG -1
+#define DEF_DISABLE_REG_NOTIFY -1
+#define DEF_DISABLE_DLG_NOTIFY -1
+#define DEF_CACHE_EXPIRE 3600
+#define DEF_CACHE_CLEANUP_INTERVAL 300
+#define DEF_CACHE_HASH_SIZE 4096
+
+#define MAX_CALLID_BUF_SIZE 1024
```
henningw: This define and the following one seems to be unused, then it should be removed

src/modules/peerstate/peerstate_mod.c
```diff
@@ -0,0 +1,1329 @@
...
+#define STR_AT "@"
+
+struct dlg_binds dlg_api;
```
henningw: If its unsed only here, make it static. Otherwise add the module name as prefix to it.
Also look to the following api variable.

src/modules/peerstate/peerstate_mod.c
```diff
@@ -0,0 +1,1329 @@
...
+	/* Prepare callback context */
+	memset(&ctx, 0, sizeof(peerstate_cb_ctx_t));
+	ctx.peer = *peer;
+	ctx.current_state = pstate;
+	ctx.previous_state = prev_state;
+	ctx.uniq_id = *uniq_id;
+	ctx.event_type = event_type;
+
+	/* Get additional info from cache */
+	if(get_peer_info(peer, NULL, &call_count, &is_registered) == 0) {
```
henningw, line 296: Should we log an error/notice/info message?

src/modules/peerstate/peerstate_mod.c
```diff
@@ -0,0 +1,1329 @@
...
+
+	if(disable_callee_notify_flag != -1
+			&& (request
+					&& (request->flags & (1 << disable_callee_notify_flag))))
+		disable_callee_notify = 1;
+
+	dlginfo = get_dialog_data(
+			dlg, type, disable_caller_notify, disable_callee_notify);
+
+	if(dlginfo == NULL)
```
henningw: Should we log an error/notice/info message?

src/modules/peerstate/peerstate_mod.c
```diff
@@ -0,0 +1,1329 @@
...
+
+	if(request == NULL || request->REQ_METHOD != METHOD_INVITE)
+		return;
+
+	LM_INFO("INVITE dialog loaded: from=%.*s\n", dlg->from_uri.len,
+			dlg->from_uri.s);
+
+	dlginfo = get_dialog_data(dlg, type, 0, 0);
+
+	if(dlginfo == NULL)
```
henningw: Should we log an error/notice/info message?

src/modules/peerstate/peerstate_mod.c
```diff
@@ -0,0 +1,1329 @@
...
+		buf.len += it->s.len;
+		count++;
+		it = it->next;
+	}
+	if(count > 1)
+		buf.len += (count - 1);
+
+	buf.s = (char *)pkg_malloc(buf.len + 1);
+	if(!buf.s)
+		return -1;
```
henningw, line 844: Error logging, e.g. PKG_MEM_ERROR

src/modules/peerstate/docs/peerstate_admin.xml
```diff
@@ -0,0 +1,595 @@
...
+]>
+
+<chapter>
+    
+    <title>&adminguide;</title>
+    
+    <section>
+	<title>Overview</title>
+	<para>
+	    The peerstate module provides real-time peer state tracking for the
```
henningw: As discusses, should be re-phrased to not mixup with other projects:
"The peerstate module provides peer state tracking in real-time for the.."

src/modules/peerstate/peerstate_cache.c
Serdar51: struct peer_state only used here it is not also exposed on header file and it is file-local. But still, changing the name makes sense to avoid confusion.

Serdar51:
Hi Henning,
I’ve consolidated all modifications into a single commit following your advice.
Thanks for taking the time to review it.

henningw:
Thank you, I will review again and merge it.

miconda:
@henningw: ok to merge it? Just to start getting a rough estimation on possibility to create the branch 6.1 in a matter of days or not (like 2 weeks since the freeze time).

henningw:
Thank you, merged.

PR #4539, opened 2025-12-31
permissions: add locks support

zenichev:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Move permissions module to work with hash tables instead of directly working with lists.

Use locks to cover these hash tables.
This allows to avoid using crutches alike 0039c50e83916,
and gives a race-free usage of tables regardless the amount of data inserted into tables and regardless how many competitive reloads are pending at a time (e.g. 2-3 RPC commands sent at a time to reload same tables).

Use hash indexes for a quicker lookup in buckets.

Improve the mechanism of duplicate entries detection,
use: the source ip, ruri pattern, from pattern and the protocol to detect existing similar entries.

Add more safety checks in functions manipulating hash tables,
like checking NULL pointers, empty patterns for matching purposes etc.

Align return values in trusted/hash implementations so that return values are everywhere considered in the similar way
(everything != 0 is considered as failure, 0 is considered as fine).

Split tables allocation and initialization.

Add destruction handlers for tables instead of manually handling
entries freeing.

Some other minor improvements.

### DIFF
--- a/src/modules/permissions/hash.c
+++ b/src/modules/permissions/hash.c
@@ -357,40 +357,6 @@ struct trusted_hash_table * trusted_table_allocate(unsigned int hash_table_size)
 
 
 /*
- * Create and initialize a hash table
- */
-struct trusted_list **new_hash_table(void)
-{
-	struct trusted_list **ptr;
-
-	/* Initializing hash tables and hash table variable */
-	ptr = (struct trusted_list **)shm_malloc(
-			sizeof(struct trusted_list *) * PERM_HASH_SIZE);
-	if(!ptr) {
-		LM_ERR("no shm memory for hash table\n");
-		return 0;
-	}
-
-	memset(ptr, 0, sizeof(struct trusted_list *) * PERM_HASH_SIZE);
-	return ptr;
-}
-
-
-/*
- * Release all memory allocated for a hash table
- */
-void free_hash_table(struct trusted_list **table)
-{
-	if(!table)
-		return;
-
-	empty_hash_table(table);
-	shm_free(table);
-}
-
-
-
-/*
  * Add <src_ip, proto, pattern, ruri_pattern, tag, priority> into hash table, where proto is integer
  * representation of string argument proto.
  */
@@ -761,34 +727,6 @@ int hash_table_rpc_print(struct trusted_hash_table *trusted_table, rpc_t *rpc, v
 		lock_release(trusted_table->row_locks[i]);
 	}
 	return 0;
-}
-
-/*
- * Free contents of hash table, it doesn't destroy the
- * hash table itself
- */
-void empty_hash_table(struct trusted_list **table)
-{
-	int i;
-	struct trusted_list *np, *next;
-
-	for(i = 0; i < PERM_HASH_SIZE; i++) {
-		np = table[i];
-		while(np) {
-			if(np->src_ip.s)
-				shm_free(np->src_ip.s);
-			if(np->pattern)
-				shm_free(np->pattern);
-			if(np->ruri_pattern)
-				shm_free(np->ruri_pattern);
-			if(np->tag.s)
-				shm_free(np->tag.s);
-			next = np->next;
-			shm_free(np);
-			np = next;
-		}
-		table[i] = 0;
-	}
 }
 
 
--- a/src/modules/permissions/hash.c
+++ b/src/modules/permissions/hash.c
@@ -389,128 +389,187 @@ void free_hash_table(struct trusted_list **table)
 }
 
 
+
 /*
  * Add <src_ip, proto, pattern, ruri_pattern, tag, priority> into hash table, where proto is integer
  * representation of string argument proto.
  */
-int hash_table_insert(struct trusted_list **table, char *src_ip, char *proto,
+int hash_table_insert(struct trusted_hash_table *hash_table, char *src_ip, char *proto,
 		char *pattern, char *ruri_pattern, char *tag, int priority)
 {
-	struct trusted_list *np;
-	struct trusted_list *np0 = NULL;
-	struct trusted_list *np1 = NULL;
-	unsigned int hash_val;
+	struct trusted_list *cur_entry, *prev_entry;
+	struct trusted_list *new_entry = NULL;
 
-	np = (struct trusted_list *)shm_malloc(sizeof(*np));
-	if(np == NULL) {
-		LM_ERR("cannot allocate shm memory for table entry\n");
+	unsigned int hash_index;
+
+	new_entry = (struct trusted_list *)shm_malloc(sizeof(*new_entry));
+	if(new_entry == NULL) {
+		LM_ERR("cannot allocate shm memory for new table entry\n");
 		return -1;
 	}
 
 	if(strcasecmp(proto, "any") == 0) {
-		np->proto = PROTO_NONE;
+		new_entry->proto = PROTO_NONE;
 	} else if(strcasecmp(proto, "udp") == 0) {
-		np->proto = PROTO_UDP;
+		new_entry->proto = PROTO_UDP;
 	} else if(strcasecmp(proto, "tcp") == 0) {
-		np->proto = PROTO_TCP;
+		new_entry->proto = PROTO_TCP;
 	} else if(strcasecmp(proto, "tls") == 0) {
-		np->proto = PROTO_TLS;
+		new_entry->proto = PROTO_TLS;
 	} else if(strcasecmp(proto, "sctp") == 0) {
-		np->proto = PROTO_SCTP;
+		new_entry->proto = PROTO_SCTP;
 	} else if(strcasecmp(proto, "ws") == 0) {
-		np->proto = PROTO_WS;
+		new_entry->proto = PROTO_WS;
 	} else if(strcasecmp(proto, "wss") == 0) {
-		np->proto = PROTO_WSS;
+		new_entry->proto = PROTO_WSS;
 	} else if(strcasecmp(proto, "none") == 0) {
-		shm_free(np);
+		shm_free(new_entry);
 		return 1;
 	} else {
 		LM_CRIT("unknown protocol\n");
-		shm_free(np);
+		shm_free(new_entry);
 		return -1;
 	}
 
-	np->src_ip.len = strlen(src_ip);
-	np->src_ip.s = (char *)shm_malloc(np->src_ip.len + 1);
-
-	if(np->src_ip.s == NULL) {
+	/* source IP */
+	new_entry->src_ip.len = strlen(src_ip);
+	new_entry->src_ip.s = (char *)shm_malloc(new_entry->src_ip.len + 1);
+	if(new_entry->src_ip.s == NULL) {
 		LM_CRIT("cannot allocate shm memory for src_ip string\n");
-		shm_free(np);
+		shm_free(new_entry);
 		return -1;
 	}
+	(void)strncpy(new_entry->src_ip.s, src_ip, new_entry->src_ip.len);
+	new_entry->src_ip.s[new_entry->src_ip.len] = 0;
 
-	(void)strncpy(np->src_ip.s, src_ip, np->src_ip.len);
-	np->src_ip.s[np->src_ip.len] = 0;
-
+	/* pattern */
 	if(pattern) {
-		np->pattern = (char *)shm_malloc(strlen(pattern) + 1);
-		if(np->pattern == NULL) {
+		new_entry->pattern = (char *)shm_malloc(strlen(pattern) + 1);
+		if(new_entry->pattern == NULL) {
 			LM_CRIT("cannot allocate shm memory for pattern string\n");
-			shm_free(np->src_ip.s);
-			shm_free(np);
+			shm_free(new_entry->src_ip.s);
+			shm_free(new_entry);
 			return -1;
 		}
-		(void)strcpy(np->pattern, pattern);
+		(void)strcpy(new_entry->pattern, pattern);
 	} else {
-		np->pattern = 0;
+		new_entry->pattern = 0;
 	}
 
+	/* R-URI pattern */
 	if(ruri_pattern) {
-		np->ruri_pattern = (char *)shm_malloc(strlen(ruri_pattern) + 1);
-		if(np->ruri_pattern == NULL) {
+		new_entry->ruri_pattern = (char *)shm_malloc(strlen(ruri_pattern) + 1);
+		if(new_entry->ruri_pattern == NULL) {
 			LM_CRIT("cannot allocate shm memory for ruri_pattern string\n");
-			shm_free(np->src_ip.s);
-			shm_free(np);
+			shm_free(new_entry->src_ip.s);
+			shm_free(new_entry);
 			return -1;
 		}
-		(void)strcpy(np->ruri_pattern, ruri_pattern);
+		(void)strcpy(new_entry->ruri_pattern, ruri_pattern);
 	} else {
-		np->ruri_pattern = 0;
+		new_entry->ruri_pattern = 0;
 	}
 
+	/* tag */
 	if(tag) {
-		np->tag.len = strlen(tag);
-		np->tag.s = (char *)shm_malloc((np->tag.len) + 1);
-		if(np->tag.s == NULL) {
+		new_entry->tag.len = strlen(tag);
+		new_entry->tag.s = (char *)shm_malloc((new_entry->tag.len) + 1);
+		if(new_entry->tag.s == NULL) {
 			LM_CRIT("cannot allocate shm memory for pattern or ruri_pattern "
 					"string\n");
-			shm_free(np->src_ip.s);
-			shm_free(np->pattern);
-			shm_free(np->ruri_pattern);
-			shm_free(np);
+			shm_free(new_entry->src_ip.s);
+			shm_free(new_entry->pattern);
+			shm_free(new_entry->ruri_pattern);
+			shm_free(new_entry);
 			return -1;
 		}
-		(void)strcpy(np->tag.s, tag);
+		(void)strcpy(new_entry->tag.s, tag);
 	} else {
-		np->tag.len = 0;
-		np->tag.s = 0;
+		new_entry->tag.len = 0;
+		new_entry->tag.s = 0;
 	}
 
-	np->priority = priority;
+	/* priority */
+	new_entry->priority = priority;
 
-	hash_val = perm_hash(np->src_ip);
-	if(table[hash_val] == NULL) {
-		np->next = NULL;
-		table[hash_val] = np;
-	} else {
-		np1 = NULL;
-		np0 = table[hash_val];
-		while(np0) {
-			if(np0->priority < np->priority)
-				break;
-			np1 = np0;
-			np0 = np0->next;
-		}
-		if(np1 == NULL) {
-			np->next = table[hash_val];
-			table[hash_val] = np;
-		} else {
-			np->next = np1->next;
-			np1->next = np;
-		}
+	/* determinde the indexation */
+	hash_index = perm_hash(new_entry->src_ip);
+
+	/* ensure to have a correct index within boundaries */
+	if (hash_index >= hash_table->size) {
+		trusted_table_free_entry(new_entry);
+		LM_ERR("hash index out of bounds [%d]\n", hash_index);
+		return -1;
 	}
 
-	return 1;
+	/* now new entry is ready for insertion */
+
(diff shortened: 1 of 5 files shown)


zenichev:
@juha-h or @henningw  can you guys please have a look?

P.S.:
The failed format check (due to the Clang style) is actually fulfilled with the latest of the commits in the list. I don't know why it still shows this check as failed tbh.

miconda:
This is quite a big change for one of the most used modules out there, so it has to be carefully reviewed if decided to be merged.

Is the ability to do reloads without time interval limiting the main benefit? How is the reload performed, changing the buckets of the hash table?

Are the changes only for the `trusted` table, or also for the `address` table?

zenichev:
@miconda 

> Is the ability to do reloads without time interval limiting the main benefit? 

**The main problem out there is that:**
the bigger is the trusted table, the bigger must be the reload delta (modparam) that controls the time between two rpc triggered reloads. As it takes more time to reload the table when each new chunk of new sources is provisioned.

As our experience showed, on the heavily loaded platforms the reload delta should regularly be increased to not let two or even more competing RPC processes suddenly manipulate the same linked list. Moreover the 0039c50e83916c6 commit doesn't really solve the problem, probably it slightly improved the situation, but still somehow we still keep on crashing while doing the insert, an example (part of the BT trace, kamailio 5.8.6):
```
#0  hash_table_insert (table=table@entry=0x7f3a655f3cb0, src_ip=<optimized out>, proto=<optimized out>, pattern=pattern@entry=0x561a5e01b569 ".*username111.*", ruri_pattern=ruri_pattern@entry=0x0,
    tag=tag@entry=0x561a5e01b583 "uuid-was-here-but-is-substituted", priority=0) at ./src/modules/permissions/hash.c:231
231     ./src/modules/permissions/hash.c: No such file or directory.
(gdb) bt full
#0  hash_table_insert (table=table@entry=0x7f3a655f3cb0, src_ip=<optimized out>, proto=<optimized out>, pattern=pattern@entry=0x561a5e01b569 ".*username111.*", ruri_pattern=ruri_pattern@entry=0x0,
    tag=tag@entry=0x561a5e01b583 "uuid-was-here-but-is-substituted", priority=0) at ./src/modules/permissions/hash.c:231
        np = 0x7f3a66a09490
        np0 = 0x6f5f65656c6c6163
        np1 = <optimized out>
        hash_val = <optimized out>
        __func__ = "hash_table_insert"
#1  0x00007f3bbccfc269 in reload_trusted_table () at ./src/modules/permissions/trusted.c:147
        cols = {0x7f3bbcd0fa50 <perm_source_col>, 0x7f3bbcd0fa40 <perm_proto_col>, 0x7f3bbcd0fa30 <perm_from_col>, 0x7f3bbcd0fa20 <perm_ruri_col>, 0x7f3bbcd0fa10 <perm_tag_col>, 0x7f3bbcd0fa00 <perm_priority_col>}
        res = 0x7f3bc3ff4980
        row = 0x7f3bc60f60c0
        val = 0x7f3bc2106820
        new_hash_table = 0x7f3a655f3cb0
        i = 10
        priority = <optimized out>
        pattern = <optimized out>
        ruri_pattern = <optimized out>
        tag = 0x561a5e01b583 "uuid-was-here-but-is-substituted"
        __func__ = "reload_trusted_table"
#2  0x00007f3bbcd018be in reload_trusted_table_cmd () at ./src/modules/permissions/trusted.c:712
        __func__ = "reload_trusted_table_cmd"
#3  0x00007f3bbccf9cea in rpc_trusted_reload (rpc=0x7f3bbf9f1220 <func_param>, c=0x7f3bbf9f1280 <ctx>) at ./src/modules/permissions/rpc.c:61
No locals.
#4  0x00007f3bbf9e7c64 in ki_dispatch_rpc (msg=<optimized out>) at ./src/modules/xmlrpc/xmlrpc.c:2693
        exp = 0x7f3bc60bbe80
        ret = 1
        rdata = 0
        __func__ = "ki_dispatch_rpc"
```

So from my PoV, the module itself has an architecture, so the code design, that allows such races to happen.
And therefore it should be reworked, whereas the quickest and the easiest way is to introduce locks and wrap lists into structures that really introduce real hash table concept. Please see the way how it's designed in the rtpengine module, where we never experienced issues when manipulating entries in hash buckets.

**So answering your question**: shortly said, yes, the main benefit in the short perspective is to solve the race issue of RPC reloads. But in the longer perspective actually to limit the other issues that can appear because of the unsecured buckets list.

> How is the reload performed, changing the buckets of the hash table?

I'm not sure I properly understood your question, but the reload is done on the same way, with an exception that now the table gets locked and one cannot introduce two or more processes changing the data inside of the table, while the first process is still not done.
The algo remains the same:
- select a table
- re-init the buckets inside (clean old content)
- re-insert the data adding the newly content plus older (so updated version)
- make the current table (cursor) pointing to the freshest table

The only difference is that now it manipulates the lists via the table structures having locks.
And also uses the updated hash API with newer safe methods (covered with locks).

> Are the changes only for the trusted table, or also for the address table?

For now only the trusted table gets the changes. We never run into the address table issues, but probably it's just a matter of time and luck. It has the same issues as the trusted table, totally lock-free and gives RPC processes a chance to introduce race issues like iterator invalidation.

---

One thing that I should mention apart of everything, is that the `hash_table_insert()` got a duplicates prevention mechanism. That wasn't in place before, so this must be taken into account (no duplicates in the trusted table will be allowed). I've mentioned this chance in the commit message, but probably should be added into the doc as well.
I'm not sure if this behavior change can be an obstacle for the community. I can make it optional via the modparam though. You can see it there: https://github.com/kamailio/kamailio/pull/4539/commits/5c052718482ed1764e1a89ed7f2f37a20324387b

---

Also I totally agree with you, we need a good expertise here before to actually accept such changes.
For now the only thing we can do from our side is run tests on our side, what we've done already.

henningw:
Thanks for the additional comments @zenichev, it sounds resonable to me, understand the motivation. But I think that we probably should migrate the module completely to the new approach, and not having a mix of new (trusted) and old (address) data structures logic. About the duplicate entries, if they are not needed from a functional point of view, we don't need to keep this capability IMHO.

As a side note, the permission module is probably a bit of a special case, as its data its much more volatile as e.g. LCR or carrierroute. The carrierroute module uses a simple approach with access locking and swapping the old and new data structures. But as there are no concurrent reloads done, this is of course fine.

zenichev:
@henningw 

> But I think that we probably should migrate the module completely to the new approach, and not having a mix of new (trusted) and old (address) data structures logic.

Just in the meanwhile, before this rework is actually done, I suggest still to cover the existing solution with locks.
Regarding the address table, I can update the commit and send additionally the coverage of address table with locks as well. I don't think this would be a big work.

> About the duplicate entries, if they are not needed from a functional point of view, we don't need to keep this capability IMHO.

So if I understand this correct, we want the duplicate entries to be allowed, or did I misunderstand you?

> As a side note, the permission module is probably a bit of a special case, as its data its much more volatile as e.g. LCR or carrierroute. The carrierroute module uses a simple approach with access locking and swapping the old and new data structures. But as there are no concurrent reloads done, this is of course fine.

Exactly, and that's why it is so much important to cover the trusted and address tables with locks.
We've been suffering from this drawback since a while. 

Let me know if I can add the address table rework here as well. Thanks.

henningw:
> @henningw
> 
> > But I think that we probably should migrate the module completely to the new approach, and not having a mix of new (trusted) and old (address) data structures logic.
> 
> Just in the meanwhile, before this rework is actually done, I suggest still to cover the existing solution with locks. Regarding the address table, I can update the commit and send additionally the coverage of address table with locks as well. I don't think this would be a big work.
> 
> > About the duplicate entries, if they are not needed from a functional point of view, we don't need to keep this capability IMHO.
> 
> So if I understand this correct, we want the duplicate entries to be allowed, or did I misunderstand you?

If there is no need for having duplicate entries, I think they are not necessary, we just need to document it.

> > As a side note, the permission module is probably a bit of a special case, as its data its much more volatile as e.g. LCR or carrierroute. The carrierroute module uses a simple approach with access locking and swapping the old and new data structures. But as there are no concurrent reloads done, this is of course fine.
> 
> Exactly, and that's why it is so much important to cover the trusted and address tables with locks. We've been suffering from this drawback since a while.
> 
> Let me know if I can add the address table rework here as well. Thanks.

For me its fine, maybe other developers want to add something as well. Just to understand it correctly, the mentioned rework would be a smaller change for the address, just adding some locking improvements? Or using basically the same approach as for the trusted? For me the using the same approach would be fine for both, as mentioned.

miconda:
Trying to understand the architecture of the changes .. on reload, you reuse the hash table: first you destroy the items on the buckets, keeping the first one as `dummy`, then insert new items with data loaded from db? It is no new hash table built in parallel and swapping the pointer to the active hash table.

What is supposed to happen in between 'destroying the items on all buckets' and 'inserting the new items'? If a Kamailio worker needs to match on trusted table, that is empty for some short time and it fails?

What is the role of the first `dummy` item per bucket?

zenichev:
@miconda 

> on reload, you reuse the hash table: first you destroy the items on the buckets, keeping the first one as dummy, then insert new items with data loaded from db?
> It is no new hash table built in parallel and swapping the pointer to the active hash table.

Current commit(s) keep the same logic as it's been before. There is no functional change in this regard.
From what I can see, permissions module implements the trusted table reload in the following way:
- it uses the global pointer to the current (so actual) tursted table
- two table versions are kept for the parallelism reasons (two or more processes updating the table at a time)
- a temporary pointer (in `reload_trusted_table()`) takes the one which is less actual (not the latest)
- and re-initializes that (so cleans the content fully)
- now it goes to the db backend, grabs actual entries, and inserts the actual list of rows one by one, this one will be the most actual table from now
- then the global pointer gets updated with the latest pointed table (within `reload_trusted_table()`) and hence starts using the very last updated linked list

This is the previously existing implementation, which is now in all of the kamailio branches including the master.
The newer implementation that I offered, does exactly the same in this regard, with the exception that it operates on the hash tables (new abstraction object), which are covered with the locks and doesn't allow any other 2d or 3d process in parallel accidentally to take the very same table and update it in parallel.

This is so far was the problem we've been experiencing since a while. And this a lot depends on the size of the table (in our case tables are quite big), and on the amount of RPC processes trying to trigger the reload, in our case it can be up to 3 processes trying to update the trusted table at a time.

> What is supposed to happen in between 'destroying the items on all buckets' and 'inserting the new items'?

The reload worker takes non-actual table when he destroys and then inserts new items.
It's the one which is not used at the moment by the module, that's why the already existing implementation keeps two tables for this purpose.
After it's updated it gets pointed by the global pointer, in other words becomes a new actual table. And the previously actual gets non-actual, and will be used next time in the same way to prepare a new table version.

> If a Kamailio worker needs to match on trusted table, that is empty for some short time and it fails?

This cannot happen, see my explanation above.
Again, I didn't introduce any functional/algorithm change. It is in all kamailio branches already, and it works fine.
What I do, is that I introduce safer way covered with locks, for cases when 2-3 or even more processes trying to reload table(s).
Moreover it's more elegant, because it uses an abstraction layer for representing a hash table, and not purely working with linked lists, what makes it more efficient and easy to work with (as well when coding).

> What is the role of the first dummy item per bucket?

The hash buckets can be implemented with a dummy head or without it.
On the heavy loaded things, like the permissions module is, it should be more efficient to operate with a dummy head, because:

- we are always sure that the bucket is initialized and ready to use (fast)
- the pointer to the bucket never changes, it stays the same, so only the `->next` part is an actual beginning
- this gives more stability (locks granularity, concurrent readers/writers, re-initialization)
- partial cleanup. We do not need to reallocate the bucket when doing a re-init (like during the reload): no allocation, no pointer churn, no locks re-init etc.
- not the last thing is that dummy heads based implementation simplifies the code (less code branching, more predictable iteration etc.)

For example, when doing the re-init, having no dummy head, I would have to free the buckets pointer, allocate it again, reassign.
What all in common gives a little bit more race windows increase and processing time. Even being covered with locks.

To summarize, I can re-write the code to have no dummy head in the bucket(s), but i suggest to keep it for the reasons mentioned above.

Please let me know if that explains your questions.

---

@henningw 

> If there is no need for having duplicate entries, I think they are not necessary, we just need to document it.

Yes, I agree. I will update the doc then.

> For me its fine, maybe other developers want to add something as well. Just to understand it correctly, the mentioned rework would be a smaller change for the address, just adding some locking improvements? Or using basically the same approach as for the trusted? For me the using the same approach would be fine for both, as mentioned.

Yes, exactly, I use the same approach for the address table, hence it's going to be within the same PR.
I would say the approach should be kept as much similar as possible, so that it's simpler for the code maintenance.

The address table rework is in progress, I haven't had much time yesterday/today, but it should be ready this/next week.

miconda:
@zenichev: thanks for the detailed explanation!

So you introduced a reload lock/mutex, that practically blocks other processes doing reloading once such operation is in progress. The switch to hash table is not related to the reload issue, just considered a better design to store the items?

Although not the main author of the module, I think the secondary table was kept (at least for a while) to ensure that SIP worker processes finish matching on the old active list when a reload is completed.

Probably I have to look at the code, since I still have some doubts: with your new approach, when you expect 2-3 processes to do reload immediately one after the other, if 1st reloading process swaps the active (old data) with secondary (new data) and there is a SIP worker still walking through the old data, then the 2nd reloading process completes the reload and swaps again, could it happen that the SIP worker is still walking the old-old-data, which has to be destroyed by the 2nd reloading process more or less at the same time?

Also, I haven't gotten (yet) the idea with the dummy item. Normally the bucket is its own structure, where the bucket lock is kept  as well as the pointer to the first item (maybe along with some other fields/statistics, like number of items) and the items are just holding the data. So the bucket structures are initialized only once, at startup, only the `first-item` field changes at runtime (concerting the stored data there, other fields might change as well, like the mentioned example of number of items).

zenichev:
@miconda 

Sorry for the late reply, I had to re-consider certain things.

> The switch to hash table is not related to the reload issue, just considered a better design to store the items?

hash tables abstraction in common with the locks fulfill two goals:
- design improvement
- robust and convenient way to introduce locks (we control them on the upper level)

I could introduce the locks on the linked lists directly, but that would have been much more cumbersome.
The model running on the hash tables is already common, used in some other modules.

> Although not the main author of the module, I think the secondary table was kept (at least for a while) to ensure that SIP worker processes finish matching on the old active list when a reload is completed.

Yes, this is not the least point. I assume they introduced the concept of cursor (pointer to the actual table) and two possible versions of them latest (newest) and outdated: to be able to write to one table, meanwhile the other can be used by SIP. Then tables get swapped. It feels a little bit oldish, but that works.

**The biggest drawback of that is**: it doesn't consider the multi-process (e.g. RPCs) system.
Whereas multiple reloads can hit the system and nothing stops those processes to have a race.

> could it happen that the SIP worker is still walking the old-old-data, which has to be destroyed by the 2nd reloading process more or less at the same time?

**Short answer**: currently the permissions module already has an issue, when two or more processes start simultaneously the reload.

**Longer answer**:
Let me try to explain, how it's working now (even without the locks patch).

The case:
- DB backend gets updated with the new provisioned trusted sources
- P1 fires the RPC reload, is ongoing
- DB backend gets now more updates, so other new provisioned trusted sources
- P2 fires the RPC reload and collides with the P1

What is going to happen:
- before all the updates started: table 1 is the newest version (global pointer on that), table 2 is the oldest version
- now updates started (first P1)
- P1 sees that table 1 is the actual one, takes table 2 (for updates), no locks, it starts the update (the global pointer is not touched yet)
- P2 has arrived with a newer reload, has the same situation as the P1, no locks, and nothing prevents it to empty the table (so cleaning the list), meanwhile P1 hasn't finished its update yet

This is a seg.fault, either because of the changed list (iterator invalidation) or because of hitting the corrupted pointer (re-written by parallel process), while trying to do read/write operations via it. So far, we experienced this case pretty much regularly. (to reproduce it, one needs to have quite big trusted tables, whereas each reload takes a sensible amount of time)

**What this PR does not do:**

This thing still not solved with the current PR, because this PR only introduces the locks for the scope of hash tables themselves, but does not cover the reload as whole (upper level). So it only solves the cases, when two processes simultaneously try to iterate the content (within one table).

**What we should better do to improve this change:**

Now, when I understood, that the solution I've uploaded is not ideal, I will have to extend this a little bit.

I will have to introduce additionally **an exclusive lock** for the reload function (`reload_trusted_table()`).
This one is going to serve the purpose of not letting the P2 select the very same table, when P1 already does updates.
Hence P2 will select always the older table, without having to suddenly clean the latest (actual) table.

What I mean:
- P1 acquired the locks of table 1, cleans the content of it and re-inits the buckets
- P1 starts filling in the table with a new content
- P2 arrives and waits for the lock to be acquired
- P1 releases the lock after the table is filled and makes the global pointer towards table 1
- P2 acquires the lock and cleans the content (<- here, one step before P2 selects the table for updates, we need the exclusive lock, which will serve only the `reload_trusted_table()`)

F.y.i: `reload_trusted_table()` is the only multi-process weak point of the module, as I can see.

Additionally it's worth having an **atomic bool**, for the cases when P3, P4 etc. are coming and also trying to acquire the exclusive lock.
If there is a process already waiting to acquire a lock and provide fresher updates, there is no need to let P3, P4.. do the same. So having the **atomic bool** is not obligatory, but that will make the reload process more efficient, because having less duplicated work.

> Also, I haven't gotten (yet) the idea with the dummy item. Normally the bucket is its own structure

I think I should have clarified better, but shortly said I wanted to stay with the dummy-head design
(when buckets always owns a permanent head node, which is pre-initilized and list items are coming always after it)
for the sake of stable head pointers in buckets, which gives simpler iteration implementation,
and reduced re-allocs.. so it's only about the speed. But this is definitely not obligatory to have it.

I get your idea, you're talking rather about the semantic clarity.
The bucket item must always keep a data, never dummy etc. 

This is possible, I can just rework it, will take some time, but I don't see it as a big deal.

---

**My current proposal for the further improvements is:**
- introduce an exclusive lock only for the scope of  `reload_trusted_table()` (solves sudden table clean)
- introduce an atomic bool (improves the work, no duplicated reloads)
- introduce the same concept for the address table (so that both trusted and address tables implementations are unified)

Please let me know what you think about my proposal, and let's find out what we should do about this rework.
Then I can invest my time to finalize this PR.

miconda:
My comments were more from the perspective of trying to figure out if this PR should be merged before branching 6.1, being a commonly-used module, I would prefer not to go with a half-solution.

Definitely things can be improved, MI/RPC was initially designed to be a serial interface, with a single process executing RPC commands, but later some rpc-transports allow parallel connections for rpc execution.

When a new data model to store and reload records is decided, it should be used across the module (address+trusted), that makes it easier to maintain/troubleshoot.

I will try find some time to look over the changes introduced by this PR and comment more.

zenichev:
@miconda All clear, thank you.
Also in parallel, I could finalize the address table rework and introduce changes I've mentioned (improvements to the reload function of the trusted table).

miconda:
@zenichev: thanks for you disposition to work further, but wait a bit with the address table and the other changes, we should first get to the conclusion of what is the best approach, otherwise it could be a waste of time.

zenichev:
@miconda hi! can we please have any update on this? Thank you.

miconda:
During the past few weeks the main focus was on testing and preparing the release of new major version Kamialio 6.1.0. We will come back to this one soon.

zenichev:
@miconda should we try to proceed?
I think I will have to actualize the code, because already a few months here.

henningw:
@zenichev The PR needs indeed a rebase, as there were some other changes introduced in the last weeks.

henningw:
@miconda The discussion mentioned some coordination about the next steps to take between you and @zenichev . The PR got a bit stalled, not sure if this needs still to be done?

zenichev:
@henningw Good time of the day.

Sorry I was busy last months, getting back here.
I see an amount of commits to the permissions module lately.

Would that make sense to re-send a new PR?
I would review the code again and adapt to the actual code, then probably also adding a new commit for the address table handling.

henningw:
Hi @zenichev thanks for the update. This PR need to definitely be rebased on the current status of the git master. You could do it here, or we can close this one and create a new PR, if this is easier, sure.

zenichev:
@henningw I feel better to start with a new PR, this one is quite stale.
I will try to send this out this week/beginning of the next week.

henningw:
Ok, I will close this one then. If some older discussion needs to be checked again, we can link them.

zenichev:
@henningw done here: https://github.com/kamailio/kamailio/pull/4787

PR #4540, opened 2026-01-01
db_redis: protect len overflow

zenichev:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description
Use `db_redis_key_add_str()` where possible
No sense to use `db_redis_key_add_string()`, if there is `str` object and not the char array.

Add string key length protection
Introduce a list of improvements for size overflow protection:
- change func signatures to work with `size_t` only
- check the raw pointer and its length

This measure will protect inserts from extremely large key size (e.g. possible negative length that then gets into integer overflow).

Add support of `max_key_length`
Introduce the modparam `max_key_length` which, if used, controls the max possible key length
to be used when adding/prepending strings into the redis key.

By default disabled (means 0). If used will be always compared against the given string key length.
This change provides full backwards compatibility.

### DIFF
--- a/src/modules/db_redis/redis_dbase.c
+++ b/src/modules/db_redis/redis_dbase.c
@@ -884,8 +884,7 @@ static int db_redis_scan_query_keys_pattern(km_redis_con_t *con,
 					goto err;
 				}
 			}
-			if(db_redis_key_add_str(&query_v, index_key)
-					!= 0) {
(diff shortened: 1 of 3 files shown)


zenichev:
@henningw or @agranig hey guys, could you please have a look?

P.S.: latter commit formats everything into the required clang format.

miconda:
Thanks! Merging it.

PR #4544, opened 2026-01-07
pv: Add sha3/keccak transformation

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Add sha3/keccak transformation for strings

### DIFF
--- a/src/modules/pv/pv_trans.c
+++ b/src/modules/pv/pv_trans.c
@@ -51,6 +51,7 @@
 
 #include "../../core/strutils.h"
 #include "../../core/crypto/shautils.h"
+#include "../../core/crypto/sha3.h"
 #include "pv_trans.h"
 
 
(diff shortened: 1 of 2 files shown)


xkaraman:
Hey @miconda,

We introduced the sha3 transformation and keccak but the interface is slightly different than the sha2 variant.

This PR uses a second parameter to determine the size like `$(var(x){s.sha3,bitSize}` where `bitSize=256|384|512` else it fails. Same for `keccak` variant with `$(var(x){s.keccak,512}`. One can also use it with no extra parameter like `$(var(x){s.sha3}` and will default to 256.

Would you prefet to have it like the sha2 interface with each one a different transformation like a `sha3-256` or something similar?

Not sure if order and size of `enum _tr_s_subtype` also matters, if you want me to add them to the end.

thanks

miconda:
Thanks, fine for me, merging!

PR #4559, opened 2026-01-19
db_redis: revert support of NULL values

zenichev:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

Noticed internally after introducing NULL pointer / 0 length
guards in d79540f7aba4f8c2bab1227bbe724609535b0d5d
for the `db_redis_key_add_string()`.

db_redis module implicitly supports conversion
of <null> key values into empty "" strings,
see: `db_redis_val2str()`.

Remove empty-string/zero-length guard introduced
previously (for adding keys), because they break this support.

E.g.: dialog and usrloc modules can in fact
handle <null> key values.

The guard for the prepend can be left intact because doesn't break anything.

Additionally: introduce the memcpy() guard,
because running it on the NULL pointer / 0 length
is in fact technically a typical UB.

The thing about memcpy() was never noticed before, because modern
libc implementations do not de-reference pointers
when the size is zero.

### DIFF
--- a/src/modules/db_redis/redis_table.c
+++ b/src/modules/db_redis/redis_table.c
@@ -34,11 +34,6 @@ int db_redis_key_add_string(redis_key_t **list, const char *entry, size_t len)
 {
 	redis_key_t *k;
 
-	if(!entry || !len) {
-		LM_ERR("Empty entry or zero length\n");
-		return -1;
-	}
-
 	if(db_redis_max_key_len > 0 && len > db_redis_max_key_len) {
 		LM_ERR("Too big length for key being added: allowed '%u' / given "
 			   "'%zu'\n",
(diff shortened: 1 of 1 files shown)


zenichev:
@miconda please have a look. Thanks.

miconda:
The https://github.com/kamailio/kamailio/commit/d79540f7aba4f8c2bab1227bbe724609535b0d5d seems to have introduced two blocks `if(!entry || !len)`, only one needs to be removed now?

zenichev:
@miconda i've introduced the guards for the add and for the prepend functionality.
But it turned out only the add functionality gets affected by it, because new redis key members can in fact be `<null>`. 

Meanwhile with the prepend functionality, even if the value to be prepended got to be `<null>`, we don't care, because the eventual redis key value will not be changed (either we return right away or pretend to prepend nothing). So we can return from the function right away.

Which is not the case with the adding of the new key, whereas returning right away upon having NULL pointer or 0 length, means never convert `<null>` redis keys into empty "" one. So that's the main regression noticed so far.

It turns out such modules like dialog or usrloc want `<null>` members of redis keys to be allowed.

Issue #2024, opened 2019-08-04
topos module does not update topos_d timestamp at re-INVITE


henningw:
The topos module fails to update the dialog data in the topos_d table after the processing of re-INVITEs.

I have setup a call at 11:55, and triggered a re-INVITE at 12:08. The respective topos_d table entry is not updated.

This causes the entry to expire after topos dialog_timeout value.

I would expect that the timestamp of the call in the topos_d table is updated. Is this assumption correct? If the current behavior is expected, it should documented in the module README.

No error messages from topos module in kamailio.log. Output from tables:

```
MariaDB [kamailio]>  select id, rectime, s_method, a_callid from topos_t; select id, rectime, s_method, a_callid from topos_d;
+-----+---------------------+----------+----------------------------------+
| id  | rectime             | s_method | a_callid                         |
+-----+---------------------+----------+----------------------------------+
| 305 | 2019-08-04 12:08:12 | INVITE   | q0Jeh9g49GeYdJXqcjCo-G0CW4dQ0iav |
| 306 | 2019-08-04 12:08:12 | ACK      | q0Jeh9g49GeYdJXqcjCo-G0CW4dQ0iav |
| 307 | 2019-08-04 12:08:14 | INVITE   | q0Jeh9g49GeYdJXqcjCo-G0CW4dQ0iav |
| 308 | 2019-08-04 12:08:14 | ACK      | q0Jeh9g49GeYdJXqcjCo-G0CW4dQ0iav |
+-----+---------------------+----------+----------------------------------+

+-----+---------------------+----------+----------------------------------+
| id  | rectime             | s_method | a_callid                         |
+-----+---------------------+----------+----------------------------------+
| 156 | 2019-08-04 11:55:22 | INVITE   | q0Jeh9g49GeYdJXqcjCo-G0CW4dQ0iav |
+-----+---------------------+----------+----------------------------------+
```

henningw:
$ kamailio -v
version: kamailio 5.2.3 (x86_64/linux)

miconda:
That is supposed to be the lifetime of the dialog, not updated on re-INVITE or other requests within dialog.

You can add a new param to control this behaviour if you want to change this value on requests within dialog.

henningw:
Thanks Daniel. So basically something like the dialog timeout_noreset [parameter](https://kamailio.org/docs/modules/5.2.x/modules/dialog.html#dialog.p.timeout_noreset)

Ok, I have added some documentation for now:

```
[...]

PR #4561, opened 2026-01-19
topos: Add new modparam for enabling updating rectime

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally
- [x] Related to issue #2024 (replace XXXX with an open issue number)

#### Description

Add a new modparam that enables updating topos table when invite/re-invites are received.

### DIFF
--- a/src/modules/topos/doc/topos_admin.xml
+++ b/src/modules/topos/doc/topos_admin.xml
@@ -634,6 +634,25 @@ modparam("topos", "version_table", 0)
 		</programlisting>
 		</example>
 	</section>
+	<section id="topos.p.methods_update_time">
+		<title><varname>methods_update_time</varname> (str)</title>
+		<para>
+		List of SIP methods (comma separated) for which the dialog timestamp (rectime) should be updated in the topos_d table when processing in-dialog requests. This allows extending the dialog lifetime for specific methods beyond the initial INVITE.
+		</para>
+		<para>
+		<emphasis>
+		Default value is <quote>SUBSCRIBE</quote>.
+		</emphasis>
+		</para>
+		<example>
+		<title>Set <varname>methods_update_time</varname> parameter</title>
+		<programlisting format="linespecific">
+...
+modparam("topos", "methods_update_time", "SUBSCRIBE,UPDATE,INVITE")
+...
+</programlisting>
+		</example>
+	</section>
 	</section>
 
 	<section>


henningw:
Docs will be added later, after eventual name changes or refactorings.

miconda:
As I could understand the commit, the new parameter enables/disables for both invite/subscriber at the same time. It is no longer possible to have the current behaviour (update only for subscribe).

Maybe it is better to introduce a generic parameter like `update_mode` which acts on flags/bits. If set to 0, do not update at all, if bit 1 is set update for subscriber, if bit 2 is set update for invite. Such parameter could be reused in the future if other update cases/conditions are wanted.

Update: the default value can be set to 1 to preserve current behaviour.

xkaraman:
Yes the commit as is, will enable or disable all updates. 

Ok that was one of our discussions as well but though we might go like dialog `noreset` modparam.

xkaraman:
@miconda pr updated to use flag system like the other modparams found in this module

src/modules/topos/topos_mod.c
```diff
@@ -260,6 +263,14 @@ static int mod_init(void)
...
 		}
 	}
+	if(_tps_methods_update_time_list.len > 0) {
+		if(parse_methods(
+				   &_tps_methods_update_time_list, &_tps_methods_update_time)
+				< 0) {
+			LM_ERR("failed to parse methods_update_time parameter\n");
+			return -1;
+		}
+	}
```
miconda, line 278: If the `methods_update_time` is set to empty, it is still going to match the SUBSCRIBE because of:

```
unsigned int _tps_methods_update_time = METHOD_SUBSCRIBE;
```

Is it intended to be so? No way to disable updating record time for all requests.

src/modules/topos/tps_storage.c
```diff
@@ -1854,11 +1854,12 @@ int tps_db_update_dialog(
...
-					&& ((md->direction == TPS_DIR_UPSTREAM)
-							&& (msg->first_line.type == SIP_REQUEST))
-					&& (msg->first_line.u.request.method_value
-							== METHOD_SUBSCRIBE))) {
+			&& ((sd->b_tag.len > 0) && (msg->first_line.type == SIP_REQUEST)
+					&& (((md->direction == TPS_DIR_UPSTREAM)
+								&& (msg->first_line.u.request.method_value
+										== METHOD_SUBSCRIBE))
+							|| ((msg->first_line.u.request.method_value
+									== METHOD_INVITE))))) {
```
miconda: With this condition, even the modparam can include any method (e.g., UPDATE, INFO), the record time update is still limited to INVITE and SUBSCRIBE.

As a matter of fact, I am not sure why SUBSCRIBE direction has to UPSTREAM, is it done for the response processing? Otherwise the SUBSCRIBE goes DOWNSTREAM. I haven't coded the part related to presence, and didn't look in the code right now.

Probably it can be done without direction and just bitmask-matching of the  method id with the parsed parameter value.

src/modules/topos/topos_mod.c
xkaraman: You are right i think. What if we remove the len check. Does the parse_methods allow empty string to set also the update_time?

Just to verify, if the user does not provide the modparam at all we want the old behavior of updating the SUBSCRIBE right? That was the intention of intitializing it to METHOD_SUBSCRIBE.

Otherwise, i can set to 0, and have a default_update_methods = SUBSCRIBE var that is assigned only if the user does not provide the modparam.

If the user provided the modparam, and it's empty it should respect it and update nothing, correct?

src/modules/topos/tps_storage.c
xkaraman: I agree on this, and it was one of my concerns as well. even if the user provide more methods, only those two are curretnly supported.

SUBSCRIBE was upstream before and didn't want to change the behavior from my side. 

If all methods have common UPSTREAM and DOWNSTREAM update, it would simplify the condition and allow bitmasking as well to be more extensible for other methods as well.

src/modules/topos/topos_mod.c
miconda: I would consider to check if len <= 0, then `_tps_methods_update_time = 0`. The string parameter is defaulting to `SUBSCRIBE`, so by parsing it in mod init because the len > 0, the default is to update on SUBSCRIBEs like so far.

src/modules/topos/tps_storage.c
miconda: I would say that the direction does not matter and I would propose to remove the condition on it and simplify the check to the bitmask matching.

xkaraman:
After the comments now PR should allow to:

1. no updates when provided with empty modparam. 
2. No modparam, should respect the default to be only subscribe, didn't test it properly.
3. Allow any provided method to change the time

src/modules/topos/tps_storage.c
```diff
@@ -1855,11 +1856,8 @@ int tps_db_update_dialog(
 	}
 	if((mode & TPS_DBU_TIME)
 			&& ((sd->b_tag.len > 0) && (msg->first_line.type == SIP_REQUEST)
-					&& (((md->direction == TPS_DIR_UPSTREAM)
-								&& (msg->first_line.u.request.method_value
-										== METHOD_SUBSCRIBE))
-							|| ((msg->first_line.u.request.method_value
-									== METHOD_INVITE))))) {
+					&& (msg->first_line.u.request.method_value
+							== _tps_methods_update_time))) {
```
miconda: The match on `method_value` has to be a bitwise `&`, not `==`, otherwise it won't match any if it is more than one method type in the module parameter.

xkaraman: Ohh good catch, while testing this, it somehow worked the first time (probably old binary). Fixed it.

miconda:
Docs for the new param have to be added.

xkaraman:
Also updated a core parse_methods.h doc to match the one in parse_methods.c as IDE reported different result behavior

xkaraman:
> Docs for the new param have to be added.

Added

miconda:
Changes being minimal/not-very-intrusive, I am merging it now, before branching.

Issue #4231, opened 2025-05-06
http_client leaks TLS shared memory (probably related to curl version/openssl version)


alexbakker:
### Description

Performing an HTTPS request using the ``http_client`` module results in a shared memory leak, seemingly related to TLS.

#### Reproduction

Our use case for ``http_client`` is to update an htable with the latest banned IP's from APIBAN as described here: https://github.com/apiban/apiban?tab=readme-ov-file#integration-into-kamailio. Normally, we refresh the list of banned IP's once every 15 minutes, so the shared memory usage grows fairly slowly.

The issue can be reproduced using the Kamailio configuration file below. To demonstrate the issue and make the shared memory usage grow faster, we set up a timer that calls ``http_client_query`` in a tight loop. To prevent unnecessary load on the APIBAN infrastructure, we request an example file from my personal server instead.

```
#!KAMAILIO

debug=2
log_stderror=yes
fork=yes
enable_tls=1

loadmodule "tls.so"
loadmodule "cfg_rpc.so"
loadmodule "pv.so"
loadmodule "xlog.so"
loadmodule "ctl.so"
loadmodule "rtimer.so"
loadmodule "http_client.so"

modparam("tls", "certificate", "")
modparam("tls", "private_key", "")

modparam("rtimer", "timer", "name=apiban;interval=100u;mode=1;")
modparam("rtimer", "exec", "timer=apiban;route=APIBAN;")

route[APIBAN] {
    xinfo("running apiban refresh\n");
    http_client_query("https://alexbakker.me/u/7bvjn9jfas.txt", "$var(banned)");
}
```

Start Kamailio:

```
kamailio -f kamailio.cfg -DD
```

And then watch the shared memory usage grow rapidly:

```
watch -n1 -- kamcmd core.shmmem
```

#### Log Messages

Eventually, Kamailio will run out of shared memory and print the following messages to the log:

```
21(184184) INFO: <script>: running apiban refresh
21(184184) ERROR: http_client [functions.c:471]: curL_request_url(): failed to perform curl (56) (url: https://alexbakker.me/u/7bvjn9jfas.txt)
21(184184) INFO: <script>: running apiban refresh
21(184184) ERROR: http_client [functions.c:471]: curL_request_url(): failed to perform curl (56) (url: https://alexbakker.me/u/7bvjn9jfas.txt)
21(184184) INFO: <script>: running apiban refresh
[...]

PR #4566, opened 2026-01-23
core/qm: fix bookkeeping of used and real_used in case of fragment split

nbruning:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4231

#### Description

When debugging memory leaks for Kamailio, it seems when splitting a fragment, the "used memory" counters are not updated correctly. This leads to Kamailio *indicating* less and less available memory, while in fact, it's just the bookkeeping.

Used the docker setup of #4231 for testing. Could reproduce that issue on my setup (Ubuntu Noble + Kamailio 5.8) and with this patch, problem is fixed.

Root cause: checking for `!=0` to indicate success; but `split_frag` returns 0 in case of success.


henningw:
Good catch, thanks for the PR. This could indeed explain the linked issue. Looks good to me, merged.

xkaraman:
backported to 5.8 and 6.0

PR #4572, opened 2026-01-28
uac: copy uac_flag to transaction's uas request in replace_callback

grumvalski:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

When using uac_replace_to/uac_replace_from with dialog matching, replace_callback(), after operating on the message being processed, doesn't set the flags (FL_USE_UAC_FROM | FL_USE_UAC_TO) on the trasaction's uas request. The effect is that replies to in-dialog messages don't get their headers restored.


miconda:
Thanks!

PR #4573, opened 2026-01-28
cmake: Remove man pages for modules

xkaraman:
- Outdated.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description

Remove build generation and install for man7 pages for modules and remove outdated docs the were used to generate them.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -76,7 +76,7 @@ else()
   )
 endif()
 
-option(BUILD_DOC "Build documentation" ON)
+option(BUILD_DOC "Build documentation in the default target" OFF)
 # Readme file and man page
 find_program(XSLTPROC_EXECUTABLE xsltproc QUIET)
 find_program(LYNX_EXECUTABLE lynx QUIET)


xkaraman:
The build now fails due to missing dependency for missing documentation `docbook-xsl`. We discussed it with @linuxmaniac and he will add it as build dependency for the packaging.

As the default is to build documentation the generation of READMEs fail. Should i set the build of docs default to `OFF` so it's not included in the default target?

Therefore anyone who wants to create the docs will have to provide the option `-DBUILD_DOC=ON` on the configuration phase of cmake to be included in the default target.

miconda:
The generation of the READMEs should not be done at compile/build time. It is anyhow done on kamailio.org server and pushed to github repo from there.

It is fine to provide an extra option for those that want to do it.

miconda:
I guess this can me merged.

xkaraman:
Hey @sergey-safarov and @linuxmaniac,

Just pinging you to also suggest to enable docs generation in you packaging workflows.

Since now build docs are going to be disabled by default, you must provide `-DBUILD_DOC=ON` in your cmake configuration phase to re-enable them in the default target. 

Alternatively, you can build them as a seperate `kamailio_docs` target. `cmake --build build -t kamailio_docs`.

Issue #4574, opened 2026-01-30
[BUG] xhttp_prom: xhttp_prom_tags not added to prom_counter metrics


joelsdc:
## Description

Given the following [xhttp_prom_tags](https://www.kamailio.org/docs/modules/stable/modules/xhttp_prom.html#xhttp_prom.p.xhttp_prom_tags) config:

```
modparam("xhttp_prom", "xhttp_prom_uptime_stat", 1)
modparam("xhttp_prom", "xhttp_prom_tags", "hostname=\"kamailio-test\"");
modparam("xhttp_prom", "prom_counter", "name=kamailio_test_metric_one; label=label1:label2:label3");
modparam("xhttp_prom", "prom_counter", "name=kamailio_test_metric_two; label=label1:label2");
modparam("xhttp_prom", "xhttp_prom_stats", "all")
modparam("xhttp_prom", "xhttp_prom_pkg_stats", 1)
```

When you scrape the metrics endpoint, the custom metrics (in this case: `kamailio_test_metric_one` and `kamailio_test_metric_two`)  don't include the tags configured in **xhttp_prom_tags**.

Example:
```
kamailio_test_metric_one{label1="XXX", label2="YYY", label3="ZZZ"} 3 1769758020393
kamailio_test_metric_two{label1="XXX", label2="YYY"} 68 1769758020393
kamailio_core_drop_requests{hostname="kamailio-test"} 177 1769758020393
kamailio_core_err_requests{hostname="kamailio-test"} 0 1769758020393
kamailio_core_fwd_requests{hostname="kamailio-test"} 0 1769758020393
```

Expected:
```
kamailio_test_metric_one{hostname="kamailio-test", label1="XXX", label2="YYY", label3="ZZZ"} 3 1769758020393
kamailio_test_metric_two{hostname="kamailio-test", label1="XXX", label2="YYY"} 68 1769758020393
kamailio_core_drop_requests{hostname="kamailio-test"} 177 1769758020393
kamailio_core_err_requests{hostname="kamailio-test"} 0 1769758020393
kamailio_core_fwd_requests{hostname="kamailio-test"} 0 1769758020393
```

## Reproduction

Load xhttp_prom.

Configure:

```
modparam("xhttp_prom", "xhttp_prom_uptime_stat", 1)
modparam("xhttp_prom", "xhttp_prom_tags", "hostname=\"kamailio-test\"");
[...]

PR #4575, opened 2026-01-30
xhttp_prom: include global tags in user-defined metrics

joelsdc:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #4574

#### Description

This pull request updates the Prometheus metric label printing logic in `prom_metric.c` to ensure that global tags are always included in the metric output, even when no metric-specific labels are present. The changes also improve how global tags are appended when both user-defined labels and global tags exist.

Label and tag printing improvements:

* Updated `prom_label_print` to print global tags when metric-specific labels are absent, ensuring consistent inclusion of global tags in all metrics.
* Modified both `prom_label_print` and `prom_label_print_le` to append global tags before closing the label braces, so that global tags are always present and correctly formatted. [[1]](diffhunk://#diff-3234c6053b9b93dc271b84392d97cc1c0900a4b4a5b1d62edf71bd48918449c3L1776-R1776) [[2]](diffhunk://#diff-3234c6053b9b93dc271b84392d97cc1c0900a4b4a5b1d62edf71bd48918449c3L1835-R1835)
* Refactored `prom_metric_lvalue_print` to handle the presence of user-defined labels and global tags more robustly: when both are present, labels and tags are combined properly; when only tags are present, only tags are printed in braces.

### DIFF
--- a/src/modules/xhttp_prom/prom_metric.c
+++ b/src/modules/xhttp_prom/prom_metric.c
@@ -1735,13 +1735,12 @@ static int prom_label_print(
 		goto error;
 	}
 
-	if(plval->n_elem == 0) {
-		/* Nothing to print. */
-		return 0;
-	}
-
-	if(!lb_name || lb_name->n_elem == 0) {
-		/* Nothing to print. */
+	if(plval->n_elem == 0 || !lb_name || lb_name->n_elem == 0) {
+		/* No metric-specific labels, but print global tags if present. */
+		if(prom_body_printf(ctx, "%s", xhttp_prom_tags_braces) == -1) {
+			LM_ERR("Fail to print\n");
+			goto error;
+		}
 		return 0;
 	}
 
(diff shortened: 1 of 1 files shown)


joelsdc:
I forced-pushed to apply `clang-format` changes.

henningw:
Thank you for the PR. Just to double-check, does this PR introduce some user visible changes? Or is this more like a bug fix, correcting the output that otherwise was sometimes corrupt? If the first, we probably should also add a short update to the module docs.

joelsdc:
Hi @henningw, 

Sorry for the delay! This is a bugfix, not a new user-visible feature. The `xhttp_prom_tags` parameter already documents and implies that global tags apply to all metrics, and they do work correctly for core/built-in metrics. The bug was that user-defined custom counters were silently omitting those global tags from their output.

So from a user's perspective, anyone who had configured `xhttp_prom_tags` and was relying on those tags appearing consistently across all metrics (as documented behavior would imply) was getting incorrect/incomplete output for their custom metrics. This PR restores the expected behavior.

No documentation update should be needed, the intent was always that global tags apply globally. This just makes the implementation match that intent.

joelsdc:
@vhernando I'm tagging you here in case you can take a look at this PR 🙏

henningw:
Thank you for the explanation, looks fine. I am merging it, beeing more then one months old. If there are other comments, they can be addressed directly in the repository.

joelsdc:
@henningw can you cherry-pick this into 6.1 and 6.0 branchs?

henningw:
Sure, done.

PR #4577, opened 2026-02-03
core: dns - Fix lookup of multiple NAPTR entries

kilianwww:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
Previously when doing DNS NAPTR lookup, incorrect detection of first lookup was causing only the first NAPTR entry to be checked.
This has been fixed by storing NAPTR entry information in dns_srv_handle, allowing us to correctly differentiate between first lookup and following lookups.

### DIFF
--- a/src/core/dns_cache.c
+++ b/src/core/dns_cache.c
@@ -3430,7 +3430,7 @@ inline static int dns_naptr_sip_resolve(struct dns_srv_handle *h, str *name,
 		}
 		return -E_DNS_NO_NAPTR;
 	}
-	if(((h->srv == 0) && (h->a == 0)) && /* first call */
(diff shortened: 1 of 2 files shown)


henningw:
Thanks for the PR. I did a quick review, it looked fine so far. The structure and naming of variables is a bit confusing, beeing an older part of the code, like having a srv structure that now also contains a naptr value. It should be also reviewed from another developer, e.g. @miconda before merging, I think.

miconda:
I think it is fine to have a naptr field inside the srv structure, being like the list from where the item is derived -- it is also the generic dns_hash_entry type.

I haven't implemented the dns srv/naptr/cache either, but looks ok for me as well.

henningw:
Just a note, as I am looking into some other NAPTR issues, this was so far only added to the master branch.

Issue #4325, opened 2025-07-15
stuck of processing messages on UDP localhost:5060 to TLS target on ARM arch


sem32:
### Description

The issue: in 15 second after start of kamailio, it stops to read the UDP socket localhost:5060

### Troubleshooting

<img width="683" height="406" alt="Image" src="https://github.com/user-attachments/assets/b93ed7d3-8944-477b-867b-ca7fcae832f3" />

#### Reproduction

Preconditions:
- More than one worker for UDP localhost:5060
- Kamailio receives SIP OPTION from the Asterisk by UDP localhost
- In the Kamailio's script we have a logic to send the SIP OPTION by TLS. We modify $ru and add to it ";transport=TLS"
- 25+ sip trunks which generates SIP OPTIONS with transport TLS
- all the trunks are resolvable but not reachable (don't listen the port)
- libssl3

#### Debugging Data

[gdb_239054.txt](https://github.com/user-attachments/files/21236064/gdb_239054.txt)
[gdb_239053.txt](https://github.com/user-attachments/files/21236065/gdb_239053.txt)

### Possible Solutions

We have no issue in case we have:
- only one worker for UPD localhost and any version of libssl (libssl3 or libss1.1)
- or use libssl1.1 and any amount of workers

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
# kamailio -v
version: kamailio 5.8.6 (aarch64/linux) a14e49
flags: , EXTRA_DEBUGUSE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, MEM_JOIN_FREE, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT-NOSMP, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLOCKLIST, HAVE_RESOLV_RES
ADAPTIVE_WAIT_LOOPS 1024, MAX_RECV_BUFFER_SIZE 262144, MAX_SEND_BUFFER_SIZE 262144, MAX_URI_SIZE 1024, BUF_SIZE 65535, DEFAULT PKG_SIZE 8MB
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
[...]

PR #4583, opened 2026-02-05
core: tls - preserve outbound SNI across tcp main

sem32:
We started using tcp_main_threads=1 (with tls_threads_mode=2) to avoid stuck UDP workers (the initial issue described [here](https://github.com/kamailio/kamailio/issues/4325)) under high load with a lot of TLS connections. However, now we realized (not really realized, but receive a complaint that calls are completely broken with an operator that requires SNI) that SNI is not added anymore.

#### Reproduction

We've been using the following configuration for years:

```
modparam("tls", "xavp_cfg", "tls")
...
$xavp(tls=>server_name) = "domain.com";
```

When tcp_main_threads=1 doesn't present, SNI is added to outgoing TLS connection as it should. When it's added, Kamailio stops adding SNI at all.

#### Log Messages

```
Feb 04 16:58:33 myhost kamailio[654955]: DEBUG: tls [tls_server.c:202]: tls_get_connect_server_name(): xavp with outbound server name not found
```

### Additional Information

  * **Kamailio Version** - I repeated tests using the latest Kamailio from master and it works the same

```
# kamailio -v
version: kamailio 6.2.0-dev0 (aarch64/linux) e5e0e9-dirty
flags: , EXTRA_DEBUGUSE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, MEM_JOIN_FREE, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, DBG_QM_MALLOC, DBG_F_MALLOC, DBG_TLSF_MALLOC, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLOCKLIST, HAVE_RESOLV_RES
ADAPTIVE_WAIT_LOOPS 1024, MAX_RECV_BUFFER_SIZE 262144, MAX_SEND_BUFFER_SIZE 262144, MAX_URI_SIZE 1024, BUF_SIZE 65535, DEFAULT PKG_SIZE 8MB
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
id: e5e0e9 -dirty
compiled on 13:14:23 Feb  4 2026 with gcc 12.2.0
```
* **Operating System**:
```
# lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 12 (bookworm)
Release:        12
Codename:       bookworm

# uname -a
Linux myhost 6.1.0-42-cloud-arm64 #1 SMP Debian 6.1.159-1 (2025-12-30) aarch64 GNU/Linux
```

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

In multi‑process mode (tcp_main_threads=1), XAVP list head is process‑local, so values set in worker routes are not visible in tcp main when TLS handshake occurs. This caused missing SNI even when $xavp(tls=>server_name) was set correctly. Storing the values in the shared tcp_connection resolves this while keeping behavior unchanged in single‑process mode.

Changes:
- store outbound TLS server_name/server_id from XAVP in tcp_connection (shm)
- use stored values during TLS init in tcp main
- free stored strings on connection cleanup

### DIFF
--- a/src/core/tcp_conn.h
+++ b/src/core/tcp_conn.h
@@ -30,6 +30,7 @@
 #include "tcp_options.h"
 
 #include "ip_addr.h"
+#include "str.h"
 #include "locking.h"
 #include "atomic_ops.h"
 #include "timer_ticks.h"
@@ -283,6 +284,8 @@ typedef struct tcp_connection
 	enum tcp_conn_states state; /* connection state */
 	enum tcp_conn_states initstate; /* initial connection state */
 	void *extra_data; /* extra data associated to the connection, 0 for tcp*/
+	str tls_sni; /* outbound SNI (TLS) */
+	str tls_srvid; /* outbound TLS server_id */
 	struct timer_ln timer;
 	time_t timestamp; /* connection creation timestamp */
 	ticks_t timeout;  /* connection timeout, after this it will be removed*/
--- a/src/modules/tls/tls_server.c
+++ b/src/modules/tls/tls_server.c
@@ -130,6 +130,75 @@ extern str sr_tls_xavp_cfg;
 
 static str _ksr_tls_connect_server_id = STR_NULL;
 
+static int tls_shm_str_set(str *dst, const str *src)
+{
+	if(dst == NULL) {
+		return -1;
+	}
+	if(dst->s) {
+		shm_free(dst->s);
+		dst->s = NULL;
+		dst->len = 0;
+	}
+	if(src == NULL || src->s == NULL || src->len <= 0) {
+		return 0;
+	}
+	dst->s = (char *)shm_malloc(src->len + 1);
+	if(dst->s == NULL) {
+		SHM_MEM_ERROR;
+		return -1;
+	}
+	memcpy(dst->s, src->s, src->len);
+	dst->s[src->len] = '\0';
+	dst->len = src->len;
+	return 0;
+}
+
+static void tls_store_outbound_xavp(struct tcp_connection *c)
+{
+	sr_xavp_t *vavp = NULL;
+	str sname = {"server_name", 11};
+	str sid = {"server_id", 9};
+
+	if(c == NULL || (c->flags & F_CONN_PASSIVE)) {
+		return;
+	}
+	if(sr_tls_xavp_cfg.s == NULL) {
+		return;
+	}
+
+	LM_DBG("tls: store outbound xavp (root=%.*s pid=%d proc=%d %s conn=%p)\n",
+			sr_tls_xavp_cfg.len, sr_tls_xavp_cfg.s, my_pid(), process_no,
+			my_desc(), c);
+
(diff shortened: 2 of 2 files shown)


henningw:
You should not get crashes on TLS with the tls_thread_mode = 2 set, this should the OpenSSL issue usually. Can you provide more context about the crashes, e.g. in an issue?

I am not sure if adding TLS specific information to a TCP core structure is the right way to go, it would be probably better to solve it in the tls module.
I also don't understand your description, first you are describing the problem occurs with tcp_main_threads = 1, but later you are saying it happens in tcp_main_thread = 0.

sem32:
> You should not get crashes on TLS with the tls_thread_mode = 2 set, this should the OpenSSL issue usually. Can you provide more context about the crashes, e.g. in an issue?
> 
> I am not sure if adding TLS specific information to a TCP core structure is the right way to go, it would be probably better to solve it in the tls module. I also don't understand your description, first you are describing the problem occurs with tcp_main_threads = 1, but later you are saying it happens in tcp_main_thread = 0.

You’re right - that was my mistake. I’ve updated the description and added a link to the original issue (UDP workers getting stuck due to an OpenSSL deadlock). BTW the issue was reproduced also in AMD platform. With tcp_main_thread=1, we don’t observe OpenSSL deadlocks. However, SNI extension via `$xavp(tls=>server_name) = "domain.com";` doesn’t work in this mode, so the TLS handshake fails.

miconda:
Thanks for the report and PR! I will review and eventually push the changes manually or do an alternative is considered to suit better. Next time be sure you follow the contributing guidelines, so the commits are per component and the commit message is formatted accordingly:

 - https://github.com/kamailio/kamailio/blob/master/.github/CONTRIBUTING.md#contributing-code-or-content

miconda:
I pushed a slightly different alternative implementation, trying to reuse some core function for str shm duplicate, and added the new fields in the `ksr_coninfo_t` (`cinfo` field), instead of directly in the tcp connection structure.

Can you test? If not working, open an issue. The commits will be backported to 6.1 branch.

sem32:
> I pushed a slightly different alternative implementation, trying to reuse some core function for str shm duplicate, and added the new fields in the `ksr_coninfo_t` (`cinfo` field), instead of directly in the tcp connection structure.
> 
> Can you test? If not working, open an issue. The commits will be backported to 6.1 branch.

sure, I'll test and let you know. Thank you!

sem32:
@miconda, We’ve tested your changes and confirmed they work. Thank you!

PR #4584, opened 2026-02-06
jwt3: initial implementation for libjwt v3.2.0+

wkampich:
- ported functionality from the jwt module to support libjwt v3.2.0+ API changes
- added support for new JWT algorithms available in the latest library
- removed key_mode and added leeway_sec parameter
- included documentation

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR introduces jwt3, a new module providing JSON Web Token support using libjwt version 3.2.0 or higher.

The existing jwt module relies on older versions of the library; however, libjwt v3.2.0 introduced significant API changes and new features that are not backward compatible. jwt3 was created to support the latest  library version.

Changes
- Library Compatibility: Ported core functionality to support the libjwt v3.2.0+ API.
- Algorithm Support: Added support for JWT signing/verification algorithms made available in the latest library releases.
- Parameter Updates:
  - Removed the `key_mode` parameter.
  - Added a new `leeway_sec` parameter to allow for clock drift during token validation (sets the global leeway for exp and nbf claims).
- Enhanced Key Management: libjwt v3.2.0+ natively uses JWK/JWKS formats, this module includes an OpenSSL 3-based conversion layer to provide seamless backward compatibility by allowing keys to be loaded via:
  - PEM strings (direct configuration parameters)
  - PEM files (legacy public/private key files)
  - JWK/JWKS files (native JSON format)
- KEMI Support: Export of functions to the Kamailio Embedded Interface (KEMI).
- Documentation: DocBook documentation included, covering all exported functions and parameters.

Testing
- Verified compilation on Alpine 3.19.
- Validated against libjwt v3.2.3.
- Tested jwt3_verify() and jwt3_generate() using e.g. ES256 algorithms.

```
    jwt3_generate("/etc/kamailio/certs/ec_key.pem", "ES256", "caller='abcd';callee='xyz';callid='abc123';index=100", "kid=wk123");
    xlog("L_INFO", "$jwt3(val)\n");

    if(!jwt3_verify("/etc/kamailio/certs/ec_key_pub.pem", "ES256", "caller='abcd';callee='xyz';callid='abc123';index=100", "$jwt3(val)")) {
        xlog("L_WARN", "INVALID -\n");
        xlog("L_WARN", "status: $jwt3(status)\n");
    } else {
        xlog("L_INFO", "VALID ---\n");
        xlog("L_INFO", "status: $jwt3(status)\n");
    }
```

### DIFF
--- a/src/modules/jwt3/Makefile
+++ b/src/modules/jwt3/Makefile
@@ -1,0 +1,25 @@
+#
+# 
+# WARNING: do not run this directly, it should be run by the main Makefile
+
+include ../../Makefile.defs
+auto_gen=
+NAME=jwt3.so
+
+ifeq ($(CROSS_COMPILE),)
+JWT_BUILDER=$(shell \
+	if pkg-config --exists libjwt; then \
+		echo 'pkg-config libjwt'; \
+	fi)
+endif
+
+ifneq ($(JWT_BUILDER),)
+	DEFS += $(shell $(JWT_BUILDER) --cflags)
+	LIBS += $(shell $(JWT_BUILDER) --libs)
+else
+	DEFS += -I$(LOCALBASE)/include
+	LIBS += -L$(LOCALBASE)/lib -ljwt
+endif
+
+include ../../Makefile.modules
+
--- a/src/modules/jwt3/doc/Makefile
+++ b/src/modules/jwt3/doc/Makefile
@@ -1,0 +1,4 @@
+docs = jwt3.xml
+
+docbook_dir = ../../../../doc/docbook
+include $(docbook_dir)/Makefile.module
--- a/src/modules/jwt3/doc/jwt3.xml
+++ b/src/modules/jwt3/doc/jwt3.xml
@@ -1,0 +1,45 @@
+<?xml version="1.0" encoding='ISO-8859-1'?>
+<!DOCTYPE book PUBLIC "-//OASIS//DTD DocBook XML V4.4//EN"
+"http://www.oasis-open.org/docbook/xml/4.4/docbookx.dtd" [
+
+<!-- Include general documentation entities -->
+<!ENTITY % docentities SYSTEM "../../../../doc/docbook/entities.xml">
+%docentities;
+
+]>
+
+<book xmlns:xi="http://www.w3.org/2001/XInclude">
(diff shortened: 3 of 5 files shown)


henningw:
Thank you for the PR. Before doing more reviews, could you please do a clang-format on the code to adjust the formatting? Just do a force-push to this PR to update it then.

sergey-safarov:
Please before merge let me check also build on Alpine and RPM build on different arch.

@wkampich could you also clarify.
Is the module API which can be used in Kamailio config equal old module API?
Should we package both modules (old and new) or we can just package new module version?

P.S. I see module use `jw3_` prefix in functions names. It looks like need to package both module, old and new.

wkampich:
> Thank you for the PR. Before doing more reviews, could you please do a clang-format on the code to adjust the formatting? Just do a force-push to this PR to update it then.

Done. I've re-run clang-format and force-pushed the changes.

wkampich:
> Please before merge let me check also build on Alpine and RPM build on different arch.
> 
> @wkampich could you also clarify. Is the module API which can be used in Kamailio config equal old module API? Should we package both modules (old and new) or we can just package new module version?
> 
> P.S. I see module use `jw3_` prefix in functions names. It looks like need to package both module, old and new.

The module API can be used in Kamailio config equal to old module API. Please note that the `key_mode` parameter has been removed.

Unlike the legacy jwt module, jwt3 does not statically cache key objects. This aligns with libjwt v3.2+'s native support for JWK Sets (JWKS) and key rotation. Since keys are selected dynamically based on the token's kid (Key ID) header, strict static caching would not support key rotation events.

wkampich:
I noticed the CI build failed. I suspect this is because the CI runners have no version of libjwt installed, and my current Makefile does not correctly skip compilation when the version is < 3.2.0 or the library is missing.

@henningw: I’ve left the Makefile as-is for your review. Could you advise on the preferred way to exclude the module from the build if the correct libjwt version isn't found?

miconda:
Thanks! I am merging, it has 2 weeks here -- further updates can be done in the git repo.

henningw:
It needs to also get a cmake file.

PR #4588, opened 2026-02-13
remove makefiles

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally


sergey-safarov:
make files do not used for rpm packaging and alpine image.

miconda:
I would prefer to go for a two stage removal of old makefiles, first relocate them in the source tree, maybe `misc/makefiles/` with the same directory structure, then at a later time remove that folder.

The main reason is that right now the server side cronjobs/scripts/tasks rely on makefiles and migrating them can take some time. Having a way to quickly and temporarily get the old makefiles is most probably going to help. I can take care of creating the scripts to relocate in git with proper directory strucutre (and also restore when needed, just on local clone) after the release of 6.1.0.

xkaraman:
> I would prefer to go for a two stage removal of old makefiles, first relocate them in the source tree, maybe `misc/makefiles/` with the same directory structure, then at a later time remove that folder.

I dont mind this but i dont think it will be straightforward to just run the makefiles from the new folder, since makefiles have hardcoded folders i believe. Only if they were re-moved back to their original path they will work.

> The main reason is that right now the server side cronjobs/scripts/tasks rely on makefiles and migrating them can take some time. Having a way to quickly and temporarily get the old makefiles is most probably going to help. I can take care of creating the scripts to relocate in git with proper directory strucutre (and also restore when needed, just on local clone) after the release of 6.1.0.

I see... I can of course help migrating those cron/scripts to proper cmake when possible. It would be good to adopt them sooner than later, for more testing!

xkaraman:
Hey @linuxmaniac,

Regarding the build fail on gcc, it seems to be due to https://github.com/kamailio/kamailio/blob/6ca2cf813d38f17a24307def8a0ddff1cb9b0910/test/travis/build_travis.sh#L31-L38

Is there a reason why clang does not install or why gcc needs to install it?

miconda:
@xkaraman: like I said before in the comment above, just removing the makefiles is not the option I am looking at right now. Not to waste more time, you should not focus on it and  do not add more related commits, this PR is likely not to merged.

I see some cmake commits, probably not related to makefiles -- they should be done separately.

xkaraman:
Yes the cmake commits will be pushed seperately when confirm they working with this way.

Sure, but for this PR it's good to have them removed completely to be sure that is not using them in any way by mistake.

henningw:
> @xkaraman: like I said before in the comment above, just removing the makefiles is not the option I am looking at right now. Not to waste more time, you should not focus on it and do not add more related commits, this PR is likely not to merged.

@miconda We don't need to rush it, if the adaption of the web server scripts takes some weeks, we remove it after it was finished. Keeping over 500 Makefiles temporary somewhere in the code base, and writing even some new scripts to copy them around just because a few scripts are not fully migrated does not sound like the easiest way from the outside. We don't know all the details about the scripts of course. Xenofon is working on getting the last missing cmake parts in place right now that all make targets should have an equivalent cmake target. And as said, if there is some help possible for the scripts, just let us know.

xkaraman:
> Hey @linuxmaniac,
> 
> Regarding the build fail on gcc, it seems to be due to
> 
> https://github.com/kamailio/kamailio/blob/6ca2cf813d38f17a24307def8a0ddff1cb9b0910/test/travis/build_travis.sh#L31-L38
> 
> Is there a reason why clang does not install or why gcc needs to install it?

https://bugs-devel.debian.org/challenge.html?original=%2fcgi-bin%2fbugreport.cgi%3fbug%3d1016936 this is probably the reason.

@miconda Are the doc/tutorial stuff used in any way? I dont see them published anywhere and the makefile seem to fail to build as well due to deprecate options.

linuxmaniac:
> Hey @linuxmaniac,
> 
> Regarding the build fail on gcc, it seems to be due to
> 
> https://github.com/kamailio/kamailio/blob/6ca2cf813d38f17a24307def8a0ddff1cb9b0910/test/travis/build_travis.sh#L31-L38
> 
> Is there a reason why clang does not install or why gcc needs to install it?

The reason behind was to check the install on ``gcc`` build only. The install phase was independent of the compiler used. No need to test it twice.

linuxmaniac:
Alarm Alarm... those commits are now in master!! @xkaraman

xkaraman:
Oh damn... only the last one was intended for this. sorry for this. let me start reverting

xkaraman:
Revert the bad commits. sorry for this! thanks for the alert!

henningw:
The oss-fuzz was moved to cmake. This PR needs an update/rebase due to the conflicts.

Issue #4592, opened 2026-02-16
dlgs module don't track forked INVITE requests that are send in short time intervals


henningw:
### Description

The dlgs module is used on a load-balancer class system for traffic statistics. Another system sends INVITE requests to the load-balancer, which are balanced to different carriers. If carrier A rejects the traffic, the traffic source creates another INVITE (basically in a serial forking scenario) which only differs from the branch-id to the previous INVITE. The load-balancer forwards this to carrier B, which accept the call.

The dlgs module will not track the second call correctly, as the first calls is already in the list and not yet expired due to the timer. Setting the timer to e.g. 1s will not solve it, as the serial forking happens faster if the first carrier just rejects the call for example.. This issue will misreporting of caller statistics, as established calls which tooks one or more tries to get setup are not tracked anymore.

Sometimes the call is accepted only at the third or fourth attempt, but the problem is basically the same.

#### Reproduction

Send serial forked INVITE calls to a system with dlgs enabled.

#### Log Messages

```
Feb 13 18:23:26 lab.WWW.com /usr/local/sbin/kamailio[1675553]: DEBUG: dlgs [dlgs_records.c:321]: dlgs_add_item(): call-id already in hash table [d3c5ba80-04b8-4d5c-a249-50220ab6eea5].
Feb 13 18:23:26 lab.WWW.com /usr/local/sbin/kamailio[1675553]: DEBUG: dlgs [dlgs_mod.c:191]: ki_dlgs_init(): added item return code: 1
```

#### SIP Traffic

Examples for two SIP messages, IPs anonymized below. The log messages quoted above is from another call (call-id), but the patters is the same. The only difference of the INVITEs is the branch-id in the first Via header.

```
INVITE sip:+123YYY@172.0.0.3 SIP/2.0
Record-Route: <sip:172.0.0.3;lr;ftag=c75ecd74-6610-46bb-9258-3ce68105aa25>
Record-Route: <sip:172.0.0.2;lr=on;ftag=c75ecd74-6610-46bb-9258-3ce68105aa25;dlgcor=748c2.b0b1>
Record-Route: <sip:172.0.0.1;r2=on;lr;ftag=c75ecd74-6610-46bb-9258-3ce68105aa25>
Record-Route: <sip:Z.Z.Z.Z;r2=on;lr;ftag=c75ecd74-6610-46bb-9258-3ce68105aa25>
[...]

PR #4595, opened 2026-02-20
dlgs: use branch id for tracking as well for parallel/serial forking support

xkaraman:
- Use branch member to identify same call-id messages

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4592

### DIFF
--- a/src/modules/dlgs/doc/dlgs_admin.xml
+++ b/src/modules/dlgs/doc/dlgs_admin.xml
@@ -21,7 +21,9 @@
 	</para>
 	<para>
 		The dialogs can be tagged and the number of dialogs with the same tag
-		can be retrieved in configuration file.
+		can be retrieved in configuration file. Serial forking is supported,
+		but no spiraled calls or parall forking. It will track the first call
+		attempt and ignore eventual later attempts with the same Call-ID.
 	</para>
 	<para>
 		The module aims to be a lightweight alternative to dialog module, to


henningw:
A new version will be pushed soon, there are some issues in this draft.

src/modules/dlgs/dlgs_records.c
```diff
@@ -89,6 +89,7 @@ int dlgs_destroy(void)
 int dlgs_sipfields_get(sip_msg_t *msg, dlgs_sipfields_t *sf)
 {
 	memset(sf, 0, sizeof(dlgs_sipfields_t));
+	via_body_t *via1 = msg->h_via1 ? (msg->h_via1)->parsed : NULL;
```
miconda: Variables should be declared at the beginning of functions or blocks.

src/modules/dlgs/dlgs_records.c
```diff
@@ -423,6 +463,7 @@ dlgs_item_t *dlgs_get_item(sip_msg_t *msg)
 	while(it != NULL && it->hashid == hid) {
 		if(cid->len == it->callid.len
 				&& strncmp(cid->s, it->callid.s, cid->len) == 0) {
+			/* Don't check branch-id, spiraling or parallel forking not supported */
```
miconda, line 467: The comment is confusing, I guess that if an item is added because it has same call-id, but different Via branch, it is because of serial/parallel forking. Or is it added if a previous dlgs item is marked as terminated due to a >=300 code of a prior serial step?

henningw: It was added to indicate that for the dlgs_get_item the Via branch id will be not checked. If it would be checked, it would break the tracking for spiralled calls according to our tests. But we can of course rephrase it.

src/modules/dlgs/dlgs_records.c
henningw: it was adressed

henningw:
Any more changes or comments, or could it be merged? The one discussed comment above can be rephrased of course.

miconda:
Fine for me.

PR #4611, opened 2026-03-01
ims_qos: fix infinite loop in case of video call downgraded to audio call

herlesupreeth:
I accidentally introduced an infinite loop in this PR - https://github.com/kamailio/kamailio/pull/4522 which caused the downgrade of video call to audio to fail. This PR addresses the following fixes:
1. Comparing whether flow description already exists without considering dialog direction (compares purely based on IP and port and media type)
2. Fix the infinite loop by moving the incrementing of sdp_stream_num outside of if condition

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally


linuxmaniac:
backported to 6.1 too. Thanks!

PR #4625, opened 2026-03-05
add udp_accept_proxy feature

rfuchs:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
(I've left the `su2a()` change in a separate commit even though it's also for the core, as it really is a separate change)

This adds a new core variable `udp_accept_proxy` which serves a purpose similar to `tcp_accept_haproxy` but for UDP.

Two protocols are supported: The HAproxy v2 protocol (v1 is not available for UDP), and the "simple" proxy protocol used by CloudFlare. Bit value 1 enables the HAproxy protocol, bit value 2 enables the "simple" protocol, and so value 3 enables both. The default is zero which disables the support and leaves existing behaviour unchanged.

If a proxy protocol header is detected on a received UDP packet, the reported "real" address is extracted and substituted as source address for this UDP packet.

Since UDP is connectionless, sending responses or any other packets in the opposite direction requires the reverse operation (substituting the real address for the proxy peer address). For this purposes, real-to-proxy mappings and proxy-to-real mappings are kept in a hash table. This is hooked into `udp_send()` when either protocol is enabled.

The proxy-to-real mapping hash table is needed because the HAproxy protocol (optionally) only sends the HAproxy header in the first packet of a UDP flow. Therefore, for any subsequent packet, substitution of the real address must be done from the hash table based on the proxy source address.

Each hash table entry has a "last accessed" time stamp and entries older than 2 hours are removed when encountered during a lookup.

Four new RPC commands are included, one to dump the full contents of the hash tables, one to flush out all entries and reset them to empty, one to just remove expired entries, and one to give a stats overview.

Protocol refs:
https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt
https://developers.cloudflare.com/spectrum/reference/simple-proxy-protocol-header/

### DIFF
--- a/src/core/cfg.lex
+++ b/src/core/cfg.lex
@@ -172,6 +172,7 @@ UDP_RECEIVER_MODE "udp_receiver_mode"
 UDP4_RAW		"udp4_raw"
 UDP4_RAW_MTU	"udp4_raw_mtu"
 UDP4_RAW_TTL	"udp4_raw_ttl"
+UDP_ACCEPT_PROXY	"udp_accept_proxy"
 SETFLAG		setflag
 RESETFLAG	resetflag
 ISFLAGSET	isflagset
@@ -710,6 +711,7 @@ IMPORTFILE      "import_file"
 <INITIAL>{UDP4_RAW_MTU}	{ count(); yylval.strval=yytext; return UDP4_RAW_MTU; }
 <INITIAL>{UDP4_RAW_TTL}	{ count(); yylval.strval=yytext; return UDP4_RAW_TTL; }
 <INITIAL>{UDP_RECEIVER_MODE}	{ count(); yylval.strval=yytext; return UDP_RECEIVER_MODE; }
+<INITIAL>{UDP_ACCEPT_PROXY}	{ count(); yylval.strval=yytext; return UDP_ACCEPT_PROXY; }
 <INITIAL>{IF}	{ count(); yylval.strval=yytext; return IF; }
 <INITIAL>{ELSE}	{ count(); yylval.strval=yytext; return ELSE; }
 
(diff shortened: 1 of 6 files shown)


miconda:
Having it controlled by a core parameter is good, therefore I haven't done right now a deep analysis, but the returned buffer that is linked inside sip_msg_t has to be at least BUF_SIZE in size, because some functions expects that. Because of that, the raw_buf should BUF_SIZE + max(haproxy headers size) + 1.

Also, to cope with some older C specs/older OSes, the variables should be declared at the beginning of the block/function, in the PR, the `buf` was replaced by `raw_buf` and `buf` declared later in the middle of the function.

rfuchs:
> Having it controlled by a core parameter is good, therefore I haven't done right now a deep analysis, but the returned buffer that is linked inside sip_msg_t has to be at least BUF_SIZE in size, because some functions expects that. Because of that, the raw_buf should BUF_SIZE + max(haproxy headers size) + 1.

Ok wasn't aware of that. Fixed.

> Also, to cope with some older C specs/older OSes, the variables should be declared at the beginning of the block/function, in the PR, the `buf` was replaced by `raw_buf` and `buf` declared later in the middle of the function.

Good catch, I tried to keep that in mind, but missed that instance. Would be nice to have that as a build check somewhere 😀

miconda:
Fine for me to merge.

rfuchs:
@miconda  I suppose the docs for the new RPC commands should go into the `kex` docs?

miconda:
Yes, kex if fine -- it was decided to document them in a module to be able to index them.

The new core param will be in the wiki core cookbook.

Issue #4534, opened 2025-12-20
tls: coredump on Alpine


sergey-safarov:
### Description

On Alpine Linux th `tls` module load is crashing Kamailio with a coredump
```
#0  0x00007fcbdf45f8d2 in __pthread_getspecific (k=0) at src/thread/pthread_getspecific.c:7
        self = 0x7fcbdf4ac588 <builtin_tls+136>
#1  0x00007fcbdeb6201c in mod_init () at tls_mod.c:406
        method = <optimized out>
        verify_client = <optimized out>
        rand_buf = "\270\f=\337\313\177\000\000\230X\201\203\000\001\000\000\000\001\000\000\0011\0010\00217\003172\a"
        k = 0
        __func__ = "mod_init"
        __llevel = <optimized out>
        __kld = <optimized out>
#2  0x00005557fdcdf2a7 in init_mod (m=0x7fcbded4a390) at core/sr_module.c:1037
        ret = <optimized out>
        __func__ = "init_mod"
#3  0x00005557fdcdf060 in init_mod (m=0x7fcbded4d9d0) at core/sr_module.c:1032
        ret = <optimized out>
        __func__ = "init_mod"
#4  0x00005557fdcdf060 in init_mod (m=0x7fcbded4e380) at core/sr_module.c:1032
        ret = <optimized out>
        __func__ = "init_mod"
#5  0x00005557fdcdf060 in init_mod (m=0x7fcbded4f100) at core/sr_module.c:1032
        ret = <optimized out>
        __func__ = "init_mod"
#6  0x00005557fdcdf060 in init_mod (m=0x7fcbded51380) at core/sr_module.c:1032
        ret = <optimized out>
        __func__ = "init_mod"
#7  0x00005557fdcdf060 in init_mod (m=0x7fcbded51b90) at core/sr_module.c:1032
        ret = <optimized out>
        __func__ = "init_mod"
#8  0x00005557fdcdf060 in init_mod (m=0x7fcbded52af0) at core/sr_module.c:1032
        ret = <optimized out>
        __func__ = "init_mod"
#9  0x00005557fdcdf060 in init_mod (m=0x7fcbded5d010) at core/sr_module.c:1032
        ret = <optimized out>
        __func__ = "init_mod"
#10 0x00005557fdcdf060 in init_mod (m=0x7fcbded5d650) at core/sr_module.c:1032
        ret = <optimized out>
        __func__ = "init_mod"
[...]

PR #4628, opened 2026-03-08
tls: fix segfault on musl libc due to uninitialized pthread TSD

joeygo:
## Summary

The diagnostic loop in `mod_init()` at [tls_mod.c:413](https://github.com/kamailio/kamailio/blob/6.1.1/src/modules/tls/tls_mod.c#L413) calls `pthread_getspecific()` with raw integer keys 0–31. On musl libc (Alpine Linux), the main thread's pthread TSD array (`self->tsd`) is `NULL` until `pthread_key_create()` is called at least once, so `pthread_getspecific()` dereferences a NULL pointer → segfault.

This loop runs unconditionally before any `tls_threads_mode` or `tls_disable` check, so no configuration setting can prevent the crash.

### Fix

Create and immediately delete a pthread key before the diagnostic loop to force TSD initialization. This is a no-op on glibc and resolves the crash on musl.

### Reproduction

```
docker run --rm --entrypoint sh ghcr.io/kamailio/kamailio-ci:6.1.1-alpine -c '
cat > /tmp/tls_test.cfg << "CFGEOF"
enable_tls = 1
tcp_main_threads = 1

loadmodule "sl.so"
loadmodule "tm.so"
loadmodule "tls.so"

listen=tls:0.0.0.0:5061

request_route {
  sl_send_reply("200", "OK");
}
CFGEOF
kamailio -f /tmp/tls_test.cfg -DD -E; echo "EXIT CODE: $?"'
```

Output: `EXIT CODE: 139` (SIGSEGV)

### References

- kamailio/kamailio-ci#20
- #4534
- [musl mailing list analysis](https://www.openwall.com/lists/musl/2025/12/21/2) — Rich Felker (musl author) confirms iterating raw integers through `pthread_getspecific` is undefined behavior

### DIFF
--- a/src/modules/tls/tls_mod.c
+++ b/src/modules/tls/tls_mod.c
@@ -410,6 +410,11 @@ static int mod_init(void)
 	}
 #endif
 #if OPENSSL_VERSION_NUMBER >= 0x10101000L
+	{
+		pthread_key_t _ksr_tsd_init;
+		pthread_key_create(&_ksr_tsd_init, NULL);
+		pthread_key_delete(_ksr_tsd_init);
+	}
 	for(k = 0; k < 32; k++) {
 		if(pthread_getspecific(k) != 0) {
 			LM_WARN("detected initialized thread-locals created before tls.so; "


miconda:
Thanks!

Issue #4626, opened 2026-03-06
TM: Memory leak in 6.1.1


whosgonna:
### Description

There appears to be a memory leak in Kamailio 6.1.1's TM module (although perhaps it originates in the Dispatcher module).  

#### Reproduction

I've created a trivial example of a server using the dispatcher module to send `OPTIONS` keep alive messages and nothing else. Given no additional traffic, the shared memory used by the TM module will continually decrease.

#### Debugging Data

I've created a docker compose project to reproduce the issue in a trivial example as described above.

https://github.com/whosgonna/6.1.1_memleak

`kamcmd mod.mem_stats tm all` can be used to observe the memory steadily increasing over time.

#### Log Messages

No accompanying log message

#### SIP Traffic

I'm not sure if this can occur with traffic other than the SIP OPTIONS, but from what I observed in our production environment the rate of the leak was somewhat slow overnight (when traffic volume was low) and increased during business hours, so I'm assuming it's any transactions.

### Possible Solutions

```
version: kamailio 6.1.1 (aarch64/Linux)
flags: USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, MEM_JOIN_FREE, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, DBG_QM_MALLOC, DBG_F_MALLOC, DBG_TLSF_MALLOC, USE_PTHREAD_MUTEX, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLOCKLIST, HAVE_RESOLV_RES, TLS_PTHREAD_MUTEX_SHARED
MAX_RECV_BUFFER_SIZE 262144, MAX_SEND_BUFFER_SIZE 262144, MAX_URI_SIZE 1024, BUF_SIZE 65535, DEFAULT PKG_SIZE 8MB
poll method support: poll.
id: unknown
compiled with gcc 14.2.0
```

### Additional Information

  * **Kamailio Version** - output of `kamailio -v`

```
(paste your output here)
```

* **Operating System**:

Reproduced using the 6.1.1 packages from deb.kamailio.org running on Debian 13 containers on arm64 hosts - both Amazon linux 2023 and Macintosh (arm64 CPU).

henningw:
[...]

PR #4638, opened 2026-03-10
core: fix swapped comparison in atomic_stdatomic.h dec/inc_and_test

NormB:
## Summary

- Fix swapped comparison values in `atomic_stdatomic.h` `dec_and_test` and `inc_and_test` functions
- `atomic_fetch_sub_explicit()` returns the **pre-decrement** value, so testing "result == 0" requires comparing the old value against `1`, not `-1`
- The symmetric bug exists in `inc_and_test` (compares `== 1` instead of `== -1`)
- All 8 `_and_test` functions are fixed (int/long × relaxed/seq_cst)

### Impact

This header is only used on aarch64 (ARM64). The bug causes `atomic_dec_and_test` to always return false, so any refcount-guarded `free()` never fires. Affected subsystems include TM transactions, DNS cache, TCP connections, config framework, and WebSocket connections — all leak shared memory monotonically on aarch64.

### Reproduction

Reproduced on Apple Silicon (aarch64) using Docker with @whosgonna's [reproduction project](https://github.com/whosgonna/6.1.1_memleak). Unpatched 6.1.1 leaks ~175 KB/30s (linear, never returns to baseline). Patched 6.1.1 is stable with zero sustained growth.

Fixes: #4626

## Test plan

- [x] Reproduced leak on unpatched 6.1.1 aarch64 (20 samples, linear growth)
- [x] Verified patched 6.1.1 aarch64 has zero sustained memory growth (14 samples, returns to baseline)
- [x] Verified x86 semantics match: `decl` + `setz` tests post-decrement ZF (result == 0)
- [x] Verified `atomic_unknown.h` semantics match: `ret = --(*var); return (ret == 0)`
- [x] Confirmed all 8 _and_test functions have the same bug pattern

### DIFF
--- a/src/core/atomic/atomic_stdatomic.h
+++ b/src/core/atomic/atomic_stdatomic.h
@@ -84,14 +84,14 @@ inline static int atomic_inc_and_test_int(volatile int *var)
 {
 	return atomic_fetch_add_explicit(
 				   (_Atomic int *)var, 1, memory_order_relaxed)
-		   == 1;
+		   == -1;
 }
 
 inline static int atomic_dec_and_test_int(volatile int *var)
 {
 	return atomic_fetch_sub_explicit(
 				   (_Atomic int *)var, 1, memory_order_relaxed)
-		   == -1;
+		   == 1;
 }
 
 inline static int atomic_get_and_set_int(volatile int *var, int v)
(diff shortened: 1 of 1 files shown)


miconda:
This should be ok.

miconda:
@n2p-litvinov: do you want to double check?

henningw:
There seems to be no more comments, merge it for better testing.

PR #4640, opened 2026-03-12
RFC: tls - remove shared memory and pthreads shims in multi-threaded mode

space88man:
#### Pre-Submission Checklist

- [X ] Commit message has the format required by CONTRIBUTING guide
- [X ] Each component has a single commit (if not, squash them into one commit)

#### Type Of Change

#### Checklist:

- [X ] Tested changes locally

#### Description
This PR is based on the multi-threaded RPC OpenSSL work done by @miconda and released in 6.1. In that work - there are still many accessors (pv, sel) that peek into shared memory SSL* - that is why shared memory was retained.

This PR provides the next step for multi-threaded RPC OpenSSL
- disable CRYPTO_set_mem_functions - no more shared memory for OpenSSL(!)
- disable pthreads PROCESS_SHARED fixups(!)

To provide SSL metadata to all workers - the metadata is eagerly cached in the tcp connection; workers access the metadata (e.g. SSL_get_servername) as strings or bytes and do not peek at SSL* themselves. Based on looking through pv/sel - the following is cached

```
    tls_c->ssl_servername 
    tls_c->ssl_cipher_name
    tls_c->ssl_my_cert
    tls_c->ssl_peer_cert
    tls_c->ssl_cert_chain
    tls_c->ssl_verify_result
    tls_c->ssl_cipher_bits
    tls_c->ssl_version
    tls_c->ssl_cipher_desc
```
- this is sufficient to cover all known pv/sel and module functions.

This PR also provides the single SSL_CTX implementation - instead of being duplicated across all workers.

Update:
- this PR does not remove shared memory mode which is configured by  `ksr_tcp_main_threads = 0`. The change that affects both modes is sharing of `SSL  *` metadata by caching into serialized formats. The is due to  `X509` and `STACK_OF(X509)` being opaque types by OpenSSL design — `sizeof(X509)` is unavailable and they contain process-private heap pointers. In this PR they are serialized to DER format for storage in shm.
-  the rpc `tls.reload` is implemented with `tcp_mtops`

**Limitations**

- eager caching - if the routing script does not access pv/sel it is waste of memory: TODO: replace eager caching with lazy caching - in this case we would need an RPC to PROC_TCP_MAIN for metadata
- ~~RPC tls.reload: this currently cannot execute as it is in the JSONRPC process; TODO: use an RPC to PROC_TCP_MAIN for reload~~

Co-authored-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/modules/tls_wolfssl/tls_domain.c
+++ b/src/modules/tls_wolfssl/tls_domain.c
@@ -155,7 +155,7 @@ void tls_free_domain(tls_domain_t *d)
 {
 	if(!d)
 		return;
-	if(d->ctx && ksr_tcp_main_threads != 2) {
+	if(d->ctx && ksr_tcp_main_threads == 0) {
 		do {
 			if(d->ctx[0])
 				wolfSSL_CTX_free(d->ctx[0]);
--- a/src/modules/tls_wolfssl/tls_init.c
+++ b/src/modules/tls_wolfssl/tls_init.c
@@ -371,7 +371,7 @@ int tls_pre_init(void)
 	mf = NULL;
 	rf = NULL;
 	ff = NULL;
-	if(ksr_tcp_main_threads != 2
+	if(ksr_tcp_main_threads == 0
 			&& wolfSSL_SetAllocators(ser_malloc, ser_free, ser_realloc)) {
 		LM_ERR("Unable to set the memory allocation functions\n");
 		// CRYPTO_get_mem_functions(&mf, &rf, &ff);
--- a/src/modules/tls_wolfssl/tls_util.c
+++ b/src/modules/tls_wolfssl/tls_util.c
@@ -37,8 +37,7 @@
 
 /* compatibility for < v5.5.2-stable */
 #if LIBWOLFSSL_VERSION_HEX < 0x05005002
-    /* In 5.2.0, 'new' takes no args and 'new_null' doesn't exist */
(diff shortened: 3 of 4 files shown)


miconda:
It is indeed an approach that can be considered for the next major release, which also means the old style will no longer work. I am not sure at this moment if this impacts older Linux distro that might still have libssl 1.x. So it needs some careful review.

For the tls cfg reload, the approach can be switched to something similar done with other modules like app_lua, where each worker has to load again the lua script file on rpc reload: a version number is kept in shared memory, initialized to 0 at startup, then the rpc reload increases that one and the tcp main, before accepting or connecting, checks if a local version number variable is lower that the shared one, and in that case performs the file reload, setting the local version variable to the value found in shared memory.

space88man:
> I am not sure at this moment if this impacts older Linux distro that might still have libssl 1.x. 

This coexists with shared memory OpenSSL and depends only on  `ksr_tcp_main_threads` at runtime. Updated the PR to make this clear.

The main change that affects shared memory OpenSSL is the eager caching of certs, and other SSL metadata in `struct tls_extra_data`. So even in shared memory mode (`ksr_tcp_main_threads = 0`) the getters(pv/sel) don't parse raw `SSL *` objects.

The code should also still work on OpenSSL ≥ 1.1.1 (RHEL 8 / Debian 11)

PR #4643, opened 2026-03-13
Sanity add some Via related validations

xkaraman:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This PR add two validation tests related to VIA.

1. Check if VIA is present through the `required_headers` function. the docs suggest to check for it but code does not (not in this function anyway) (considered a bug?) 

Another note though is that sanity module may be not really acting on this at all due to some sanity check found in (https://github.com/kamailio/kamailio/blob/41c30c244ca0ca8b3c201e0c02602b2554f57cb3/src/core/receive.c#L448-L463) and that kamailio cfg will drop the msg much earlier

2. Check if branch paramater is present in the Via header which is mandatory according to RFC. (new feature)

Also increase a dbg to err log and check for host len more stricly. 0-len host is not valid.

### DIFF
--- a/src/modules/sanity/doc/sanity_admin.xml
+++ b/src/modules/sanity/doc/sanity_admin.xml
@@ -122,6 +122,12 @@
 			available, can be parsed and has an address value.
 			</para>
 		</listitem>
+		<listitem>
+			<para>
+			first via header branch paramater (32768) - checks if the branch parameter of the first Via header is present and
+			parseable by the &kamailio; parsers and starts with magic cookie of z9hG4bK.
+			</para>
+		</listitem>
 		</itemizedlist>
 	</para>
     </section>


miconda:
I haven't checked the PR yet, just a comment on specs: Via branch is mandatory in a RFC3261 implementation, but that RFC also requires backward compatibility with RFC2543, where it is not required. So this check must not be done implicitly, maybe a new test option (or function) for sanity module.

henningw:
> I haven't checked the PR yet, just a comment on specs: Via branch is mandatory in a RFC3261 implementation, but that RFC also requires backward compatibility with RFC2543, where it is not required. So this check must not be done implicitly, maybe a new test option (or function) for sanity module.

The RFC actually specify that its mandatory in Section 8.1.1.7
```
   When the UAC creates a request, it MUST insert a Via into that
   request.  The protocol name and protocol version in the header field
   MUST be SIP and 2.0, respectively.  The Via header field value MUST
   contain a branch parameter.  This parameter is used to identify the
   transaction created by that request.  This parameter is used by both
   the client and the server.
```
The grammar defines it as not mandatory, correct. Are you sure we need to care still about RFC 2543? I have never seen any device that uses SIP 1.0 in almost 20 years. 

Some places in the code assume that its there without checking. We observed a crash due to a malformed message.

miconda:
I would like to not care about rfc2543, that could have simplified a lot in the code. Iirc, the restore of To/From URI (uac module)is a "constraint" of coping with rfc2543, where the dialog matching was done on URIs, not the Tags of those headers. To indicate that UA is rfc3261, the Via branch has to start with the cookie, it is not just any value.

Anyhow, tm module still has code to match transaction with or without via branch with cookie. rr still has code for strict routing ... so, as it is right now, the check should not be mandatory. A decision to be only rfc3261 without backward compatibility needs more impact analysis and code changes.

rfc2543 is also SIP 2.0, there was never a SIP 1.0, afaik -- the version in the first line cannot be used to differentiate between a rfc2543 UA and rfc3261.

xkaraman:
Changed the PR to intorduce a new check for via branch parameter. if good i ll add the the docs for it.

miconda:
Thanks!

PR #4644, opened 2026-03-13
tm: fix transaction leak on drop; in branch_route

### LINKED ISSUE #4503: Potential SHM memory leak in drop; ?
Hello,

on my Kamailio installation I am experiencing a significant memory leak in SHM.
Here are the details of my system:

```
version: kamailio 6.0.3 (x86_64/linux) 
flags: USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST,
       NO_SIG_DEBUG, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, MEM_JOIN_FREE, 
       Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX,
       FAST_LOCK-ADAPTIVE_WAIT, USE_DNS_CACHE, USE_DNS_FAILOVER, 
       USE_NAPTR, USE_DST_BLOCKLIST, HAVE_RESOLV_RES, TLS_PTHREAD_MUTEX_SHARED
ADAPTIVE_WAIT_LOOPS: 1024
MAX_RECV_BUFFER_SIZE: 262144
MAX_SEND_BUFFER_SIZE: 262144
MAX_URI_SIZE: 1024
BUF_SIZE: 65535
DEFAULT PKG_SIZE: 8MB
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select
compiled with gcc 14.2.0
```

The memory leak appears only when the presence feature is enabled.

When presence is active, Kamailio starts running dialog_publish(), whose code is here:

https://github.com/kamailio/kamailio/blob/9dc160d1d2bdf0542d3d9d8ae090bb1352520a1c/src/modules/pua_dialoginfo/dialog_publish.c#L291

This function does not send the PUBLISH directly: it calls pua_send_publish(), which is a function pointer referring to the send_publish() implementation in the pua module.
Then send_publish() eventually calls set_uac_req() and tmb.t_request():

https://github.com/kamailio/kamailio/blob/9dc160d1d2bdf0542d3d9d8ae090bb1352520a1c/src/modules/pua/send_publish.c#L617

Digging further, tmb.t_request() maps to request() in the TM module, which calls t_uac_with_ids() and then t_uac_prepare().

Now comes the suspicious part:

If I comment out the call to t_uac_prepare(), the memory leak disappears.
This doesn’t necessarily mean the bug is inside t_uac_prepare(), but it’s a strong hint.

t_uac_prepare() allocates a new struct cell and returns it:

https://github.com/kamailio/kamailio/blob/9dc160d1d2bdf0542d3d9d8ae090bb1352520a1c/src/modules/tm/uac.c#L511

My concern is: is this cell always freed?

The matching cleanup function is free_cell(), used only here:

https://github.com/kamailio/kamailio/blob/9dc160d1d2bdf0542d3d9d8ae090bb1352520a1c/src/modules/tm/uac.c#L683

From what I can tell, free_cell() is called only if all these conditions are true:

- dst_cell == 0
- is_ack == 1
- dst_req == 0

In my situation no ACK is involved (Kamailio is a proxy that sends PUBLISH and immediately gets a 200 OK).
Therefore, is_ack is always false meaning the free_cell() cleanup logic is skipped entirely.

I tried forcing free_cell() unconditionally, but it leads to crashes, so clearly other parts of the code still rely on this structure.

Does the current free_cell() logic look correct to you?
Is it expected that struct cell allocated by t_uac_prepare() remains unfreed in cases where PUBLISH → 200 OK occurs without ACK?

Any guidance on how to proceed or where else to look would be greatly appreciated.

Thanks

NormB:
Fixes #4503.

### Problem

Transactions leak when `drop;` executes in a `branch_route` and modules
like `pua_dialoginfo` are loaded.  The leaked transactions persist
indefinitely in shared memory with `ref_count: 1`, `outgoings: 0`, no
timers running.  The only recovery is `kamcmd tm.clean`.

### Cause

There are two code paths that lead to the leak.  Both require a module
that calls `tmb.t_request()` inline during INVITE processing (such as
`pua_dialoginfo`) to contaminate the process-local `_tm_kr` variable
with `REQ_FWDED` via `t_uac_prepare()`.

**Path A — `REQ_ERR_DELAYED` set but masked by contamination:**

When `t_relay()` is used without `T_DISABLE_INTERNAL_REPLY`, the error
path in `t_relay_to()` calls `set_kr(REQ_ERR_DELAYED)`.  Because
`set_kr()` uses `|=`, the result is `REQ_FWDED | REQ_ERR_DELAYED` (17).
The first check in `t_unref()` uses exact equality (`kr == 16`), which
fails for 17.  The other two checks also fail for this value.

**Path B — `REQ_ERR_DELAYED` never set:**

When `T_DISABLE_INTERNAL_REPLY` is active — set explicitly via
`t_set_disable_internal_reply(1)`, or implicitly via `t_relay_to()` with
flag 2 — the error path skips both `set_kr(REQ_ERR_DELAYED)` and
`kill_transaction()`.  The `_tm_kr` value remains just `REQ_FWDED` (1).
All three checks in `t_unref()` fail for this value.

In both paths, `UNREF(T)` decrements `ref_count` from 2 to 1 (releasing
the process reference from `new_t()`), but the hash table reference is
never released because neither `kill_transaction()` nor
`t_release_transaction()` ran.

### Fix

Two changes to `t_unref()`:

**Change 1:** The first check uses exact equality to bitwise test.
This handles path A — `kill_transaction()` now fires when
`REQ_ERR_DELAYED` is present regardless of other accumulated flags.

**Change 2:** A fallback after the existing three checks catches
orphaned transactions with no forwarded branches.  This handles path B
and any future scenario where `_tm_kr` contamination causes all
existing checks to fail.  It uses `t_release_transaction()` rather
than `kill_transaction()` — it does not generate a SIP error reply,
consistent with configs that set `T_DISABLE_INTERNAL_REPLY` to
suppress automatic replies.

```diff
--- a/src/modules/tm/t_lookup.c
+++ b/src/modules/tm/t_lookup.c
@@ -N,7 +N,7 @@ int t_unref(struct sip_msg *p_msg)
 	if(p_msg->first_line.type == SIP_REQUEST) {
 		kr = get_kr();
-		if(unlikely(kr == REQ_ERR_DELAYED)) {
+		if(unlikely(kr & REQ_ERR_DELAYED)) {
 			LM_DBG("delayed error reply generation(%d)\n", tm_error);
@@ after the REQ_ERR_DELAYED BUG check block:
 			t_release_transaction(T);
+		} else if(unlikely(T->nr_of_outgoings == 0
+						   && !(T->flags & T_IN_AGONY))) {
+			LM_WARN("dropping orphaned transaction without branches"
+				   " T=%p kr=%d\n", T, kr);
+			t_release_transaction(T);
 		}
 	}
 	tm_error = 0; /* clear it */
```

### Test results

Kamailio 6.0.3, upstream tag, 8 x86_64 cores, 8 workers, 10,000
INVITEs per test at 500/sec.  All configs load `pua_dialoginfo` and
`rtpengine` with an unreachable daemon so every call hits
`rtpengine_offer()` failure followed by `drop;` in `branch_route`.

Values are `tm.stats current` after all timers expire (0 = no leak):

**Without `T_DISABLE_INTERNAL_REPLY`:**

| Config | No patch | `==` to `&` only | This PR |
|--------|:--------:|:----------------:|:-------:|
| `t_relay()` | 3548 | 0 | 0 |
| `t_relay_to()` | 3546 | 0 | 0 |
| `t_relay_to_udp("host","port")` | 3548 | 0 | 0 |
| `t_relay_to("proxy","0")` | 3547 | 0 | 0 |
| `t_relay_to("proxy","1")` no auto 100 | 3707 | 0 | 0 |
| `t_relay_to("proxy","4")` no dns fo | 3549 | 0 | 0 |
| `t_relay_to("proxy","5")` 1+4 | 3703 | 0 | 0 |

**With `T_DISABLE_INTERNAL_REPLY`:**

| Config | No patch | `==` to `&` only | This PR |
|--------|:--------:|:----------------:|:-------:|
| `t_relay_to("proxy","2")` no int reply | 3549 | **3547** | 0 |
| `t_relay_to("proxy","3")` 1+2 | 3703 | **3703** | 0 |
| `t_relay_to("proxy","6")` 2+4 | 3547 | **3544** | 0 |
| `t_relay_to("proxy","7")` 1+2+4 | 3702 | **3708** | 0 |
| `t_set_disable_internal_reply(1)` + `t_relay()` | 3545 | **3547** | 0 |
| `t_set_disable_internal_reply(1)` + `t_relay_to()` | 3542 | **3544** | 0 |
| `t_set_disable_internal_reply(1)` + `t_relay_to_udp()` | 3548 | **3548** | 0 |

The `==` to `&` change alone fixes configs without
`T_DISABLE_INTERNAL_REPLY`.  The orphan fallback is required for configs
that use it — including any config using `t_relay_to()` with flag 2, or
calling `t_set_disable_internal_reply(1)` for serial forking.

Documented flags: 1 (no auto 100), 2 (no internal reply), 4 (no DNS
failover).  Flags 3, 5, 6, 7 are undocumented combinations tested for
completeness.

### DIFF
--- a/src/modules/tm/t_lookup.c
+++ b/src/modules/tm/t_lookup.c
@@ -2057,7 +2057,7 @@ int t_unref(struct sip_msg *p_msg)
 		return -1;
 	if(p_msg->first_line.type == SIP_REQUEST) {
 		kr = get_kr();
-		if(unlikely(kr == REQ_ERR_DELAYED)) {
+		if(unlikely(kr & REQ_ERR_DELAYED)) {
 			LM_DBG("delayed error reply generation(%d)\n", tm_error);
 			if(unlikely(is_route_type(FAILURE_ROUTE))) {
 				LM_BUG("called w/ kr=REQ_ERR_DELAYED in failure"
@@ -2074,13 +2074,11 @@ int t_unref(struct sip_msg *p_msg)
 								   && !(kr & REQ_RLSD)))) {
 			LM_WARN("script writer didn't release transaction\n");
 			t_release_transaction(T);
-		} else if(unlikely((kr & REQ_ERR_DELAYED)
-						   && (kr
-								   & ~(REQ_RLSD | REQ_RPLD | REQ_ERR_DELAYED
-										   | REQ_FWDED)))) {
-			LM_BUG("REQ_ERR DELAYED should have been caught much"
-				   " earlier for %p: %d (hex %x)\n",
-					T, kr, kr);
+		} else if(unlikely(T->nr_of_outgoings == 0
+						   && !(T->flags & T_IN_AGONY))) {
+			LM_WARN("dropping orphaned transaction without branches"
+					" T=%p kr=%d\n",
+					T, kr);
 			t_release_transaction(T);
 		}
 	}


gaaf:
Please don't keep unreachable code, tm's code is already hard enough to grasp.

Is it really safe/wanted to omit **all** the extra checks from path 3 and now call `kill_transaction()` instead of just `t_release_transaction()` in those cases? If so, please explain why in the commit message and remove the dead code. Note that this code is 16+ years old, maybe the real cause is somewhere else?

Maybe 2812bd9c426d was wrong and the extra condition had to be handled in path 1? Like so: `if (unlikely(kr == REQ_ERR_DELAYED || (kr == (REQ_ERR_DELAYED | REQ_FWDED)))`

NormB:
Good points, thanks for the review.

**Dead code removal** -- agreed, updated patch removes the path 3 block entirely.

**kill_transaction() vs t_release_transaction()** -- safe. `kill_transaction()` checks `has_reply_route()` and `FL_FINAL_REPLY` first (t_funcs.c), and if a final reply was already sent, it degrades to `t_release_transaction()` -- identical to what path 3 did. When no final reply was sent (our case: all branches dropped), the UAC is waiting with no response, so generating a 500 is the correct behavior.

**On the explicit `||` alternative** -- the problem is wider than just `REQ_FWDED`. Because `set_kr()` uses `|=` on process-local state, any combination of flags can accumulate. The path 3 mask itself proves this:

```c
(kr & ~(REQ_RLSD | REQ_RPLD | REQ_ERR_DELAYED | REQ_FWDED))
```

This explicitly *excludes* those flags from triggering cleanup, so path 3 never fires for any of these combinations either:

- `ERR_DELAYED | FWDED` (17): `17 & ~23 = 0` → path 3 skips → **leak**
- `ERR_DELAYED | RPLD` (18): `18 & ~23 = 0` → path 3 skips → **leak**
- `ERR_DELAYED | RLSD` (20): `20 & ~23 = 0` → path 3 skips → **leak**

Path 3 only ever caught `REQ_EXIST` contamination (`kr & ~23 != 0`), which is unlikely in practice. The `||` approach would fix one known case; the `&` test handles all of them and is robust against future flag additions.

Tested the updated patch (bitwise `&` + path 3 removed) on 6.2.0-dev0 master (4767914):

| Test | Patch | current | 5xx | freed | Verdict |
|------|-------|---------|-----|-------|---------|
| Baseline (no patch) | no | 500 | 0 | 500/1000 | **LEAK** |
| Updated patch | yes | **0** | 500 | 2000/2000 | **OK** |
| Updated patch + `delayed_reply=0` | yes | **0** | 500 | 2000/2000 | **OK** |

Same results as the original 10-test matrix across 6.0.3/6.1.0/master -- removing path 3 has no effect since it was already unreachable after the `&` change.

Commit updated with path 3 removed and expanded commit message.

NormB:
One more note on gaaf's point about whether the real cause is somewhere else -- he's right that the deeper architectural issue is `set_kr()` using `|=` to accumulate flags across unrelated TM operations within the same process. Fixing that properly would mean reworking every `set_kr()` caller (or scoping `_tm_kr` per-transaction instead of per-process), which is a much larger change.

The `&` check is the minimal defensive fix at the decision point -- it tolerates flag contamination from inline TM operations rather than trying to prevent it.

henningw:
These tool generated text blocks are really hard to read. Just compare your last comment to your previous one, @NormB. 
It would be better if it would be actually you, who is commenting, like in the last one.

> kill_transaction() vs t_release_transaction() -- safe. kill_transaction() checks has_reply_route() and FL_FINAL_REPLY first (t_funcs.c), and if a final reply was already sent,[..]

Also there is no has_reply_route() function in the whole tm module:

```
henning@app01.dev:~/repositories/kamailio/src/modules/tm$ ack has_reply_route
henning@app01.dev:~/repositories/kamailio/src/modules/tm$
```

NormB:
Sorry @henningw , fair point all around. has_reply_route() doesn't exist.

I was referring to the FL_FINAL_REPLY check at the top of kill_transaction(), which falls back to t_release_transaction() if a final reply was already sent.

So for @gaaf question: in the drop-all-branches case, no final reply exists, so it generates a 500 via t_reply(). If a reply had already gone out, it degrades to the same t_release_transaction() that path 3 used.

gaaf:
I'm not familiar with that part of tm. I'm just worried about possible side-effects of changing long-standing code for multiple cases while fixing only a single failure case. Given the path 3 conditions, I do think the original `==`  in path 1 was a deliberate choice and it should be checked very carefully if all path 3 conditions can be merged in path 1. That's why I suggested changing it only for the case at hand (`REQ_FWDED`) and evaluating against 2812bd9c426.
Unfortunately, 2812bd9c426 does not provide a explanation on why it was added to path 3 instead of 1.

NormB:
> I'm not familiar with that part of tm. I'm just worried about possible side-effects of changing long-standing code for multiple cases while fixing only a single failure case. Given the path 3 conditions, I do think the original `==` in path 1 was a deliberate choice and it should be checked very carefully if all path 3 conditions can be merged in path 1. That's why I suggested changing it only for the case at hand (`REQ_FWDED`) and evaluating against [2812bd9](https://github.com/kamailio/kamailio/commit/2812bd9c426d7a186bdc9b1d5fc4766315a64290). Unfortunately, [2812bd9](https://github.com/kamailio/kamailio/commit/2812bd9c426d7a186bdc9b1d5fc4766315a64290) does not provide a explanation on why it was added to path 3 instead of 1.

I've got the entire history of t_unref() path 3 from day 1 in 2003. After reading the entire document, it does make sense and details each commit, when it was made along with the author and commit message. If you like, I can append the markdown of the entire document here.

This is the Summary:

This document traces the git history of the three cleanup paths in t_unref() to answer the question: was the == in path 1 a deliberate choice, and is it safe to replace it with &?

The answer: path 1's == was correct when written because REQ_ERR_DELAYED couldn't coexist with other flags at that time. Path 3 was added two days later as a diagnostic BUG catcher, not as an alternative cleanup path. Two subsequent commits narrowed path 3's trigger to suppress false positive BUG messages. The == in path 1 was never revisited during those changes — understandably, since the focus each time was on the noisy BUG log rather than the broader cleanup logic.

henningw:
It seems we are still discussing with AI generated summaries. I have spend now several hours going through this PR, validating the assumptions and call paths to make sure there is not another hallucination. This approach is not helpful for me, I must admit. I am setting this PR to draft, as it could not be verified yet and also did not solve the issue reported in #4503, it seems.

Here some short page I came around yesterday, which I found interesting regarding the usage of AI tools: [ai-manifesto](https://ai-manifesto.dev/) . It was created because some issues other open source projects had in this situation.

Ozzyboshi:
One side note, what I tested is https://github.com/kamailio/kamailio/issues/4503#issuecomment-4125345286

This is a sum of https://github.com/kamailio/kamailio/pull/4644/changes/55ae00455a8772aa3997809b02f491315f860a97
plus what you mentioned as "Change 2" at message https://github.com/kamailio/kamailio/pull/4644#issue-4073052522

Please update the commit before merging it.

miconda:
@NormB: does the commits of PR need to be updated according to previous comment? If yes, can you do it so the PR can be reviewed with latest changes.

henningw:
Thanks for the update, format checks are failing on an unrelated earlier commit.

miconda:
Thanks, merging it!

Not yet sure if it should be immediately backported, willing to watch it for a while...

PR #4659, opened 2026-03-24
textopsx: fix segfault in pv_get_hf_iterator_hname()

jchavanton:
Use correct array index [k] instead of [i] after loop exit. Variable 'i' contains undefined value after the search loop completes, causing segfault when dereferencing _hf_iterators[i].it->name.

The loop finds the iterator at index i and stores it in k (line 2476), but then incorrectly uses _hf_iterators[i] instead of _hf_iterators[k] after the loop (lines 2486, 2489).

Crash backtrace:
#0  pv_get_strval() - dereferencing invalid pointer #1  pv_get_hf_iterator_hname() at textopsx.c:2489
    Passing garbage pointer &_hf_iterators[i].it->name where i is undefined

This bug causes crashes when processing SIP replies with header iterators.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description
This is not tested, I was getting the crash when I was doing something like this.
This patch was made by AI and is not tested, I changed to another simpler solution but this new feature is great.
Would be nice to patch it to work well in reply_route !

@miconda 
Can you have a look ?

This is what I was testing at that time :

```
onreply_route[DEFAULT] {
    hf_iterator_start("i1");
    while(hf_iterator_next("i1")) {
        if ($hfitname(i1) == "Record-Route") {
                if (is_ip_rfc1918("$si")) {
                        $var(regex) = $_s(/$shv(private)/$shv(public)/sg);
                } else {
                        $var(regex) = $_s(/$shv(public)/$shv(private)/sg);
                }
                hf_iterator_rm("i1");
                $avp(rr) = $hfitbody(i1);
                $avp(rr) =  $(avp(rr){re.subst,$var(regex)});
                hf_iterator_append("i1", "Record-Route: $avp(rr)\r\n");
                xinfo("rr hdr[$hfitname(i1)] is: $avp(rr) [$dd] [$var(regex)]\n");
                msg_apply_changes();
        }
    }
    hf_iterator_end("i1");
```

### DIFF
--- a/src/modules/textopsx/textopsx.c
+++ b/src/modules/textopsx/textopsx.c
@@ -2483,10 +2483,10 @@ static int pv_get_hf_iterator_hname(
 		return pv_get_null(msg, param, res);
 	}
 
-	if(_hf_iterators[i].it == NULL) {
+	if(_hf_iterators[k].it == NULL) {
 		return pv_get_null(msg, param, res);
 	}
-	return pv_get_strval(msg, param, res, &_hf_iterators[i].it->name);
+	return pv_get_strval(msg, param, res, &_hf_iterators[k].it->name);
 }
 
 /**


miconda:
Thanks!

PR #4460, opened 2025-11-05
dispatcher: Trigger event_routes when old_state its not present

bundasmanu:
### Description

During some tests, i saw that `event_routes` on `dispatcher` module are not triggered, in some specific situations like:

- When we do a `dispatcher.reload`;
- On startup, when destinations are `AP` in the first attempt;

In both cases, we don't have an `old state`, so the `event_routes` are not triggered.

But, in a lot of cases, makes sense to do some log, or some other operations, on these kind of situations.

So, i have introduced a small patch, to make possible to trigger these routes, when we don't have the `old state` available.

#### Pre-Submission Checklist

- [x] Commit message has the format required by CONTRIBUTING guide
- [x] Commits are split per component (core, individual modules, libs, utils, ...)
- [x] Each component has a single commit (if not, squash them into one commit)
- [x] No commits to README files for modules (changes must be done to docbook files
in `doc/` subfolder, the README file is autogenerated)

#### Type Of Change
- [ ] Small bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would change existing functionality)

#### Checklist:
- [x] PR should be backported to stable branches
- [x] Tested changes locally
- [ ] Related to issue #XXXX (replace XXXX with an open issue number)

#### Additional Information

Tested on my own, old behaviour works as is, and now when `old_state` its not present these `event_routes` are now triggered.

PR #4661, opened 2026-03-25
dmq: add peer-up and peer-down event_routes

bundasmanu:
#### Description

This PR introduces `event_routes` that will trigger when `dmq peers` go up and down during the time.

Really useful to detect problems between `peers` and allows addition of prometheus metrics for example, to add automatic visibility.

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Tests

Tested locally simulating crashes and startups to evaluate the trigger of routes.

Couldn't found any problems or strange cases, on ping intervals the status validation, gives the correct output.

### DIFF
--- a/src/modules/dmq/dmq.c
+++ b/src/modules/dmq/dmq.c
@@ -533,13 +533,13 @@ static sr_kemi_t sr_kemi_dmq_exports[] = {
 /* clang-format on */
 
 /**
- * Run peer event route. @a evt is which route to run (DMQ_NODE_ACTIVE ->
- * dmq:peer-up, DMQ_NODE_NOT_ACTIVE -> dmq:peer-down, DMQ_NODE_DISABLED ->
- * dmq:peer-disabled). Pass the transition decided under dmq_node_list lock (or
- * when queuing extract callbacks); do not derive from node->status after
- * lock_release — other workers may change SHM before this runs.
+ * Run peer event route from @a node->status (DMQ_NODE_ACTIVE -> dmq:peer-up,
+ * DMQ_NODE_NOT_ACTIVE -> dmq:peer-down, DMQ_NODE_DISABLED -> dmq:peer-disabled).
+ * Callers must set node->status to the intended announcement before calling
+ * (e.g. under dmq_node_list lock, or after applying a transition).
+ * Duplicate for the same status are suppressed using last_peer_evt.
  */
-void dmq_peer_run_event_route(int evt, dmq_node_t *node)
+void dmq_peer_run_event_route(dmq_node_t *node)
 {
 	const char *rtname;
 	int rt, backup_rt;
@@ -547,11 +547,14 @@ void dmq_peer_run_event_route(int evt, dmq_node_t *node)
 	sip_msg_t *fmsg = NULL;
 	sr_kemi_eng_t *keng = NULL;
 	str evname;
+	int evt;
 
 	if(node == NULL || node->local)
 		return;
 	if(node->orig_uri.s == NULL || node->orig_uri.len <= 0)
 		return;
+
+	evt = node->status;
 
 	/* Same announcement twice (timer/body oscillation). */
 	if(node->last_peer_evt == evt) {
--- a/src/modules/dmq/dmq.c
+++ b/src/modules/dmq/dmq.c
@@ -42,6 +42,10 @@
 #include "../../core/cfg/cfg_struct.h"
 #include "../../core/rpc_lookup.h"
 #include "../../core/kemi.h"
+#include "../../core/route.h"
+#include "../../core/dset.h"
+#include "../../core/fmsg.h"
+#include "../../core/action.h"
 
 #include "dmq.h"
 #include "dmq_funcs.h"
@@ -73,6 +77,7 @@ int dmq_fail_count_enabled = 0;
 int dmq_fail_count_threshold_not_active = 0;
 int dmq_fail_count_threshold_disabled = 1;
 int dmq_init_with_single = 0;
+str dmq_event_callback = STR_NULL;
 
 /* TM bind */
 struct tm_binds _dmq_tmb = {0};
@@ -135,6 +140,7 @@ static param_export_t params[] = {
 	{"fail_count_threshold_not_active", PARAM_INT, &dmq_fail_count_threshold_not_active},
 	{"fail_count_threshold_disabled", PARAM_INT, &dmq_fail_count_threshold_disabled},
 	{"init_with_single", PARAM_INT, &dmq_init_with_single},
+	{"event_callback", PARAM_STR, &dmq_event_callback},
 	{0, 0, 0}
 };
 
@@ -525,6 +531,105 @@ static sr_kemi_t sr_kemi_dmq_exports[] = {
 	{ {0, 0}, {0, 0}, 0, NULL, { 0, 0, 0, 0, 0, 0 } }
 };
 /* clang-format on */
+
+/**
+ * Run peer event route. @a evt is which route to run (DMQ_NODE_ACTIVE ->
+ * dmq:peer-up, DMQ_NODE_NOT_ACTIVE -> dmq:peer-down, DMQ_NODE_DISABLED ->
+ * dmq:peer-disabled). Pass the transition decided under dmq_node_list lock (or
+ * when queuing extract callbacks); do not derive from node->status after
+ * lock_release — other workers may change SHM before this runs.
+ */
+void dmq_peer_run_event_route(int evt, dmq_node_t *node)
+{
+	const char *rtname;
+	int rt, backup_rt;
+	struct run_act_ctx ctx;
+	sip_msg_t *fmsg = NULL;
+	sr_kemi_eng_t *keng = NULL;
+	str evname;
+
+	if(node == NULL || node->local)
+		return;
+	if(node->orig_uri.s == NULL || node->orig_uri.len <= 0)
+		return;
+
+	/* Same announcement twice (timer/body oscillation). */
+	if(node->last_peer_evt == evt) {
+		LM_DBG("peer event %d skipped (duplicate) for %.*s\n", evt,
+				STR_FMT(&node->orig_uri));
+		return;
+	}
+	/* After peer-disabled, only peer-up is a valid next announcement. */
+	if(node->last_peer_evt == DMQ_NODE_DISABLED && evt != DMQ_NODE_ACTIVE) {
+		LM_DBG("peer event %d skipped (post-disabled) for %.*s\n", evt,
+				STR_FMT(&node->orig_uri));
+		return;
+	}
+
+	/* Logical transition recorded even if no event_route (suppression still applies). */
+	node->last_peer_evt = evt;
+
+	switch(evt) {
+		case DMQ_NODE_ACTIVE:
+			rtname = "dmq:peer-up";
+			break;
+		case DMQ_NODE_NOT_ACTIVE:
+			rtname = "dmq:peer-down";
+			break;
+		case DMQ_NODE_DISABLED:
+			rtname = "dmq:peer-disabled";
+			break;
+		default:
+			LM_BUG("dmq peer event: unexpected evt %d\n", evt);
+			return;
+	}
+
+	LM_DBG("executing event_route[%s] for %.*s\n", rtname,
+			STR_FMT(&node->orig_uri));
+
+	rt = -1;
+	if(dmq_event_callback.s == NULL || dmq_event_callback.len <= 0) {
+		rt = route_lookup(&event_rt, (char *)rtname);
+		if(rt < 0 || event_rt.rlist[rt] == NULL) {
+			LM_DBG("event_route[%s] not defined - skip\n", rtname);
+			return;
+		}
+	} else {
+		keng = sr_kemi_eng_get();
+		if(keng == NULL) {
+			LM_DBG("dmq event_callback set but no kemi engine\n");
+			return;
+		}
+	}
+
+	if(faked_msg_init() < 0) {
+		LM_ERR("faked_msg_init failed\n");
+		return;
+	}
+	fmsg = faked_msg_next();
+	if(rewrite_uri(fmsg, &node->orig_uri) < 0) {
+		LM_ERR("rewrite_uri failed\n");
+		return;
+	}
+
+	if(rt >= 0 || dmq_event_callback.len > 0) {
+		backup_rt = get_route_type();
+		set_route_type(REQUEST_ROUTE);
+		init_run_actions_ctx(&ctx);
+		if(rt >= 0) {
+			run_top_route(event_rt.rlist[rt], fmsg, 0);
+		} else if(keng != NULL) {
+			evname.s = (char *)rtname;
+			evname.len = strlen(rtname);
+			if(sr_kemi_route(
+					   keng, fmsg, EVENT_ROUTE, &dmq_event_callback, &evname)
+					< 0)
+				LM_ERR("dmq kemi event route failed\n");
+		}
+		set_route_type(backup_rt);
+	}
+	reset_uri(fmsg);
+}
 
 int mod_register(char *path, int *dlflags, void *p1, void *p2)
 {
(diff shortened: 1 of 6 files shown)


henningw:
Thanks for the PR. Just had a quick look to the code. One question, what is the purpose of the DMQ_EXTRACT_PEER_UP_MAX 128 ?

bundasmanu:
> Thanks for the PR. Just had a quick look to the code. One question, what is the purpose of the DMQ_EXTRACT_PEER_UP_MAX 128 ?

Hi @henningw

DMQ_EXTRACT_PEER_UP_MAX is basically a safety cap for one extract_node_list() run.
It limits how many peers are queued for peer-up callbacks after parsing a notification body (the callback runs after unlocking the list).
So it avoids unbounded stack/callback bursts; if more than 128 peers need peer-up in one pass, the extra ones are skipped and a warning is logged.

bundasmanu:
@henningw, @miconda can you check, please?

vlitvinov:
Hello, Gustavo and thank you for the PR!

I see two new even routes were added, but there are currently 3 different statuses for dmq node: ACTIVE, NOT_ACTIVE, DISABLED. Wouldn't it be better to have single event route for any status change with some new "status" variable available in it?
Also naming is confusing. "node_up\down, node_status_change" sounds more reasonable for me because "peer" is used mainly as a role definition in module itself (for example htable or usrloc are dmq peers).

henningw:
I agree that "node" is a better fit for the naming of the event routes, e.g. dmq:node-up, dmq:node-down. The same logic is also used in the dispatcher module, so would be fine for me to have two routes like this.

miconda:
The overview of the module docs says: `replication between multiple instances, referred as "nodes" (or "peers").`

And the apps are referred as: `data flow between nodes is grouped in a logical entity referred as "channel" (or "bus")`.

Based on these, peer is considered same as node.

Just for clarification, I don't have a preference for a particular name.

bundasmanu:
Hi guys,

Thanks for the feedback, really appreciate it.

The topic regarding `node` vs `peer` was also discussed in my head :)

But, in my conscious, i prefer to use `peer` to avoid misunderstandings. `Node` seems to be more generic, and `peer` points to my notification address, except me (actual node), and the events manage only peers. In contrast with `dispatcher`, that i will not include actual server address there, here in `dmq` i can do that on `notification address`.

But, i can switch the naming, if you guys prefer.

Regarding `disable` state, if you guys prefer, i also can include a new `event_route` for it. But i prefer to keep isolated `event_routes` instead of aggregating into a single one, i think its the main logic on other modules.

Please, let me now, about your preferences.

bundasmanu:
Included support for `disabled` state (and tested).

`peer` naming was kept.

**Note:** 2 checks failed, but are not related to my commits.

bundasmanu:
@miconda Its ok, to be merged?

src/modules/dmq/dmq.c
```diff
@@ -524,6 +530,92 @@ static sr_kemi_t sr_kemi_dmq_exports[] = {
...
+	}
+	reset_uri(fmsg);
+}
+
+/**
+ * Run peer event route for a notification keyed by DMQ_NODE_* status:
+ * DMQ_NODE_ACTIVE -> dmq:peer-up, DMQ_NODE_NOT_ACTIVE -> dmq:peer-down,
+ * DMQ_NODE_DISABLED -> dmq:peer-disabled.
+ */
+void dmq_peer_run_event_route(int status, dmq_node_t *node)
```
vlitvinov: Why do we need "status" here?
I see only one case when status != node->status, in dmq_node_del_filter, but I think it is simplest to update node status there before calling this function. 
Then looks like dmq_peer_run_event_route_name is not needed and all logic can be combined in single function. 
And also you will not need all this node_up/down/disabled variables.

src/modules/dmq/dmqnode.c
```diff
@@ -367,8 +367,15 @@ int dmq_node_del_filter(dmq_node_list_t *list, dmq_node_t *node, int filter)
 	while(cur) {
 		if(cmp_dmq_node(cur, node)) {
 			if(filter == 0 || cur->local == 0) {
+				int down = (!cur->local && cur->status == DMQ_NODE_ACTIVE);
+
 				*prev = cur->next;
+				lock_release(&list->lock);
+				LM_DBG("released dmq_node_list->lock\n");
+				if(down)
+					dmq_peer_run_event_route(DMQ_NODE_NOT_ACTIVE, cur);
```
vlitvinov: due to README and code delete node status is DMQ_NODE_DISABLED

vlitvinov:
```suggestion
*prev = cur->next;
				lock_release(&list->lock);
				LM_DBG("released dmq_node_list->lock\n");
				if(cur->status != DMQ_NODE_DISABLED)
					dmq_peer_run_event_route(DMQ_NODE_DISABLED, cur);
```

src/modules/dmq/dmqnode.c
```diff
@@ -427,6 +434,8 @@ dmq_node_t *add_dmq_node(dmq_node_list_t *list, str *uri)
 	list->count++;
 	lock_release(&list->lock);
 	LM_DBG("released dmq_node_list->lock\n");
+	if(!no_peer_evt && !newnode->local && newnode->status == DMQ_NODE_ACTIVE)
```
vlitvinov: why is it only for "active" status?

src/modules/dmq/notification_peer.c
```diff
@@ -438,21 +449,69 @@ int extract_node_list(dmq_node_list_t *update_list, struct sip_msg *msg)
...
 				  && ret->status != find->status
 				  && ret->status != DMQ_NODE_DISABLED) {
 			/* don't update the node if it is in ending state */
+			int old_status = ret->status;
+
 			LM_DBG("updating status on %.*s from %d to %d\n", STR_FMT(&tmp_uri),
 					ret->status, find->status);
 			ret->status = find->status;
 			total_nodes++;
+			if(old_status != DMQ_NODE_ACTIVE
```
vlitvinov: What is the purpose of checking old and new statuses? This code is executing only if old_status != new_status.

src/modules/dmq/dmqnode.c
bundasmanu: `DMQ_NODE_DISABLED` in the docs is the in-list state (“will be deleted from the node list”), not a guarantee that every unlink path has that flag set at removal time. `dmq_node_del_filter` can remove a node that is still `DMQ_NODE_ACTIVE` (e.g. timeout delete). We intentionally fire `dmq:peer-down` there: the peer is no longer an active cluster member. If we should fire `dmq:peer-disabled` on delete instead (or both), we can align it.

bundasmanu: Thanks for the snippet—there’s a logic issue: with `!= DMQ_NODE_DISABLED` we would call `peer-disabled` for active and not_active nodes, and skip it when the node is already disabled. We’ll keep explicit semantics: down when removed while active (after setting status to not_active for the event), disabled only when the state transition is disabled.

src/modules/dmq/dmqnode.c
bundasmanu: `dmq:peer-up` means the peer is active in the DMQ view. New entries from `build_dmq_node` are created `active`; if we added a non-active node on purpose, firing up would be wrong. We only trigger the route when `status == DMQ_NODE_ACTIVE`

src/modules/dmq/notification_peer.c
bundasmanu: The outer check is only `ret->status != find->status` (something changed). We still need `old_status vs new` to choose which event: e.g. `active → not_active is dmq:peer-down`, `active → disabled is dmq:peer-disabled`, `not_active → disabled is peer-disabled`, → active is peer-up. The new status alone doesn’t encode that distinction when classifying the transition.

src/modules/dmq/dmq.c
bundasmanu: I will do a commit for this, and after tests are done on my end, i will reply here again.

src/modules/dmq/dmqnode.c
bundasmanu: `add_dmq_node()` only inserts the struct; local is set in the caller after `add_dmq_node()` returns. Inside `add_dmq_node`, `newnode->local` is still `0` at that point, so without `no_peer_evt == 1` the function would treat this like a remote active node and could fire `peer-up` for your own DMQ address which is wrong.

`no_peer_evt` means: “this add is internal / don’t treat as a new remote peer coming up.” Remote peers from `server_list` use 0 so they still get a proper peer-up when first added.

bundasmanu:
A re-test was applied and events are triggering in the right way, also no event duplications on the same state were observed.

PS: It will be useful to me, to get this new feature available in the next: 6.1.2 release.

henningw:
> A re-test was applied and events are triggering in the right way, also no event duplications on the same state were observed.
> 
> PS: It will be useful to me, to get this new feature available in the next: 6.1.2 release.

Thank you for the update. Can you please have a look and to the clang format error in this commit?
Commit b2729b6 dmq: fix disabled timer state trigger

About the 6.1.2 release - this is not how the Kamailio project manages the stable branches. This feature will be included after the merge in the master branch and then part of the upcoming 6.2.0 release.

bundasmanu:
> > A re-test was applied and events are triggering in the right way, also no event duplications on the same state were observed.
> > PS: It will be useful to me, to get this new feature available in the next: 6.1.2 release.
> 
> Thank you for the update. Can you please have a look and to the clang format error in this commit? Commit [b2729b6](https://github.com/kamailio/kamailio/commit/b2729b663502f5259c2f1de35f8daa367dfd0af6) dmq: fix disabled timer state trigger
> 
> About the 6.1.2 release - this is not how the Kamailio project manages the stable branches. This feature will be included after the merge in the master branch and then part of the upcoming 6.2.0 release.

Thank you @henningw

The problems on the commits, are not related to my commit, but for something from `tls` module.

henningw:
> > > A re-test was applied and events are triggering in the right way, also no event duplications on the same state were observed.
> > > PS: It will be useful to me, to get this new feature available in the next: 6.1.2 release.
> > 
> > 
> > Thank you for the update. Can you please have a look and to the clang format error in this commit? Commit [b2729b6](https://github.com/kamailio/kamailio/commit/b2729b663502f5259c2f1de35f8daa367dfd0af6) dmq: fix disabled timer state trigger
> > About the 6.1.2 release - this is not how the Kamailio project manages the stable branches. This feature will be included after the merge in the master branch and then part of the upcoming 6.2.0 release.
> 
> Thank you @henningw
> 
> The problems on the commits check, are not regarding to my commit, but for something from `tls` module.

Yes, it complains about unrelated other commits. Your commits introduced in this PR are now fine regarding the clang checks.

miconda:
@henningw @vlitvinov :  you still want to review more, or it should be considered for merging if no other comments soon?

henningw:
The naming was discussed earlier, but I don't mind to have it either way. Just one thing - there is now a merge conflict due to the merging of other PRs related to DMQ. @bundasmanu can you please re-base this and force-push to the PR?

vlitvinov:
I'll check it later today and comment here. Thanks.

src/modules/dmq/dmq.c
vlitvinov: This question is still open. Why do we need "status" when we have it in "node" structure.

src/modules/dmq/dmqnode.c
vlitvinov: yes, in my opinion we should align this and consider that node removing done for DISABLED status always.

vlitvinov: Are this new routes executed for node status change? If so on deleting only ACTIVE and NOT_ACTIVE nodes will change their status to DISABLED. 
If node is already in DISABLED status, then this event route was already fired.

src/modules/dmq/dmqnode.c
vlitvinov: uri can contain params, for now only one parameter is allowed and this parameter is status. So node in any available status can be added by this function. 

By the way, were these comments generated by AI ?

src/modules/dmq/notification_peer.c
vlitvinov: Sorry, I didn't get your comment. 
We want to execute event route when status changed, there are 3 status, if we are firing event route execution we node new status, node->status.

vlitvinov:
@bundasmanu was AI involved in writing the code (and comments) in this PR? I almost sure that it was. Please check my comments above.

By the way I agree with @henningw and would rename event routes to be node instead of peer, one more reason is module's rpc commands, we already have "node_list" there.

bundasmanu:
> @bundasmanu was AI involved in writing the code (and comments) in this PR? I almost sure that it was. Please check my comments above.
> 
> By the way I agree with @henningw and would rename event routes to be node instead of peer, one more reason is module's rpc commands, we already have "node_list" there.

Hi @vlitvinov, yes i have used AI (in mixed terms, AI + manual changes) to guide me, as this is my first contribution to dmqueue module, and i'm using that module only a few time ago.

The PR's objective is to detect changes and trigger status changes, yes, like in dispatcher logic. So, old state logic, from my point of view makes sense to allow comparison between states, knew in advance which transaction to trigger and also help avoid retrigger same state in some ocassions. Dispatcher uses some related logic: https://github.com/kamailio/kamailio/pull/4460/changes.

But feel free to also contribute here, as you have more experience with the module.

--------

I have pushed a new commit, with some changes related to some of your suggestions.
For some points, i still kept as is, as from my view, makes sense to keep them, and i declared some more comments in the code (with some explanations).
Re-tested again.

vlitvinov:
Hello, Gustavo!

Looks like I misunderstood your original idea. After reading dispatcher module implementation I see that there is only two event routes: up and down, even for more statuses.

I'm sorry for moving your PR to the wrong direction. I prepared some changes based on your feature branch. 
Renamed event-routes and left only two of them: node-up\node-down.

I just built it but not tested yet. It would be great if you can test it.

Here is the link to my branch (there is only one commit on top of your changes): 
https://github.com/vlitvinov/kamailio/tree/dmq_event_routes_peer_up_down

miconda:
@vlitvinov: the renaming from peer to node is out of the scope for the review. As noted in a comment above, the documentation of the module refer node and peer as the same concept, while channel would be more like application/module endpoint.

  - https://github.com/kamailio/kamailio/pull/4661#issuecomment-4168706914

I haven't checked, but if your review/follow up patch fixes some other issues, that is fine, if it is just renaming the event routes, then it is just fine what @bundasmanu has in the PR.

bundasmanu:
> Hello, Gustavo!
> 
> Looks like I misunderstood your original idea. After reading dispatcher module implementation I see that there is only two event routes: up and down, even for more statuses.
> 
> I'm sorry for moving your PR to the wrong direction. I prepared some changes based on your feature branch. Renamed event-routes and left only two of them: node-up\node-down.
> 
> I just built it but not tested yet. It would be great if you can test it.
> 
> Here is the link to my branch (there is only one commit on top of your changes): https://github.com/vlitvinov/kamailio/tree/dmq_event_routes_peer_up_down

Hi @vlitvinov,

Not a problem, maybe my initial explanation could be better :(

I’ve reviewed your commit on the branch and have a quick question: do you intend to remove the `disabled event_route`? I don’t need it for my use case, but it might still be useful for others.

And absolutely, i can help on testing.

vlitvinov:
Hello @miconda . There are other changes, not only routes renaming. 

@bundasmanu yes, I removed disabled route and kept only two event routes to make it similar to dispatcher module. No matter how many internal statuses we have the node can be up or down.
I would appreciate if you can test it. I'll be able to do this later this week.

bundasmanu:
> Hello @miconda . There are other changes, not only routes renaming.
> 
> @bundasmanu yes, I removed disabled route and kept only two event routes to make it similar to dispatcher module. No matter how many internal statuses we have the node can be up or down. I would appreciate if you can test it. I'll be able to do this later this week.

Hi @vlitvinov, sorry for the delay. I only had time to test it today.

Unfortunately, your PR is not working as expected. It does not detect state changes properly. 

Also, I think the idea of including the `disabled` route makes sense in this module, unlike in the `dispatcher` module. Here in `dmqueue`, we have more control over the `inactive` state because we have `modparam's` that interact with it, and for some users it makes sense to expose it.

miconda:
I am merging it because the code is going to diverge due to long time holding the PR. Further improvements or redesign can be done as follow up PRs.

PR #4664, opened 2026-03-26
core: q_malloc: add optional backtrace debugging

rfuchs:
- New compile-time debugging flag (off by default) to add backtraces to q_malloc fragments
- Only available when `backtrace()` is available
- Print backtraces when listing fragments via `shm.rprint`

#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
Adds a new compile-time option `MEMDBG_BACKTRACE` to enable storing backtrace information in `q_malloc` fragments. This requires the `backtrace()` function from `execinfo.h`. The change to the `cmake` file makes it check for the presence of that function and exposes the option only if it is.

The new option is disabled by default as the overhead can be non trivial.

The `shm.rprint` RPC call is extended to return these backtraces.

Backtraces for memory fragments are useful to debug leaks that occur in frequently-used leaf functions. For these functions, simply printing the function name or file name and line number is not enough, as the root cause for the leak is in a calling function, which without a backtrace remains hidden.


miconda:
Thanks! Fine for me to merge it, if no other comments.

PR #4676, opened 2026-03-31
fix tls_connection_match_domain

furmur:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

fix wrong connections reuse in tls_connection_match_domain=yes mode and add test for it

* change related places for tls and tls_wolfssl modules
* fix protocol checking in _tcpconn_find
* add test/unit/62.sh

test/unit/62.sh execution time:
```
test/unit$ time make run UNIT=62.sh
Run test 62: checks tls_connection_match_domain=yes creates new connection to the same dst ip:port for another tls domain
Test unit file 62.sh: ok

real    0m1.191s
user    0m0.048s
sys     0m0.019s
```

was wrongly reusing the same TLS connection even between different TLS domains because:
* invalid proto check in _tcpconn_find was skipping tls hook match_domain when it called with proto:0
* passing connection ptr to the get_tls_domain_str in tls_h_match_domain_f was causing tls_get_connect_server_name,tls_get_connect_server_id to check connection against itself instead of XAVPs lookup for the new connection context

### DIFF
--- a/src/core/tcp_main.c
+++ b/src/core/tcp_main.c
@@ -1781,7 +1781,7 @@ struct tcp_connection *_tcpconn_find(int id, struct ip_addr *ip, int port,
 						a->parent->id);
 
 #ifdef USE_TLS
-				if(tls_connection_match_domain && proto == PROTO_TLS
+				if(tls_connection_match_domain && a->parent->type == PROTO_TLS
 						&& !tls_hook_call(match_domain, 1, a->parent, ip, port))
 					continue;
 #endif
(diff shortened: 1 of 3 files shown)


miconda:
Thanks for this PR, I will review the code changes. One thing at this moment: can you adapt and create the test unit in the repository:

The ones in the source code are no longer maintained, they should be moved to archive.

furmur:
* rm commit with unit test
* mv it to kamailio-tests: https://github.com/kamailio/kamailio-tests/pull/21

miconda:
Thanks!

PR #4677, opened 2026-04-01
dmq: added init_with_single parameter

vlitvinov:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Right now synchronization during dmq init is done with all available nodes in dmq node list.
In this PR the new parameter init_with_single introduced that force sync with single node that first answered on ping or is first pinged.

### DIFF
--- a/src/modules/dialog/dlg_dmq.c
+++ b/src/modules/dialog/dlg_dmq.c
@@ -35,7 +35,7 @@ dmq_api_t dlg_dmqb;
 dmq_peer_t *dlg_dmq_peer = NULL;
 
 int dmq_send_all_dlgs(dmq_node_t *dmq_node);
-int dlg_dmq_request_sync();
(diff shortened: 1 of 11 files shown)


henningw:
Thanks for the PR. Just a generic question first - what is the motivation behind this change, some performance benefit etc..?

vlitvinov:
> Thanks for the PR. Just a generic question first - what is the motivation behind this change, some performance benefit etc..?

Hi Henning! Yes, mainly performance. Imagine that you have cluster with 5 nodes in it and spinning up 3 new kam nodes. In current implementation all 3 new nodes will trigger sync with al 5 currently running nodes that is 15 sync processes in cluster. Add here that sync is a quite heavy operation with locks\transactions creation. Also it doesn't make a lot of sense to sync with all nodes because each node in normal state should have all data.

miconda:
Thanks!

Issue #4653, opened 2026-03-20
Allow customizing the contact user in uac reg requests


flaviogrossi:
### Description

Currently the uac module uses the value of the `l_uuid` variable as the user part in the contact header for the REGISTER request.
Some providers, especially when registering trunks, requires the user part to be the registration username though, and, while it is possible to do it using the config, it may get cumbersome and error prone to remap the contact back in the reply.

### Expected behavior

To be able to customize the contact in the register

#### Actual observed behavior

The register hardcodes the `l_uuid` value in the register, see [here](https://github.com/kamailio/kamailio/blob/d9a05f9781d2d9cddf485e938ea9f721c3aed139/src/modules/uac/uac_reg.c#L1022)

This causes the registration to never be in the registered status when it is queried via the reg_dump rpc, unless the contact is rewritten via config both in the request (to make the SIP provider happy) and in the reply (to replace it back to use the uuid so that the check [here](https://github.com/kamailio/kamailio/blob/d9a05f9781d2d9cddf485e938ea9f721c3aed139/src/modules/uac/uac_reg.c#L797) matches).

### Possible Solutions

I'm currently testing a patch which:

- adds an optional new field `contact_user` which, if present, is used in the contact user part instead of the uuid;
- adds to the outgoing REGISTER the contact parameter `line=<l_uuid>` (only when `contact_user` is configured, returns an error if `contact_addr` already contains a line parameter to avoid it to be silently overwritten)
- the REGISTER reply match is extended to check the uuid in the line parameter instead of the user part (only if `contact_user` is configured) to select the correct binding.

This means that if the new parameter is configured, the contact will be in the form `sip:<contact_user>@<contact_addr>;line=<l_uuid>`. If not, the usual `sip:<l_uuid>@<contact_addr>` is sent.

This should allow the new feature while also being retro compatible. 

Do you think this is correct? Any feedback? If it is, i can send a PR for review.

### Additional Information

This applies to the latest develop version of kamailio and is OS independent

miconda:
Why not setting the value of `contact_user` in the `l_uuid`?

flaviogrossi:
> Why not setting the value of `contact_user` in the `l_uuid`?

in my use case because kamailio is used as a frontend for a multi-tenant application, so different tenant could have the same contact_user (for example two trunks using the same extension on different providers)

flaviogrossi:
opened a draft PR in #4679 just as a reference for the idea discussed here

[...]

PR #4679, opened 2026-04-01
Allow customizing the contact user in uac reg requests #4653

flaviogrossi:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #4653

#### Description

This is the proposed pull request for issue #4653

Marked as draft because:
- still seeking for suggestions if this is the correct way to do it
- the PR still misses the database migration related code and documentation changes, which i will add if the general idea of the PR is accepted
- i am not an expert in kamailio internals

currently the uac module uses the value of the l_uuid variable as the user part in the contact header for the REGISTER request.

this change allows to customize the contact user part by providing an optional new field contact_user which, if present, used in the contact user part instead of the uuid.

it also adds to the outgoing REGISTER the contact parameter line=<l_uuid> (only when contact_user is configured, returns an error if contact_addr already contains a line parameter to avoid it to be silently overwritten)

the REGISTER reply match is extended to check the uuid in the line parameter instead of the user part (only if contact_user is configured) to select the correct binding.

Open points:

- [ ]  Documentation changes missing
- [ ] Db changes missing
- [ ] should uac_reg_lookup() also be fixed to lookup by the line parameter?

### DIFF
--- a/src/lib/srdb1/schema/entities.xml
+++ b/src/lib/srdb1/schema/entities.xml
@@ -1,6 +1,7 @@
 <!ENTITY user_len "64">
 <!ENTITY domain_len "64">
 <!ENTITY contact_addr_len "255">
+<!ENTITY contact_user_len "255">
 <!ENTITY socket_len "128">
 <!ENTITY from_tag_len "128">
 <!ENTITY to_tag_len "128">
--- a/src/lib/srdb1/schema/uacreg.xml
+++ b/src/lib/srdb1/schema/uacreg.xml
@@ -9,7 +9,7 @@
 
 <table id="uacreg" xmlns:db="http://docbook.org/ns/docbook">
     <name>uacreg</name>
-    <version>5</version>
+    <version>6</version>
     <type db="db_mysql">&MYSQL_TABLE_TYPE;</type>
     <description>
 		<db:para>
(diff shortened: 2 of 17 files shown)


miconda:
Is this still a draft as indicated in the title of the PR?

Note that the commit message does not follow the contributing guidelines, first line not starting with the component name.

Also, the changes in the schema have to be done as a separate commit (in the same PR).

flaviogrossi:
> Is this still a draft as indicated in the title of the PR?

Yes, i guess i needed some feedback about the approach as noted in #4653. Is this the correct solution? If so i can finish the remaining missing pieces and fix the things you noted.

Thanks for your feedback

flaviogrossi:
> Is this still a draft as indicated in the title of the PR?

I removed the draft tag to signal it is ready for review, but the feedback requests in the PR text still apply

> Note that the commit message does not follow the contributing guidelines, first line not starting with the component name.
> 
> Also, the changes in the schema have to be done as a separate commit (in the same PR).

Fixed the issues:
- the module name is in the commit subject line
- added the database changes in the schema in a different commit
- added the commit for the autogenerated kamctl db changes
- updated to the latest master

miconda:
Thanks, merging it!

PR #4688, opened 2026-04-10
tm: support custom callid and fromtag in req_outside

Mattcazz:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description
This PR checks if uac_r contains a custom callid or fromtag before generating new ones in req_outside(). If they are provided and have a valid length, they are used directly. Otherwise, it falls back to generating new ones.

### DIFF
--- a/src/modules/tm/uac.c
+++ b/src/modules/tm/uac.c
@@ -1181,8 +1181,16 @@ int req_outside(uac_req_t *uac_r, str *ruri, str *to, str *from, str *next_hop)
 	if(check_params(uac_r, to, from) < 0)
 		goto err;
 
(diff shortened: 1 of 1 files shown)


IvanRibakov:
Hi @miconda , I appreciate that the PR review process could take a while, but could you please provide some quick early indication whether you would even be interested in accepting a change/feature like this and whether you seen any issues with following the approach that this PR proposes? This will help us greatly in case of a negative answer as we would then have to pursue alternative workarounds asap.

miconda:
Thanks!

PR #4696, opened 2026-04-15
core: ksr_tcp_main_threads default to 1

space88man:
  from source code folder

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description
This sets the default value of `ksr_tcp_main_threads` to 1. This means that OpenSSL (tls) or wolfSSL(tls_wolfssl), will by default, run in tcp_mtops mode(single process). I have tested that normal functions and tls.reload are working for both TLS modules. In addition running with OpenSSL 4 ~~betas~~ has been verified. The niche use cases of using externalized keys (engine, pkcs11-provider, wolfSSL PKCS#11) have also been verified, including tls.reload.

With this flag on, the main changes are:

- no use of shared memory allocators in OpenSSL/wolfSSL; the functions are confined to PROC_TCP_MAIN
-  the pthread monkey patches become no-ops; these are needed in multi-process mode so that `SSL *` objects can be shared by workers

Multi-process mode is still fully supported (including tls.reload) and also tested with OpenSSL 4 ~~beta~~.

The user can revert to the previous behaviour by setting `tcp_main_threads` to 0 in the config. 

This is part of the long term strategy that Kamailio be less exposed to the (fragile) internals of OpenSSL.

### DIFF
--- a/src/core/tcp_main.c
+++ b/src/core/tcp_main.c
@@ -173,7 +173,7 @@ int tcp_main_max_fd_no = 0;
 int tcp_max_connections = DEFAULT_TCP_MAX_CONNECTIONS;
 int tls_max_connections = DEFAULT_TLS_MAX_CONNECTIONS;
 int tcp_accept_unique = 0;
-int ksr_tcp_main_threads = 0;
+int ksr_tcp_main_threads = 1;
 int ksr_tcp_listen_backlog = TCP_LISTEN_BACKLOG;
 int tcp_connection_match = TCPCONN_MATCH_DEFAULT;
 


henningw:
@space88man Thanks for the PR. Quick questions - does the change of the ksr_tcp_main_threads affects other module besides the tls and tls_wolfssl ?

space88man:
> @space88man Thanks for the PR. Quick questions - does the change of the ksr_tcp_main_threads affects other module besides the tls and tls_wolfssl ?

@henningw -  ~~this does not affect other modules; only tls / tls_wolfssl  use tcp_mtops at the moment~~.

Oh wait - this means the thread workers will still be created if you do not use TLS although they are idle. If one does not use TLS then one must set `ksr_tcp_main_threads = 0` to avoid this. That is not so ergonomic for existing configs in non-TLS mode.

I think it would be best to revert this - and recommend that if one uses TLS:
1. Recommend to use `ksr_tcp_main_threads = 1`
2. For `tls_wolfssl` it is mandatory anyway.

henningw:
> > @space88man Thanks for the PR. Quick questions - does the change of the ksr_tcp_main_threads affects other module besides the tls and tls_wolfssl ?
> 
> @henningw - ~this does not affect other modules; only tls / tls_wolfssl use tcp_mtops at the moment~.
> 
> Oh wait - this means the thread workers will still be created if you do not use TLS although they are idle. If one does not use TLS then one must set `ksr_tcp_main_threads = 0` to avoid this. That is not so ergonomic for existing configs in non-TLS mode.
> 
> I think it would be best to revert this - and recommend that if one uses TLS:
> 
>     1. Recommend to use `ksr_tcp_main_threads = 1`
>     2. For `tls_wolfssl` it is mandatory anyway.

Thank you, sounds good. We probably should also add some content to the TLS module docs about this.

PR #4699, opened 2026-04-16
topos: restore port/transport in contact URI for contact_mode=0

thanhloi2603:
In 6.1 the contact-building logic in tps_storage_fill_contact was refactored into helper functions. The else branch (ctmode=0 SKEYUSER and ctmode=3 XAVPHOST) was unified into a single append_host_info call with append_port_proto=0, which silently dropped port and transport from the generated contact URI for ctmode=0.

In 6.0 the equivalent inline code always appended port and transport, producing contacts like <sip:bXXX@host:port;transport=tcp>. ctmode=3 (XAVPHOST) did not exist in 6.0, so its omission of port/transport is intentional and is kept as-is.

Fix: pass append_port_proto=1 for TPS_CONTACT_MODE_SKEYUSER only.

#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to pull request https://github.com/kamailio/kamailio/pull/4246

### DIFF
--- a/src/modules/topos/tps_storage.c
+++ b/src/modules/topos/tps_storage.c
@@ -317,7 +317,9 @@ int tps_storage_fill_contact(
 			vavu = get_xavu_host(dir);
 		}
 
-		append_host_info(td, &puri, vavu, 0);
+		/* ctmode=0 preserves port/transport from original URI; ctmode=3 does not */
+		append_host_info(
+				td, &puri, vavu, (ctmode == TPS_CONTACT_MODE_SKEYUSER) ? 1 : 0);
 	}
 
 	/* Finalize the contact header */


xkaraman:
Hey @thanhloi2603,

The clang check is performed per commit therefore, you will need to adopt your first commit to match the expected format and then force push the formatted commit.

Thanks for the PR, i think it's reasonable fix, let's see if other have something to say as well.

thanhloi2603:
thanks @xkaraman for the instruct, done it.

celliso1:
> Thanks for the PR, i think it's reasonable fix, let's see if other have something to say as well.
+1

Anecdotally, I was troubleshooting why 200 OK in my dev environment weren't getting ACK over TLS, and found the port and transport had gone missing. I had Claude Code check the source, and it spotted the problem before it found this PR.

henningw:
Thanks, merged

Issue #4701, opened 2026-04-20
http_async_client [http_multi.c:88]: event_cb(): Cell for handler not found in table


braams:
### Description

When using http_async_client, I'm getting the following error:

`4(123386) INFO: http_async_client [http_multi.c:88]: event_cb(): Cell for handler 0xffff7adeddb0 not found in table`

It occurs repeatedly and causes a brief spike in CPU usage.
This is mostly a duplicate of https://github.com/kamailio/kamailio/issues/4431 but I've figured out how to reproduce and fix the issue

#### Reproduction

Here is the minimal configuration that allows me to reproduce the issue:

```
#!KAMAILIO

debug=2

loadmodule "tm.so"
loadmodule "pv.so"
loadmodule "http_async_client.so"
loadmodule "timer.so"
loadmodule "xlog.so"

modparam("timer", "declare_timer", "TIMER1=ON_TIMER,10,fast,enable");

request_route {
   return;
}

route[ON_TIMER] {
   $http_req(all) = $null;
   $http_req(suspend) = 0;
   http_async_query("http://localhost/", "HTTP_REPLY");
}

route[HTTP_REPLY] {
   $shv(cnt)=$shv(cnt)+1;
   xlog("L_INFO", "cnt: $shv(cnt)\n");
}

```

A standard nginx server with the default configuration is used as the HTTP server 

#### Log Messages

```
 4(123386) INFO: <script>: cnt: 1
 4(123386) INFO: <script>: cnt: 2
 4(123386) INFO: <script>: cnt: 3
 4(123386) INFO: <script>: cnt: 4
 4(123386) INFO: <script>: cnt: 5


 4(123386) INFO: <script>: cnt: 997
 4(123386) INFO: <script>: cnt: 998
 4(123386) INFO: <script>: cnt: 999
 4(123386) INFO: <script>: cnt: 1000
 4(123386) INFO: <script>: cnt: 1001
 4(123386) INFO: http_async_client [http_multi.c:88]: event_cb(): Cell for handler 0xffff7adeddb0 not found in table
 4(123386) INFO: <script>: cnt: 1002
 4(123386) INFO: http_async_client [http_multi.c:88]: event_cb(): Cell for handler 0xffff7adeddb0 not found in table
 4(123386) INFO: <script>: cnt: 1003
 4(123386) INFO: http_async_client [http_multi.c:88]: event_cb(): Cell for handler 0xffff7adeddb0 not found in table
[...]

Issue #4431, opened 2025-10-13
http_async_client [http_multi.c:88]: event_cb(): Cell for handler 0x7f78f8c79b10 not found in table


fulhade:
Testing 6.0.3 noticed that log is filled with INFO log messages from title (every 5s). No errors in logs, usual stuff except info on tcpcon

```
Oct 12 00:40:24 kamailio[4350]: INFO: <core> [core/tcp_main.c:3124]: tcpconn_1st_send(): quick connect for 0x7f78f8c79b10 sock 38
Oct 12 22:42:26 kamailio[4355]: INFO: <core> [core/tcp_main.c:3124]: tcpconn_1st_send(): quick connect for 0x7f78f8c79b10 sock 32
...
Oct 13 07:42:12 kamailio[4355]: INFO: http_async_client [http_multi.c:88]: event_cb(): Cell for handler 0x7f78f8c79b10 not found in table
Oct 13 07:42:17 kamailio[4355]: INFO: http_async_client [http_multi.c:88]: event_cb(): Cell for handler 0x7f78f8c79b10 not found in table
Oct 13 07:42:22 kamailio[4355]: INFO: http_async_client [http_multi.c:88]: event_cb(): Cell for handler 0x7f78f8c79b10 not found in table
Oct 13 07:42:27 kamailio[4355]: INFO: http_async_client [http_multi.c:88]: event_cb(): Cell for handler 0x7f78f8c79b10 not found in table
Oct 13 07:42:32 kamailio[4355]: INFO: http_async_client [http_multi.c:88]: event_cb(): Cell for handler 0x7f78f8c79b10 not found in table
Oct 13 07:42:37 kamailio[4355]: INFO: http_async_client [http_multi.c:88]: event_cb(): Cell for handler 0x7f78f8c79b10 not found in table
```

Using docker image built on trixie: https://pastebin.com/0wZ0X9wb

```
ii  libcurl3t64-gnutls:amd64           8.14.1-2                             amd64        easy-to-use client-side URL transfer library (GnuTLS flavour)
ii  libcurl4t64:amd64                  8.14.1-2                             amd64        easy-to-use client-side URL transfer library (OpenSSL flavour)
```

Not sure how to reproduce, it just appear on testing machine after couple of days.

Unfortunately we used 5.6.3 before this, so not sure when did it break for us. http_async_client usage:

```
$http_req(hdr) = "Content-Type: application/json";
$http_req(hdr) = "Expect: ";
[...]

PR #4702, opened 2026-04-20
http_async_client: freeing event upon request completion

braams:
#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue [#4701](https://github.com/kamailio/kamailio/issues/4701) 

#### Description

check_multi_info does not free the libevent event upon normal request completion.

However, the event is registered with the EV_PERSIST flag (setsock, line 703) — it is not deleted automatically. As a result, libevent continues to call event_cb with the same pointer, but the cell no longer exists in the table.
This leads to a large number of the following messages:
`4(123386) INFO: http_async_client [http_multi.c:88]: event_cb(): Cell for handler 0xffff7adeddb0 not found in table`

### DIFF
--- a/src/modules/http_async_client/http_multi.c
+++ b/src/modules/http_async_client/http_multi.c
@@ -682,6 +682,13 @@ void check_multi_info(struct http_m_global *g)
 
 			if(cell != 0) {
 				LM_DBG("cleaning up cell %p\n", cell);
+				if(cell->evset && cell->ev) {
+					LM_DBG("freeing event %p\n", cell->ev);
+					event_del(cell->ev);
+					event_free(cell->ev);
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the PR. @grumvalski Any comments from your side? Its interesting that it (according to the #4431) seems to be broken for newer distributions.

grumvalski:
Thank you! It looks good even if it's not clear to me why this behaviour started happening with this curl PR: https://github.com/curl/curl/pull/14561. Let's merge it!

henningw:
Thank you, can you also cherry-pick to stable branches?

PR #4704, opened 2026-04-21
sanity: Add more RFC3261 checks

xkaraman:
#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [] Related to issue #XXXX (replace XXXX with an open issue number)

#### Description

Add RFC3261 compliance to Route headers by checking lr paramater

### DIFF
--- a/src/modules/sanity/sanity.c
+++ b/src/modules/sanity/sanity.c
@@ -407,7 +407,6 @@ int check_rfc3261_compliance(sip_msg_t *msg)
 	if(msg->via1->branch->value.len < 7
 			|| memcmp(msg->via1->branch->value.s, "z9hG4bK", 7) != 0) {
 		LM_WARN("Via1 branch parameter does not start with the magic cookie\n");
-		msg->msg_flags |= FL_MSG_NOREPLY;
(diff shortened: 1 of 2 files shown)


src/modules/sanity/sanity.c
```diff
@@ -443,6 +443,38 @@ int check_rfc3261_compliance(sip_msg_t *msg)
 		hf = next_sibling_hdr(hf);
 	}
 
+	/* Check lr parameter for Route */
+	LM_DBG("check Route lr parameter for all Route headers\n");
+
+	if(parse_route_headers(msg) != 0) {
+		LM_WARN("failed to parse the Route headers\n");
+		msg->msg_flags |= FL_MSG_NOREPLY;
+		return SANITY_CHECK_FAILED;
```
miconda, line 446: Any reason to set FL_MSG_NOREPLY here and below? This header should have no impact in generating a failure reply.

I think also for Via cookie check, if not-RFC3261 compliant, a reply can still be generated if wanted.

xkaraman:
Updated and resolved remarks

PR #4729, opened 2026-05-08
rtpengine: Ping all the rtpengine in parallel.

rajneeshksoni:
- In periodic ping of rtpengine, all the configured rtpengines are pinged sequentially. For small number of rtpengine it is not a problem But for large number of rtpengine where some of them are not responding this periodic timer can take long time.sequential ping takes N * timeout seconds for N dead nodes.

- use poll() to wait for responses from all the nodes simultaneously. and then retry only the nodes which are dead.

- It also helps in kamailio start up time, where child_init pings all the nodes during startup.

#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -3103,11 +3103,15 @@ static int child_init(int rank)
 	/* Iterate known RTPEngine instances - create sockets */
 	if(rank == PROC_SIPINIT) {
 		/* probe rtpengines only in first worker */
-		if(build_rtpp_socks(0, rtpengine_ping_mode))
+		if(build_rtpp_socks(0, 0))
 			return -1;
 		else {
 			if(build_rtpp_socks(0, 0))
 				return -1;
+		}
+		if(rtpengine_ping_mode) {
+			/* test the rtpengine with ping */
+			rtpengine_ping_check_timer(0, NULL);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/modules/rtpengine/rtpengine.c
```diff
@@ -4070,6 +4074,274 @@ static bencode_item_t *rtpp_function_call_ok(bencode_buffer_t *bencbuf,
...
+		/* Setup pollfd */
+		fds[j].fd = rtpp_socks[nodes[idx].idx];
+		fds[j].events = POLLIN;
+		fds[j].revents = 0;
+
+		/* Drain stale data from socket (once) */
+		while((poll(&fds[j], 1, 0) == 1) && (fds[j].revents & POLLIN)) {
+			recv(fds[j].fd, recv_buf, sizeof(recv_buf), 0);
+			fds[j].revents = 0;
+		}
```
rfuchs, line 4176: I'm confused about this part - what purpose does this have?

rajneeshksoni: I want to flush out any pending data from the socket, which is present before sending the ping request. Say some delayed response for some previous command.
This avoid "stale" response to be considered the response of the "ping".

src/modules/rtpengine/rtpengine.c
```diff
@@ -4070,6 +4074,274 @@ static bencode_item_t *rtpp_function_call_ok(bencode_buffer_t *bencbuf,
...
+ * @param nodes Local copy of rtpp_node array (safe to modify without locks).
+ * @param total_nodes Total number of entries in the nodes array.
+ */
+static void rtpengine_ping_nodes_udp_parallel(
+		struct rtpp_node *nodes, int total_nodes)
+{
+	struct rtpe_ping_ctx *ctx = NULL;
+	struct pollfd *fds = NULL;
+	bencode_item_t *dict;
+	char recv_buf[0x10000]; /* 64KB - pong replies are small */
```
rfuchs, line 4106: 64k is quite a lot for something that is advertised as "small" 😁

src/modules/rtpengine/rtpengine.c
rfuchs: Yes ok, but why? 😄 The processing below is able to distinguish between stray data and actual ping responses thanks to the cookie. I mean, I don't mind if this is here, but it just seems redundant?

src/modules/rtpengine/rtpengine.c
```diff
@@ -4070,6 +4074,274 @@ static bencode_item_t *rtpp_function_call_ok(bencode_buffer_t *bencbuf,
...
+		}
+		ctx[j].v = bencode_iovec(dict, &ctx[j].vcnt, 1, 0);
+		if(!ctx[j].v) {
+			LM_ERR("bencode iovec failed for node %s\n", nodes[idx].rn_url.s);
+			bencode_buffer_free(&ctx[j].bencbuf);
+			ctx[j].responded = -1;
+			fds[j].fd = -1;
+			j++;
+			continue;
+		}
```
rfuchs: Since you're doing the sending and receiving manually, if you want you could hard-code the format of the ping message. It would make this whole section much simpler. Up to you.

src/modules/rtpengine/rtpengine.c
```diff
@@ -4070,6 +4074,274 @@ static bencode_item_t *rtpp_function_call_ok(bencode_buffer_t *bencbuf,
...
+			nodes[idx].rn_recheck_ticks =
+					get_ticks()
+					+ cfg_get(rtpengine, rtpengine_cfg, rtpengine_disable_tout);
+		}
+	}
+
+	/* CLEANUP: free bencode buffers and tracking arrays */
+	for(j = 0; j < udp_count; j++) {
+		if(ctx[j].v) /* bencode was initialized */
+			bencode_buffer_free(&ctx[j].bencbuf);
```
rfuchs: This could be moved into the loop just above.

rfuchs (commented):
Overall LGTM, just a few comments

src/modules/rtpengine/rtpengine.c
rajneeshksoni: Thanks @rfuchs 
As suggested i have removed the dynamic bencode buffer and replaced with hardcoded value.
It simplified the code a lot.

Issue #4743, opened 2026-05-14
siputils: tel2sip() may return success with non-parsable SIP URI


Ozzyboshi:
### Description

While testing tel2sip(), I observed that the function may return success (1) even when the generated SIP URI is not valid and cannot be parsed by internal parsing functions such as parse_uri().

After validating the input parameters, tel2sip() constructs the resulting SIP URI but does not perform any validation on the final output before returning success.

This can lead to inconsistent behavior in later routing stages, where the malformed URI is used assuming it is valid.

### Reproduction

The issue can be reproduced with the following snippet (e.g. in request_route):

```
$var(tmp) = "tel:playcfannounce@217.15.95.99;user=phone";
$var(tmp_domain) = "10.10.151.23";
tel2sip("$var(tmp)", "$var(tmp_domain)", "$var(tmp)");
xlog("L_NOTICE", "tel2sip launched with «$var(tmp)» and domain «$var(tmp_domain)» returned $retcode\n");
```

This produces output similar to:

```
tel2sip launched with «sip:playcfannounce@217159599;user=phone@10.10.151.23;user=phone» and domain «10.10.151.23» returned 1
```

The function returns 1 (success) even though the resulting SIP URI is malformed
The generated URI is not valid and cannot be parsed correctly

What I would expect:
  - The function should return an error (-1)
  - the function should ensure the generated URI is parsable before returning success

### Additional Information 

```
lsb_release -a
Linux sp2 6.12.85+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.85-1 (2026-04-30) x86_64 GNU/Linux
root@sp2:~# lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 13 (trixie)
Release:	13
Codename:	trixie

 ```
 
### Operating System: 
```
uname -a
Linux sp2 6.12.85+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.85-1 (2026-04-30) x86_64 GNU/Linux
```

miconda:
Referenced PR was merged.

PR #4744, opened 2026-05-14
siputils: validate SIP URI produced by tel2sip and tel2sip2

Ozzyboshi:
#### Pre-Submission Checklist
  from source code folder

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #4743 

#### Description
After constructing the SIP URI, call parse_uri() to verify the result is well-formed before writing it to the output pvar. 
If validation fails, free the allocated pkg memory and return -1.


linuxmaniac:
Would be nice to add the same validation to tel2sip2() too

Ozzyboshi:
I did not know about tel2sip2, they are almost identical and the same problem is there, I tested locally even if I do not use it.
My pr is now updated.

linuxmaniac:
Thanks!

PR #4746, opened 2026-05-17
Extended HAproxy protocol support

sergey-safarov:
#### Pre-Submission Checklist
  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [ ] Related to issue #xxxx

#### Description
In PR extended HAproxy protocol support.
For server and client IP addresses used values from HAproxy header.
Via headers generated by the values from Haproxy protol.
Adjusted `path`, `rr`, `siptrace` module to use IP address values from HAproxy protocol.

For PR recommended review by commits.

### DIFF
--- a/src/core/msg_translator.c
+++ b/src/core/msg_translator.c
@@ -115,6 +115,7 @@
 
 extern char version[];
 extern int version_len;
+extern int ksr_tcp_accept_haproxy;
 
 str _ksr_xavp_via_params = STR_NULL;
 str _ksr_xavp_via_fields = STR_NULL;
@@ -588,6 +589,9 @@ static inline int lumps_len(
 	int recv_proto_id = PROTO_NONE;
 	int recv_af = 0;
 	struct socket_info *send_sock;
+	char ip_buf[MAX_URI_SIZE] = {0};
+	str _recv_address_str = STR_NULL;
+	str _recv_port_str = STR_NULL;
 
 
 #ifdef USE_COMP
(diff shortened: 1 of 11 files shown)


miconda:
The changes are quite big, probably the PR will take some time to review, maybe others interested in HAProxy protocol can take a look at it, too.

sergey-safarov:
Hello @rfuchs 
Probably you use HAproxy on your servers and HAproxy extended  support may useful for you also.
Could you look this PR?

sergey-safarov:
@miconda does it make sense to create dedicated PR for first commit and split this commit to more 3-5 smaller commits.
Probable it will make change more easy to review and more understandable.

miconda:
The PR is quite big, I started a few time, but I haven't finished.

One thing that has to be update is the condition to check if HAPROXY protocol is enabled, it has to be:

`(ksr_tcp_accept_haproxy || (ksr_tcp_accept_protocols & KSR_TCPAP_HAPROXY))`

Besides the old global variable, there is also a flag added meanwhile. Probably a macro/define can be created for the condition.

Then try to fix the existing conflicts, maybe soon the PR can be reviewed, others are welcome to do it as well.

miconda:
@sergey-safarov: there seems to be a risk of a double tcp connection put in siptrace at lines 1361 and 1367, if the condition there is not true to return:

- https://github.com/sergey-safarov/kamailio/blob/e4d03614909d8784e077bd4affabe38523686b3b/src/modules/siptrace/siptrace.c#L1361

The HAPROXY protocol is not for UDP in this PR, right? If yes, then is_haproxy() should check the `send_info->proto` -- it is very rare but a tcp target can have same ip and port as a udp target.

Furthermore, _tcpconn_find() seems to invoke _tcpconn_find_haproxy_rcv() after every alias-table miss, even when HAProxy support is disabled or no HAProxy connections exist, walking every ID hash bucket and connection -- this can add quite some performance penalty when a lot of tcp/tls connections are open.

sergey-safarov:
> The HAPROXY protocol is not for UDP in this PR, right?

Main use case in AWS. Yes AWS does not support HAProxy over UDP protocol.
I make search for other cloud providers who support HAProxy over UDP and was not able find any one.

If somebody will find then we can check UDP case also.

I want split first commit to set of smaller commit.
I want review my PR again.
I will mark PR ass draft.

PR #4751, opened 2026-05-21
[5.0] backport CVE-2020-27507: duplicate headers in tm local requests

dkgkdfg65:
`ada3701d22b1` makes `build_local_reparse` track which of CSeq/To/From/Call-ID it's already emitted and bail on a duplicate. without it a crafted request with duplicate headers is rebuilt with the duplicates intact — header injection / smuggling. missing on 5.0.

modeled the header-copy loop with an INVITE carrying two From headers: pre-fix both copy through, post-fix the duplicate is detected and the request rejected. (modeled the control flow rather than a full daemon build.) clean cherry-pick.

upstream: https://github.com/kamailio/kamailio/commit/ada3701d22b1
CVE: https://nvd.nist.gov/vuln/detail/CVE-2020-27507

old CVE; companion PR for 5.1 alongside this.


miconda:
5.0 is out of maintenance.

PR #4752, opened 2026-05-21
[5.1] backport CVE-2020-27507: duplicate headers in tm local requests

dkgkdfg65:
same issue as the 5.0 PR, for the 5.1 branch. `ada3701d22b1` teaches `build_local_reparse` to track which of CSeq/To/From/Call-ID it has already emitted and skip duplicates; without it a crafted request with duplicate headers gets rebuilt with the duplicates intact (header injection / smuggling). missing on 5.1.

checked the header-copy path with an INVITE carrying two From headers: pre-fix both copy through, post-fix the second is dropped and the request rejected. clean cherry-pick.

upstream: https://github.com/kamailio/kamailio/commit/ada3701d22b1
CVE: https://nvd.nist.gov/vuln/detail/CVE-2020-27507


miconda:
5.1 is out of maintenance

PR #4755, opened 2026-05-25
cplc: trace level and crash

fgaisnon:
#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

cplc:
- modify trace level when a 3XX response is managed and ignore3xx is activated
- crash: Sometimes cplc module wants to unlink lump on current sip message but the pointer is no more valid and a crash occurs. Moreover the lump created is also destroyed when freply is called like sip message (it was the root cause of this crash). So this patch does not introduce any memory leak as freply already have done the stuff by deleting lump.

### DIFF
--- a/src/modules/cplc/cpl_proxy.h
+++ b/src/modules/cplc/cpl_proxy.h
@@ -149,7 +149,12 @@ static void reply_callback(struct cell *t, int type, struct tmcb_params *ps)
 	int rez;
 
 	if(intr == 0) {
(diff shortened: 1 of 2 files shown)


fgaisnon:
this patch is in production since more than one year on our platform without any memory leak and any crash

miconda:
Merged manually.

PR #4757, opened 2026-05-28
rtpengine:Fix UDP socket leak when rtpengine is removed.

rajneeshksoni:
build_rtpp_socks() was creating connected UDP sockets for all nodes in the linked list, including soft-deleted ones (rn_displayed=0). Since rtpengine_delete_node() never removes nodes from the list, every reload after a removal would re-create a live ESTAB socket to the removed instance — one per worker process.

Fix: skip nodes with rn_displayed=0 in build_rtpp_socks(), avoiding both socket creation and the rtpp_test ping for deleted nodes.

Note: after a reload, worker processes rebuild their sockets lazily (on next request). So a stale socket may remain briefly on idle workers until they process their next request and detect the version change — at which point the socket is closed and not re-created.

#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

### DIFF
--- a/src/modules/rtpengine/rtpengine.c
+++ b/src/modules/rtpengine/rtpengine.c
@@ -2858,6 +2858,9 @@ static int build_rtpp_socks(int lmode, int rtest)
 
 		rtpe_reload_lock_get(rtpp_list->rset_lock);
 		for(pnode = rtpp_list->rn_first; pnode != 0; pnode = pnode->rn_next) {
+			if(!pnode->rn_displayed) {
+				continue;
+			}
 			char *hostname;
 			char *hp;
 


rfuchs:
I wouldn't call this an actual leak, the extra sockets would be closed again at the next iteration. But a good improvement nevertheless.

rajneeshksoni:
> I wouldn't call this an actual leak, the extra sockets would be closed again at the next iteration. But a good improvement nevertheless.

The disabled “rtpengine” always remains in the link-list, so the port gets opened again in the next iteration. In scenarios where rtpengine is added/removed frequently, I am observing that the number of bound ports keeps increasing in production box.

Keeping disabled rtpengine in the link-list is kind of a memory leak as well, but I believe removing nodes from the list is a bit bigger change.

rfuchs:
> The disabled “rtpengine” always remains in the link-list, so the port gets opened again in the next iteration. In scenarios where rtpengine is added/removed frequently, I am observing that the number of bound ports keeps increasing in production box.

That shouldn't really happen because the ports get closed right at the beginning of the function: https://github.com/kamailio/kamailio/blob/master/src/modules/rtpengine/rtpengine.c#L2841

If you see the used ports going up, then probably something else is happening.

PR #4759, opened 2026-05-28
cmake: fix build race condition on machine with many cores

mtryfoss:
#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description
Got a random error from build when building on a machine with 8 cores. Seems to fix the issue, but please have a look.

gmake[3]: *** [CMakeFiles/dbschema_pi_framework_mod.dir/build.make:177: dbschema_pi_framework_mod] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:6001: CMakeFiles/dbschema_pi_framework_mod.dir/all] Error 2

### DIFF
--- a/cmake/dbschema.cmake
+++ b/cmake/dbschema.cmake
@@ -246,10 +246,11 @@ if(XSLTPROC_EXECUTABLE
 )
   add_custom_target(dbschema_xhttp_pi)
   add_dependencies(dbschema_xhttp_pi dbschema_pi_framework_table dbschema_pi_framework_mod)
+  add_dependencies(dbschema_pi_framework_mod dbschema_pi_framework_table)
 
   add_custom_target(dbschema_xhttp_pi_clean)
   add_dependencies(
-    dbschema_xhttp_pi dbschema_pi_framework_table_clean dbschema_pi_framework_mod_clean
+    dbschema_xhttp_pi_clean dbschema_pi_framework_table_clean dbschema_pi_framework_mod_clean
   )
 
   add_dependencies(dbschema dbschema_pi_framework_table dbschema_pi_framework_mod)


xkaraman:
Hey @mtryfoss,

Thanks for bringing this up!

I believe this one change is enough

   ```
    dbschema_xhttp_pi_clean dbschema_pi_framework_table_clean dbschema_pi_framework_mod_clean
   ```
as before the `dbschema_xhttp_pi` target had dependencies also on the clean targets which of course doesnt make sense and might clean just after it was build.

Can you verify with only that change?

i dont think this change `  add_dependencies(dbschema_pi_framework_mod dbschema_pi_framework_table)` add something as both are independent i believe.

mtryfoss:
Tried to remove the line you suggested, but then the error appeared again on first attempt.

The only change then is this:
```
-    dbschema_xhttp_pi dbschema_pi_framework_table_clean dbschema_pi_framework_mod_clean
+    dbschema_xhttp_pi_clean dbschema_pi_framework_table_clean dbschema_pi_framework_mod_clean
```

xkaraman:
Can you provide how you are calling cmake and a bit more of the logs?

mtryfoss:
Have a look at the relevant output. It's an RPM build based on the shipped spec on a fully up to date Rocky 10 server.
The error doesn't trigger on all runs, so there's some timing involved.

[build-error.txt](https://github.com/user-attachments/files/28347575/build-error.txt)

xkaraman:
I can't seem to reproduce this all. Can it be that any permissions are related on this? 

I can see that the builds fail when trying to copy the files into the source dir `utils/kamctl/xhttp_pi`. 

Somehow the cmake command `copy_directory` fails, might be that permisions are changed during the copying (ie one docker compilation using volume for src code and copies as root) and then host compilation tries to overwrite but cant? 

(this one i replicated. inside a docker rhel based container as root, configured and building dbschema target, copied the files succesfully on the source directory which was binded on the container. hence now host directory had some root owned files in `utils/kamctl/db*` folders and then trying a host based compilation produced the same results as you)

other than this [copy_directory](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-E-arg-copy_directory) states `Added in version 3.15: The command now fails when the source directory does not exist. Previously it succeeded by creating an empty destination directory.` I think they meant destination directory does not exist, but i can't see how this is relevant to us are directory should be there from kamailio repo

mtryfoss:
Processing uri_db for pi_framework_mod
cd /home/mtryfoss/rpmbuild/BUILD/kamailio-6.1.3/src/lib/srdb1/schema && /usr/bin/cmake -E env XML_CATALOG_FILES=/home/mtryfoss/rpmbuild/BUILD/kamailio-6.1.3/doc/stylesheets/dbschema_k/catalog.xml /usr/bin/xsltproc --xinclude --stringparam dir /home/mtryfoss/rpmbuild/BUILD/kamailio-6.1.3/redhat-linux-build/utils/kamctl/xhttp_pi/'' --stringparam prefix uri_db- --stringparam db pi_framework_mod /home/mtryfoss/rpmbuild/BUILD/kamailio-6.1.3/doc/stylesheets/dbschema_k/xsl/pi_framework_mod.xsl kamailio-uri_db.xml

..this part of that line seems bogus (the single quotes): 
`xhttp_pi/''`

Some var not being ready or with correct value, and the time of copy.

Really seems like just a standard race condition to me, when running on many cpus.
dbschema_pi_framework_mod must wait for dbschema_pi_framework_table to finish.

mtryfoss:
Did some more experiments. Seems like it's working if building with -j4 or less.

xkaraman:
> ..this part of that line seems bogus (the single quotes):
> `xhttp_pi/''`

Yeah i see your point, but actually that's handled just fine and traversed fine. path is expanded just fine to `xhttp_pi/` and doing the command manually with various variations like `xhttp_pi/""` or `xhttp_pi/` work as expected and copies to correct location. I don't like it either but couln't find an easy way to tell cmake that some `dbs` need the `kamailio` sub-folder and some other are not (var defaults to `''`, hence what you are seeing)

> Really seems like just a standard race condition to me, when running on many cpus.

I know, it looks like it, and i am not saying it isn't. I just tried on multiple docker images like `rhel8`, `ubuntu`, `rocky-10`, with `-j8` and i didn't replicate it, so i am trying to remove any other possibilities.

> dbschema_pi_framework_mod must wait for dbschema_pi_framework_table to finish.

This is not true, the files are generated independently and can be generated in the same time (they produce different files also). they are produced though in the same folder which is `build_folder/utils/kamctl/xhttp_pi/` as you said.

Each of these are a target by their own, and they have the `custom_command(... POST_BUILD)` to copy the result files. 

This is the only place i can see the race come into place, when one of the target has completed and tries to copy the whole folder but the other target still produces it's own files. Then copying the files that already exist, even some of the other target is fine. it might some files are still being written and can't be accessed there maybe? i guess so. 

i ll try to isolate this and if not, the PR looks good, it won't mess anything else i think, since it just adds a dependency to build each relevant target sequentially.

xkaraman:
> Did some more experiments. Seems like it's working if building with -j4 or less.

This also provide some race condition hint...

xkaraman:
Can you compile with [latest cmake 4.3.x +](https://cmake.org/download/#latest)  that includes this https://gitlab.kitware.com/cmake/cmake/-/merge_requests/11401 fix to have some more verbose information why the copy_directory failed?

Just download the .sh file, run it to extract to a subfolder (or whatever you are more comfortable) and use that cmake binary to configure and build?

mtryfoss:
/home/mtryfoss/cmake/cmake-4.3.3-linux-x86_64/bin/cmake -E copy_directory /home/mtryfoss/rpmbuild/BUILD/kamailio-6.1.2/redhat-linux-build/utils/kamctl/db_postgres/'' /home/mtryfoss/rpmbuild/BUILD/kamailio-6.1.2/utils/kamctl/postgres/''
Error copying directory from "/home/mtryfoss/rpmbuild/BUILD/kamailio-6.1.2/redhat-linux-build/utils/kamctl/xhttp_pi/" to "/home/mtryfoss/rpmbuild/BUILD/kamailio-6.1.2/utils/kamctl/xhttp_pi/": No such file or directory
gmake[3]: *** [CMakeFiles/dbschema_pi_framework_mod.dir/build.make:180: dbschema_pi_framework_mod] Error 1
gmake[3]: Leaving directory '/home/mtryfoss/rpmbuild/BUILD/kamailio-6.1.2/redhat-linux-build'
gmake[2]: *** [CMakeFiles/Makefile2:7359: CMakeFiles/dbschema_pi_framework_mod.dir/all] Error 2

xkaraman:
I couldn't replicate reliable but it happened for a me once or twice after lot of repetitions. 

Using the same folder for producing and copying the files from, or even creating that folder, probably is the cause.

The fix is reasonable, just add a dependency for one of related targets (`dbschema_pi_framework_mod` or `dbschema_pi_framework_table`) , so that the don't collide.

I think this can be merged as is, thanks for the patience!

henningw:
Thank you both, merged.

PR #4762, opened 2026-06-01
Adding support for vendor specific radius attributes.

neutrino38:
#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Adding support radius dictionnaries with vendor specific attribute: e.g.

#
# SIP RADIUS attributes for IVeS
#
VENDOR    Ives  22737

BEGIN-VENDOR    Ives
ATTRIBUTE           Sip_RPId                213     string      Ives
ATTRIBUTE       Context                 104     string  Ives
ATTRIBUTE       LastApp                 108     string  Ives
ATTRIBUTE       UserField               118     string  Ives
ATTRIBUTE       Remote-IP-Address        23     string  Ives

END-VENDOR      Ives

Otherwise, the vendor ID is not sent into the RADIUS packet which is rejected.

### DIFF
--- a/src/modules/misc_radius/radius.h
+++ b/src/modules/misc_radius/radius.h
@@ -47,6 +47,16 @@
 #ifndef VENDOR
 #define VENDOR(x) (((x) >> 16) & 0xffff)
 #endif
+#ifndef ATTRID
+#define ATTRID(x) ((x) & 0xffff)
+#endif
+#endif
+
+#ifndef VENDOR
+#define VENDOR(x) (((x) >> 16) & 0xffff)
+#endif
+#ifndef ATTRID
+#define ATTRID(x) ((x) & 0xffff)
 #endif
 
 
@@ -55,6 +65,7 @@ struct attr
 	const char *n;
 	int v;
 	unsigned int t; /* type of attribute */
+	unsigned vendor;
 };
 
 struct val
@@ -124,8 +135,9 @@ struct val
 						fn, at[i].n);                 \
 				return e1;                            \
 			}                                         \
-			at[i].v = da->value;                      \
+			at[i].v = ATTRID(da->value);              \
 			at[i].t = da->type;                       \
+			at[i].vendor = VENDOR(da->value);         \
 		}                                             \
 		for(i = 0; i < nr_vl; i++) {                  \
 			if(vl[i].n == NULL)                       \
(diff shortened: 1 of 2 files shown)


src/modules/misc_radius/radius.h
linuxmaniac, line 1: On ``radcli/radcli.h``:
```
#define RADCLI_VENDOR_MASK 0xffffffff
#define VENDOR_BIT_SIZE         32
#define RADCLI_VENDOR_ATTR_SET(attr, vendor) ((attr)|((uint64_t)((vendor)&RADCLI_VENDOR_MASK)) << VENDOR_BIT_SIZE)
 
#define VENDOR(x)               (((x) >> VENDOR_BIT_SIZE) & 0xffffffff)
#define ATTRID(x)               ((x) & 0xffffffff)
```
And on the old ``radiusclient-ng.h``:
```
#define VENDOR(x)               (((x) >> 16) & 0xffff)
#define ATTRID(x)               ((x) & 0xffff)
```
So, 55-60 lines are not necessary, since the definitions are only missing on ``freeradius-client.h``

linuxmaniac (changes requested):

src/modules/misc_radius/radius.h
neutrino38: How would this work if acc_radius is compiled against freeradius-client.h ?

src/modules/misc_radius/radius.h
```diff
@@ -47,6 +47,16 @@
 #ifndef VENDOR
 #define VENDOR(x) (((x) >> 16) & 0xffff)
 #endif
+#ifndef ATTRID
+#define ATTRID(x) ((x) & 0xffff)
+#endif
```
linuxmaniac, line 49: What I'm trying to say is that only this is needed in case of ``freeradius-client.h``
adding ``ATTRID`` definition

src/modules/misc_radius/radius.h
```diff
@@ -47,6 +47,16 @@
...
+#endif
+
+#ifndef VENDOR
```
linuxmaniac: these two definitions are not needed outside ifndef blocks in my opinion

linuxmaniac (changes requested):

linuxmaniac:
pushed 399980d955 with my suggested changes. Thanks!

Issue #4671, opened 2026-03-30
P-CSCF crash in ims_registrar_pcscf during concurrent REGISTER and de-REGISTER


kongjun586999:
### Description
Under high-rate concurrent registration activity for the same UE, Kamailio P-CSCF may crash while processing IMS REGISTER requests.
Based on the observed logs, the failure happens in the ims_registrar_pcscf path, where sec_agree.c:parse_sec_agree() first reports abnormal warnings, and shortly afterwards a child process exits with signal 11 and a core dump is generated.
The issue appears when REGISTER and de-REGISTER style requests for the same contact are processed concurrently. In this situation, one execution path is still handling security-agreement related state, while another path is updating or expiring the same pending registration state. This may lead to inconsistent internal state and eventually a crash.

### Troubleshooting
Deployment mode: Docker
Role: Kamailio as IMS P-CSCF
No core dump was available
#### Observed log sequence
1.Multiple REGISTER requests for the same UE/contact are processed concurrently by P-CSCF.
2.ims_registrar_pcscf reports warnings from sec_agree.c:parse_sec_agree(), including:  qm_free(): WARNING: free(0) called from ims_registrar_pcscf: sec_agree.c: parse_sec_agree(...)
4.The affected requests continue into:
     request_route=[REGISTER]
     pcscf_save_pending
     save_pending()
5.At least some requests are processed as regular registration updates with a non-zero expiry value.
6.Another concurrent request for the same contact is processed with expiry 0, corresponding to a de-registration/unregister-style update.
7.A Kamailio child process then exits with signal 11, and a core dump is generated.
#### Reproduction
The issue can be reproduced by sending high-rate concurrent REGISTER traffic for the same UE against P-CSCF, including both:
repeated REGISTER / re-REGISTER requests for the same contact
concurrent de-REGISTER style requests for the same contact, using expiry 0
In my setup, reproduction was performed in a Docker-based deployment by rapidly and repeatedly issuing registration, re-registration, and de-registration for the same UE so that these requests overlap in processing on the P-CSCF side.

#### Debugging Data

```
(paste your debugging data here)
```

#### Log Messages

```
 3(44) NOTICE: {1 100003144 REGISTER 00000000000000000000@192.168.55.147 REGISTER} <script>: Received IP and Port: (172.22.0.21:5060)
[...]

PR #4764, opened 2026-06-02
ims_pcscf: fix sec-agree shared memory leaks on (re-)registration

qasimloay:
#### Pre-Submission Checklist

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #4670 and #4671

#### Description
Follow-up to the double-free fixes in #4670 / #4671. Those fixed the crash in
`parse_sec_agree()`'s error path, but the P-CSCF sec-agree code still leaks
shared memory on the normal (re-)registration paths:

**ims_registrar_pcscf — `save_pending()`**
- `sec_verify_params` (Security-Verify) is only used to copy SPI/port values
  into `sec_params` and is never stored on the contact → leaked on every
  REGISTER carrying a Security-Verify header.
- `sec_params` (Security-Client) is only handed to the contact on the initial
  registration; on re-REGISTER and on the DB_ONLY de-registration early returns
  it is never stored → leaked on every re-REGISTER.

**ims_ipsec_pcscf — `ipsec_create()`**
- Initial registration overwrites `pcontact->security_temp->data.ipsec` with the
  new `ipsec_t` without freeing the previous one (stored by `save_pending()`) →
  orphaned on every registration.
- On re-registration `req_sec_params->data.ipsec` is only used to build the
  tunnel and is never stored → leaked on every re-REGISTER.

Ownership is tracked (`sec_params_stored`, `ipsec_swapped`) so structures are
freed only on the paths where they are not handed to the contact; no double-free.

Verified with `kamcmd mod.stats all shm`: on a fresh start `ims_registrar_pcscf`
is 0 and `ims_ipsec_pcscf` is just the fixed SPI pool. Before the fix both grew
on every re-registration cycle and never returned to baseline after UEs
de-registered; with the fix the totals plateau under load and return to baseline.

### DIFF
--- a/src/modules/ims_ipsec_pcscf/cmd.c
+++ b/src/modules/ims_ipsec_pcscf/cmd.c
@@ -823,6 +823,34 @@ int add_security_server_header(struct sip_msg *m, ipsec_t *s)
 	return 0;
 }
 
+// Free an ipsec_t (the struct and its sec-agree string fields). Used to release
+// the old data.ipsec that gets replaced on the contact at registration, and the
+// request ipsec_t that is not stored on re-registration - both were previously
+// orphaned and leaked.
+static void free_ipsec_data(ipsec_t *p)
+{
+	if(!p) {
+		return;
+	}
+	if(p->ealg.s)
+		shm_free(p->ealg.s);
+	if(p->r_ealg.s)
+		shm_free(p->r_ealg.s);
+	if(p->ck.s)
+		shm_free(p->ck.s);
+	if(p->alg.s)
+		shm_free(p->alg.s);
+	if(p->r_alg.s)
+		shm_free(p->r_alg.s);
+	if(p->ik.s)
+		shm_free(p->ik.s);
+	if(p->prot.s)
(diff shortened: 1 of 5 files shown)


linuxmaniac:
@qasimloay merge must be for master branch not for 6.0

qasimloay:
@linuxmaniac  Done — rebased onto master and retargeted this PR to master. Happy to have it backported to 6.0 after merge.

linuxmaniac:
please create a commit updating comment at ``ims_usrloc_pcscf/pcontact.c`` pointing to the right function. Right now is:
```
// The same piece of code also lives in modules/ims_registrar_pcscf/sec_agree.c
// Function - parse_sec_agree()
// goto label - cleanup
// Keep them in sync!
void free_security(security_t *_p)
```

qasimloay:
Added a commit updating the free_security() sync comment in ims_usrloc_pcscf/pcontact.c to point to free_security_t().

linuxmaniac:
Thanks!

PR #4780, opened 2026-06-12
core: update max length of path header

bundasmanu:
#### Description

Small update on the maximum length of `Path` header. Update from `256` hard limit to `512`.

For some possible cases, `256` is not enough.
Enabling `r2` + `socket names` + `received` + `parameters`, we can easily surpass that value.

With this, we are more safe.

#### Pre-Submission Checklist
  from source code folder

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

### DIFF
--- a/src/core/config.h
+++ b/src/core/config.h
@@ -61,7 +61,7 @@
 
 #define MAX_URI_SIZE 1024 /*!< Max URI size used when rewriting URIs */
 
-#define MAX_PATH_SIZE 256 /*!< Maximum length of path header buffer */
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the PR. We probably also should check for patch storage size in the database (XML sources).

bundasmanu:
> Thanks for the PR. We probably also should check for patch storage size in the database (XML sources).

It's already 512 on `registrar` and `usrloc` tables.

miconda:
Then this can be probably merged, is a small change.

PR #4781, opened 2026-06-15
websocket: avoid double-initialization of structs

dilyanpalauzov:
This avoids setting twice the values of `ws_event_info_t wsev ` - first to zero, then to something else.

### DIFF
--- a/src/modules/websocket/ws_handshake.c
+++ b/src/modules/websocket/ws_handshake.c
@@ -369,15 +369,9 @@ static void ws_mark_failed_client(ws_connection_t *wsc)
 
 static int ws_send_buffer(int conid, char *buf, unsigned int len)
 {
-	ws_event_info_t wsev;
-	sr_event_param_t evp = {0};
-
-	memset(&wsev, 0, sizeof(ws_event_info_t));
-	wsev.type = SREV_TCP_WS_FRAME_OUT;
-	wsev.buf = buf;
-	wsev.len = len;
-	wsev.id = conid;
-	evp.data = (void *)&wsev;
+	ws_event_info_t wsev = {
+			.type = SREV_TCP_WS_FRAME_OUT, .buf = buf, .len = len, .id = conid};
+	sr_event_param_t evp = {.data = &wsev};
 
 	return (sr_event_exec(SREV_TCP_WS_FRAME_OUT, &evp) < 0) ? -1 : 1;
 }


dilyanpalauzov:
Why is check-commit failing?
```
Run /home/runner/work/kamailio/kamailio/actions/.github/scripts/check-commit.sh
From https://github.com/kamailio/kamailio
 * branch            master     -> FETCH_HEAD
 * [new branch]      master     -> origin/master
fatal: shallow file has changed since we read it
Error: Process completed with exit code 128.
```

miconda:
I am not against PR, but afaik this style of initialisation does not bring any runtime speed improvements. memset() should be usually preferred when one wants everything at byte level to be 0, not sure in this case that would be necessary.

Maybe other devels can have opinions here if makes sense to "prefer" designated fields initialisations instead of memset() and fields assignment.

dilyanpalauzov:
My understanding is that this style of initialization:
- avoids a function call (`memset()`)
- avoids setting twice a value, before reading it.

For the latter, before this change, wsev.buf was first set to zero, and then to `wsev.buf = buf;`.  With this change the step to set to zero is skipped, because it is unnecessary.

miconda:
The memset() is used in many places across the code to ensure that the fields which are not explicitly  set after it are 0/NULL, being safer if new fields are added to a structure that can be used/initialised in many parts of the code. There are many initialisations like that all over the code.

Designated field initialisation on declaration (introduced in C99) should ensure that the other fields (not initialised explicitly by name) are initialised by the compiler, but that might not be as a block of zeros. I haven't checked if all compilers comply and from what version.

I was calling for other developers to express their opinion before deciding to merge. I see it more like a cosmetic change (that can be applied in many other parts of code), rather than a real necessity.

rfuchs:
> Designated field initialisation on declaration (introduced in C99) should ensure that the other fields (not initialised explicitly by name) are initialised by the compiler, but that might not be as a block of zeros. I haven't checked if all compilers comply and from what version.

To my knowledge, this behaviour (zero-initialise of any struct members that aren't listed) is part of the standard and therefore required. Any compiler not doing it would be broken.

> I was calling for other developers to express their opinion before deciding to merge. I see it more like a cosmetic change (that can be applied in many other parts of code), rather than a real necessity.

Theoretically, initialising the struct with a compound literal would be faster than doing a memset and then overwriting the zeroed members, but in practice a good compiler optimises this anyway, and the resulting instructions are very similar or even identical:  https://godbolt.org/z/od8rzMvcK

Personally I prefer direct initialisation with a compound literal as I find it easier to read, but yeah, this is really cosmetic.

dilyanpalauzov:
It is correct, that `-O1` and `-O2` abolish the `memset()` function call. I tried compiling a small program without `-O1`/`-O2` and then observed, that it invokes `memset()`.

PR #4784, opened 2026-06-17
IMS_REGISTRAR_PCSCF: Features-Device-specific call routing (GRUU),Multiple phone numbers per user, Barred identity, Reliable return path to the user handling, Survives restarts 

toharish:
#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

###Features added
1. Device-specific call routing (GRUU) — The proxy can route calls to one specific phone or app on a user’s account, instead of ringing every registered device. Standards: RFC 5627, 3GPP TS 24.229, 3GPP TS 23.228

2. Multiple phone numbers per user — Supports IMS users who have several public identities (e.g. a SIP address and a mobile number). Calls to any of those identities find the right registration. Standards: 3GPP TS 23.228, 3GPP TS 23.003, 3GPP TS 24.229, RFC 3966

3. Barred identity handling — Remembers which identities are “barred” (not allowed for certain services) and keeps that list updated when the network sends notifications. Standards: 3GPP TS 24.229, 3GPP TS 23.008, RFC 3680

4. Reliable return path to the user — Stores the routing path from registration so the proxy can reach the user correctly when someone calls them (mobile-terminated traffic).Standards:  RFC 3327, 3GPP TS 24.229,RFC 3608

5. Survives restarts — GRUU and related registration data are saved in the database, so registrations are not lost after a server restart or reload. 3GPP TS 23.228, 3GPP TS 24.229, 3GPP TS 33.203


miconda:
This PR seems to have quite large changes, maybe people working more with IMS can review it.

@toharish: the `test` commit probably belongs to `kamailio-tests` repository?

toharish:
> This PR seems to have quite large changes, maybe people working more with IMS can review it.
> 
> @toharish: the `test` commit probably belongs to `kamailio-tests` repository?
> 
> * https://github.com/kamailio/kamailio-tests

These test are the unit test done for the module changes probabily should be in kamailio-tests.
there are some more pcscf/iscsf/scscf features as per 3gpp latest features i would like to commit after this,

henningw:
Thanks for the PR. This is indeed a larger one, it needs some closer review. Lets start with some generic remarks:

- the tests should be moved to the kamailio-tests directory
- the new insert_path_on_register parameter should probably default to 0 (disabled), unless there is a good reason for it to enable for everybody
- please add the common GPL licence header to the new files for path and gruu in ims_registrar_pcsf and ims_usrloc_pscf
- the name of the "pcscf_temp_gruu_history" table is a bit misleading, maybe we can just name it "pcscf_gruu_history". A lot of existing tables (e.g. location) contains by design only temporary valid values.
- Why we need the temp_gruu value in the existing table pcscf_location? Just double-checking if its necessary if we are having the pcscf_gruu_history already
- there seems to be no user documentation for the changes in the ims_usrloc_pscf (e.g. RPC commands)
- there seems to be some support for compilation against some unit tests in certain files. Is this something temporary for development and you will remove it later? This is quite a lot of code, you re-define also some core parts etc..

toharish:
### Review follow-up (pushed to `feature/gruu`)

Addressed PR review feedback in follow-up commits (tip `d4f80c07ee`):

**Behavior / docs**
- `insert_path_on_register` default is now **0** (opt-in Path insert); admin XML matches
- GPL/SPDX headers on new registrar/usrloc sources (Harish S / toharishs@gmail.com)
- History table renamed `pcscf_temp_gruu_history` → **`pcscf_gruu_history`** (schema, kamctl, C, tests); `pcscf_location.temp_gruu` kept
- Usrloc admin XML: GRUU/multi-IMPU storage + `ulpcscf.lookup_temp_gruu`
- `#ifdef UNIT_TEST` removed from production `ims_usrloc_pcscf`; unit tests use stubs under `test/modules/ims_usrloc_pcscf/stubs/`

**Path correctness (post-review fix)**
- Path header allocated with `pkg_malloc` (lump ownership; no stack buffer)
- Path URI built as `%.*s;lr` so `pcscf_uri` is not doubled to `sip:sip:`
- Removed stale “default 1” paragraph from admin XML

**Verify:** `bash test/modules/verify_gruu.sh` passed; registrar unit tests (incl. `test_path_build`) pass.

**Note for maintainers:** this branch also still contains some older non-GRUU commits (`ctl`, `websocket`, GH Actions bump) that predate the GRUU work. Happy to split those into a separate PR if preferred.

toharish:
> Thanks for the PR. This is indeed a larger one, it needs some closer review. Lets start with some generic remarks:
> 
> * the tests should be moved to the kamailio-tests directory
> * the new insert_path_on_register parameter should probably default to 0 (disabled), unless there is a good reason for it to enable for everybody
> * please add the common GPL licence header to the new files for path and gruu in ims_registrar_pcsf and ims_usrloc_pscf
> * the name of the "pcscf_temp_gruu_history" table is a bit misleading, maybe we can just name it "pcscf_gruu_history". A lot of existing tables (e.g. location) contains by design only temporary valid values.
> * Why we need the temp_gruu value in the existing table pcscf_location? Just double-checking if its necessary if we are having the pcscf_gruu_history already
> * there seems to be no user documentation for the changes in the ims_usrloc_pscf (e.g. RPC commands)
> * there seems to be some support for compilation against some unit tests in certain files. Is this something temporary for development and you will remove it later? This is quite a lot of code, you re-define also some core parts etc..

 kamailio-tests directory donot contain the actual unit test. In the kamailio repository there is already a test folder which contains the unit tests similar to the test I have added is simiar, and hence it is added under this folder. Other changes required are updated.

PR #4787, opened 2026-06-19
permissions: rework trusted, address and domain tables to use locked hash tables

zenichev:
#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

#### Description

This commit targets the locks safety in scope of processing of address, trusted and domain tables of permissions module.
It doesn't change existing functionality, but improves existing module to be race safe.

This commit introduces the usage of:
- hash-bucket level locks
- upper level locks for the tables themselves

Locks of different levels do not have any contradiction, and fit each other well because:
- upper level locks guarantee that concurrent reloads don't hit the re-init of tables at the same time
- hash-bucket level locks guarantee that at a time of manipulation the content there is no chance for t hings like: iterator invalidation, pointer corruption etc.
- locks are used in the precedence of: 1. First upper level lock; 2. Secondly hash-buckets level locks;

This commit improves the structures organization and from using linked lists (that were considered as hash tables, but indeed were not) re-organizes them into real hash tables, which handle hash buckets (using previously existed linked lists).
This is true for trusted, address and domain tables. As they are the most vulnerable.

This commit deprecates workarounds done previously to make the reload-delta dependent on hardcoded time values in the source code. See `0039c50`.

Reason of the change: gives a race-free usage of tables regardless the amount of data inserted into tables and regardless how many competitive reloads are pending at a time (e.g. 2-3 RPC commands sent at a time to reload same tables).

Other improvements:
- use hash indexes for a quicker lookup in buckets
- improve the mechanism of duplicate entries detection
(use the source ip, ruri pattern, from pattern and the protocol to detect existing similar entries)
- more safety checks in functions manipulating hash tables
(like checking NULL pointers, empty patterns for matching purposes etc.)
- align return values in trusted/hash implementations so that return values are everywhere considered in the similar way
(everything != 0 is considered as failure, 0 is considered as fine)
- split tables allocation and initialization
- add destruction handlers for tables instead of manually handling entries freeing
- some other minor improvements (see more details in commit bodies themselves)


zenichev:
@henningw @miconda 

I was eventually able to get back to this subject, it took only half a year for me to get focused on this again.
I kept previous changes, just improved some handling of tables as discussed previously with @miconda in https://github.com/kamailio/kamailio/pull/4539 

But now additionally introduced refactored structuring of tables for address and domain.
So now we target all those vulnerable tables that suffer from race reloads and other kind of accesses.
Trusted, address and domain tables are the target.

Please review the code and let me know if we can merge it.
My personal ask is to try this time to merge if possible, because last time the code was getting stale, while permissions module was getting some commits and this made previous PR stale.

I did basic tests on my local setup for tables.
Most focus was on the race access during reloads (reload_delta was explicitly disabled to see the actual introduced locks in work).

Let me know if you have any questions.

zenichev:
@henningw Hi, any update here?

miconda:
@zenichev: thanks to putting a lot of effort into this one, the PR is quite large to review, so I am thinking if a different approach should be discussed and decided on it, respectively to get this as a new module `permissions2` or a different name.

Because the module is really used and also security related, I would try to avoid bringing any instability to it. Then the match has to be done linear anyhow, therefore I am not sure about relevant benefits of the hash over the list. Also, in terms of many reloads in very short time, either has to be a global lock that can block SIP workers or there will be some inconsistency on gradual reload. 

Then, people can decide what module to use and if the new one proves to be better based on community feedback over a couple of major releases, then the old one can be archived.

Again, just as a proposal now to have a discussion about this variant...

zenichev:
@miconda I've introduced two level of locks, low level (per bucket) and global (per table).
They ensure that concurrent access to the hash tables will never lead to unexpected races and hence invalidation of data.

What I can offer:
- I let introducing a global lock for tables (this PR already has it, just please review)
- for the per bucket locks and hash tables introduction, we extend the module to have an alternative mode, which works using them (then everyone can decide which mode to take, by default we keep the older one)

Drawback: more code to maintain in the module. But I will be there, so you can always poke me to do any changes if any maintenance required.

henningw:
My suggestion would be to not duplicate the module, if the functionality stays the same and just the internal implementation is different. We would probably end up with two modules, as people would not switch because of inertia. The "old" one would be still get fixes and improvements, we can not deprecate it anymore, and we duplicated the maintenance effort for it.

miconda:
I don't particularly like duplicating, but it is about a potential trading of solid stability of a well established module with a key role in security.

If the benefits are worth, then people will switch. I see it more like the tls variants with openssl or wolfssl, the functionality is the same, but one or the other could be more suitable case by case.

Again, the role is to see the benefits or drawbacks, then decide. So far is preserving a stable option vs duplicated maintenance.

zenichev:
I have two approaches that I can offer.

First approach:
1. I keep the global locks per table (all of them: trusted, address, domains), this is by default (not harmful).
2. I introduce a modparam that controls whether the implementation of containers is older (locks-free linked lists) or the newer one with hash tables which are locks safe.

Second approach:
1. Second module with all of what's been introduced here by default and the co-exists in parallel.

Second approach is less attractive tbh, because this means we will have to keep both of them consistent over the course of time. Which gives a space for human mistakes while maintaining both modules.

Also the first approach is smoother for those people who want to do the switch, meanwhile the second way requires them to load another new module, which will sound to them as a heavier step.

miconda:
Having both old and new (lists and hash tables) types of structure will make the module more complex, rather have different modules than making the existing one heavier to maintain.

I do not see the real benefits (maybe not yet), at least for `address` table that I use to worth such a large change that is not easier to review and test for stability.  Also, for subnet matching, there is also match the longest one, which means going over all records, so hash table vs list makes no difference.

Anyhow, I do not think the matching speed was an issue so far, and regarding reloads, from the design of the PR I think that lookups can keep using an old table while another process (e.g., rpc) frees it.

I recognize that you put a lot of effort in the PR and probably you see benefits for your use case, the proposed variant with a different module was to offer quicker the chance to people to test (eventually in parallel to compare).

Back to the design, with two level locks, one per table and one per bucket:

- if table-lock is acquired by each process using the table, then the bucket-lock should not be necessary, but then only one process can use the table at that moment, practically matching cannot be done in parallel anymore
- if table-lock is NOT acquired by each process using the table, then reloading in parallel can result in races, with a process still accessing an old table that is freed by another one

Have I understood it wrong? If yes, can you provide more details about what locks are acquired on IP matching by a SIP worker process and what happens when a RPC does a reload?

zenichev:
> I do not see the real benefits (maybe not yet), at least for address table that I use to worth such a large change that is not easier to review and test for stability. Also, for subnet matching, there is also match the longest one, which means going over all records, so hash table vs list makes no difference.

@miconda regarding the address table, it was requested by your right here:
https://github.com/kamailio/kamailio/pull/4539#issuecomment-3786639567

> Anyhow, I do not think the matching speed was an issue so far, and regarding reloads, from the design of the PR I think that lookups can keep using an old table while another process (e.g., rpc) frees it.

This is not true, multiple reloads at a time create a race, and this was explained well in details in the older version of PR here:  https://github.com/kamailio/kamailio/pull/4539#issuecomment-3785133654

> if table-lock is acquired by each process using the table, then the bucket-lock should not be necessary, but then only one process can use the table at that moment, practically matching cannot be done in parallel anymore

This is not a true and I described this well in the commit:
https://github.com/kamailio/kamailio/pull/4787/changes/4ebb47d1adc10b5eb2a6a2dfc95249e6cfc51365

> if table-lock is NOT acquired by each process using the table, then reloading in parallel can result in races, with a process still accessing an old table that is freed by another one

The table level lock (the upper lock) is acquired always by each process who has to do either with reload or hr timer, and I wrote about it as well in my commits bodies..

I see that most of what I wrote in commits bodies wasn't ever read.
I abandon the PR.

BR, Donat

miconda:
In many cases I look at the diff of the PRs, because if there are many commits, one can amend another one and it is hard to track the final result by looking at subsequent commits. It is a significant diff and trust me I looked at it more than to many other PRs, but it is not humanly possible to cover all the cases there at such size.

The races between match and reload are a matter of design and with the one I understood being in this PR, the races cannot be avoided.

So, the `match_hash_table()` starts with `*current_trusted_table`, but uses only bucket lock, not table lock. A reload can swap the active pointer (table lock not being acquired by the match) and the timer routine starts freeing the table that can still be used by already-started `match_hash_table()`. Am I wrong here? If yes, how an already started match is protecting that a reload does not destroy the table it is using?

Up to you if you don't want to go with this PR further, it is not that I didn't want to accept your changes, I suggested an alternative as well. It is about the fact that a module used for critical operations should be protected as much as possible and a massive change has to show its benefits. If you don't want to discuss and clarify the concerns, the variant with a different module stands as an option.

miconda:
And to show an example that popped up in my eyes of how small mistakes/overlooks can lead to issues, see:

- https://github.com/zenichev/kamailio/blob/50d9f1056c8fd3cfaa69079087b12d53e9bd762e/src/modules/permissions/trusted.c#L152

That leads to a deadlock, as the acquired  lock by `lock_get(trusted_table_reload_lock);` done a couple of lines above is not released.

henningw:
> > I do not see the real benefits (maybe not yet), at least for address table that I use to worth such a large change that is not easier to review and test for stability. Also, for subnet matching, there is also match the longest one, which means going over all records, so hash table vs list makes no difference.
> 
> @miconda regarding the address table, it was requested by your right here: [#4539 (comment)](https://github.com/kamailio/kamailio/pull/4539#issuecomment-3786639567)
> 
> > Anyhow, I do not think the matching speed was an issue so far, and regarding reloads, from the design of the PR I think that lookups can keep using an old table while another process (e.g., rpc) frees it.
> 
> This is not true, multiple reloads at a time create a race, and this was explained well in details in the older version of PR here: [#4539 (comment)](https://github.com/kamailio/kamailio/pull/4539#issuecomment-3785133654)
> 
> > if table-lock is acquired by each process using the table, then the bucket-lock should not be necessary, but then only one process can use the table at that moment, practically matching cannot be done in parallel anymore
> 
> This is not a true and I described this well in the commit: [4ebb47d](https://github.com/kamailio/kamailio/commit/4ebb47d1adc10b5eb2a6a2dfc95249e6cfc51365)
> 
> > if table-lock is NOT acquired by each process using the table, then reloading in parallel can result in races, with a process still accessing an old table that is freed by another one
> 
> The table level lock (the upper lock) is acquired always by each process who has to do either with reload or hr timer, and I wrote about it as well in my commits bodies..
> 
> I see that most of what I wrote in commits bodies wasn't ever read. I abandon the PR.
> 
> BR, Donat

Thank you @zenichev for taking the time and creating two versions of the PR. Maybe somebody else like to pick it up and continue working on it.

zenichev:
@henningw good day, tbh I don't get.
I see that the PR had no chance to proceed (at least I get it so that the PR got no-go, because of multiple reasoning why it can't be part of the project). How do you see that it's still possible that someone can pick that up and proceed?
For me it would be really pity to loose this work and give it away, because of enough time invested.

zenichev:
> In many cases I look at the diff of the PRs, because if there are many commits, one can amend another one and it is hard to track the final result by looking at subsequent commits. It is a significant diff and trust me I looked at it more than to many other PRs, but it is not humanly possible to cover all the cases there at such size.

Clear, thanks for the explanation.

> The races between match and reload are a matter of design and with the one I understood being in this PR, the races cannot be avoided.

I didn't cover the match part with the upper level locks indeed (that would have been an over-kill), but they aren't the problem. Match is only locking the buckets for the read time, there is no need to lock the whole table, because match works based on the calculated hash index.

The problem happens only when there is a change of the content, and this is namely **only**: hr timer and reload (reload either with RPC or any other way triggered).

> So, the match_hash_table() starts with *current_trusted_table, but uses only bucket lock, not table lock. A reload can swap the active pointer (table lock not being acquired by the match) and the timer routine starts freeing the table that can still be used by already-started match_hash_table(). Am I wrong here? If yes, how an already started match is protecting that a reload does not destroy the table it is using?

This is a good point.
The upper lock has a precedence over the buckets level lock, but that only prevails in case of reloads and hr timer.
As soon as the match process starts, it uses the hash index to find the required bucket, and locks it while the matching is being done. But it locks only the bucket, not the whole table, then when by the hash index the bucket (the list if to call it simpler) is selected, it's locked for write operations. Hence any reload who wants to get to the buckets and change anything there will comes through: 1. upper level lock (acquiring the whole table); 2. buckets level lock (bucket specific lock) while trying to change anything and will be waiting to acquire it until the match is done.

That just parallelizes the work of two competing things: write operations one side (e.g. reload) and read operations other side (e.g. match). Since the most what happens is match and RPC reloads or the hr timer are much rare thing in comparison to the match, this isn't a hit to the processing time.

>  Am I wrong here? If yes, how an already started match is protecting that a reload does not destroy the table it is using?

When the reload happens, the swap will trigger the `trusted_table_reinit()` on the `new_trusted_table` whereas the `trusted_table_free_buckets()` is covered with the bucket locks, and it will not touch the concerned bucket used for matching. Please remember that the swap of the global cursor happens at the point of real swap at the end of the reload.

So if P1 does the match based on the most actual table (let say taking the table X), P2 tries to reload and takes the less actual table (let's say Y). P1 will lock the bucket of the X table and not allow any other process coming ahead (P3) to touch the content, and such a new process will have to wait.
P2 will finish the reload (based on the older table Y content) and swap the global cursor to it. X table gets stale.

Even if two competitive processes come to run the reload, the second process will be blocked by the upper lock acquired previously, and hence not allow to overlap in write operations (main issue in the architecture of the module right now).

That means that potentially only one single process get into the situation when it tries to re-fill the content of the selected table, but there is a match being done. As soon as the match is managed, the swap of tables is allowed. The lock preventing things to be messed in this case is on the buckets level.

**I want to underline,** now nothing is locks safe, the processes are chaotically doing the read and write operations on tables and the users have quite a good chance to fall into our situation with iterator invalidation issues. This all depends on how huge the loading is on the system.

> Up to you if you don't want to go with this PR further, it is not that I didn't want to accept your changes, I suggested an alternative as well. It is about the fact that a module used for critical operations should be protected as much as possible and a massive change has to show its benefits. If you don't want to discuss and clarify the concerns, the variant with a different module stands as an option.

Tbh, I had a perception of things that the change had no go.
And most of the argumentation was not meeting the actual commits content (especially what's been described), what made me surprised and leading to understanding that invest time into something that won't have a green light.

zenichev:
@miconda 

> That leads to a deadlock, as the acquired lock by lock_get(trusted_table_reload_lock); done a couple of lines above is not released.

yes, this is a good catch, that meant to be using the alias.

This is good to have a review from maintainers of the project, because I can do human mistakes as such
And still I think that the proposed concept goes entirely right direction.

PR #4789, opened 2026-06-23
path: add advertised_address functions

bundasmanu:
#### Description

Support to declare an `advertised_address` in the `Path` header(s).
The approach is similar to `rr` module.

`socket name (sn)` support is also included, when `advertised_address` is used. And in a lot of scenarios, makes sense to use both at same time.

`transport` mapping also applied.

Quiet useful, allowing the preload of `lr` in a pool of instances, instead of pointing to static IP's.

Support for `received`, `r2` tested.

Example of usage:

```sh
add_path_received_advertised_address("$env(INST_FQDN)", "", "ob");
```

#### Pre-Submission Checklist
  from source code folder

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

### DIFF
--- a/src/core/lump_struct.h
+++ b/src/core/lump_struct.h
@@ -57,7 +57,9 @@ enum lump_subst
 	SUBST_RCV_ALL,
 	SUBST_SND_ALL, /* ip:port;transport=proto */
 	SUBST_RCV_ALL_EX,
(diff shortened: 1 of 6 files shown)


miconda:
Thanks, merging!

PR #4791, opened 2026-06-24
Fix heap overflow when rewriting multipart SIP bodies

SABITHSAHEB:
#### Description

- Found a heap overflow in `check_boundaries()` while poking at the multipart body code.
- It rebuilds the body into a `buf.len + 2` buffer, but those 2 bytes aren't enough once the last boundary gets rewritten to its closing form (`--boundary--\r\n`) — the final `memcpy()` writes past the end.
- The body and the boundary value both come straight off the wire, so anything doing multipart (e.g. `set_body_multipart()`) can trip it.
- Fix is small: reserve `fb.len` extra bytes instead of 2, and bounds-check the copies so it can't run over.
- Valid bodies come out exactly the same as before.
- Also added a fuzz target at `misc/fuzz/fuzz_check_boundaries.c` to keep this path covered.
- Tested with ASan: fires on master, clean with the patch, and a normal build throws no new warnings.

### DIFF
--- a/misc/fuzz/fuzz_check_boundaries.c
+++ b/misc/fuzz/fuzz_check_boundaries.c
@@ -1,0 +1,65 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "../config.h"
+#include "../globals.h"
+#include "../dprint.h"
+#include "../parser/msg_parser.h"
+#include "../parser/parse_hname2.h"
+#include "../msg_translator.h"
+
+/* check_boundaries() is a core function not exposed via a public header */
+int check_boundaries(struct sip_msg *msg, struct dest_info *send_info);
+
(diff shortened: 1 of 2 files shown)


henningw:
Thanks for the PR. Please adapt your commit message by adding a proper prefix (e.g. core: or misc/fuzz:). I noticed that you removed the fuzzing target, it should be fine to include it, if there are error we need to look into it. After you change the commit messages please do a force-push to this PR.

SABITHSAHEB:
Updated..

SABITHSAHEB:
CI came back green on the Fuzzing job, so the re-added target builds and runs clean on this path, no errors there. The only red is check-commit, and that one died on a transient "fatal: shallow file has changed since we read it" during its git fetch rather than anything in the commits, so a re-run should clear it.

PR #4792, opened 2026-06-24
sipcapture: validate HEPv3 chunk boundaries against packet length

SABITHSAHEB:
### Description
Noticed the HEPv3 parser in `sipcapture` takes `total_length` from the packet header and walks chunks up to that value without verifying it against the number of bytes actually received. The HEPv2 path right above already performs this validation, making HEPv3 the exception. As a result, a malformed HEP packet can cause out-of-bounds reads while parsing chunks, including a small underflow in type-15 chunk handling. Since HEP is received over UDP, the issue is remotely reachable.

### Changes
Added two helper functions, `hepv3_total_length()` and `hepv3_chunk_ok()`, and reused them across the three HEPv3 parsers that iterate over chunks. The helpers ensure the advertised packet length and chunk boundaries stay within the received buffer. Also added `misc/fuzz/fuzz_hep.c` to exercise malformed HEP inputs.

### Result
Valid HEP packets continue to parse exactly as before, while malformed packets are safely rejected. ASan reproduces the issue on master and runs clean with this patch applied.

### DIFF
--- a/misc/fuzz/fuzz_hep.c
+++ b/misc/fuzz/fuzz_hep.c
@@ -1,0 +1,193 @@
+/*
+ * libFuzzer / OSS-Fuzz target for the HEPv3 capture-packet chunk parser in
+ * src/modules/sipcapture/hep.c.
+ *
+ * The HEPv3 parsers (parsing_hepv3_message, hepv3_message_parse,
+ * hepv3_get_chunk) are sipcapture module functions that pull in the whole
+ * module and the SIP receive path, so they cannot be linked into the core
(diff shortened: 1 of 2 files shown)


abalashov:
Looks vibe-coded.

SABITHSAHEB:
That's on me — the description was from a different fix and didn't match
this change. Fixed now.

Short version: the HEPv3 parser trusts total_length from the header and
walks chunks up to it without checking it against the bytes received. The
HEPv2 path right above already does that check, so HEPv3 was the one missing
it — a malformed packet reads past the buffer, and it's UDP so reachable
remotely. Can paste the ASan trace if useful.

PR #4794, opened 2026-06-25
sipcapture: validate HEPv3 chunk boundaries against packet length

SABITHSAHEB:
The HEPv3 parsers (`parsing_hepv3_message`, `hepv3_message_parse`,
`hepv3_get_chunk`) take the total length straight from the packet's control
header and walk the chunk list with it, without ever checking it against the
number of bytes actually received. Each chunk header and field is then
dereferenced directly from the buffer.

A malformed HEP datagram that advertises more bytes than were received (e.g.
header says 200, only 20 arrive) makes the loop read chunks past the end of the
received data — an out-of-bounds read reachable straight from the network.

The fix adds two small helpers used at all three parse points: one rejects a
control header that claims more than was received, the other bounds-checks each
chunk (header inside the buffer, advertises at least its 6-byte header, whole
chunk fits) before it is dereferenced. It also replaces the old
`chunk_length == 0` guard. No behaviour change for well-formed packets.

Tested locally and `clang-format` clean.

### DIFF
--- a/src/modules/sipcapture/hep.c
+++ b/src/modules/sipcapture/hep.c
@@ -251,6 +251,44 @@ int hepv3_received(char *buf, unsigned int len, struct receive_info *ri)
 	return -1;
 }
 
+/*
+ * Return the advertised HEPv3 total length after checking the control header
+ * fits and does not claim more bytes than were received, so the chunk loop
+ * stays inside the datagram. Returns -1 on a malformed/oversized header.
+ */
+static int hepv3_total_length(const char *buf, unsigned int len)
+{
+	int total_length;
+
+	if(len < sizeof(hep_ctrl_t))
+		return -1;
+	total_length = ntohs(((const hep_ctrl_t *)buf)->length);
+	if((unsigned int)total_length > len)
+		return -1;
+	return total_length;
+}
+
+/*
+ * Verify the chunk at offset off is safe to read before it is cast and its
+ * fields are dereferenced: the 6-byte generic header must lie within the
+ * blen-byte buffer, it must advertise at least that header (so the loop
+ * advances), and the whole chunk must fit. Returns 0 if in bounds, -1 if not.
+ */
+static int hepv3_chunk_ok(const char *buf, unsigned int blen, unsigned int off)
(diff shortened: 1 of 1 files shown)


henningw:
Thanks. Does it replace #4792 ?

SABITHSAHEB:
Yes, this replaces #4792. It's the same HEPv3 bounds fix, reworked to be
minimal and easier to review: I dropped the per-chunk-type length table and
the fuzz harness, keeping only the checks needed to stop the out-of-bounds
read — validating the control-header length against the bytes received and
bounds-checking each chunk before it's dereferenced.

I'll close #4792 in favor of this one.

miconda:
I am merging it, but probably sipcapture module is obsolete. It was designed as a component for Homer SIP Capture server application, but, afaik, for many versions Homer has its own server application written in Go. Maybe @adubovikov can shed some light if sipcapture is still used by Homer project.

whosgonna:
@miconda @adubovikov Does the go application capture and decode TLS?  If not, then there is still a need for sipcapture.

miconda:
@whosgonna: the module mirroring to homer is siptrace. The sipcapture is the module supposed to receive mirrored traffic from siptrace and save to homer database. In other words: siptrace is homer client app and sipcapture is homer server app. Likely you use only the first one these days, unless you have very old Homer server version.

adubovikov:
> @miconda @adubovikov Does the go application capture and decode TLS? If not, then there is still a need for sipcapture.

@whosgonna 

Some people are still using the old configuration (H5) based on Kamailio/SipCapture. And yes, heplify-server or homer11 can decode TLS (HEP transport). It’s not entirely clear what you mean by “capture.” SIP over TLS? No, that’s not possible, but you can use the siptrace module for that.

whosgonna:
@adubovikov My apologies here.  As Daniel points out, i had the `sipcapture` module conflated with the `siptrace` module.

SABITHSAHEB:
Thanks for merging. Good to know about sipcapture's current status - this fix should at least help any existing deployments still using it.

PR #4795, opened 2026-06-25
pua_dialoginfo: crash correction, adding null pointer protection

fgaisnon:
#### Pre-Submission Checklist

- [x Commit message has the format required by CONTRIBUTING guide
  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

crash correction

### DIFF
--- a/src/modules/pua_dialoginfo/pua_dialoginfo.c
+++ b/src/modules/pua_dialoginfo/pua_dialoginfo.c
@@ -582,8 +582,10 @@ static void __dialog_sendpublish(
 	}
 
 	if(use_pubruri_avps && (refresh_pubruri_avps_flag > -1) && (request != NULL)
-			&& (request->flags
-					& (1U << (unsigned int)refresh_pubruri_avps_flag))) {
+		&& (request
+				&& request->flags
+						& (1U << (unsigned int)
+								refresh_pubruri_avps_flag))) {
 		lock_release(&dlginfo->lock);
 	}
 }


src/modules/pua_dialoginfo/pua_dialoginfo.c
```diff
@@ -582,8 +582,10 @@ static void __dialog_sendpublish(
 	}
 
 	if(use_pubruri_avps && (refresh_pubruri_avps_flag > -1) && (request != NULL)
-			&& (request->flags
-					& (1U << (unsigned int)refresh_pubruri_avps_flag))) {
+		&& (request
```
jcolp, line 585: I believe this change is redundant. Line 584 already has "request != NULL" which is equivalent to this.

abalashov: @jcolp "You're right to push back on that, this is a slop PR."

abalashov: On second thought, I withdraw my comment. This one may be redundant, but looks legit. Sorry, a lot of slop PRs coming in lately.

henningw: It looks indeed redundant. @fgaisnon Can you provide a bit more context why you believe this fixes a crash?

fgaisnon: Hi, I have checked the code where I had a crash and this condition (request != NULL) was not present.
So I am agree that this patch is now redundant.
My patched code in my production version:
if(use_pubruri_avps && (refresh_pubruri_avps_flag > -1
                || (request && request->flags & (1U<<(unsigned int)refresh_pubruri_avps_flag))))

Thanks ans sorry for the noise

henningw: Thanks, close it then.

PR #4796, opened 2026-06-26
core: strutils - fix out of bounds read in replace()

SABITHSAHEB:
### what i changed 
`replace()` in `src/core/strutils.c` uses a `\d` backreference digit straight
as an index into `pmatch[]`, but the array only has 6 entries (indices 0–5,
`SR_RE_MAX_MATCH`). A `\6`–`\9` token reads past the end of it.

The garbage `regmatch_t` that comes back gives a bad source offset and a
`size = rm_eo - rm_so` that can be negative — and a negative `size` slips
through the `j + size < result->len` check straight into `memcpy()`, where it
turns into a huge length and overflows `result->s`.

Fix just bounds the index (`digit < 6`) and rejects a negative size before the
copy. Valid backreferences behave exactly as before.

Tested with an ASan build — a `\6` replacement trips stack-buffer-overflow
before the patch, clean after. Builds fine with gcc `-Wall`, clang-format clean.

### DIFF
--- a/src/core/strutils.c
+++ b/src/core/strutils.c
@@ -768,6 +768,8 @@ int cmp_aor_str(str *s1, str *s2)
 	return cmp_aor(&uri1, &uri2);
 }
 
+#define SR_RE_MAX_MATCH 6
+
 /*! \brief Replace in replacement tokens \\d with substrings of string pointed by
  * pmatch.
  */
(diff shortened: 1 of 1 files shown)


henningw:
Thank you for the PR, it will be reviewed. In case of other findings from your side that might cause an issue for a running Kamailio service, please follow the process described here: https://www.kamailio.org/wikidocs/security/policy/ instead of directly open a PR.

SABITHSAHEB:
Thanks ,on the security process — I'll use security@kamailio.org
for any future findings of that kind. Leaving this one here for review.

src/core/strutils.c
```diff
@@ -783,9 +783,9 @@ int replace(regmatch_t *pmatch, char *string, char *replacement, str *result)
 			if(i < len - 1) {
 				if(isdigit((unsigned char)replacement[i + 1])) {
 					digit = replacement[i + 1] - '0';
-					if(pmatch[digit].rm_so != -1) {
+					if(digit < 6 && pmatch[digit].rm_so != -1) {
```
miconda: Shouldn't the condition be `digit < SR_RE_MAX_MATCH`?

SABITHSAHEB: Done — switched to SR_RE_MAX_MATCH. Moved the #define above replace() so it's
in scope, since it was defined just below the function.

abalashov:
As a code reviewer, one has the distinct sense that one is interacting with an LLM, on this and other PRs. We can do that already, if that's what we want to do.

SABITHSAHEB:
Thank you for the review and merge

PR #4800, opened 2026-07-01
RFC: TCP Reactor: own all TCP network I/O in `PROC_TCP_MAIN`

space88man:
## Pre-Submission Checklist
  from source code folder

## Type Of Change

## Checklist:

- [X] Tested changes locally

## Description
Update: the code change and description was conducted together with Claude code.

This PR adds a new operating mode for kamailio's TCP layer, selected by the
existing global `tcp_main_threads`:

- `0` — fd-passing (unchanged)
- `1` — fd-passing + OpenSSL trampoline threads (was `>0`)
- `2` — **full TCP reactor (new):** `PROC_TCP_MAIN` owns every TCP fd and performs
  all network I/O; the TCP worker processes become pure SIP engines.

In mode 2 there is **no fd-passing**. Connections are accepted, read,
reassembled, written, opened, closed, and timed out entirely inside
`PROC_TCP_MAIN`. Reassembled SIP messages are dispatched to worker processes as
small shared-memory task objects over a kernel-load-balanced datagram socket.
The mode is opt-in and additive; modes 0 and 1 are unchanged.

The motivation is architectural: **handling all network I/O in one process is
simpler than shuttling connection file descriptors
between processes with `sendmsg(SCM_RIGHTS)`.**

---

### Background: the fd-passing model

In the current TCP architecture, `PROC_TCP_MAIN` owns the listen sockets and
accepts connections, then **hands each connection's file descriptor to a TCP
worker** via `sendmsg()`/`SCM_RIGHTS` over a UNIX socket. The worker then owns
that fd and does the reads, reassembly, `receive_msg()`, and writes for that
connection.

This has served kamailio well, but the fd-passing at its core carries recurring
structural costs:

1. **Every handoff is a control-message syscall.** Passing an fd is a
   `sendmsg`/`recvmsg` pair carrying ancillary `SCM_RIGHTS` data; the kernel
   installs (dups) the fd into the receiver's descriptor table. This happens on
   every inbound accept→worker handoff, and the async write path adds further
   fd movement to let a process write on a connection it does not currently hold.

2. **A passed fd is duplicated across processes.** After a pass, the same
   connection fd exists in *both* processes until closed in both. Descriptor
   accounting doubles for in-flight connections, and close/teardown must be
   coordinated across processes to avoid leaking or closing too early.

3. **Connection ownership is smeared across processes.** Because an fd can move,
   more than one process can end up wanting to write to the same connection, and
   **outbound connection creation to the same peer can race across workers**
   (multiple workers independently opening a socket to the same destination).

4. **The write path is the most complex part of the subsystem.** Writing on a
   connection you do not own requires either passing the fd back or driving a
   fd-shuffling async write queue. This is a large, subtle body of code precisely
   because ownership is distributed.

5. **Connection state is scattered.** Read buffers, partial-message reassembly
   state, write queues, and per-connection timers live in whichever worker
   currently holds the fd. Reasoning about, monitoring, and reaping connections
   means reasoning about distributed state across N workers.

---

### What this PR introduces (mode 2)

`PROC_TCP_MAIN` runs one `io_wait` thread plus a small pool of I/O worker
threads (hardcoded to 8 at the moment). It owns all TCP fds for their entire lifetime.

- **Reads:** `PROC_TCP_MAIN` reads and reassembles complete SIP messages. Each
  assembled message is copied into a shared-memory task object and its pointer is
  written to an `AF_UNIX`/`SOCK_DGRAM` socketpair. All TCP workers `recvfrom()`
  that one socket; **the kernel load-balances** delivery across them — no explicit
  worker assignment or affinity is needed.
- **Writes:** any process that wants to send posts a *write request* to
  `PROC_TCP_MAIN` (`CONN_WRITE_REQ`) with the payload in shared memory.
  `PROC_TCP_MAIN` queues and transmits it. **No fd is passed for writes.**
- **Outbound:** a worker with no usable connection posts a *connect request*
  (`CONN_CONNECT_REQ`); `PROC_TCP_MAIN` performs the `socket()`+`connect()`,
  hashes the connection, and sends the first payload. Creation is serialized in
  one process.
- **Workers:** own no fds, run no io_wait loop, do no raw socket I/O, and (see
  below) run no OpenSSL. They are stateless SIP processors fed by the dispatch
  socket.

Inside `PROC_TCP_MAIN`, per-connection work (read/reassemble, encrypt/decrypt,
write) is handed to the thread pool under a single-owner-per-connection rule, so
the io_wait thread never blocks on a slow read, a large write, or a TLS handshake.

---

#### 1. No `SCM_RIGHTS` shuttling
The per-connection `sendmsg`/`recvmsg` fd handoff disappears from the hot path.
Inbound connections are never passed; writes are never passed; the async write
path no longer moves descriptors. What crosses a process boundary instead is a
single pointer (to a shm task) over a datagram socket — a fixed, tiny,
ancillary-data-free transfer that the kernel fans out for free.

#### 2. No cross-process fd duplication or split lifecycle
Each connection fd exists in exactly one process. There is no window where the
same descriptor lives in two descriptor tables, and no cross-process protocol for
"who closes it." Descriptor accounting and connection teardown become local and
unambiguous.

#### 3. Multi-writer races are eliminated by construction
Only `PROC_TCP_MAIN` ever opens, writes, or closes a connection. The
outbound-connect-to-same-peer race is gone because creation is serialized in one
process. "Two workers writing the same socket" cannot occur because workers hold
no sockets.

#### 4. The write path collapses
"Write on a connection you don't own" becomes "post a write request to the one
owner." The distributed async-write / fd-shuffle machinery is replaced by a
single queue owned by a single process. Less code, fewer states, fewer corners.

#### 5. Connection lifecycle is centralized
One hash table, one timer wheel, one reaper, in one process. Idle timeouts,
partial-read timeouts, and error teardown are local decisions on local state
rather than a cross-process dance.

#### 6. Clean separation of concerns → independent scaling
The reactor does network I/O; the workers do SIP logic. Workers become CPU-bound,
stateless message processors with no I/O responsibilities, fed by a
kernel-balanced queue. The number of I/O threads and the number of SIP processors
are now independent knobs rather than both being tied to "one fd owner per
worker."

#### 7. TLS: the single-process constraint, resolved rather than worked around
OpenSSL's per-connection `SSL` objects and the shared `SSL_CTX` carry
process-local heap pointers and assume in-process locking. They **cannot be
shared across processes** without installing custom shared-memory crypto
allocators *and* `PROCESS_SHARED` pthread-mutex hooks — a fragile arrangement.
The practical consequence under fd-passing is that a TLS
connection's cryptographic state cannot follow its fd to an arbitrary worker.

`tcp_main_threads=1` already acknowledges this: its trampoline relays each
OpenSSL operation from the worker to a dedicated thread *inside* `PROC_TCP_MAIN`
over a per-worker socketpair — keeping the socket in the worker but the crypto in
`PROC_TCP_MAIN`. It works, but it is inherently a **two-mechanism** design: the
fd lives in one process and its crypto in another, coordinated on every call, with
one relay thread and socketpair pair per worker.

Mode 2 dissolves the split. The fd **and** its `SSL` object live and are used in
the same process, so the single-process constraint is satisfied *by construction*
— no relay, no cross-process crypto, no allocator/mutex hooks. Since the crypto
had to be centralized in `PROC_TCP_MAIN` anyway, centralizing the I/O alongside it
is the natural simplification, and it removes the trampoline relay entirely. One
pool thread owns a connection (and thus its `SSL`) at a time, so OpenSSL is driven
correctly with nothing more than modern OpenSSL's own in-process thread safety.

---

### Design notes

- **Dispatch:** assembled messages are shm task objects (receive-info + message
  buffer); the pointer is a single `uintptr_t` written to the dispatch socketpair
  — well within `PIPE_BUF`, so delivery is atomic and lossless. Workers cast back
  and call `receive_msg()`, then free the task.
- **In-process pool:** a `io_wait` thread owns file descriptors/timers/hash/lifecycle
  exclusively; pool threads run read/write/crypto on a connection that has been
  *shielded* (removed from io_wait and the timer) for the duration of a job, then
  signal completion back to the io_wait thread. This keeps a strict, small
  concurrency contract: connection metadata mutation stays on one thread.
- **Cross-process wakeups** use a `PROCESS_SHARED`, robust condition variable with
  owner-died recovery.

---

### Compatibility and safety

- **Opt-in.** Behaviour is unchanged unless `tcp_main_threads=2` is set.
- **Modes 0 and 1 are unchanged** — verified: the reworked paths are gated on
  `tcp_main_threads == 2`; the legacy fd-passing and trampoline code paths are not
  modified.
- **Incremental.** The series builds the mode step by step (dispatch socket, task
  struct, read path, write path, outbound path, then the in-process thread pool,
  then TLS offload), each step preserving the mode-0 path.

---

### Testing / validation

- **No regression** on the plain-TCP path in mode 2, including the outbound
  dispatcher probe socket.
- **Inbound TLS** (TLS 1.2 and 1.3): handshake + REGISTER + in-dialog request/
  response handled entirely through the reactor and its pool.
- **Writes from a non-TCP process:** a config that relays inbound TLS to UDP and
  back exercises a UDP receiver issuing a TLS write — validating that *any* process
  can write without owning the fd.
- **Outbound TLS:** connection opened in `PROC_TCP_MAIN`, handshake driven by the
  pool, OPTIONS ping answered (dispatcher marks the target active).
- **Concurrency soak:** 256 simultaneous TLS UACs, repeated. No crash; connections
  drain back to baseline; shared memory and fragmentation flat across repeated
  runs (no leak); no TLS record corruption under contention (confirms per-thread
  crypto isolation).
- TODO: add a container to the kamailio test repo for this use case

---

### Known limitations / follow-ups (tracked)

- **WS/WSS** are not yet offloaded to the pool in mode 2 (handled inline on the
  io_wait thread). Functional, but the websocket framing layer still needs a
  thread-safety audit before offload.
- **`event_route[tls:connection-out]`** does not fire under `tcp_main_threads > 0`
  (the handshake completes where the outbound-send context is not populated). It
  is documented as unsupported in this mode pending a rework that runs the event
  on a worker. This is a pre-existing property of TLS-in-`PROC_TCP_MAIN` (modes 1
  and 2), not introduced here.
- Mode 2 relies on OpenSSL ≥ 1.1.0 (in-process thread safety); this should be made
  an explicit requirement.

---

### Risk & rollback

The feature is gated behind a single global that defaults off. Setting
`tcp_main_threads` back to `0` (or `1`) fully restores the prior behaviour with no
residual effect, because those code paths are untouched. The blast radius of a
problem is therefore bounded to deployments that explicitly opt into mode 2.

### DIFF
--- a/src/core/tcp_cond.c
+++ b/src/core/tcp_cond.c
@@ -1,0 +1,142 @@
+/*
+ * Copyright (C) 2025 S-P Chan
+ *
+ * This file is part of Kamailio, a free SIP server.
+ *
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ * Kamailio is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * Kamailio is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+/* pthread_mutexattr_setrobust()/pthread_mutex_consistent() need _GNU_SOURCE
+ * (or _POSIX_C_SOURCE >= 200809) on glibc; must be set before any include. */
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
+#include <errno.h>
+#include <pthread.h>
+#include <string.h>
+
+#include "dprint.h"
+#include "tcp_cond.h"
+
+int tcp_cond_init(tcp_cond_t *cond)
+{
+	pthread_mutexattr_t mattr;
+	pthread_condattr_t cattr;
+
+	/* process-shared + robust mutex */
+	if(pthread_mutexattr_init(&mattr) != 0) {
+		LM_ERR("failed to init mutex attributes\n");
+		return -1;
+	}
+	if(pthread_mutexattr_setpshared(&mattr, PTHREAD_PROCESS_SHARED) != 0
+			|| pthread_mutexattr_setrobust(&mattr, PTHREAD_MUTEX_ROBUST) != 0
+			|| pthread_mutex_init(&cond->m, &mattr) != 0) {
+		LM_ERR("failed to set up process-shared robust mutex\n");
+		pthread_mutexattr_destroy(&mattr);
+		return -1;
+	}
+	pthread_mutexattr_destroy(&mattr);
+
+	/* process-shared condition variable */
+	if(pthread_condattr_init(&cattr) != 0) {
+		LM_ERR("failed to init cond attributes\n");
+		pthread_mutex_destroy(&cond->m);
+		return -1;
+	}
+	if(pthread_condattr_setpshared(&cattr, PTHREAD_PROCESS_SHARED) != 0
+			|| pthread_cond_init(&cond->c, &cattr) != 0) {
+		LM_ERR("failed to set up process-shared cond\n");
+		pthread_condattr_destroy(&cattr);
+		pthread_mutex_destroy(&cond->m);
+		return -1;
+	}
+	pthread_condattr_destroy(&cattr);
+
+	return 0;
+}
+
+void tcp_cond_destroy(tcp_cond_t *cond)
+{
+	pthread_cond_destroy(&cond->c);
+	pthread_mutex_destroy(&cond->m);
+}
+
+/* Recover a robust mutex whose previous owner died while holding it: we now own
+ * the lock, but the data it protects may be inconsistent. Marking it consistent
+ * keeps the mutex usable; without this it would become ENOTRECOVERABLE and every
+ * future lock attempt would fail. */
+static void tcp_cond_recover_owner_dead(tcp_cond_t *cond)
+{
+	LM_WARN("robust mutex %p owner died; recovering (state may be partial)\n",
+			(void *)cond);
+	if(pthread_mutex_consistent(&cond->m) != 0)
+		LM_ERR("failed to make robust mutex %p consistent\n", (void *)cond);
+}
+
+void tcp_cond_lock(tcp_cond_t *cond)
+{
+	int ret;
+
+	ret = pthread_mutex_lock(&cond->m);
+	if(ret == EOWNERDEAD) {
+		tcp_cond_recover_owner_dead(cond);
+	} else if(ret != 0) {
+		LM_ERR("pthread_mutex_lock failed: %s (%d)\n", strerror(ret), ret);
+	}
+}
+
+void tcp_cond_unlock(tcp_cond_t *cond)
+{
+	int ret;
+
+	ret = pthread_mutex_unlock(&cond->m);
+	if(ret != 0)
+		LM_ERR("pthread_mutex_unlock failed: %s (%d)\n", strerror(ret), ret);
+}
+
+void tcp_cond_wait(tcp_cond_t *cond)
+{
+	int ret;
+
+	/* re-acquires the mutex on wakeup; that re-acquire can also report a dead
+	 * previous owner on a robust mutex */
+	ret = pthread_cond_wait(&cond->c, &cond->m);
+	if(ret == EOWNERDEAD) {
+		tcp_cond_recover_owner_dead(cond);
+	} else if(ret != 0) {
+		LM_ERR("pthread_cond_wait failed: %s (%d)\n", strerror(ret), ret);
+	}
+}
+
+void tcp_cond_signal(tcp_cond_t *cond)
+{
+	int ret;
+
+	ret = pthread_cond_signal(&cond->c);
+	if(ret != 0)
+		LM_ERR("pthread_cond_signal failed: %s (%d)\n", strerror(ret), ret);
+}
+
+void tcp_cond_broadcast(tcp_cond_t *cond)
+{
+	int ret;
+
+	ret = pthread_cond_broadcast(&cond->c);
+	if(ret != 0)
+		LM_ERR("pthread_cond_broadcast failed: %s (%d)\n", strerror(ret), ret);
+}
--- a/src/core/tcp_cond.h
+++ b/src/core/tcp_cond.h
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2025 S-P Chan
+ *
+ * This file is part of Kamailio, a free SIP server.
+ *
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ *
+ * Kamailio is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * Kamailio is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+#ifndef _TCP_COND_H_
+#define _TCP_COND_H_
+
+#include <pthread.h>
+
+/*
+ * Cross-process condition variable for the TCP reactor thread pool
+ * (tcp_main_threads == 2).
+ *
+ * The struct is meant to live in shared memory: TCP worker processes signal it
+ * to submit write jobs, while the pool threads inside PROC_TCP_MAIN wait on it.
+ * Because waiters and signalers are different PIDs sharing one shm page, both
+ * the mutex and the condvar are PTHREAD_PROCESS_SHARED.
+ *
+ * The mutex is additionally PTHREAD_MUTEX_ROBUST: if a process dies while
+ * holding the lock, the next tcp_cond_lock()/tcp_cond_wait() recovers it
+ * (EOWNERDEAD -> pthread_mutex_consistent) instead of leaving every future
+ * locker stuck at ENOTRECOVERABLE.
+ */
+typedef struct tcp_cond
+{
+	pthread_mutex_t m;
+	pthread_cond_t c;
+} tcp_cond_t;
+
+/* Initialize a tcp_cond_t (typically allocated in shared memory).
+ * Returns 0 on success, -1 on error. */
+int tcp_cond_init(tcp_cond_t *cond);
+
+/* Destroy a tcp_cond_t. Does not free the memory holding it. */
+void tcp_cond_destroy(tcp_cond_t *cond);
+
+/* Lock / unlock the embedded mutex. tcp_cond_lock() transparently recovers a
+ * robust mutex abandoned by a crashed owner. */
+void tcp_cond_lock(tcp_cond_t *cond);
+void tcp_cond_unlock(tcp_cond_t *cond);
+
+/* Wait on the condition. Must be called with the mutex held (via
+ * tcp_cond_lock()). Recovers the robust mutex if it was abandoned while being
+ * re-acquired on wakeup. */
+void tcp_cond_wait(tcp_cond_t *cond);
+
+/* Wake one / all waiters. */
+void tcp_cond_signal(tcp_cond_t *cond);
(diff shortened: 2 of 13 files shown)


henningw:
Thank you for the PR. Please specify if you used a LLM to generate the description and or code changes? There are also some build failures from the pipeline that should be checked before doing a more thorough review.

space88man:
> Thank you for the PR. Please specify if you used a LLM to generate the description and or code changes? 
- description has been updated.

> There are also some build failures from the pipeline that should be checked before doing a more thorough review 

- I see the build failure on macos -  this is due to pthreads difference on macos - I'll look into the equivalent. Is macos mandatory ?

abalashov:
IMHO, this slop PR should be closed out of hand for being slop. The cognitive burden of reviewing should not be born by the project.

Practically, there is way too much going on here. This is a very significant and invasive change, even if it is, strictly speaking, additive. If one wanted to approach this in a way that does not vandalise the project with slop, one would likely make a feature branch, proceed in small feedback loops, and request intermediate review throughout those steps, along with validation that new constructs adhere to the norms, vocabulary, conventions and culture of the existing code base. 

It's hard work, and it takes time. If the author doesn't want to do it, I don't think we need to accept this slop.

fredposner:
> IMHO, this slop PR should be closed out of hand for being slop. The cognitive burden of reviewing should not be born by the project.
> 
> Practically, there is way too much going on here. This is a very significant and invasive change, even if it is, strictly speaking, additive. If one wanted to approach this in a way that does not vandalise the project with slop, one would likely make a feature branch, proceed in small feedback loops, and request intermediate review throughout those steps, along with validation that new constructs adhere to the norms, vocabulary, conventions and culture of the existing code base.
> 
> It's hard work, and it takes time. If the author doesn't want to do it, I don't think we need to accept this slop.

I think this is a little harsh, especially considering @space88man is part of the dev team and also introduced this in an email as well as his past contributions to TLS and the project. I understand the current bias, but believe this comment is unwarranted.

henningw:
I also think this is not the way we should change this core part. To add - the whole description looks that it was generated directly out of an LLM. So every claim there needs to be verified if its actually valid, and of course also the code. Setting it to draft for now.

abalashov:
> I think this is a little harsh, especially considering @space88man is part of the dev team and also introduced this in an email as well as his past contributions to TLS and the project. I understand the current bias, but believe this comment is unwarranted.

I don't think it is. The most one can say based on this pedigree is that he might be in a better position to do it than the average drive-by slop-slinger, but it's still not organised in a way that takes into account the modalities of ingestion by the open-source organism.

There is a legitimate argument for adding a “distributor thread” to the TCP side of Kamailio (UDP already has this in 6.0+), versus just having forked processes nibble at the kernel trough. But this is not the way to do it. Large slop PRs of this nature are anti-collaborative; I don't care if Hawking or Einstein or Dijkstra wrote them.

fredposner:
> > I think this is a little harsh, especially considering @space88man is part of the dev team and also introduced this in an email as well as his past contributions to TLS and the project. I understand the current bias, but believe this comment is unwarranted.
> 
> I don't think it is. The most one can say based on this pedigree is that he might be in a better position to do it than the average drive-by slop-slinger, but it's still not organised in a way that takes into account the modalities of ingestion by the open-source organism.
> 
> There is a legitimate argument for adding a “distributor thread” to the TCP side of Kamailio (UDP already has this in 6.0+), versus just having forked processes nibble at the kernel trough. But this is not the way to do it. Large slop PRs of this nature are anti-collaborative; I don't care if Hawking or Einstein or Dijkstra wrote them.

There is a method of expressing in a more respectful manner, which is what is [expected](https://github.com/kamailio/kamailio/blob/master/CODE_OF_CONDUCT.md) here to both encourage participation and maintain a healthy community. I'm asking that you _please_ find a more respectful way of expressing your concerns.

space88man:
@abalashov - I absolutely hear your concerns - may I gently push back. What is the purpose of this in the first place?

The goal is to consider the viability of a reactor style IO (by this I mean all network is confined to one process including OpenSSL) while being well aware that SIP processing has to be farmed out to all the workers.

For robust use of OpenSSL we have to trampoline all OpenSSL calls to a central process - but this begs the question - why then not just do all network IO in that process? 

In fact you would observe that on the SSL read path today (tcp_main_threads > 0) **all** reads go through PROC_TCP_MAIN (via mtops and the `_tconfd(...)` macro) - so the socket clone in the TCP workers is used only for io_wait notification and then we ask PROC_TCP_MAIN to do the real read - this is really awkward. (Side note: in fact the `_tconfd(...)` macro is really a bit of wizardry and I did not realise what it was doing until I converted  tls_wolfssl to use mtops.)

There is also a strange asymmetry - on TCP we read in the worker and on TLS we read in PROC_TCP_MAIN. 

This PR takes attempts to its logical extreme by centralising all IO in PROC_TCP_MAIN. 

Each commit is designed to compile and  to cause no regression in kamailio fd-passing mode. I explicitly chose to leave it raw and crude (unsquashed)  with all the dirty laundry for viewing.

Over this current development cycle I have considered how best to work with TLS+mtops and disable the use of OpenSSL crypto shared memory hooks and us having to patch pthreads at runtime. This current work was motivated particularly by the complex and delicate network socket handling that Kamailio has to perform. 

kamailio devs - I hear the echoes of all your concerns. If you can indulge me and leave this up here in draft mode  let me refactor my approach by providing a design document and what each step was intended to achieve. Further it would be easier to evaluate when I provide a docker container in the kamailio test repo.

abalashov:
@space88man This is a sensible rationale, and I appreciate the in-depth analysis. I don't doubt that your intentions are good, but I don't think this is the way to do work of this scope. 

@fredposner, I take to heart your concerns.

All the same:

1. This PR is monstrous in scope, and introduces lots of new constructs. 
2. The description is lazy, and replete with LLM clichés. 
3. In general, it tries to do too many things at once while externalising the costs of review. 

The third point was less of a problem when humans had to manually type the code, since submitting a monstrous PR also functioned as proof of a mental stake in the work and an implicit commitment to ongoing maintenance. Now, of course, the capacity to generate is infinite. Mere LoC -- or other proxy for "ambition" -- of contribution cannot any longer be the bar.

I've seen a lot of other open-source projects become overrun with slop PRs, at least some of which are sure to come from very well-intentioned people. It's not ultimately up to me. I'm just expressing my view.

space88man:
Let me at least address this:
> This PR is monstrous in scope, and introduces lots of new constructs.

First let me highlight existing behaviour of Kamailio:

- async write queue - this is using PROC_TCP_MAIN as  the writer
- SSL read path in mtops mode - this is using PROC_TCP_MAIN as sole reader

So it is not the case at all that this
> introduces lots of new constructs

- it heavily reuses everything that kamailio does well - the io_wait loop and the SIP parser
- it fact in the TCP worker it simplifies it: the workers no longer have to deal with - am I receiving a file descriptor or am I receiving network data. The worker only receives a completely prepared SIP message task (the actual SIP message lies in shared memory).
- instead of chunking SIP data in the worker  we do that in PROC_TCP_MAIN using exactly the same code from `tcp_read.c`
-  note that the write path always had 1. worker writes or 2. PROC_TCP_MAIN writes - this was  awkward. Now there is only one write path - PROC_TCP_MAIN is doing the writes.
-  there is a lot of complexity around sendmsg and fd-passing and bookkeeping. All this is gone in reactor mode - only PROC_TCP_MAIN ever touches the network socket.

On the contrary a reactor significantly simplifies the cognitive load of what io_wait is supposed to be doing.

abalashov:
@space88man I wasn't referring to the complexity of the message processing workflow, but to the complexity of reviewing your PR. 

What I see in this PR is lots of LLM-rationalised changes -- as expressed in the LLM-generated comments -- for lots of different kinds of things, which someone has to tease out, separate and weigh in a logically independent way, as @henningw gestured toward.

abalashov:
> kamailio devs - I hear the echoes of all your concerns. If you can indulge me and leave this up here in draft mode let me refactor my approach by providing a design document and what each step was intended to achieve. Further it would be easier to evaluate when I provide a docker container in the kamailio test repo.

@space88man To be fair, I do not claim to represent the position or anything like a consensus view of "kamailio devs", so please do not construe my comments that way. 

However, so far as my opinion counts for anything, I don't know that adding more design documents will help here. This is like adding serving more discovery paperwork upon a respondent to a lawsuit whose main complaint is that they're drowning in paperwork.

space88man:
Good (if some startling) feedback here - food for thought indeed.  I think it makes more sense for me to close this PR as is . Let me tilt at my windmill again in another PR where the intent and design is more transparent. This branch will not go away soon so you can try the "reactor" mode if you are curious: just remember `tcp_main_threads = 2`. A new PR will be based off a more palatable sequence of commits.

abalashov:
> A new PR will be based off a more palatable sequence of commits.

Music to _my_ ears. Not sure about anyone else's. ;-)

henningw:
> Good (if some startling) feedback here - food for thought indeed. I think it makes more sense for me to close this PR as is . Let me tilt at my windmill again in another PR where the intent and design is more transparent. This branch will not go away soon so you can try the "reactor" mode if you are curious: just remember `tcp_main_threads = 2`. A new PR will be based off a more palatable sequence of commits.

Thank you @space88man. Your work on improving Kamailio is appreciated. One requirement for acceptance for a PR (for me) is that the author is able to explain why the change is necessary like this and how it works. This needs to be done in his or her own words. Just pasting 4 pages of LLM generated text description into a PR does not show this understanding to me. 
Looking forward to your next contribution.

Issue #4809, opened 2026-07-02
permissions: concurrent RPC reloads corrupt shared memory and crash workers


kuzaxak:
### Description

Concurrent `permissions.addressReload` (or `permissions.trustedReload`) RPC calls can crash kamailio: shared memory holding the permissions hash tables gets double-freed, and a worker process eventually dies with signal 11/6 (with the default `mem_safety=1` the allocator first logs double-free BUGs and keeps running while corruption accumulates).

It happens when the RPC transport is served by more than one process (e.g. `jsonrpcs` datagram workers) and two reloads overlap. We hit it in production on 6.0.x under frequent automated reloads; it reproduces on current `master` and on `6.0.7`. Expected behavior: overlapping reload requests may be rejected or serialized, but should not corrupt memory.

#### Reproduction

Self-contained Docker reproducer: [kamailio-permissions-reload-race-repro-with-pcap.zip](https://github.com/user-attachments/files/29596047/kamailio-permissions-reload-race-repro-with-pcap.zip)

```sh
./run.sh                # unpatched master  -> double free within seconds
./run.sh --patched      # + reload lock     -> same stress, survives
./run.sh --base 6.0.7   # same crash on the 6.0.7 release
```

It builds kamailio with `-x q_malloc` (DBG_QM_MALLOC), 16 `jsonrpcs` datagram workers and `reload_delta=0`, then floods `permissions.addressReload` from 32 senders for 25 s. Unpatched run: 905 `qm_free(): BUG: freeing already freed pointer` lines, then a worker dies with signal 11. Patched run: 0 BUG lines, no crash. Full logs, core backtraces and pcaps of the RPC flood are in the archive.

#### Analysis

The reload swaps between the module's two shm hash-table slots without a lock. In `reload_address_table()` (`src/modules/permissions/address.c`):

```text
read *perm_addr_table              (pick the inactive slot)
empty_addr_hash_table(inactive)    (shm_free its contents)
repopulate the inactive slot
*perm_addr_table = inactive        (publish)
```

Two workers can read the same "current" pointer, pick the same inactive slot and both empty/repopulate it concurrently — the double free above. `reload_trusted_table()` (`trusted.c`) has the same pattern. The `reload_delta` gate in `rpc_check_reload()` (`rpc.c`) does not prevent this: it is a non-atomic check-then-set, so two RPC calls can both pass it.

#### Debugging Data

[...]

PR #4819, opened 2026-07-10
carrierroute: improved threading safety

rfuchs:
The current lock/refcount handling of the global data table is (very slightly) racy during reloads and could theoretically lead to a crash.

- the *global_data pointer itself must be accessed and swapped out atomatically to eliminate possible race conditions
- reading the pointer _and_ increasing the reference count for the pointed-to table atomically is only possible with a global lock
- usually there are only one or two tables in existence at any given time, so we can drop the per-table lock and use the global lock to also protect the per-table reference count (this could be done with an atomic int as well, but we already have a lock anyway)
- the global lock implicitly guarantees both memory fences and compiler fences

Came up in a (marginally) related discussion in #4809

#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

### DIFF
--- a/src/modules/carrierroute/cr_data.c
+++ b/src/modules/carrierroute/cr_data.c
@@ -29,6 +29,7 @@
 
 #include <stdlib.h>
 #include "../../core/mem/shm_mem.h"
+#include "../../core/locking.h"
 #include "cr_data.h"
 #include "carrierroute.h"
 #include "cr_config.h"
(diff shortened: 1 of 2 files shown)


miconda:
@lbalaceanu: any comment on this PR? If nothind against, probably it can be merged soon.

henningw:
Thank you for the PR. Looks good to me, merged.

PR #4825, opened 2026-07-15
TCP threaded receiver mode - v2

space88man:
#### Pre-Submission Checklist
- [] Each component has a single commit (if not, squash them into one commit)
- [X] LLM/AI Coding Assistants were involved in creating the C code or other included content
- [X]Code is formatted with `clang-format` using the config file `.clang-format`
  from source code folder

#### Type Of Change

#### Checklist:

- [X] Tested changes locally

## Threaded TCP receiver (`tcp_main_threads = 2`)

This PR adds a threaded TCP receiver — the TCP counterpart to the UDP `udp_receiver_mode = 1` setting. Unlike UDP, this mode also confines TCP writes to a single process.

The mode is selected with `tcp_main_threads`:

- `0` — classic multi-process TCP network I/O and TLS (unchanged).
- `1` — multi-process TCP network I/O, but the read for TLS and all TLS run in a single process. This was previously `tcp_main_threads > 0`.
- `2` (new) — all TCP network I/O and all TLS run in `PROC_TCP_MAIN`.

### What changes in mode 2

Transport is separated from SIP message handling. `PROC_TCP_MAIN` is the only process that holds TCP sockets, and sockets are no longer passed between processes with `sendmsg`/`SCM_RIGHTS`. This removes the file descriptor book-keeping and the borrow/return ceremony around TCP sockets.

- all reads are done in `PROC_TCP_MAIN` and SIP messages (as shm pointers) are passed to the workers (as with `udp_receiver_mode = 1`)
- Writes are always asynchronous. A worker never writes to a socket; it hands the payload to `PROC_TCP_MAIN` via `CONN_WRITE_REQ`.
- Outbound connections with no existing `tcp_connection_t` are opened in `PROC_TCP_MAIN` (via `CONN_CONNECT_REQ`) instead of in a worker.
- `PROC_TCP_MAIN` runs a thread pool for socket reads and async writes. The pool size is set with `tcp_reactor_threads` (default 8).

In mode 2: conceptually a TCP worker is identical to a UDP async worker - it only deals with  SIP messages. The TCP worker `io_wait` loop simplifies to the `recvfrom`-loop of UDP; in this PR we keep `io_wait` to coexist with mode-0/1.

### Notes for reviewers

- update: during the development of this PR some bugs were found on master regarding event handling. This PR depends on commits that have landed in master so the original commits that contained workarounds have been cleaned-up.
- The `src/core` commits are not squashed, so each step is a smaller read.
- `pkg_malloc` is lock-free and assumes one thread per process, which mode 2 breaks. If Kamailio is built against the system `malloc` (glibc, musl, …) this is a non-issue, since those are thread-safe. With Kamailio's own allocators, the `pkg` path is serialized inside `PROC_TCP_MAIN`; that serialization is confined to `PROC_TCP_MAIN` and touches nothing else. This contention was exposed by the websocket module - by fluke TLS, HEP, MSRP did not use `pkg_malloc` in any way that resulted in multi-threaded contention. (I have ongoing work to see if websocket can use `pkg_malloc` in a thread safe manner - but that is incidental to the thrust of this PR.)
- Modes 0 and 1 are untouched. Every new code path is gated on `tcp_main_threads = 2`.

### DIFF
--- a/cmake/lock_methods.cmake
+++ b/cmake/lock_methods.cmake
@@ -159,3 +159,49 @@ set(LOCK_METHOD_FINAL
     "${_SELECTED_LOCK_METHOD}"
     CACHE INTERNAL "Final locking method selected"
 )
+
+# Robust process-shared mutexes: PTHREAD_MUTEX_ROBUST / pthread_mutex_consistent()
+# are a POSIX option glibc and musl provide but some platforms (notably macOS) do
+# not. src/core/tcp_cond.c uses them for the tcp reactor condvar and guards them
+# on HAVE_PTHREAD_MUTEX_ROBUST. Probe the capability instead of guessing by OS.
+include(CheckSymbolExists)
+set(CMAKE_REQUIRED_DEFINITIONS -D_GNU_SOURCE)
+set(CMAKE_REQUIRED_LIBRARIES pthread)
+check_symbol_exists(pthread_mutex_consistent "pthread.h" HAVE_PTHREAD_MUTEX_ROBUST)
+unset(CMAKE_REQUIRED_DEFINITIONS)
+unset(CMAKE_REQUIRED_LIBRARIES)
+if(HAVE_PTHREAD_MUTEX_ROBUST)
+  target_compile_definitions(common INTERFACE HAVE_PTHREAD_MUTEX_ROBUST)
+  message(STATUS "Robust process-shared mutex: supported")
+else()
+  message(STATUS "Robust process-shared mutex: NOT supported (tcp_cond uses non-robust fallback)")
+endif()
+
+# Per-thread names: pthread_setname_np() lets the tcp reactor label its io_wait
+# and pool threads so profilers / top -H / gdb show distinct threads instead of
+# one opaque "kamailio" comm. The symbol exists on Linux and macOS but with
+# different arity - Linux/glibc/musl take (pthread_t, name); macOS takes (name)
+# and renames only the caller. Probe the actual signature rather than guess by OS.
+include(CheckCSourceCompiles)
+set(CMAKE_REQUIRED_DEFINITIONS -D_GNU_SOURCE)
+set(CMAKE_REQUIRED_LIBRARIES pthread)
+check_c_source_compiles(
+  "#include <pthread.h>
+int main(void) { pthread_setname_np(pthread_self(), \"x\"); return 0; }"
+  HAVE_PTHREAD_SETNAME_NP_2ARG
+)
+check_c_source_compiles(
+  "#include <pthread.h>
+int main(void) { pthread_setname_np(\"x\"); return 0; }" HAVE_PTHREAD_SETNAME_NP_1ARG
+)
+unset(CMAKE_REQUIRED_DEFINITIONS)
+unset(CMAKE_REQUIRED_LIBRARIES)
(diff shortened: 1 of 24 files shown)


space88man:
After a few weeks of hammering on this, I plan to merge this soon so that it can be  tested before the next major Kamailio release — unless there are further objections. It can be reverted easily if it is problematic as it is based on file sets in core and module changes are confined to their own commits.

Existing multi-process code is untouched and this multi-threaded TCP receiver  mode requires explicit opt-in. 

This PR does not claim that multi-threaded TCP is "better" than multi-process TCP; instead it provides a genuinely new axis for profiling of Kamailio workloads by splitting network handling from core SIP handling. The design follows udp_receiver_mode,  diverging only where TCP actually differs from UDP (TLS, WS, HEP, MSRP framing; async writes). As previously when tcp_main_threads > 0,  this mode confines SSL to a single-process and obviates the need for pthreads shims and OpenSSL shared memory hooks.
 
LLM was used in this PR:  most importantly to locate all the "seams" where file descriptors were used in TCP workers so that the work could be handled in the thread pool instead and to handle locking of TCP connection structures.  

As a side-effect the investigation surfaced pre-existing bugs in core's event and error code handling (events not firing in some cases, error codes overwritten by a later syscall before being used, tls:connection-out  event not firing) — fixed separately and already in master

henningw:
Thank you for your persistent work on this topic, really appreciated.
Please give it a bit time (e.g. at least one week) before merging it, to allow other developers possible reviewing it. We are still in holiday season, and some people might be away for some days.

stefan-mititelu-idt:
@space88man I have a question related to this refactoring: 

Have you considered/done something special for the half closed TCP connections? E.g. socket blocked in LAST_ACK, but not cleaned. Have you encountered the need for such thing?

space88man:
@stefan-mititelu-idt 
> Have you considered/done something special for the half closed TCP connections? E.g. socket blocked in LAST_ACK, but not cleaned. Have you encountered the need for such thing?

This PR (`tcp_main_threads = 2`) does not change close semantics - it reuses the same TCP support as multi-process TCP: functions like `tcpconn_close_main_fd()`. Think of it as operating at a slightly higher level than the low level TCP plumbing.

In my test harness I did not observe `LAST_ACK` pileup either in modes 0 or 2. However the test harness is limited  to < 10K TCP connections and < 100 cps.

Separately:  if you observe half-closed pile-up in production could you explain a bit more about your setup?  There is an obscure knob in Kamailio `tcp_close_rst = 1`: obscure because I could not find it in the docs but it's code from 2022. It sets `SO_LINGER` behaviour with `l_onoff=1`, `l_linger=0`. The original motivation was `TIME_WAIT` but it covers your `LAST_ACK` case too. 

This knob applies identically to `tcp_main_threads = 2` since the close path wasn't touched.

Reference:
PR#3101: https://github.com/kamailio/kamailio/pull/3101

stefan-mititelu-idt:
> This PR (`tcp_main_threads = 2`) does not change close semantics - it reuses the same TCP support as multi-process TCP: functions like `tcpconn_close_main_fd()`. Think of it as operating at a slightly higher level than the low level TCP plumbing.

Thanks for the comprehensive answer and for the hints! Will definitely check them out.

space88man:
Closed with 0bc6630b.

PR #4826, opened 2026-07-16
core: fix case insensitivity and bounds in _strnistr

Karticoder:
While reviewing the core string utility functions in src/core/str.c, I noticed two issues with the _strnistr implementation:

Case-Insensitivity Bug: The first character comparison sc != c was done case-sensitively. If the search substring and target buffer differed in case on the very first character (such as searching for "Alias" inside a contact parameter "alias="), it would fail to match and miss the substring entirely.
Buffer Overread: The bounds checking sequence in the inner loop (sc = *s++) == '\0' || slen-- < 1 was ordered incorrectly. The dereference *s++ occurred before checking if slen had reached zero, potentially reading one byte past the buffer's designated length constraint.
This patch refactors the function to:

Validate input parameters (s, find, and slen) to prevent NULL dereferences.
Calculate a safe sliding-window limit pointer (last) to avoid any out-of-bounds pointer increments.
Perform a fully case-insensitive comparison using tolower on both characters.

### DIFF
--- a/src/core/str.c
+++ b/src/core/str.c
@@ -24,6 +24,7 @@
  * Module: \ref core
  */
 
+#include <ctype.h>
 #include <string.h>
 #include "str.h"
 #include "mem/mem.h"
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the PR. This is a refinement of #4824, so the other one can be closed? Or are there alternative implementations?

Karticoder:
Yes, this PR is a refinement of #4824 that corrects the bounds-checking and case-insensitivity logic. The other PR (#4824) can be closed. Thank you

henningw:
Please don't change the github workflow files in this PR. If you like to propose an enhancement there, please create a dedicated PR.

Karticoder:
Understood. I have reverted the changes to the GitHub workflow files in this PR. If we want to move forward with those enhancements, I'll submit a dedicated PR for them. Thanks for the heads up!"

henningw:
Please do a force-push to properly remove this commits. Otherwise they will be merged to the public tree, which is not what we want from this PR.

Karticoder:
Done. I've  force-pushed to clean up the commit history. The PR now only contains the clean C code changes for _strnistr and no longer includes any workflow-related commits. Thanks

henningw:
#4830

PR #4830, opened 2026-07-17
core: fix case-insensitive search and buffer overread in _strnistr

Karticoder:
While checking _strnistr() in src/core/str.c, I noticed it performs a case-sensitive check on the first character of the search target, which breaks the case-insensitive search. Additionally, the boundary check evaluates after the pointer increment, leading to a 1-byte buffer overread on non-null-terminated strings. I've reordered the condition to prevent the overread and pre-computed the uppercase/lowercase characters for an efficient, case-insensitive match.

### DIFF
--- a/src/core/str.c
+++ b/src/core/str.c
@@ -25,6 +25,7 @@
  */
 
 #include <string.h>
+#include <ctype.h>
 #include "str.h"
 #include "mem/mem.h"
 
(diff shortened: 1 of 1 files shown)


henningw:
Are you creating this PRs automatically? This looks like the third PR in this area. Please explain the relationship to the previous PR #4826 and stop doing this automatically.

Karticoder:
"I apologize for the confusion and for opening a new PR instead of updating the branch for #4826.

The relationship is that this PR is a corrected and optimized version of the changes in #4826. In the implementation of #4826, there were a few issues:

It was inefficient because it called tolower() inside the inner scanning loop.
It had a correctness bug where it would return NULL early on a partial mismatch (like searching "ab" in "aab").
It lacked (unsigned char) casts for safety.
This PR pre-computes the lowercase/uppercase characters outside the loop to keep the scanning hot path efficient, ensures correct outer-loop continuation, and adds character casts.

I will close PR #4826 in favor of this one. Sorry again for the noise!"

abalashov:
@Karticoder I am uplifted that you pasted your LLM response in with double quotation marks, as appropriate. Very "AI-native". 

Please, please stop vandalising our project with this crap.

PR #4861, opened 2026-08-04
dispatcher: refcount ds_set_t table

### LINKED ISSUE #4807: dispatcher: race between concurrent reloads corrupts shared memory (double-free in ds_avl_destroy)
### Description

Concurrent dispatcher reloads can crash kamailio: the module keeps its destination table in two shared-memory slots, and two worker processes reloading at the same time can rebuild the same slot, double-freeing its contents. A worker eventually dies with signal 11/6 (with the default `mem_safety=1` the allocator first logs double-free BUGs and keeps running while corruption accumulates).

It happens whenever reloads can run in more than one process concurrently — `ds_reload()` from the routing script (as in the reproducer below) or the RPC reload served by multiple processes. We hit it in production; it reproduces on current `master`, and the same code is on the `6.0` branch. Expected behavior: overlapping reloads may be rejected or serialized, but should not corrupt memory. Same class of issue as #4809 (permissions).

#### Reproduction

Self-contained Docker reproducer: [kamailio-dispatcher-reload-race-repro-with-pcap.zip](https://github.com/user-attachments/files/29594512/kamailio-dispatcher-reload-race-repro-with-pcap.zip)

```sh
./run.sh             # unpatched master -> shm double free / corruption within seconds
./run.sh --patched   # + reload lock    -> same stress, survives
```

It builds kamailio with `-x q_malloc` (DBG_QM_MALLOC), `children=16`, a 4000-destination file-based dispatcher list and `reload_delta=0`, calls `ds_reload()` on every received OPTIONS, then floods OPTIONS from 32 UDP senders for 25 s. Unpatched: one archived run logs 2 `fragm. ... overwritten` BUG lines and a worker dies with signal 11; another logs 4 `freeing already freed pointer` BUG lines and a worker dies with signal 6. Patched run: 0 BUG lines, no crash. Full logs, core backtraces and pcaps are in the archive.

#### Analysis

The reload swaps between the module's two shm table slots without a lock. In `ds_load_list()` and `ds_load_db()` (`src/modules/dispatcher/dispatch.c`):

```text
if (*ds_crt_idx != *ds_next_idx) abort     (reload-in-progress guard)
*ds_next_idx = (*ds_crt_idx + 1) % 2       (pick the inactive slot)
ds_avl_destroy(&ds_lists[*ds_next_idx])    (shm_free its contents)
add_dest2list() / reindex_dests()          (repopulate the inactive slot)
*ds_crt_idx = *ds_next_idx                 (publish)
```

The guard is a non-atomic check-then-set on shared memory: two workers can both pass it, pick the same inactive slot and destroy/repopulate it concurrently — the double free above. It does fire when timing allows (16 and 3876 `load command already generated` warnings in the two archived runs) but cannot close the window. `ds_add_dst()` / `ds_remove_dst()` do the same slot swap with no guard at all, and the `reload_delta` gate in `ds_reload()` (`dispatcher.c`) is likewise a non-atomic check-then-set.

#### Debugging Data

Backtrace from the archived core (`logs/coredump-bt-unpatched.txt`, full `bt full` inside): the main process crashing at shutdown while destroying the already-corrupted table, after child 8 died with signal 11 during the flood:

```text
#2  qm_free (..., func="ds_avl_destroy", line=5141, ...) at core/mem/q_malloc.c:569
#4  ds_avl_destroy (node_ptr=0xffffb5a19518) at dispatch.c:5141
#7  ds_destroy_list () at dispatch.c:1731
#8  destroy () at dispatcher.c:589
```

#### Log Messages

```text
CRITICAL: <core> [core/mem/q_malloc.c:578]: qm_free(): BUG: freeing already freed pointer (0xffff99c9e870), called from dispatcher: dispatch.c: ds_avl_destroy(5136), first free dispatcher: dispatch.c: ds_set_attrs(383) - ignoring
CRITICAL: <core> [core/mem/q_malloc.c:130]: qm_debug_check_frag(): BUG: qm: fragm. 0xffffb68bc530 (address 0xffffb68bc570) beginning overwritten (0)! Memory allocator was called from dispatcher: dispatch.c:1201. Fragment marked by (null):0. Exec from core/mem/q_malloc.c:439.
ALERT: <core> [main.c:824]: handle_sigs(): child process 8 exited by a signal 11
```

#### SIP Traffic

Not message-dependent — the flood is identical plain OPTIONS requests, only the rate matters. Generator (`stress/flood.py`) and captures (`pcap/`) are in the archive.

### Possible Solutions

Serialize the rebuild critical section (guard/select/destroy/repopulate/publish) with a shm lock in `ds_load_list()` / `ds_load_db()` / `ds_add_dst()` / `ds_remove_dst()`; the per-request selection hot path stays lock-free and keeps reading the active slot as before. Tested under the same stress (the `--patched` run above): 0 BUG lines, no crash. The patch is in the archive (`build/dispatcher-reload-lock.patch`); I can open a pull request if the approach is acceptable.

### Additional Information

- [x] LLM/AI Assistants were involved in discovery, analysis or submitting of the issue
  (analysis and write-up were AI-assisted; the reproducer and all runs were executed and verified by me)
- [x] It happened in a production deployment
- [x] It happened in a testing or development deployment
- [x] Testing or fuzzing tools were used to generate SIP traffic when it happened

  * **Kamailio Version** - output of `kamailio -v`

```text
version: kamailio 6.2.0-dev1 (aarch64/linux)
flags: USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, MEM_JOIN_FREE, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, DBG_QM_MALLOC, DBG_F_MALLOC, DBG_TLSF_MALLOC, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLOCKLIST, HAVE_RESOLV_RES, TLS_PTHREAD_MUTEX_SHARED
compiled on 09:55:20 Jul  2 2026 with gcc 14.2.0
```

Reproduced on `master` (`b3a8384d9d10343317b3df83c595f5cfe570d14a`); the same unguarded code is on the `6.0` branch (`5709418e6a47768e4f729e44530722bc42fedf1c`). Originally observed in production.

  * **Operating System**:

```text
Debian GNU/Linux 13 (trixie)
Linux 6.10.14-linuxkit aarch64 (Docker)
```

### LINKED ISSUE #4845: dispatcher: segfault in algorithm 10, ds_ht_timer() - corrupted cell pointer in call-load hash table chain
### Description

Kamailio 6.1.3 segfaults intermittently in `ds_ht_timer()` (dispatcher call-load hash table expiry timer) under production traffic. The timer crashes while walking a hash-slot chain: the `ds_cell_t *` it is iterating contains ASCII text originating from SIP message data instead of a valid address.

This happens roughly 2 times per day on a single instance under live traffic. Each crash restarts the process and causes a ~30s service outage.

`ds_ht_timer()` itself locks correctly (it holds `entries[i].lock` for the entire traversal), so the corruption appears to come either from a path that does not respect that lock, or from a use-after-free of a cell that is still linked into the chain. A few candidate issues in `ds_ht.c` are noted under "Possible Solutions" below.

#### Reproduction

Not reproducible on demand. It occurs sporadically under real traffic with the call-load distribution algorithm enabled. Relevant configuration:

```
modparam("dispatcher", "flags", 2)
modparam("dispatcher", "ds_hash_size", 8)
modparam("dispatcher", "ds_hash_expire", 3600)
modparam("dispatcher", "ds_timer_mode", 1)
modparam("dispatcher", "force_dst", 1)
```

```
ds_select_domain("1", "10");   # algorithm 10 = call load distribution
```

The configuration does not call `ds_load_update()` / `ds_load_unset()` explicitly; call-load accounting relies on the dialog module callbacks. The `dialog` module is loaded and active.

#### Debugging Data

Backtrace from the core dump:

```
#0  ds_ht_timer ()
#1  sr_wtimer_exec ()
#2  fork_sync_timer ()
#3  sr_wtimer_start ()
#4  main_loop ()
#5  main ()
```

The faulting instruction is `it0 = it->next;` at the top of the `while(it)` loop:

```asm
ds_ht_timer+159:  test   %rbx,%rbx           ; while(it)
ds_ht_timer+162:  je     <ds_ht_timer+249>
ds_ht_timer+164:  mov    %rbx,%rbp           ; it
ds_ht_timer+167:  mov    0x48(%rbx),%rbx     ; it0 = it->next   <-- SIGSEGV
ds_ht_timer+171:  mov    0x30(%rbp),%r9      ; it->expire
ds_ht_timer+175:  cmp    %r15,%r9            ; compare against now
```

Registers at the fault:

```
rbx     = 0x7a3d68636e617262     ; the `it` pointer itself
rbp     = 0x7a3d68636e617262
r15     = 0x6a68aa24             ; 1785244196 == time(NULL) at crash
si_addr = 0x0                    ; non-canonical address -> GP fault
```

`0x48` is `ds_cell_t.next`, and `0x30` / `0x38` are `expire` / `initexpire`, which matches the loop body.

The faulting pointer value `0x7a3d68636e617262` is little-endian ASCII **`"branch=z"`** — the leading bytes of a SIP Via `branch=z9hG4bK...` parameter. So a cell pointer reachable from the slot chain contained SIP message text.

A second crash on the same instance the same day terminated inside a libc string/memory routine with a garbage pointer argument, which is consistent with the same shared-memory corruption surfacing at a different site.

#### Log Messages

No dispatcher errors precede the crash. The only logged events are:

```
child process N exited by a signal 11
core was generated
```

#### SIP Traffic

Nothing unusual around the time of the crash — ordinary inbound INVITE traffic. The crash is in the timer process, not in SIP message processing.

### Possible Solutions

While investigating I noticed the following in `src/modules/dispatcher/ds_ht.c`. This file is identical on `6.1` and `master`. I have not proven that any of these causes the crash, but they seem worth reviewing:

**1. `ds_unlock_cell()` can return without releasing the lock.**

```c
	/* head test and return */
	if(dsht->entries[idx].first == NULL)
		return 0;                 /* <-- returns with the lock still held */

	lock_release(&dsht->entries[idx].lock);
```

If the slot becomes empty between `ds_get_cell()` and `ds_unlock_cell()`, the slot lock is leaked and `ds_ht_timer()` would then block on it permanently.

**2. `ds_get_cell()` returns with the slot lock held**, and callers are required to pair it with `ds_unlock_cell()`. This contract is not obvious from the function name, and it makes every early return in a caller a potential lock leak.

**3. The "head test" in both `ds_get_cell()` and `ds_unlock_cell()` reads `dsht->entries[idx].first` without holding the lock**, concurrently with writers that modify `first` under the lock.

**4. Callers use unlock-then-delete**, for example in `dispatch.c`:

```c
	ds_unlock_cell(_dsht_load, &msg->callid->body);
	ds_del_cell(_dsht_load, &msg->callid->body);
```

The lock is dropped and reacquired between locating and deleting the cell, leaving a window in which the timer can expire and free the same cell.

### Additional Information

  * **Kamailio Version** - 6.1.3, installed from the official `deb.kamailio.org/kamailio61` packages

  * **Operating System**: Debian 12 (bookworm), x86_64, running in a container

rfuchs:
- refactor table object living in shared memory with a wrapper object (`ds_list_t` wrapping shared `ds_set_t` list)
- ds_list_t wrapper includes a reference count for thread safety
- reference count is protected by new global `ds_list_read_lock` mutex
- previously global `_ds_list_nr` count has been moved into the `ds_list_t` object
- read access to shared table must now happen through `ds_get_list()` or similar, which: obtains read lock, increases reference count, drops read lock, returns pointer to table with reference held
- calls to `ds_get_list()` must be paired with `ds_put_list()` to release the reference
- double-buffering of `ds_set_t` objects (via `ds_crt_idx` etc) is now obsolete: the global `ds_list` pointer always points to the single currently active list and is protected by the `ds_list_read_lock` mutex, and safe read access is guaranteed through the reference count even in case of a concurrent reload or write access
- write accesses continue to operate using copy-rewrite-replace and are therefore equivalent to full reloads
- write accesses are required to also obtain `ds_list_write_lock` in addition to `ds_list_read_lock`
- reloads or rewrites create a new `ds_list_t`, populate it, then swap out the global `ds_list` pointer
- after swapping out the pointer, the reload/rewrite thread waits until the old object's reference count drops to zero, and only then frees it, guaranteeing safety for readers in other threads
- `reload_delta` is now obsolete: reloads and rewrites are now safe at arbitrarily short intervals

Closes #4845
Closes #4807

#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #4845 and #4807

#### Description
Full thread safety for dispatcher with arbitrary reloads and concurrent read access

Possible caveat: Removal of `reload_delta` would throw an error after an upgrade for people who have it set

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -134,21 +134,16 @@ extern int ds_event_callback_mode;
 static db_func_t ds_dbf;
 static db1_con_t *ds_db_handle = NULL;
 
-static ds_set_t **ds_lists = NULL;
+static ds_list_t **ds_list = NULL; /* pointer to current list of sets */
 static gen_lock_t *ds_list_write_lock = NULL;
-
-static int *ds_list_nr = NULL;
-static int *ds_crt_idx = NULL;
-static int *ds_next_idx = NULL;
+static gen_lock_t *ds_list_read_lock = NULL;
+/* lock order for locking both: write lock first, read lock second */
 
 static ds_set_t *ds_strictest_node = NULL;
 static int ds_strictest_idx = 0;
 static int ds_strictest_match = 0;
 
 static sruid_t _ds_sruid = {0};
-
-#define _ds_list (ds_lists[*ds_crt_idx])
-#define _ds_list_nr (*ds_list_nr)
 
 void ds_rctx_set_uri(ds_rctx_t *rctx, str *uri);
 static void ds_run_route(
@@ -255,15 +250,15 @@ int ds_hash_load_destroy(void)
 /**
  *
  */
-static inline int ds_get_index(int group, int ds_list_idx, ds_set_t **index)
+static inline int ds_get_index(int group, ds_list_t *list, ds_set_t **index)
 {
 	ds_set_t *si = NULL;
 
-	if(index == NULL || group < 0 || ds_lists[ds_list_idx] == NULL)
+	if(index == NULL || group < 0 || list == NULL)
 		return -1;
 
 	/* get the index of the set */
-	si = ds_avl_find(ds_lists[ds_list_idx], group);
+	si = ds_avl_find(list->head, group);
 
 	if(si == NULL)
 		return -1;
@@ -306,7 +301,7 @@ void ds_log_dst_cb(ds_set_t *node, int i, void *arg)
 /**
  * Recursivly print ds_set
  */
-void ds_log_set(ds_set_t *node)
+static void ds_log_set(ds_set_t *node)
 {
 	ds_iter_set(node, &ds_log_dst_cb, NULL);
 
@@ -316,14 +311,59 @@ void ds_log_set(ds_set_t *node)
 /**
  *
  */
-int ds_log_sets(void)
+static int ds_log_sets(ds_list_t *list)
 {
-	if(_ds_list == NULL)
-		return -1;
-
-	ds_log_set(_ds_list);
+	ds_log_set(list->head);
 
 	return 0;
+}
+
+/**
+ * Construct a new list object in shared memory
+ */
+static ds_list_t *ds_new_list(void)
+{
+	ds_list_t *ret;
+
+	ret = shm_malloc(sizeof(ds_set_t *));
+	if(!ret)
+		return NULL;
+
+	memset(ret, 0, sizeof(*ret));
+	return ret;
+}
+
+/**
+ * Destroys the list object and the contained AVL tree
+ */
+static void ds_free_list(ds_list_t *list)
+{
+	if(!list)
+		return;
+	ds_avl_destroy(&list->head);
+	shm_free(list);
+}
+
+/**
+ * Wait for open references and then destroy the list
+ */
+static void ds_wait_free_list(ds_list_t *list)
+{
+	unsigned int iters;
+
+	if(!list)
+		return;
+
+	lock_get(ds_list_read_lock);
+	while(list->refs > 0) {
+		lock_release(ds_list_read_lock);
+		iters++;
+		sleep_us(iters * 10000);
+		lock_get(ds_list_read_lock);
+	}
+	lock_release(ds_list_read_lock);
+
+	ds_free_list(list);
 }
 
 /**
@@ -331,37 +371,30 @@ int ds_log_sets(void)
  */
 int ds_init_data(void)
 {
-	int *p;
-
-	ds_lists = (ds_set_t **)shm_malloc(2 * sizeof(ds_set_t *));
-	if(!ds_lists) {
+	ds_list = shm_malloc(sizeof(ds_set_t *));
+	if(!ds_list) {
 		SHM_MEM_ERROR;
 		return -1;
 	}
-	memset(ds_lists, 0, 2 * sizeof(ds_set_t *));
-
-
-	p = (int *)shm_malloc(3 * sizeof(int));
-	if(!p) {
-		shm_free(ds_lists);
-		SHM_MEM_ERROR;
-		return -1;
-	}
-	memset(p, 0, 3 * sizeof(int));
+	*ds_list = NULL;
 
 	ds_list_write_lock = lock_alloc();
 	if(!ds_list_write_lock) {
-		shm_free(ds_lists);
-		shm_free(p);
+		shm_free(ds_list);
 		SHM_MEM_ERROR;
 		return -1;
 	}
 	lock_init(ds_list_write_lock);
 
-	ds_crt_idx = p;
-	ds_next_idx = p + 1;
-	ds_list_nr = p + 2;
-	*ds_crt_idx = *ds_next_idx = 0;
+	ds_list_read_lock = lock_alloc();
+	if(!ds_list_read_lock) {
+		shm_free(ds_list);
+		lock_destroy(ds_list_write_lock);
+		lock_dealloc(ds_list_write_lock);
+		SHM_MEM_ERROR;
+		return -1;
+	}
+	lock_init(ds_list_read_lock);
 
 	return 0;
 }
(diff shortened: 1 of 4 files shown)


henningw:
Thanks for the PR, needs a closer review for sure. Did you made some tests from your side on a test setup or eventual production traffic? 

Regarding the removal of reload_delta, this is in my opinion no problem if done for a major release. Just needs some documentation, as usual.

rfuchs:
> Thanks for the PR, needs a closer review for sure. Did you made some tests from your side on a test setup or eventual production traffic?

I tested it in a lab setup with a number of different combinations of reloads and RPC add/remove interactions.

miconda:
I looked a bit over it, I am fine to merge it, just wondering if reload_delta should be kept in some way (this PR can be merged and then re-added, disabled by default) as a throttling mechanism, for some deployments it might still be useful to have kind of safety protection agains mistaken/abusive reloads done by external apps.

henningw:
Thanks for the PR, merged it. There were are few redundant error messages introduced related to memory allocation problems, will have a look probably in a second commit.

PR #4863, opened 2026-08-05
dispatcher: add per-destination ping_reply_codes attribute

fred-clement-91:
### Description

Adds a per-destination `ping_strict` attribute to the dispatcher module.

When a destination has `ping_strict=1` in its `attrs`, only a 2xx reply to the
keepalive OPTIONS keeps it active; the global `ds_ping_reply_codes` list is
ignored for that destination. This makes it possible to require a strict
`200 OK` for selected destinations while keeping the lenient global behaviour
(e.g. accepting `401`/`404`/`486`) for the others — which is currently not
possible because `ds_ping_reply_codes` is a global modparam.

The default (`ping_strict=0`) preserves the historical behaviour, so existing
configurations are unaffected.

### Implementation

- new `int ping_strict` field in `ds_attrs_t`;
- parsed in `ds_set_attrs()` like the other destination attributes;
- `ds_get_ping_strict()` helper, mirroring the `ds_get_state()` lookup;
- in `ds_options_callback()` the global reply-code check is skipped for strict
  destinations: `2xx || (!ping_strict && ds_ping_check_rplcode())`;
- documented in `doc/dispatcher_admin.xml`.

### Usage

Set it in the destination `attrs` (DB `attrs` column, dispatcher list file, or
`dispatcher.add`), e.g.:

```
1 sip:10.0.0.1:5060 0 0 ping_strict=1
```

### Testing

Built and loaded against 6.0.1. Verified on a live instance: with a lenient
global `ds_ping_reply_codes` accepting `404`, two destinations pointing at the
same 404-replying target — the one with `ping_strict=1` is marked inactive
while the one without stays active.

### DIFF
--- a/src/modules/dispatcher/dispatcher.c
+++ b/src/modules/dispatcher/dispatcher.c
@@ -1143,33 +1143,48 @@ static int w_ds_is_active_uri(sip_msg_t *msg, char *pset, char *puri)
 	return ki_ds_is_active_uri(msg, vset, &suri);
 }
 
-static int ds_parse_reply_codes()
+/**
+ * \brief Parse a reply-codes string into a newly shm-allocated int array.
+ *
+ * Accepts the same 'class=N'/'code=N' tokens as the ds_ping_reply_codes
+ * modparam; entries are 1..6 for a class and 100..699 for an exact code.
+ * Tokens may be separated by ';' or ',' (the per-destination attribute uses
+ * ',' because ';' is the attributes separator). Returns the number of parsed
+ * entries (0 if none), or -1 on error; *pcodes receives the array (NULL if 0).
+ */
+int ds_parse_reply_codes_list(str *input, int **pcodes)
 {
 	param_t *params_list = NULL;
 	param_t *pit = NULL;
 	int list_size = 0;
-	int i = 0;
 	int pos = 0;
 	int code = 0;
-	str input = {0, 0};
-	int *ds_ping_reply_codes_new = NULL;
-	int *ds_ping_reply_codes_old = NULL;
+	str in = STR_NULL;
+	char *buf = NULL;
+	int *codes_new = NULL;
 
-	/* validate input string */
-	if(cfg_get(dispatcher, dispatcher_cfg, ds_ping_reply_codes_str).s == 0
-			|| cfg_get(dispatcher, dispatcher_cfg, ds_ping_reply_codes_str).len
-					   <= 0)
+	*pcodes = NULL;
+	if(input == NULL || input->s == NULL || input->len <= 0)
 		return 0;
 
-	/* parse_params() updates the string pointer of .s -- make a copy */
-	input.s = cfg_get(dispatcher, dispatcher_cfg, ds_ping_reply_codes_str).s;
-	input.len =
-			cfg_get(dispatcher, dispatcher_cfg, ds_ping_reply_codes_str).len;
-
-	if(parse_params(&input, CLASS_ANY, 0, &params_list) < 0)
+	/* work on a private copy: parse_params() advances the str it is given
+	 * and modifies the buffer in place. Keep the original allocation in buf
+	 * to free it, as in.s is moved forward by parse_params(). */
+	buf = (char *)pkg_malloc(input->len + 1);
+	if(buf == NULL) {
+		PKG_MEM_ERROR;
 		return -1;
+	}
+	memcpy(buf, input->s, input->len);
+	buf[input->len] = '\0';
+	in.s = buf;
+	in.len = input->len;
 
-	/* get the number of entries in the list */
+	if(parse_params(&in, CLASS_ANY, 0, &params_list) < 0) {
+		pkg_free(buf);
+		return -1;
+	}
+
 	for(pit = params_list; pit; pit = pit->next) {
 		if(pit->name.len == 4 && strncasecmp(pit->name.s, "code", 4) == 0) {
 			str2sint(&pit->body, &code);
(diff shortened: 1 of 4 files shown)


henningw:
Thanks for the PR.
Have you thought about instead of adding a special "ping_strict" attribute, adding a generic "ping_reply_codes" to the dispatcher destination configuration? This way it could be completely customised per destination instead of your proposed solution which only allow 2XX requests.

fred-clement-91:
Thanks for the review, that's a much better idea — it's more flexible and consistent with the existing global parameter.

I've reworked the PR accordingly: instead of the boolean `ping_strict`, it now adds a per-destination **`ping_reply_codes`** attribute that overrides the global `ds_ping_reply_codes` for that destination. It reuses the same `class=N`/`code=N` parsing — I factored the existing parser and checker into `ds_parse_reply_codes_list()` and `ds_ping_check_rplcode_list()`, so the global modparam and the per-destination attribute share the same code.

One small detail: since `;` is already the attribute separator, the entries inside the attribute value are separated by `,` instead, e.g.:

```
1 sip:10.0.0.1:5060 0 0 ping_reply_codes=class=2,code=404
```

When a destination has no `ping_reply_codes` attribute, the global `ds_ping_reply_codes` list is used, so existing setups are unaffected. The original "strict 200 OK only" use case simply becomes `ping_reply_codes=class=2`.

I've force-pushed the update.

henningw:
Thanks, can you please have a look to the clang-format issues? Force-push as usual after you addressed it. Then probably there will be a bit more code reviews.

fred-clement-91:
I've addressed the clang-format issues and force-pushed (`b09f405`).

While testing this I also caught and fixed a bug in the new parser: `parse_params()` advances the `str` it is passed, so the subsequent `pkg_free(in.s)` was freeing an offset pointer (invalid free / heap corruption). I now keep the original allocation in a separate `char *buf` and free that instead. I also added the missing `shm_free()` of the per-destination `ping_reply_codes` array in `ds_avl_destroy()`.

Verified: builds cleanly on master, `clang-format-18 --dry-run --Werror` is clean on all three source files, and a live test behaves as expected — a destination replying `404` with `ping_reply_codes=class=2` is probed down, the same destination with `ping_reply_codes=code=404` is probed up, and the global `ds_ping_reply_codes` list is used when the attribute is absent.

src/modules/dispatcher/doc/dispatcher_admin.xml
```diff
@@ -2682,6 +2682,15 @@ kamcli dispatcher.oclist 1
 							<para>'ping_from' - used to set the From URI in OPTIONS keepalives. It
 								overwrites the general ds_ping_from parameter.</para>
 						</listitem>
+						<listitem>
+								<para>'ping_reply_codes' - accepted reply codes for the OPTIONS
+									keepalive of this destination, overriding the global
+									ds_ping_reply_codes parameter for it. It uses the same
+									'class=N'/'code=N' syntax, but the entries are separated by
+									',' instead of ';' (';' separates the attributes), e.g.
```
miconda: Is it working with `,` as separator? Because the comma is end of header value and the parse params might stop there. Just to double check.

Otherwise it might just be enough to enclose in double quotes, like:

```
ping_reply_codes="class=2;code=404"
```

src/modules/dispatcher/dispatcher.c
```diff
@@ -1182,36 +1203,56 @@ static int ds_parse_reply_codes()
...
 		}
-
-		/* Now create the list of valid reply-codes: */
 		for(pit = params_list; pit; pit = pit->next) {
 			if(pit->name.len == 4 && strncasecmp(pit->name.s, "code", 4) == 0) {
 				str2sint(&pit->body, &code);
-				if((code >= 100) && (code < 700)) {
-					ds_ping_reply_codes_new[pos++] = code;
-				}
+				if((code >= 100) && (code < 700))
```
miconda, line 1212: Removing the curly braces is not necessary, in many cases are used for code clarity even if it is a single statement. Do only the changes needed for the new feature to make the review easier, cosmetic changes can be proposed as separate PR.

src/modules/dispatcher/dispatcher.c
```diff
@@ -1182,36 +1203,56 @@ static int ds_parse_reply_codes()
...
+					codes_new[pos++] = code;
 			} else if(pit->name.len == 5
 					  && strncasecmp(pit->name.s, "class", 5) == 0) {
 				str2sint(&pit->body, &code);
-				if((code >= 1) && (code < 7)) {
-					ds_ping_reply_codes_new[pos++] = code;
-				}
+				if((code >= 1) && (code < 7))
```
miconda, line 1217: Curly braces removed without a specific purpose for this PR.

src/modules/dispatcher/doc/dispatcher_admin.xml
fred-clement-91: Good catch, you're right. I tested it and confirmed: with an unquoted `,`, `parse_params()` stops at the comma (it is a value terminator in `parse_token_param()`), so `ping_reply_codes=class=2,code=404` only keeps `class=2` and `code=404` is silently dropped.

Enclosing the value in double quotes works — `parse_quoted_param()` keeps the whole string including the `;`. On a small test bench (a destination replying `404`, with `ds_ping_reply_codes=class=2` as the global default):

- `ping_reply_codes=class=2,code=404` → destination stays probing/down (only `class=2` parsed, `404` rejected)
- `ping_reply_codes="class=2;code=404"` → destination goes active (both `class=2` and `code=404` honored)

So I switched to the quoted `;` form, which is also consistent with the global `ds_ping_reply_codes` syntax, and dropped the now-unnecessary `,`→`;` handling. Doc and code updated and force-pushed.

henningw:
@fred-clement-91 Can you also please address the previous review comments about the braces removal?

henningw:
> @fred-clement-91 Can you also please address the previous review comments about the braces removal?

commit f66277eae6

PR #4865, opened 2026-08-10
xhttp_prom: emit TYPE lines for internal statistics

draftcode:
#### Pre-Submission Checklist
  from source code folder

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description

`metric_generate()` emits the internal Kamailio statistics without a `# TYPE`
line. The Prometheus exposition format makes `# TYPE` optional and treats such
a sample as untyped, but strict parsers drop untyped samples entirely, so the
internal statistics are invisible to those scrapers. User defined metrics are
unaffected — they already emit `HELP` and `TYPE` via `prom_metric_list_print()`.

This is the same problem reported in #3001, which was closed in 2022 because
the reporter had moved on to a workaround rather than because the approach was
rejected. In that thread @vhernando explained the original reasoning (the
module re-exports Kamailio's registered counters and cannot know which are
counters and which are gauges) and then proposed exactly the fix implemented
here:

> I am thinking on a workaround. Would asigning gauge type to Kamailio's
> metrics, even if they are actually counters, suit you?

`gauge` is the safe choice rather than `counter` even though most of these
statistics only increase: Prometheus reserves the `_total` suffix for counters
and these names do not carry it (e.g. `kamailio_sl_400_replies`), so declaring
them `counter` would make the output invalid for the same strict parsers this
change is meant to satisfy. A gauge accepts a monotonically increasing series
without complaint.

##### Reproduction and verification

The scraper in our case is the Datadog Agent's `openmetrics` check. Against an
unpatched module it reaches the endpoint and reports the instance healthy, but
collects zero samples — the failure is silent, which is what makes it awkward
to diagnose.

Verified against 6.0.3 by building the module and swapping it into the official
container image. After the change all 55 internal statistics carry a `TYPE`
line, and Datadog Agent 7.73.0 collects 55 samples where it previously
collected 0.

Before:

```
# Kamailio whole internal statistics
kamailio_core_bad_URIs_rcvd 0 1640265350616
kamailio_core_bad_msg_hdr 0 1640265350616
```

After:

```
# Kamailio whole internal statistics
# TYPE kamailio_core_bad_URIs_rcvd gauge
kamailio_core_bad_URIs_rcvd 0 1640265350616
# TYPE kamailio_core_bad_msg_hdr gauge
kamailio_core_bad_msg_hdr 0 1640265350616
```

I have left the backport checkbox unticked — happy to prepare a cherry-pick for
the stable branches if maintainers consider it appropriate.

### DIFF
--- a/src/modules/xhttp_prom/prom.c
+++ b/src/modules/xhttp_prom/prom.c
@@ -222,6 +222,31 @@ static int metric_generate(
 			group->len, group->s, name->len, name->s, counter_val,
 			(uint64_t)ts);
 
+	/* Print TYPE line. The Prometheus exposition format makes it optional, but
+	 * strict parsers treat a sample with no declared type as untyped and drop
+	 * it, so statistics exported without one are invisible to those scrapers.
+	 * User defined metrics already emit HELP and TYPE (see
+	 * prom_metric_list_print); internal statistics did not.
+	 *
+	 * They are declared gauge rather than counter even though most of them only
+	 * increase: Prometheus reserves the _total suffix for counters, and these
+	 * names do not carry it, so declaring counter makes the output invalid for
+	 * the same strict parsers this is meant to satisfy. */
+	if(prom_body_printf(ctx, "# TYPE ") == -1) {
+		LM_ERR("Fail to print\n");
+		return -1;
+	}
+	if(prom_body_name_printf(ctx, "%.*s%.*s_%.*s", xhttp_prom_beginning.len,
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the PR, lets wait a bit for feedback e.g. from @vhernando or others. Regarding the eventual cherry-pick to stable branches, this is not done with a PR, so not needed to create one here.

henningw:
Merged it, to get eventual feedback from people testing master branch.

PR #4870, opened 2026-08-13
cmake: add sbom target generating a CycloneDX SBOM

neutrino38:
#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Add an opt-in `sbom` custom target that generates a CycloneDX 1.6 JSON SBOM (build/sbom/kamailio-sbom.cdx.json) describing the kamailio binary, the selected modules and the external libraries they link.

```bash
$ cmake --build build -t sbom
```

The inventory is taken from the built artifacts: 
- ldd provides the full runtime closure, 
- readelf -d the direct DT_NEEDED edges, 
- each library is resolved to the owning distribution package (dpkg-query, then rpm, then a pkg:generic fallback) so components carry exact distro versions.-
- The dependencies section expresses the real package-level graph.

python3 is required. It is looked up with find_program and the target is skipped silently when missing;


neutrino38:
The main purpose of this is to easy the detection of vulnerabilities.

henningw:
Thanks for the PR. I think that there are two main types of SBOM formats. SDPX is more orientated towards licence compliance, and CycloneDX more towards security. We are already having SPDX identifier added to the code.

So its probably useful to add a qualifier to this new addition, that we are not having a name clash when later SPDX is added as well.

For example add "-cyclonedx" as suffix to the new files and the cmake command.

oej:
SPDX license identifiers are separate from SBOM formats and used both in SPDX and CycloneDX, so our usage of SPDX license identifiers is no problem.

oej:
Please upload as an attachment a generated SBOM so I can review it. 

Why focus on an old version of Cyclone?

Have you verified the generated identifiers (CPE and/or PURL) ?

henningw:
> SPDX license identifiers are separate from SBOM formats and used both in SPDX and CycloneDX, so our usage of SPDX license identifiers is no problem.

Thanks for the clarification, this was also my understanding. I was only concerned about the naming, that we should not name it only "sbom", but e.g. "sbom-cyclonedx" in the cmake command and file names.

neutrino38:
Ok, doing that. As for the file names, they end with ".cdx.json". I believe that this is clear enough

henningw:
> Ok, doing that. As for the file names, they end with ".cdx.json". I believe that this is clear enough

Thanks. I probably could explain the naming better. Regarding the file suffix I was referring to the sbom-generate.py, sbom-metadata.json.in and sbom.cmake.

oej:
Remember, I think we're on the way to something good here. But we need to be clear about why we're creating an sbom. Is it for license compliance or for vulnerability management or both?

You will likely not find any pkg://generic in any CVE filed so far. If we want vulnerability management we need to find a better way to populate the identifiers.

For debian or redhat you can surely use the PURL syntax for respective system, but then you will have to also add the Kamailio package names for kamailio modules.

neutrino38:
This is now generating CycloneDX 1.7 and I got the id checked.

neutrino38:
> For debian or redhat you can surely use the PURL syntax for respective system, but then you will have to also add the Kamailio package names for kamailio modules.

Can you elaborate?

oej:
> > For debian or redhat you can surely use the PURL syntax for respective system, but then you will have to also add the Kamailio package names for kamailio modules.
> 
> Can you elaborate?

I think it's easier to do this in a call. 

PURLs for packaged systems (which you have special code for) have a different syntax than the generic PURL you have. But we also need to include the Kamailio packages which does not match one-to-one with the Kamailio modules for each platform.

oej:
I have done some experiments on a source-code-level SBOM for Kamailio where we tried to find the dependencies in the CVE database to find existing CPEs and PURLs for them. I can show you that work

neutrino38:
Here is an SBOM generated.

[kamailio-sbom.cdx.json](https://github.com/user-attachments/files/31058928/kamailio-sbom.cdx.json)

henningw:
Thanks for the extensions and adaptions. About the cmake specific topics probably @xkaraman should also have a quick look.

xkaraman:
CMake related code looks fine!

henningw:
Thanks, merged

henningw:
For eventual future PRs please configure your environment to remove this Co-author tags pointing to an LLM, I missed it during the review.

PR #4872, opened 2026-08-14
rtpengine: new modparam stats_reply_pv

covalschi:
#### Pre-Submission Checklist
  from source code folder

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description

Currently, when a command returning call statistics (`delete`, `query`,
`connect`) is sent to RTPEngine, the module only extracts a subset of the
reply data (the MOS values exposed via the `mos_*_pv` parameters) and
discards the rest of the statistics returned by RTPEngine.

This PR adds a new module parameter, `stats_reply_pv`, naming an AVP or
script variable where the complete command reply is stored, serialized as
JSON (reusing the JSON serializer already used by `rtpengine_query_v()`).
This gives config writers access to everything RTPEngine returns — full
per-SSRC stats, totals, tags, etc. — for CDR enrichment, logging or
accounting purposes.

The store is hooked into `parse_call_stats()`, so the variable is set after
`rtpengine_delete()`, `rtpengine_query()` or `rtpengine_connect()`; it is
also set by `rtpengine_manage()`, but only if the command resulted in a
deletion of the call (or call branch), e.g. when processing a BYE or
CANCEL. When the parameter is not set, the code path is a no-op (single
NULL pointer check), so there is no overhead for existing deployments.

Usage:

```
modparam("rtpengine", "stats_reply_pv", "$avp(RTPENGINE_STATS_REPLY)")
```

Tested locally:
- module builds cleanly with gcc (no warnings), changed code verified with
  clang-format 18
- invalid PV specs (non-AVP/VAR) are rejected at startup in `mod_init()`
- end-to-end test against a simulated RTPEngine ng-protocol endpoint:
  after `rtpengine_query()` and `rtpengine_delete()`, the full reply of the
  respective command is available as JSON in the configured AVP, while the
  existing MOS PV behavior is unchanged

### DIFF
--- a/src/modules/rtpengine/doc/rtpengine_admin.xml
+++ b/src/modules/rtpengine/doc/rtpengine_admin.xml
@@ -804,6 +804,42 @@ modparam("rtpengine", "media_duration", "$avp(MEDIA_DURATION)")
 	</section>
 
 
+	<section id="rtpengine.p.stats_reply_pv">
+		<title><varname>stats_reply_pv</varname> (string)</title>
+		<para>
+			The name of a pseudovariable (AVP or script variable) to store the
+			complete reply of the commands returning call statistics,
+			serialized as JSON. This gives access to all the data returned by
+			RTPEngine (e.g., the full call statistics), not only the values
+			extracted into the mos_*_pv parameters.
+		</para>
+		<para>
+			The variable is set after invoking <quote>rtpengine_delete</quote>,
+			<quote>rtpengine_query</quote> or <quote>rtpengine_connect</quote>.
+			It is also set by <quote>rtpengine_manage</quote>, but only if the
+			command resulted in a deletion of the call (or call branch), e.g.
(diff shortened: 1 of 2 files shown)


rfuchs:
OK for me.

A suggestion though. The `query` command returns the same stats. Might be worth to provide the same mechanism there, and not have it exclusive to `delete`?  In fact, could just integrate it with `parse_call_stats()`

covalschi:
Updated the PR: the parameter was generalized from `delete_reply_pv` to `stats_reply_pv`. The reply store is now hooked into `parse_call_stats()`, so besides `rtpengine_delete()` it also covers `rtpengine_query()` and `rtpengine_connect()`, which return the same statistics data. Title, commit and docs updated accordingly.

rfuchs:
Thank you

Untested but LGTM

henningw:
Thank you, merged

PR #4875, opened 2026-08-19
permissions: refcount table objects

rfuchs:
- similar approach to 1f1cbcb022 (dispatcher)
- all existing table objects are encapsulated in higher-level struct which includes a reference count (e.g. `struct addr_list *` is encapsulated in `struct addr_table`, etc)
- reference counts are protected by a mutex (one for `struct trusted_table` and a second one for the tables in `address.c`)
- access to all table objects is only allowed through a call to the respective `get_*()` function, which returns a pointer to the table and increases the reference count
- every call to `get_*()` must be paired with a matching call to `put_*()` to release the reference
- the only write operation to all tables is a complete reload, which rebuilds brand new table objects and the swaps out the global pointers
- after swapping out the pointers, the reload process waits until all reference counts to the old table objects drops to zero, and only then frees the old tables
- with all this, concurrent read access is possible and safe while allowing for concurrent safe reloads
- the standalone `empty_*()` functions have become obsolete and have been integrated into the `free_*()` functions
- functions in address.c generally need access to all 3 tables, so we use the existing `address_tables_group_t` together with matching get/put functions to facilitate this
- the subnet table previously used an extra element to keep track of the number of subnets - this count was integrated into the higher-level struct, therefore obsoleting the need for the extra element
- the subnet table has a runtime-configurable (but not runtime-mutable) size: we use the empty-array syntax (`[0]`) in the struct together with an adjusted malloc size for that
- all `get_*()` and `put_*()` functions follow the same schema, so a generic macro is used for the body of these functions
- the cleanup function scheduled at regular intervals is not needed any more, therefore `trusted_cleanup_interval` has been removed
- the `reload_delta` option is kept, but now defaults to zero (no limit) as all concurrent reloads and read accesses are safe
- the `perm_rpc_reload_time` value is protected by an additional mutex, which is needed to eliminate a small race condition between reading and writing the value

#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [X] Tested changes locally
- [X] Related to issue #4809

#### Description

Eliminates race conditions between concurrent reloads and read accesses.

The previous double-buffering approach is now obsolete and has been replaced with proper reference-counting

### DIFF
--- a/src/modules/permissions/address.c
+++ b/src/modules/permissions/address.c
@@ -30,6 +30,7 @@
 
 #include "permissions.h"
 #include "hash.h"
+#include "address.h"
 #include "../../core/config.h"
 #include "../../lib/srdb1/db.h"
 #include "../../lib/srdb1/db_ut.h"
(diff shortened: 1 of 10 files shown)


miconda:
If no other comments, this will be merged soon.

PR #4882, opened 2026-08-24
seas: fix shm buffer overflow during header serialization

AlexisHadj:
seas: fix shm buffer overflow during header serialization encode_msg()

Add length validation inside the header processing loop in encode_msg() to prevent shared memory buffer overflow.

Previously, encode_msg() wrote into a fixed 3200-byte shared memory buffer without validating the remaining available size during each loop iteration. The bounds check occurred only after the entire header processing loop had executed, allowing a SIP request containing a large number of headers to write past the allocated shm buffer.

This addresses an incomplete fix condition related to CVE-2016-2385


henningw:
This was reported multiple times, the patch came from @tunelko

PR #3546, opened 2023-08-21
smsops: Fix conversion from UCS-2 to UTF-8 and viceversa

herlesupreeth:
#### Pre-Submission Checklist

- [ ] Commit message has the format required by CONTRIBUTING guide
- [ ] Commits are split per component (core, individual modules, libs, utils, ...)
- [ ] Each component has a single commit (if not, squash them into one commit)
- [ ] No commits to README files for modules (changes must be done to docbook files
in `doc/` subfolder, the README file is autogenerated)

#### Type Of Change
- [ ] Small bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds new functionality)
- [ ] Breaking change (fix or feature that would change existing functionality)

#### Checklist:

- [ ] PR should be backported to stable branches
- [ ] Tested changes locally
- [ ] Related to issue #XXXX (replace XXXX with an open issue number)

#### Description

This PR fixes the conversion from UCS2 to UTF-8 and viceversa. Previously the code was not handling emojis and certain characters.

PR #4883, opened 2026-08-24
smsops: fix GSM 7-bit alphabet conversion tables

0xabe-io:
The two conversion tables were not inverses of each other, and neither matched the default alphabet and extension table in 3GPP TS 23.038. Twenty-one code points were wrong.

Transcription errors:

- gsm7bit_codes mapped 0x2D to 0xAD (SOFT HYPHEN) rather than 0x2D (HYPHEN-MINUS), so a hyphen was corrupted on decode
- gsm7bit_ext_codes carried VERTICAL LINE at 0x48, which is not an extension code, leaving 0x40 unmapped even though ascii2gsm7bit_ext_codes escapes '|' to it - decoding one emitted a bare 0xFF
- ascii2gsm7bit_codes mapped CURRENCY SIGN to 0x23 (NUMBER SIGN) rather than 0x24

Decode entries that contradicted the encode table:

- gsm7bit_codes returned an unaccented ASCII letter for eighteen code points - 0x04 as 'e', 0x5B as 'A', 0x7F as 'a' and so on - while ascii2gsm7bit_codes maps the ISO-8859-1 byte for that same character onto that same code point. The folding was not applied consistently either: 0x01, 0x03, 0x06, 0x07, 0x1C, 0x1D, 0x1E, 0x24, 0x40, 0x5F and 0x60 already returned their ISO-8859-1 bytes.

decode(encode(x)) == x now holds for every character both tables cover.

#### Pre-Submission Checklist
  from source code folder

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description

`smsops` has two GSM 7-bit conversion tables, one per direction. They are not
inverses of each other, and neither matches the default alphabet and extension
table in 3GPP TS 23.038. Twenty-one code points are wrong.

##### Three transcription errors

| table | index | was | is | |
| --- | --- | --- | --- | --- |
| `gsm7bit_codes` | `0x2D` | `0xAD` | `0x2D` | `0x2D` is HYPHEN-MINUS; `0xAD` is SOFT HYPHEN |
| `gsm7bit_ext_codes` | `0x40` / `0x48` | unmapped / `0x7C` | `0x7C` / unmapped | VERTICAL LINE is extension code `0x40`; `0x48` is not an extension code |
| `ascii2gsm7bit_codes` | `0xA4` | `0x23` | `0x24` | `0x24` is CURRENCY SIGN; `0x23` is NUMBER SIGN |

The first two lose messages. `gsm_to_ascii()` writes the table value straight
into the output with no validation, so a decoded hyphen becomes a bare `0xAD`
and a decoded `|` a bare `0xFF`. Neither is valid UTF-8, so a configuration
storing `$rpdata` / `$tpdu(payload)` in a UTF-8 column drops the message:

```
db_mysql_submit_query(): driver error on query: Incorrect string value:
  '\xADsim 3...' for column `smsc`.`messages`.`text` at row 1 (1366)
```

By then the sender has had `202 Accepted` and an RP-ACK built from the same
successfully-parsed RP-DATA, so from the handset the SMS was delivered.
Nothing reports an error anywhere. The hyphen is the only affected character
in printable ASCII, which is presumably how it survived since f82c4775ef, and
only senders using `application/vnd.3gpp.sms` reach the decode table at all.

Both misplaced entries look like the same kind of slip: in the 8-wide layout
the tables originally had, `0x7C` sits at the start of the row *after* the one
it belongs to, and the hyphen sits in an otherwise pure identity run
(`( ) * + , - . /`).

##### Eighteen decode entries that contradict the encode table

`gsm7bit_codes` returns an unaccented ASCII letter for `0x04` `è`, `0x5B` `Ä`,
`0x7F` `à` and fifteen others, while `ascii2gsm7bit_codes` maps the ISO-8859-1
byte for that same character onto that same code point — so a character that
survives encoding is altered by decoding.

This does not look deliberate: the folding is not applied consistently, and
`0x01` `£`, `0x03` `¥`, `0x06` `ù`, `0x07` `ì`, `0x1C` `Æ`, `0x1D` `æ`,
`0x1E` `ß`, `0x24` `¤`, `0x40` `¡`, `0x5F` `§` and `0x60` `¿` already return
their ISO-8859-1 bytes.

**This group is separable** — happy to drop it and land only the three above if
you would rather take it on its own. One thing to weigh: it makes eighteen more
characters emit ISO-8859-1, so a deployment storing decoded text in a UTF-8
column will start seeing the error above for `é`, `ä`, `à`. That collision is
pre-existing and not introduced here — `ù` produces it today, unmodified:

```
Incorrect string value: '\xF9' for column `smsc`.`messages`.`text`
```

Whether `smsops` should emit UTF-8 rather than ISO-8859-1 on this path is a
larger question and is not addressed here. `ucs2_to_utf8()` on the DCS≠0 path
already emits real UTF-8, so `$tpdu(payload)` is two different encodings
depending on the coding scheme; given #3043, #3546 and #3690 that looks like
the intended direction, but a `unsigned char[128]` table cannot express it and
the encode side takes ISO-8859-1 input. Glad to follow up separately.

##### Testing

`make modules modules=src/modules/smsops` builds clean, and `clang-format` 18
(the version CI pins) reports the file unchanged.

Verified end to end against a live Kamailio IMS deployment (P/I/S-CSCF plus an
SMSC using `smsops`), driving each direction separately so a fault in one
cannot mask the other:

* **decode** — a UE sends `application/vnd.3gpp.sms` carrying each 7-bit code
  point as a raw septet, bypassing the alphabet entirely, and the byte the
  SMSC stored is read back.
* **encode** — a known byte is placed in the SMSC's queue and the 7-bit value
  transmitted is read out of the delivered TPDU.

All 128 default-alphabet code points and all 10 extension code points, both
directions. Before: 20 code points decode to the wrong byte and 1 byte encodes
to the wrong code point. After: none. Nothing else about SMS handling changes —
`text/plain` delivery, RP-DATA in both directions, TP-OA TON/NPI, off-net
routing and store-and-forward all behave identically.


henningw:
Thanks for the PR. The "Three transcription errors" looks indeed like a straightforward fix. Regarding the "Eighteen decode entries that contradict the encode table", lets wait a bit for other feedback e.g. from @alexyosifov @herlesupreeth. I believe we should do it consistent, and not half of the special accent characters one way and the remaining the other way.

herlesupreeth:
I am of the opinion to leave out "Eighteen decode entries that contradict the encode table from this MR". @0xabe-io thanks for the fix to transcription errors

henningw:
> I am of the opinion to leave out "Eighteen decode entries that contradict the encode table from this MR". @0xabe-io thanks for the fix to transcription errors

Thanks for the feedback. Do you have some background on why this is mapped differently for different characters? Some characters seems to return the ASCII, and some of them the ISO-8859-1 value. This seems a bit strange.

herlesupreeth:
> Do you have some background on why this is mapped differently for different characters?

Sadly no idea why it was done that way. I could trace the changes to that array back to this commit - https://github.com/kamailio/kamailio/commit/f82c4775efca0580401f4b1a48dcb4d2cf26814f

henningw:
> > Do you have some background on why this is mapped differently for different characters?
> 
> Sadly no idea why it was done that way. I could trace the changes to that array back to this commit - [f82c477](https://github.com/kamailio/kamailio/commit/f82c4775efca0580401f4b1a48dcb4d2cf26814f)

Thanks. This was done originally from @alexyosifov - lets wait some days if he is able to provide some feedback.

miconda:
I haven't seen @alexyosifov active for quite some time, so whatever people interested in this PR decide, it should be fine without waiting.

PR #4884, opened 2026-08-25
seas: sanitize address length in process_bind_action()

AlexisHadj:
Cast my_addr.len to unsigned char and add length checks to prevent stack buffer overflow
during BIND action processing in event_dispatcher.c.

Co-authored-by: Pedro J. Nunez-Cacho Fuentes <[tunelko@gmail.com](mailto:tunelko@gmail.com)>


tunelko:
This issue was also reported in Kamailio’s security backlog, and the fix applied in the commit matches the one I proposed.

Issue #3254, opened 2022-09-28
No CDR is written on failover


Earn330:
Hi,

I think there is an issue with writing CDRs on failover scenarios.
I have two servers with Kamailio. Both server running Debian 11 (bulleye) and latest Kamailio 5.6.1 from git.
I'm using DMQ to sync dialogs and htable between these servers, so both servers have the same knowledge of dialog state. Both Kamailio uses nobind option, so I can switch a VIP from one server to the other one. This is managed with keepalived. Switching the VIP from one server the other one works fine an while a call is running I can switch the VIP. I can see that the BYE Message is handled OK after I made a switch. I can see that acc is triggered but what is missing is acc_cdr in this case. 
I'm using htable to fill all necessary variables to complete the CDR an I can see that all values are synced correctly.

The acc_cdr is created fine if there is no failover so i think it can't be an configuration issue. 
Both server holds the same Kamailio configuration except the IP.

`version: kamailio 5.6.1 (x86_64/linux) bfc5c2-dirty
flags: USE_TCP, USE_TLS, USE_SCTP, TLS_HOOKS, USE_RAW_SOCKS, DISABLE_NAGLE, USE_MCAST, DNS_IP_HACK, SHM_MMAP, PKG_MALLOC, Q_MALLOC, F_MALLOC, TLSF_MALLOC, DBG_SR_MEMORY, USE_FUTEX, FAST_LOCK-ADAPTIVE_WAIT, USE_DNS_CACHE, USE_DNS_FAILOVER, USE_NAPTR, USE_DST_BLOCKLIST, HAVE_RESOLV_RES, TLS_PTHREAD_MUTEX_SHARED
ADAPTIVE_WAIT_LOOPS 1024, MAX_RECV_BUFFER_SIZE 262144, MAX_URI_SIZE 1024, BUF_SIZE 65535, DEFAULT PKG_SIZE 8MB
poll method support: poll, epoll_lt, epoll_et, sigio_rt, select.
id: bfc5c2 -dirty
compiled with gcc 10.2.1`

`root@voip-lab-proxy01:~# lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye`

`root@voip-lab-proxy01:~# lscpu 
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Byte Order:                      Little Endian
[...]

PR #4885, opened 2026-08-25
dialog: add DMQ load callbacks option

LookedPath:
#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally
- [x] Related to issue #3254

## Objective

Address #3254, where active calls replicated through DMQ may not produce a CDR after failover because the new active node has no accounting callbacks attached to the replicated dialogs.

This change adds the optional `enable_dmq_load_callbacks` dialog parameter. When enabled, dialogs created from DMQ replication run `DLGCB_LOADED` callbacks and synchronize dialog variables in state updates, allowing modules such as `acc` to restore the callbacks and state required for CDR generation.

The parameter defaults to `0`, preserving the dialog module’s existing behavior unless explicitly enabled. A startup warning and documentation describe the known impact on modules that register `DLGCB_LOADED`.

### DIFF
--- a/src/modules/dialog/dialog.c
+++ b/src/modules/dialog/dialog.c
@@ -94,6 +94,7 @@ MODULE_VERSION
 static int mod_init(void);
 static int child_init(int rank);
 static void mod_destroy(void);
+static void warn_dmq_load_callback_modules(void);
 
 /* module parameter */
 static int dlg_hash_size = 4096;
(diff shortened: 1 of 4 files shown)


miconda:
Thanks, merging it!

But I am considering to remove the `enable_` prefix in the parameter name, probably same for the other one added in this development cycle. Shorter names are better and the value should indicate if the action is enabled or not.

miconda:
For the records: new parameter renamed to remove the enable_ prefix.

LookedPath:
Perfect, thank you for merging the feature

PR #4890, opened 2026-08-27
sipcapture: add bounds check for HEPv2 time header offset

AlexisHadj:
This PR fixes a potential crash in the sipcapture module when processing incoming HEPv2 packets with malformed or truncated payloads. 
**The Issue:** In the HEPv2 receive path (hep.c, under the hp_v == 2 branch), the code incremented hep_offset += sizeof(struct hep_timehdr) (12 bytes) without re-checking if the resulting offset exceeded the actual packet length
**The Fix.** Added checks to ensure that the calculated hep_offset does not exceed the packet length before proceeding


henningw:
Thanks, merged

PR #4895, opened 2026-08-28
userblocklist: Add robust prefix length validation and fault handling

AlexisHadj:
This PR fixes a vulnerability that could allow an attacker with access to a  BinRPC listener to overwrite the stack and achieve remote code execution
**The fix:**
**Prefix Validation:** Added check to reject overlong prefixes (> MAXNUMBERLEN) 

**Fault Handling:** Introduced the error_prefix handler to return a  400 error ("Check failed: prefix exceeds %d characters")

**Memory Operations:** Replaced strncpy with  memcpy  for better handling of counted strings.

The above changes apply  in both check_list_rpc() and check_userlist_rpc()

Co-authored-by: Pedro J. Nunez-Cacho Fuentes <[tunelko@gmail.com](mailto:tunelko@gmail.com)>

### DIFF
--- a/src/modules/userblocklist/userblocklist.c
+++ b/src/modules/userblocklist/userblocklist.c
@@ -1369,12 +1369,14 @@ static void check_list_rpc(rpc_t *rpc, void *ctx, int list_type)
 	/* Sanity checks */
 	if(rpc->scan(ctx, ".S", &prefix) < 1)
 		goto error_scan;
-	if(prefix.s == NULL || prefix.len == 0)
+	if(prefix.s == NULL || prefix.len <= 0)
 		goto error_scan;
+	if(prefix.len > MAXNUMBERLEN)
+		goto error_prefix;
 	if(rpc->add(ctx, "{", &out) < 0)
 		goto error;
 
-	strncpy(req_prefix, prefix.s, prefix.len);
+	memcpy(req_prefix, prefix.s, (size_t)prefix.len);
 	req_prefix[prefix.len] = '\0';
 
 	/* Check that global blocklist exists */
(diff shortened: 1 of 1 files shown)


tunelko:
Thanks for implementing the proposed fix. 

One detail from the original recommendation appears to have been omitted.

  The proposed validation used:

```c
  if(prefix.s == NULL || prefix.len <= 0)
        goto error_scan;
```
  The PR currently retains:
```c
  if(prefix.s == NULL || prefix.len == 0)
        goto error_scan;
```
The <= 0 check is intentional. If I'm not wrong: BinRPC converts an encoded zero length to str.len = -1. This bypasses both current checks and becomes a very large size_t in memcpy().

A more general solution would be to reject encoded strings with len < 1 in the BinRPC parser, but prefix.len <= 0 remains appropriate defense-in-depth within userblocklist. 

Thanks!

AlexisHadj:
> Thanks for implementing the proposed fix.
> 
> One detail from the original recommendation appears to have been omitted.
> 
> The proposed validation used:
> 
> ```c
>   if(prefix.s == NULL || prefix.len <= 0)
>         goto error_scan;
> ```
> 
> The PR currently retains:
> 
> ```c
>   if(prefix.s == NULL || prefix.len == 0)
>         goto error_scan;
> ```
> 
> The <= 0 check is intentional. If I'm not wrong: BinRPC converts an encoded zero length to str.len = -1. This bypasses both current checks and becomes a very large size_t in memcpy().
> 
> A more general solution would be to reject encoded strings with len < 1 in the BinRPC parser, but prefix.len <= 0 remains appropriate defense-in-depth within userblocklist.
> 
> Thanks!


You are absolutely right. Missed that part. Now it is ammended.
Thank you

henningw:
For consistency we probably also should check the user.len for <= :
if(prefix.s == NULL || prefix.len <= 0 || user.s == NULL || user.len == 0)

PR #4896, opened 2026-08-28
topos_htable: escape the field separator in the stored records

MarkLee131:
#### Pre-Submission Checklist
  from source code folder

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description

The `topos_htable` records are built by joining the field values with `|`, and are parsed back by position with `strsep()`. A `|` inside a field value taken from a header therefore shifts every field that follows it.

A `|` in the Call-ID is enough. The stored dialog record becomes:

```
1787806973|probe|PIPE|callid|atpsh-...|btpsh-...|0|SIP/2.0/UDP 127.0.0.1:5080;...
```

The Call-ID now occupies three fields. On the reply, `topos` loads a record where `a_uuid` is something else, does not restore the Via set, and `tps_prepare_msg()` fails for the outgoing message:

```
ERROR: topos [topos_mod.c:528]: tps_prepare_msg(): mandatory headers missing - via1: (nil) callid: 0x...
```

`tps_msg_sent()` then reaches `done:` without touching `obuf`, so the core sends the reply as it arrived, with topology hiding off. The caller gets the callee's real address instead of a masked one.

Only the htable backend is affected. The db backend keeps each field in its own column and reads it back through `RES_ROWS(...)[0].values[_c]`, so there is no separator to inject there.

What the patch does:

- escapes the separator, and the escape character itself, when building the dialog, branch and initial method branch records
- parses the records back with a scanner that skips the escaped separators and unescapes each field in place, instead of `strsep()`
- applies the same escaping to the key of the initial method branch record. A `|` in the Call-ID or in the To-tag could otherwise build the key of a different dialog: `callid=a|b, tag=c` and `callid=a, tag=b|c` produced the same key.

Field values that contain neither `|` nor `\` are stored unchanged, so the record format stays byte-identical to what the current code writes. The bounds of the existing buffers are still enforced, and a value that no longer fits makes the insert fail rather than being truncated.

`topos_htable_base64=1` was already unaffected, since base64url values contain neither the separator nor the escape character. The new escaping is a no-op there.

Testing: three dialogs back to back through one proxy, same UAS, one kamailio instance, everything default except `storage=htable` and `contact_host`. The only variable is the Call-ID.

before:
```
clean Call-ID          -> MASKED           Contact seen by UAC: <sip:atpsh-6a9142b2-3ba5a1-1@127.0.0.1>
Call-ID with '|'       -> *** LEAKED ***   Contact seen by UAC: <sip:bob@127.0.0.1:5070>
clean Call-ID (2)      -> MASKED           Contact seen by UAC: <sip:atpsh-6a9142b2-3ba5a1-3@127.0.0.1>
```

after:
```
clean Call-ID          -> MASKED           Contact seen by UAC: <sip:atpsh-6a9142c3-3ba789-1@127.0.0.1>
Call-ID with '|'       -> MASKED           Contact seen by UAC: <sip:atpsh-6a9142c3-3ba789-2@127.0.0.1>
clean Call-ID (2)      -> MASKED           Contact seen by UAC: <sip:atpsh-6a9142c3-3ba789-3@127.0.0.1>
```

I also ran a full INVITE / 200 / ACK / BYE / 200 flow for ten Call-IDs: clean, a single `|`, 23 of them, a backslash, `a\|b|c`, leading, trailing and separator-only values, and a trailing backslash. The checks were that the caller never sees the real UAS address, and that the Call-ID is restored on both legs for the in-dialog requests and for the BYE reply. All ten pass with the patch, and again with `topos_htable_base64=1`. Six of the ten leak without it.

The record builder and the scanner were also exercised standalone under ASan and UBSan, for round-trip equality, empty fields, separator-only fields, key distinctness and oversized values.

### DIFF
--- a/src/modules/topos_htable/topos_htable_storage.c
+++ b/src/modules/topos_htable/topos_htable_storage.c
@@ -46,6 +46,289 @@ static char _tps_htable_key_buf[TPS_HTABLE_SIZE_KEY];
 static char _tps_htable_val_buf[TPS_HTABLE_SIZE_VAL];
 static char _tps_base64_buf[TPS_BASE64_ROWS][TPS_BASE64_SIZE];
 
+static str _tps_htable_xpfx = str_init("x");
+
+/***		HTABLE RECORD FORMAT FUNCTIONS		***/
+
+/**
+ * The htable records are built by joining the field values with a separator,
+ * they are parsed back by position, therefore a field value containing the
+ * separator (e.g., taken from Call-ID or another header) has to be escaped,
+ * otherwise it shifts all the following fields of the record.
+ * - the escape character escapes the separator as well as itself
+ * - values without separator or escape character are stored unchanged, keeping
+ *   the format compatible with the records built by previous versions
+ */
+#define TPS_HTABLE_FSEP '|'
+#define TPS_HTABLE_FESC '\\'
+
+typedef struct tps_htable_rec
+{
+	char *buf;
+	int size;
+	int len;
+} tps_htable_rec_t;
+
+#define TPS_REC_ADD(_rec, _v)                    \
+	do {                                         \
+		if(tps_htable_rec_add((_rec), (_v)) < 0) \
+			return -1;                           \
+	} while(0)
+
+#define TPS_REC_CAT(_rec, _v)                    \
+	do {                                         \
+		if(tps_htable_rec_cat((_rec), (_v)) < 0) \
+			return -1;                           \
+	} while(0)
+
+#define TPS_REC_ADDI(_rec, _v)                    \
+	do {                                          \
+		if(tps_htable_rec_addi((_rec), (_v)) < 0) \
+			return -1;                            \
+	} while(0)
+
+/**
+ * Initialize a record builder over the buffer buf of size bytes
+ */
+static void tps_htable_rec_init(tps_htable_rec_t *rec, char *buf, int size)
+{
+	rec->buf = buf;
+	rec->size = size;
+	rec->len = 0;
+	rec->buf[0] = '\0';
+}
+
+/**
+ * Append the escaped value of sv, without a field separator
+ * - returns 0 on success, -1 if the value does not fit in the buffer
+ */
+static int tps_htable_rec_cat(tps_htable_rec_t *rec, str *sv)
+{
+	int i;
+
+	if(sv == NULL || sv->s == NULL || sv->len <= 0) {
+		return 0;
+	}
+
+	for(i = 0; i < sv->len; i++) {
+		if(sv->s[i] == TPS_HTABLE_FSEP || sv->s[i] == TPS_HTABLE_FESC) {
+			if(rec->len + 2 >= rec->size) {
+				return -1;
+			}
+			rec->buf[rec->len++] = TPS_HTABLE_FESC;
+		} else {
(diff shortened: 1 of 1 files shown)


stefan-mititelu-idt:
Thanks for this PR!

There is indeed this old, known issue with the plaintext topos_htable storage, mitigated by enabling the base64 modparam (which is how we are using it too in production).

At the first glance, this generally looks good. We will be checking this more, internally, but with lower priority.

However, since this also does a bit of rework of the storing/loading of htable string keys and values, maybe can be considered as new feature?

Also I'd like to ask a question partially related to this PR:
Do you maybe have any future plans of adding some topos RPCs that will allow display and debugging of a particular topos stored entry? That will increase motivation of using this PR fix with plaintext topos_htable storage, for production environments too.

Thanks,
Stefan

MarkLee131:
On the type of change: the patch adds no new module parameter and no new behavior. Field values that contain neither `|` nor `\` are stored exactly as before, byte-identical to what the current code writes, so an instance keeps reading and writing the same records after the upgrade. What changes is only that a value carrying the separator no longer shifts the fields that follow it.

Most of the diff is mechanical: the three `snprintf()` format strings became three builder functions with the same field order, and `strsep()` became a scanner that skips the escaped separators. The base64 branch is left untouched, since base64url values can contain neither the separator nor the escape character.

If it helps the review, I can split it into two commits: one adding the record builder and scanner helpers, one switching the three insert and three load sites over to them.

On the RPC question: neither `topos` nor `topos_htable` registers any RPC command at the moment, so the only way to look at a stored entry today is `htable.get` / `htable.dump` on the topos tables, which returns the raw record: a single `|`-joined line with plaintext storage, and an opaque base64 value with the modparam enabled. I have no plans for a topos-level RPC at the moment.

henningw:
Thanks as well. As there is an easy workaround for the underlaying issue (enabling the base64 mode), it probably not necessary to integrate this to the stable branches. Additional feedback is of course welcome.

henningw:
Thanks, please check the code format. You also should squash the commits before the force-push, to remove the not needed merge commit.

MarkLee131:
Done. Format checked with clang-format 18 (the version used by the format workflow) against master: no changes reported. The two failing checks were caused by the merge commit, which made the diff base include unrelated commits from master.

The workflows are waiting for approval to run.

miconda:
I guess this can be merged if no other comments.

On the other hand, for the future, maybe only the encoded should be kept for code simplicity and easier long term maintenance.

henningw:
> I guess this can be merged if no other comments.
> 
> On the other hand, for the future, maybe only the encoded should be kept for code simplicity and easier long term maintenance.

Yes, we probably should just remove the un-encoded code path in the development version and not merge this PR. For the stable branches there is an easy work-around by just setting the encoding parameter.

henningw:
I have pushed a comment removing the topos_htable_base64 mode to git master. Further refinements can be done directly there. The PR will be closed, thanks again.

stefan-mititelu-idt:
Hi, in my opinion there is 1 good argument to keep the plaintext because is useful when working with topos+topos_htable storage in dev/lab environments. Allows to faster get the topos_htable storage value from the SIP uuid (e.g. bypasses the base64 encoding for the key, decoding for the value when trying to get the htable cell).

In my opinion, this PR can be merged.

PR #4899, opened 2026-08-31
ldap: improve parameter parsing in ldap_params_search

AlexisHadj:
This PR improves the parameter parsing in ldap_params_search and makes it more safe against wrong data.

### DIFF
--- a/src/modules/ldap/ldap_api_fn.c
+++ b/src/modules/ldap/ldap_api_fn.c
@@ -218,7 +218,7 @@ int ldap_url_search(char *_ldap_url, int *_ld_result_count)
 			ZSW(ludp->lud_filter));
 
 	rc = ldap_params_search(_ld_result_count, ludp->lud_host, ludp->lud_dn,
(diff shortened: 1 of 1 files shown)


henningw:
I am wondering if this is actually working.

This is the function signature:
`int ldap_params_search(int *_ld_result_count, char *_lds_name, char *_dn,
                int _scope, char **_attrs, char *_filter, ...)
{
`

Wouldn't it be then passing the %s as filter, and omitting the filter parameter?

henningw:
> I am wondering if this is actually working.
> 
> This is the function signature: `int ldap_params_search(int *_ld_result_count, char *_lds_name, char *_dn, int _scope, char **_attrs, char *_filter, ...) { `
> 
> Wouldn't it be then passing the %s as filter, and omitting the filter parameter?

I was wrong, this was the actual point of the fix. Nevertheless if somebody has a test setup related to ldap, some tests would be appreciated.

miconda:
Fine for me to merge.

PR #4900, opened 2026-08-31
WIP: tls: emit 5-tuple metadata in keylog callback

jchavanton:
> **This is a draft / WIP.** Verified end-to-end in a downstream deployment
> (kamailio TLS trunk to a carrier, voipmonitor consuming the keylog file
> via an external tool that INSERTs `ssl_sessions` rows). Opening now to
> invite early feedback on the approach and framing before I clean up.

## The gap

The NSS keylog format (`CLIENT_RANDOM`, `SERVER_HANDSHAKE_TRAFFIC_SECRET`,
etc.) carries only crypto material and no association to the TCP connection
that produced it. That's sufficient for consumers that decrypt from a
wire-captured handshake — Wireshark reads the file and matches keys to
sessions by client_random. It is **not** sufficient for consumers that need
to bind a key to a connection they did not witness (a passive sniffer
restarted mid-session, needing to decrypt records on a reused TLS session).

`SSL_get_fd()` and `BIO_get_fd()` both return `-1` in the keylog callback
because the tls module uses a memory BIO (`tls_BIO_new_mbuf`), so the fd
cannot be recovered from OpenSSL alone.

## The change

Cache the `tcp_connection` back-pointer in `tls_extra_data` at SSL creation
time, then emit a `# TUPLE` comment line before each key line:

    # TUPLE src=<ip>:<port> dst=<ip>:<port> sr=<hex> cs=<u16> v=<version>
    CLIENT_RANDOM <hex> <hex>

Fields:

| Field   | Meaning                                                                                       |
|---------|-----------------------------------------------------------------------------------------------|
| `src` / `dst` | TCP 5-tuple. `dst_port` falls back to `rcv.bind_address->port_no` for accepted sockets (where `rcv.dst_port` is 0 by convention); connected sockets use `getsockname(c->fd)` / `getsockname(c->s)` to recover the true wire ephemeral port. |
| `sr`    | `server_random` hex (from `SSL_get_server_random`).                                           |
| `cs`    | Cipher suite id (`SSL_CIPHER_get_id & 0xFFFF`, IANA-assigned).                                |
| `v`     | TLS version (`0x0303` = TLS 1.2, `0x0304` = TLS 1.3).                                         |

NSS format treats `#` lines as comments, so Wireshark and other standard
readers ignore the addition — the file remains a valid keylog file.

Uses `ip_addr2sbuf()` with caller-owned buffers because `ip_addr2a()`
returns a pointer into a single static buffer and would yield the same
string for both operands in one `snprintf`.

## Why direction-aware `dst_port` resolution

- **Accepted (server-role, `F_CONN_PASSIVE`):** `rcv.dst_port` is our listen
  port; if 0 fall back to `bind_address->port_no`.
- **Connected (client-role, kamailio dialed out):** `rcv.dst_port` is not
  the local wire port — it's the peer's port for this direction of
  accounting. The actual local ephemeral is only available via
  `getsockname` on whatever fd is valid in the current process (`c->fd` for
  worker children, `c->s` for tcp-main). Try both; take the first non-zero.

Downstream consumers matching on 5-tuple need the wire ephemeral, not
`rcv.dst_port`, or their lookup will miss.

## Verification

Verified downstream 2026-08-31 on a kamailio 6.2 build with `keylog_mode=15`
armed against Telnyx TLS trunks. Both TLS 1.2 (cipher 49196) and TLS 1.3
(cipher 4866) sessions produce well-formed TUPLE lines with real ephemeral
ports in `dst_port` (getsockname fallback works). File is consumed by an
external tool that parses `# TUPLE` groups and populates `ssl_sessions`
rows in voipmonitor's DB; voipmonitor then decrypts sessions whose
handshake it did not witness, producing B-leg CDRs that were previously
missing.

## Open questions for reviewers

- Is emitting one TUPLE line before **every** key line the right choice, or
  should it be deduped per session? Cheap and simple as-is (consumers can
  dedupe by 5-tuple); TLS 1.3 produces multiple secrets per session so
  each key line already gets its own TUPLE.
- The `tcp_connection` back-pointer field lives in `tls_extra_data` — is
  that the right home, or would a smaller dedicated struct be cleaner?
- Fresh `getsockname` call in the callback for outbound sockets — I don't
  expect it to be hot, but happy to move the resolution earlier (once at
  handshake completion) if that's a concern.

### DIFF
--- a/src/modules/tls/tls_domain.c
+++ b/src/modules/tls/tls_domain.c
@@ -25,6 +25,8 @@
  */
 
 #include <stdlib.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
 #include <openssl/ssl.h>
 #include <openssl/opensslv.h>
 #include <openssl/bn.h>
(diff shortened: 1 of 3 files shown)


henningw:
Not reviewed the code much, maybe @space88man can share his thoughts. From a user point of view I think we probably should add a new keylog_mode (e.g. flag 32) for adding this additional meta-data. This way people can stay with the old format without any risk of braking anything.

space88man:
@jchavanton - go ahead - no objections from me.

jchavanton:
Thanks, I need more time with validation.

sergey-safarov:
Could you look comment at
https://gitlab.com/wireshark/wireshark/-/work_items/20512#note_2471720051

> Messages are not simply encrypted with the same key in TLS. There is a nonce for AEAD ciphers, which is derived from the sequence number within the stream. (

As I understanding you need track packet number in the stream to decript TLS flow.
If passive capture application restarted then it may lost some packets and you not able to know current packet number in the flow.

To be able to decrypt application TLS/Websocket connections I wrote application you can try
https://gitlab.com/wisteriabg/pcap-sip
https://gitlab.com/wisteriabg/pcap-sip/-/releases

PR #4905, opened 2026-09-02
sipt: Add robust length validations in ISUP message parser

AlexisHadj:
Add robust length validations in ISUP message parser

### DIFF
--- a/src/modules/sipt/sipt.c
+++ b/src/modules/sipt/sipt.c
@@ -316,9 +316,9 @@ static int sipt_get_calling_party_nai(
 static int sipt_get_calling_party(
 		struct sip_msg *msg, pv_param_t *param, pv_value_t *res)
 {
-	static char sb_s_buf[26];
+	static char sb_s_buf[MAX_SB_BUF_LEN];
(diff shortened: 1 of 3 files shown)


henningw:
Looks good, but we probably should move the `#define MAX_SB_BUF_LEN 26` to the ss7.h and then also change the following code:

```
~/repositories/kamailio/src/modules/sipt$ ack 26 sipt.c
        static char sb_s_buf[26];
        memset(sb_s_buf, 0, 26);
        static char sb_s_buf[26];
        memset(sb_s_buf, 0, 26);
        static char sb_s_buf[26];
        memset(sb_s_buf, 0, 26);
        static char sb_s_buf[26];
        memset(sb_s_buf, 0, 26);
        static char sb_s_buf[26];
        memset(sb_s_buf, 0, 26);
        static char sb_s_buf[26];
        memset(sb_s_buf, 0, 26);

```

This way it would stay consistent when we change it later.

henningw:
Thanks, merged and also cherry-picked to stable branches.

PR #4442, opened 2025-10-20
ndb_redis: TLS/auth fixes and cluster robustness

syrcx:
### Summary
Fix several robustness issues in `ndb_redis` when connecting to Redis over TLS and handling cluster redirections:
- Do not overwrite `pass` when parsing `tls=1`
- Set command timeout before issuing any command (including AUTH)
- Bound copy for sentinel master address
- Prepare dynamic-node creation to include essential attributes (added `db`; follow-up to include `pass`, `tls`, timeouts)

### Motivation
In Redis Cluster deployments (e.g., AWS ElastiCache with TLS+AUTH), we observed intermittent reconnect failures following MOVED redirections:
- error communicating with redis server …: Resource temporarily unavailable
- redirection then fails, and the command aborts

Two root causes identified:
- `tls` parsing overwrote `pass`, corrupting auth for some connections
- In reconnect/init paths, timeouts were set after issuing AUTH/PING/SELECT
- Sentinel address copy used `len+1` and could over-copy
- Dynamic nodes created after MOVED did not inherit the original connection’s attributes (TLS/auth/db/timeouts). This PR lays groundwork by appending `db`; full inheritance will be proposed next.

Related packaging note for TLS support in Docker images: [kamailio/kamailio#4401](https://github.com/kamailio/kamailio/issues/4401)

### Changes
- TLS parsing
  - Stop writing `tls` parameter value into the `pass` buffer in both `redisc_init()` and `redisc_reconnect_server()`.
- Timeout ordering
  - Move `redisSetTimeout()` to occur immediately after context creation and before any command (`AUTH`/`PING`/`SELECT`) in both init and reconnect.
- Sentinel
  - Replace `strncpy(..., len+1)` with a safe bounded copy and explicit terminator when reading `get-master-addr-by-name`.
- Dynamic nodes (MOVED handling)
  - When constructing `spec_new`, start appending inherited attributes (added `db`). Follow-up planned to carry `pass`, `tls`, `connect_timeout`, and `command_timeout`.

### Impact
- Avoids intermittent AUTH/TLS breakage caused by `tls` clobbering `pass`.
- Ensures command timeout applies to all commands, including `AUTH`.
- Safer sentinel parsing.
- Reduces failures on MOVED by beginning to propagate connection attributes to dynamically added nodes.

### Backwards compatibility
- No behavior change for non-TLS use.
- For TLS+AUTH users, behavior becomes more reliable.
[...]

Issue #4433, opened 2025-10-15
ndb_redis: MOVED redirections create connections without TLS/auth; tls parsing overwrites pass; intermittent reconnect failures (“Resource temporarily unavailable”)


syrcx:
### Description
We observe intermittent Redis failures in `ndb_redis` when using Redis Cluster with TLS and password (AWS ElastiCache). Errors often appear right after a MOVED redirection and during reconnect:

- “Resource temporarily unavailable” while reconnecting to the redirected node
- Subsequent “MOVED … host:port”
- The command fails (e.g., GET)

From reading the module code, there seem to be two distinct issues:

- In cluster mode with `allow_dynamic_nodes=1`, when handling MOVED/ASK, `ndb_redis` adds a new server entry including only `name/addr/port`. It does not inherit `pass`, `tls`, `db`, or timeouts, so redirected connections are attempted without TLS and without AUTH.
- The `tls` parameter parsing overwrites the `pass` buffer, corrupting a previously parsed password when `tls=1` is present.

This explains the intermittent nature: only keys whose slots are redirected to “new” nodes (created dynamically without TLS/auth) fail.

### Environment
- Kamailio version: ghcr.io/sergey-safarov/kamailio-ci:master-alpine, which is from https://github.com/kamailio/kamailio/issues/4401#issuecomment-3352145843
- Redis: AWS ElastiCache Redis Cluster with TLS and AUTH

### Configuration (excerpt)
```cfg
loadmodule "ndb_redis.so"

modparam("ndb_redis", "init_without_redis", 1)
modparam("ndb_redis", "connect_timeout", 15000)
modparam("ndb_redis", "cmd_timeout", 5000)
modparam("ndb_redis", "cluster", 1)
modparam("ndb_redis", "allow_dynamic_nodes", 1)
modparam("ndb_redis", "ca_path", "/etc/ssl/certs")
modparam("ndb_redis", "server", "name=con1;addr=...;port=6379;pass=...;tls=1")
```

### Steps to Reproduce
1) Configure `ndb_redis` for Redis Cluster with TLS and password (as above), set `cluster=1` and `allow_dynamic_nodes=1`.
2) Execute commands across keys that map to different hash slots (to trigger MOVED).
3) Observe that some redirected connections report “Resource temporarily unavailable” and the command fails.

### Actual Behavior
Logs (representative):
```
[...]

PR #4908, opened 2026-09-02
ndb_redis: inherit pass/tls attrs for dynamic cluster nodes

### LINKED ISSUE #4907: ndb_redis: dynamically-created Redis Cluster nodes don't inherit pass/tls after a MOVED redirect
### Description

When `ndb_redis` follows a `-MOVED` redirect from a Redis Cluster and the redirect target is a node it has not connected to before, `check_cluster_reply()` (`src/modules/ndb_redis/redis_client.c`) builds the new server spec with only `name`, `addr`, `port` and `db`:

```c
server_len = snprintf(spec_new, sizeof(spec_new) - 1,
        "name=%.*s;addr=%.*s;port=%i;db=%d", name.len, name.s,
        addr.len, addr.s, port, 0);
```

`pass` and `tls` are never copied from the server that received the redirect. Against a Redis Cluster that requires AUTH and/or TLS (e.g. AWS ElastiCache with transit encryption and an auth token enabled), the dynamically-created connection is refused, and every subsequent command against it fails:

```
ERROR: ndb_redis [redis_client.c:748]: redisc_reconnect_server(): error communicating with redis server [...]: Resource temporarily unavailable
```

Once this happens, the broken connection is retried for the lifetime of the process, continuously failing.

### Context

GH #4433 reported the general MOVED-redirect handling gap, fixed by #4442 (merge commit 31138dd). That PR added `db=0` to the new spec explicitly "to prepare attribute inheritance" but left `pass`/`tls`/timeouts as a follow-up that was never completed, so any Redis Cluster requiring auth/TLS still breaks on redirect.

### Expected behavior

The dynamically-created node should inherit `pass`/`tls` from the server that received the `MOVED` reply.

### How to reproduce

- Configure `ndb_redis` against a Redis Cluster (`redis_cluster=1`) that requires `tls`/`pass`, e.g. ElastiCache with in-transit encryption and an AUTH token.
- Trigger a slot migration or failover so a command against an existing server gets a `-MOVED` reply pointing at a node not already configured/connected.
- Observe the new connection fail with "Resource temporarily unavailable" (or a TLS/auth-specific hiredis error) on every subsequent attempt.

### Version

Reproduced on 6.1.4. `check_cluster_reply()` is unchanged between 6.1.4 and current master, so it reproduces there too.

### Fix

I have a small patch that copies `pass`/`tls` from the original server's attrs and attaches them directly to the new server's `attrs` list (rather than interpolating them into `spec_new`'s `;`-delimited string, so a password containing `;` or `=` can't inject/override another attribute). Opening a PR shortly.

calvin7889:
#### Pre-Submission Checklist
- [ ] Code is formatted with `clang-format` using the config file `.clang-format`
  from source code folder (clang-format was not available in my environment; the
  patch was hand-matched to the surrounding function's existing style)

#### Type Of Change

#### Checklist:
- [ ] Tested changes locally
- [x] Related to issue #4907

#### Description

Closes #4907.

`check_cluster_reply()` builds the spec for a dynamically-created cluster
node (created when a `-MOVED` reply points at a node not already
connected) with only `name`/`addr`/`port`/`db`. `pass`/`tls` are never
carried over from the server that received the redirect, so against a
Redis Cluster that requires AUTH and/or TLS (e.g. AWS ElastiCache with
transit encryption and an auth token), the new connection is refused and
every subsequent command against it fails with "Resource temporarily
unavailable".

GH #4433 / PR #4442 added `db=0` to the new spec explicitly "to prepare
attribute inheritance" but left `pass`/`tls`/timeouts as an unfinished
follow-up (see #4907 for the full repro and context). This PR finishes
the `pass`/`tls` half of that: once the new server node exists (after
`redisc_add_server()` succeeds), the original server's `pass`/`tls`
attrs are copied and prepended directly onto the new server's `attrs`
linked list, before `redisc_reconnect_server()` is called.

This is done as a direct attach of `param_t` nodes rather than by
interpolating `pass`/`tls` into `spec_new`'s `;`-delimited string,
because `spec_new` gets re-parsed by `redisc_add_server()` via the
generic, unescaped `parse_params()`. Interpolating a raw password into
that string would let a password containing `;` or `=` inject or
override another attribute in the new server's config (e.g. a password
crafted to end in `;tls=0` could silently downgrade the redirect
connection to plaintext). Attaching the attrs directly after the node
already exists avoids that entirely.

**Correctness/safety notes** (traced against current module code):
- `do_free_params()` only ever frees the `param_t` node itself, never
  `name.s`/`body.s`, so using a string literal for `name.s` and pointing
  `body` at the original server's already-allocated string is safe.
- Server attrs are only freed at module shutdown (`redisc_destroy()`),
  never while servers are in active use, so there's no dangling-pointer
  risk between the old and new server sharing that string memory.
- `redisc_reconnect_server()`'s attrs consumer matches by `name` string
  (`strncmp`), not by `ptype_t`, so leaving the manually-built nodes'
  `type` at the zero value (`P_OTHER`) is fine.

**Testing:** verified via `git apply --check`/build that the patch
applies cleanly and compiles against current master. The patched module
has also been running in a production-adjacent staging deployment for
several hours against a TLS+AUTH ElastiCache Redis Cluster with no
regressions (previously it logged a stale broken cluster-node connection
continuously). That deployment has not yet naturally hit a fresh `MOVED`
redirect since the patch was deployed, so the exact new-connection code
path added here has been verified by code review/tracing (see notes
above) but not yet observed end-to-end against a live redirect. Leaving
"Tested changes locally" unchecked above to reflect that honestly.

### DIFF
--- a/src/modules/ndb_redis/redis_client.c
+++ b/src/modules/ndb_redis/redis_client.c
@@ -1057,6 +1057,8 @@ int check_cluster_reply(redisReply *reply, redisc_server_t **rsrv)
 	str addr = {0, 0}, tmpstr = {0, 0}, name = {0, 0};
 	int server_len = 0;
 	char spec_new[512];
+	param_t *pit = NULL, *pnew = NULL;
+	str orig_pass = {0, 0}, orig_tls = {0, 0};
 
 	if(redis_cluster_param) {
 		LM_DBG("Redis replied: \"%.*s\"\n", (int)reply->len, reply->str);
@@ -1102,7 +1104,26 @@ int check_cluster_reply(redisReply *reply, redisc_server_t **rsrv)
 				char *server_new;
 
 				memset(spec_new, 0, sizeof(spec_new));
-				/* For now, also include db=0 to prepare attribute inheritance */
+				/* For now, also include db=0 to prepare attribute inheritance.
+				 * pass/tls are inherited too (below, once the node exists)
+				 * so a dynamically-created node for a TLS+auth-required
+				 * Redis Cluster (e.g. ElastiCache with transit encryption
+				 * and AUTH enabled) does not silently connect in plaintext
+				 * with no credentials and get refused. They are attached
+				 * directly to the new server's attrs rather than
+				 * interpolated into spec_new, so a password containing ';'
+				 * or '=' cannot inject or override another attribute here -
+				 * e.g. a password ending in ";tls=0" must not be able to
+				 * downgrade the redirect connection to plaintext. */
+				for(pit = (*rsrv)->attrs; pit; pit = pit->next) {
+					if(pit->name.len == 4
+							&& strncmp(pit->name.s, "pass", 4) == 0) {
+						orig_pass = pit->body;
+					} else if(pit->name.len == 3
+							  && strncmp(pit->name.s, "tls", 3) == 0) {
+						orig_tls = pit->body;
+					}
+				}
 				server_len = snprintf(spec_new, sizeof(spec_new) - 1,
 						"name=%.*s;addr=%.*s;port=%i;db=%d", name.len, name.s,
 						addr.len, addr.s, port, 0);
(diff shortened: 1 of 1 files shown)


henningw:
Thanks for the PR, it will be reviewed. ~~As a note, I would like to merge https://github.com/kamailio/kamailio/pull/4912 before, as this was discussed now for several weeks. Maybe there is a rebase needed, lets see.~~

PR #4912, opened 2026-09-04
db_redis: reshuffle sentinel code to cleanly build redis cluster/ sentinel

lbalaceanu:
#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

This better decouples the code that belongs to the redis cluster vs redis with sentinels (either export WITH_SENTINELS=1 or export WITH_HIREDIS_CLUSTER=1). This ensures that such builds work without problems.

I am not savvy in the Makefile and CMake part, I relied on AI.

Whoever is interested is welcome to test the cluster scenarios as I only have a sentinel setup.

Thank you

### DIFF
--- a/src/modules/db_redis/CMakeLists.txt
+++ b/src/modules/db_redis/CMakeLists.txt
@@ -1,5 +1,16 @@
 file(GLOB MODULE_SOURCES "*.c")
 
+option(WITH_SENTINELS "Build db_redis with Redis Sentinel support" OFF)
+option(WITH_HIREDIS_CLUSTER "Build db_redis with hiredis cluster support" OFF)
+
+if(WITH_SENTINELS AND WITH_HIREDIS_CLUSTER)
+  message(FATAL_ERROR "db_redis: WITH_SENTINELS and WITH_HIREDIS_CLUSTER are mutually exclusive")
+endif()
+
+if(NOT WITH_SENTINELS)
+  list(REMOVE_ITEM MODULE_SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/redis_sentinels.c")
+endif()
+
 add_library(${module_name} SHARED ${MODULE_SOURCES})
 
 find_package(PkgConfig REQUIRED)
@@ -7,6 +18,17 @@ pkg_check_modules(hiredis REQUIRED IMPORTED_TARGET hiredis)
 add_library(hiredis::hiredis ALIAS PkgConfig::hiredis)
 
 target_link_libraries(${module_name} PRIVATE hiredis::hiredis)
+
+if(WITH_SENTINELS)
+  target_compile_definitions(${module_name} PRIVATE WITH_SENTINELS)
+endif()
+
+if(WITH_HIREDIS_CLUSTER)
+  pkg_check_modules(hiredis_cluster REQUIRED IMPORTED_TARGET hiredis_cluster)
+  add_library(hiredis::hiredis_cluster ALIAS PkgConfig::hiredis_cluster)
+  target_compile_definitions(${module_name} PRIVATE WITH_HIREDIS_CLUSTER)
+  target_link_libraries(${module_name} PRIVATE hiredis::hiredis_cluster)
+endif()
 
 # Check for hiredis SSL support
 pkg_check_modules(hiredis_ssl IMPORTED_TARGET hiredis_ssl)
--- a/src/modules/db_redis/Makefile
+++ b/src/modules/db_redis/Makefile
@@ -60,17 +60,33 @@ endif
 DEFS+=$(HIREDISDEFS)
 LIBS=$(HIREDISLIBS)
 
+# Mode switches:
+# - WITH_REDIS_CLUSTER=1 (default): enable cluster mode if hiredis_cluster exists
+# - WITH_REDIS_CLUSTER=0: force normal (non-cluster) mode
+# - WITH_SENTINELS=1: enable sentinel mode and disable cluster mode
+WITH_REDIS_CLUSTER ?= 1
+
+ifeq ($(WITH_SENTINELS),1)
+DEFS+=-DWITH_SENTINELS
+endif
+
 ifneq ($(HIREDIS_CLUSTER_BUILDER),)
-	HIREDISCLUSTERDEFS = $(shell $(HIREDIS_CLUSTER_BUILDER) --cflags)
-	HIREDISCLUSTERLIBS = $(shell $(HIREDIS_CLUSTER_BUILDER) --libs)
-	HIREDISCLUSTERLIBSPATH = $(shell $(HIREDIS_CLUSTER_BUILDER) --libs-only-L | cut -c 3-)
-	ifneq ($(shell ls $(HIREDISCLUSTERLIBSPATH) | grep libhiredis_cluster_ssl.so),)
-		HIREDISCLUSTERDEFS += -DWITH_SSL
-		HIREDISCLUSTERLIBS += -lhiredis_cluster_ssl
-	endif
-	DEFS+=-DWITH_HIREDIS_CLUSTER
-	DEFS+=$(HIREDISCLUSTERDEFS)
-	LIBS+=$(HIREDISCLUSTERLIBS)
+ifeq ($(WITH_SENTINELS),1)
+$(info db_redis: WITH_SENTINELS=1, skipping hiredis cluster mode)
+else ifeq ($(WITH_REDIS_CLUSTER),0)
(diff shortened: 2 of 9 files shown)


lbalaceanu:
Hi @xkaraman, 

Can you confirm the makefile part is ok maybe?

Thank you

lbalaceanu:
Hi @xkaraman,

How do you find the CMakeLists.txt and Makefile?
Do you have any comments?

Thank  you

xkaraman:
Hey @lbalaceanu,

They look good at first sight.

Let me also try building all the cases as well to verify that as well!

lbalaceanu:
@xkaraman, I am also giving a try removing the WITH_SENTINELS altogether as WITH_REDIS_CLUSTER or WITH_SSL actually means a new library is used vs the "normal" hiredis one. This could be cleaner somehow. Should not take long.

src/modules/db_redis/CMakeLists.txt
```diff
@@ -8,6 +19,17 @@ add_library(hiredis::hiredis ALIAS PkgConfig::hiredis)
...
+
+if(WITH_HIREDIS_CLUSTER)
+  pkg_check_modules(hiredis_cluster REQUIRED IMPORTED_TARGET hiredis_cluster)
+  add_library(hiredis::hiredis_cluster ALIAS PkgConfig::hiredis_cluster)
+  target_compile_definitions(${module_name} PRIVATE WITH_HIREDIS_CLUSTER)
+  target_link_libraries(${module_name} PRIVATE hiredis::hiredis_cluster)
+endif()
+
 # Check for hiredis SSL support
 pkg_check_modules(hiredis_ssl IMPORTED_TARGET hiredis_ssl)
```
xkaraman: Remove this line, since we are checking this if with_hiredis_cluster

src/modules/db_redis/CMakeLists.txt
```diff
@@ -1,5 +1,16 @@
 file(GLOB MODULE_SOURCES "*.c")
 
+option(WITH_SENTINELS "Build db_redis with Redis Sentinel support" OFF)
```
xkaraman: Make this the default ie ON, due to before the PR `db_redis` was compiling with no cluster and therefore sentinels were being used implicitily.

Same for Makefile. make it compile with `sentinels` as default.

src/modules/db_redis/Makefile
```diff
@@ -60,17 +60,33 @@ endif
 DEFS+=$(HIREDISDEFS)
 LIBS=$(HIREDISLIBS)
 
+# Mode switches:
+# - WITH_REDIS_CLUSTER=1 (default): enable cluster mode if hiredis_cluster exists
+# - WITH_REDIS_CLUSTER=0: force normal (non-cluster) mode
+# - WITH_SENTINELS=1: enable sentinel mode and disable cluster mode
+WITH_REDIS_CLUSTER ?= 1
```
xkaraman: if i read this line correctly, it's setting the default one. Set it to sentinels as discussed in CMakelists.txt as well.

lbalaceanu:
Hello @miconda, @henningw could you maybe advise here a bit?

In the db_redis code we have these libraries hiredis and above it hiredis-cluster. Redis with sentinels uses simple hiredis and Redis with cluster needs hiredis-cluster built on top/depends on hiredis lib.

In the db_redis code, when compiling with WITH_HIREDIS_CLUSTER then some common functions are replaced

 `#ifdef WITH_HIREDIS_CLUSTER`
`#define redisCommand redisClusterCommand`
 `#define redisFree redisClusterFree` 
 

In this PR I grouped most of the code having to do with sentinels in redis_sentinels.{c/h}. I also introduced a build define WITH_SENTINELS which actually means: no hiredis-cluster lib used + we are using sentinel code. Would you say this is ok or better still renounce at WITH_SENTINELS altogether and use ifndef WITH_HIREDIS_CLUSTER and a with_sentinels modparam check?

For example, the code in the redis_sentinels.h would be safeguarded like this: #ifndef WITH_HIREDIS_CLUSTER.
In this case, at runtime, all code that would have to do with sentinels would have to check that we are in #ifndef WITH_HIREDIS_CLUSTER + modparam with_sentinels is set.

So should I go with a build define like WITH_SENTINELS or use already existing WITH_HIREDIS_CLUSTER + use_sentinels define? Maybe the latter is the better solution?

henningw:
> In the db_redis code we have these libraries hiredis and above it hiredis-cluster. Redis with sentinels uses simple hiredis and Redis with cluster needs hiredis-cluster built on top/depends on hiredis lib.
> 
> In the db_redis code, when compiling with WITH_HIREDIS_CLUSTER then some common functions are replaced
> 
> > `#ifdef WITH_HIREDIS_CLUSTER #define redisCommand redisClusterCommand #define redisFree redisClusterFree`
> 
> In this PR I grouped most of the code having to do with sentinels in redis_sentinels.{c/h}. I also introduced a build define WITH_SENTINELS which actually means: no hiredis-cluster lib used + we are using sentinel code. Would you say this is ok or better still renounce at WITH_SENTINELS altogether and use ifndef WITH_HIREDIS_CLUSTER and a with_sentinels modparam check?
> 
> For example, the code in the redis_sentinels.h would be safeguarded like this: #ifndef WITH_HIREDIS_CLUSTER. In this case, at runtime, all code that would have to do with sentinels would have to check that we are in #ifndef WITH_HIREDIS_CLUSTER + modparam with_sentinels is set.
> 
> So should I go with a build define like WITH_SENTINELS or use already existing WITH_HIREDIS_CLUSTER + use_sentinels define?

I just had a closer look together with @xkaraman. I think we should keep the WITH_SENTINEL, maybe somebody wants to build without sentinel in the future. From the module building point of view, we will align the new functionality to the current (old) way of building, i.e. build with sentinel functionality by default. When somebody wants to have cluster, the proper define needs to be set.

lbalaceanu:
Thank you @henningw,

Do I understand correctly that I need to delete the with_sentinels modparam then?

As far as I understand, these are the possible build modes:

1. WITH_SENTINELS = 0, WITH_HIREDIS_CLUSTER = 0 
2. WITH_SENTINELS = 1, WITH_HIREDIS_CLUSTER = 0 
3. WITH_SENTINELS = 0, WITH_HIREDIS_CLUSTER = 1

You say 2. WITH_SENTINELS = 1, WITH_HIREDIS_CLUSTER = 0  becomes the default?

henningw:
> Do I understand correctly that I need to delete the with_sentinels modparam then?
> 
> As far as I understand, these are the possible build modes:
> 
>     1. WITH_SENTINELS = 0, WITH_HIREDIS_CLUSTER = 0
>     2. WITH_SENTINELS = 1, WITH_HIREDIS_CLUSTER = 0
>     3. WITH_SENTINELS = 0, WITH_HIREDIS_CLUSTER = 1
> 
> You say 2. WITH_SENTINELS = 1, WITH_HIREDIS_CLUSTER = 0 becomes the default?

There is no need to delete the parameter, this is useful for activating/deactivating it at startup. The other options are compile time of course.
I think the option 2 is the current way of building db_redis in git master and 6.1 (when no cluster library is available), so it should stay this way, I think.

lbalaceanu:
Ok, if you think that somebody might want to build without sentinel support versus just have it and disable from a modparam, I will leave the WITH_SENTINELS (does not imply a new lib is added).

Currently upstream the default is: build without HIREDIS_CLUSTER which automatically means sentinel code was available. I added the WITH_SENTINELS and then I thought it just means ( !WITH_HIREDIS_CLUSTER + modparam use_sentinels )

Agree to leave WITH_SENTINELS and modparam use_sentinels then?

henningw:
Yes, I merged it. We adjust the logic as discussed in git master directly.

src/modules/db_redis/CMakeLists.txt
xkaraman: Remove this line, since we are checking this within with_hiredis_cluster block

henningw:
Adapted logic in cc49ccf374a

henningw:
And removed redundant check 11bf41afbabc

henningw:
cherry-picked to 6.1 as well

miconda:
Does it need to go to 6.0? A release out of it is planned next week.

xkaraman:
Sentinel support for db_redis was added to 6.1 if i am reading this [commit ](https://github.com/kamailio/kamailio/commit/8b3cb09cbc2537a9c9a65d4fb6eb3e6a5d149027) right:

miconda:
Ok, thanks!

PR #4916, opened 2026-09-08
core: dns - keep the NAPTR failover state in the dns_srv_handle

### LINKED ISSUE #4911: Regression since 5.3: one failed call to a NAPTR/SRV destination makes ALL other calls to it fail (works fine on 5.2.4)
## Summary

  With `use_dns_failover=yes` on a destination resolved via NAPTR → SRV (RFC 3263, several equally-weighted SRV targets), a single call that fails to
  reach any of the SRV targets causes **every other call** to the same destination (and to any other domain sharing the same NAPTR target) to immediately
  fail too, until the cached NAPTR record's TTL expires and is refreshed (roughly 1-2 minutes in our setup).

  This is a serious regression: the exact same `kamailio.cfg`, same DNS setup, works correctly on **Kamailio 5.2.4** — a single call failing there does
  not affect other calls. The problem appears starting with 5.3 and is still present in the latest 6.1.4.

  ## DNS setup (anonymized example)

  The destination used in the R-URI is a CNAME to a shared NAPTR-published target, which resolves via SRV to several equally-weighted edge hosts:

```
  $ dig +noall +answer sip-cluster5.example.net NAPTR
  sip-cluster5.example.net.  90   IN  CNAME  sip-hub.example.net.
  sip-hub.example.net.       128  IN  NAPTR  20 100 "S" "SIPS+D2T" "" _sips._tcp.sip-hub.example.net.

  $ dig +noall +answer _sips._tcp.sip-hub.example.net SRV
  _sips._tcp.sip-hub.example.net. 90 IN SRV 1 25 5061 edge-1.sip-hub.example.net.
  _sips._tcp.sip-hub.example.net. 90 IN SRV 1 25 5061 edge-2.sip-hub.example.net.
  _sips._tcp.sip-hub.example.net. 90 IN SRV 1 25 5061 edge-3.sip-hub.example.net.
  _sips._tcp.sip-hub.example.net. 90 IN SRV 1 25 5061 edge-4.sip-hub.example.net.

```

  Several other customer-facing domains (e.g. `sip-cluster6.example.net`, `sip-cluster7.example.net`, ...) CNAME to the same `sip-hub.example.net` NAPTR
  target — so poisoning it affects calls to all of them, not just one.

  ## Impact

  Once triggered, calls to the affected destination get an immediate `478 Request Failure`:

 ```
 ERROR: tm [ut.h:277]: uri2dst2(): failed to resolve "sip-cluster5.example.net" :unresolvable A or AAAA request (-7)
  ERROR: tm [t_fwd.c:1859]: t_forward_nonack(): failure to add branches

```

  right after a log line like:

`  NOTICE: <core> [core/dns_cache.c:3355]: mark_skip_current_naptr(): Mark to skip _sips._tcp.<target> NAPTR record due to all IPs are unreachable
`

  We reproduced this twice on live production traffic within the first minute after restarting Kamailio 6.1.4 — it is not a rare edge case on our traffic
  pattern (SIP trunking gateway sending calls in quick succession to the same destination).

  ## Steps to reproduce

  1. Set up a SIP destination resolved via NAPTR → SRV with several (in our case 4) SRV targets of equal priority/weight, and `use_dns_failover=yes`.
  2. Send a call where none of the SRV targets answer in time (all edges down/slow, or blackholed for testing).
  3. Kamailio logs "Mark to skip ... NAPTR record due to all IPs are unreachable" and the call fails as expected.
  4. Immediately send a second, unrelated call to the same destination (or another domain sharing the same NAPTR target).
  5. **Bug**: the second call fails instantly with "unresolvable A or AAAA request (-7)" / 478, even though it has nothing to do with the first call's
  failure.

  ## Expected behavior

  A failed call should not affect other, independent calls to the same destination. This is confirmed working correctly on **Kamailio 5.2.4** with the
  identical configuration — one call failing there does not impact subsequent/concurrent calls to the same destination.

  ## Tried and did NOT fix it

  Lowering `dns_cache_max_ttl` (e.g. to 15s) reduces how long the problem lasts but does not prevent it: when calls to the same destination arrive
  back-to-back (our gateway sends them within a second or two of each other), the next call can still land inside the shortened window and fail.

  ## Environment

  - Broken on: 6.0.1 and 6.1.4 (current GA)
  - Working correctly on: 5.2.4, same `kamailio.cfg`
  - `use_dns_failover=yes`, NAPTR-based SIP routing, 4 SRV targets behind the NAPTR record
  - OS: Debian (trixie)

### PR DESCRIPTION
**Pull request by Gskill75**
#### Pre-Submission Checklist
  
  from source code folder

> Note on the checklist: `core` is deliberately split in **two** commits, as
> requested in review, so that the refcounting fix is clearly identified and can
> be reviewed and backported on its own. Happy to squash them if preferred.

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #4911

#### Description

Fixes #4911.

Three commits:

1. `core: dns - fix the refcounting of the entries kept in dns_srv_handle`
2. `core: dns - keep the NAPTR failover state in the dns_srv_handle` (the #4911 fix)
3. `test: regression test for the NAPTR failover state ownership`

##### Problem

`use_dns_failover=yes` + `dns_try_naptr=yes`, a destination resolved through
NAPTR -> SRV -> several unreachable edge hosts. When one transaction has tried
every destination of a NAPTR record, kamailio logs:

```
mark_skip_current_naptr(): Mark to skip ... NAPTR record due to all IPs are unreachable
```

That transaction fails correctly (408), but every *unrelated* transaction to the
same name then fails immediately:

```
uri2dst2(): failed to resolve "<domain>" :unresolvable A or AAAA request (-7)
t_forward_nonack(): failure to add branches
```

and gets a 478, without retrying the NAPTR/SRV destinations at all — until the
NAPTR cache entry expires. It is especially bad when several public domains
CNAME to the same NAPTR target, since one failed call poisons all of them.
Reproducible on 5.8.x, 6.0.x, 6.1.x and master; 5.2.4 does not show it.

##### Root cause

`28049aafc8dd06c160ce5e7b8d5e4fc728441b0c` ("core: dns - use all NAPTR records",
#2290) added `unsigned char skip_record` to `struct naptr_rdata`. That struct
belongs to a `dns_hash_entry`, i.e. to the **shared** DNS cache.
`mark_skip_current_naptr()` sets `naptr->skip_record = 1` and
`dns_naptr_sip_iterate()` then skips the record — for every transaction and
every worker, until the entry expires:

```
transaction A exhausts NAPTR A
        |
        v
cached naptr_rdata.skip_record = 1
        |
        +---- transaction B sees it
        +---- transaction C sees it
        +---- other kamailio workers see it
```

So per-transaction failover state was being written into cross-transaction
shared state.

##### Fix

Keep the list of the NAPTR records already used in the resolution handle
instead, as `dns_srv_handle.naptr_tried_rrs`, next to the existing
`srv_tried_rrs`. `struct dns_srv_handle` already carries everything else needed
to continue a DNS failover across the successive attempts of one transaction.

Lifecycle:

* `dns_srv_handle_init()` clears it, so a new transaction always sees all the
  NAPTR records again;
* `dns_srv_handle_reset()` keeps it — exactly as it already keeps `h->naptr` —
  because it is only used to drop the srv/a state while continuing the *same*
  failover sequence;
* `dns_srv_handle_cpy()` carries it over to the branch of the next destination
  (plain struct copy);
* it disappears with the handle; nothing is ever written to the DNS cache.

`mark_skip_current_naptr()`, `have_more_active_naptr()` and
`naptr_rdata.skip_record` become unnecessary and are removed.

##### Marking names as unresolvable is not affected

Raised in review — detailing it here since it is the main risk to check.

Negative caching is a different mechanism and this PR does not touch it. An
unresolvable name is marked in `dns_get_entry()` / `dns_cache_do_request()`:
when the resolver returns nothing, `dns_cache_mk_bad_entry()` creates a
`DNS_FLAG_BAD_NAME` entry with `dns_neg_cache_ttl`, and every later lookup of
that name hits it (`dns_hash_get()` -> "negative cache => not resolvable") and
returns without querying the DNS. None of those functions, flags or config
variables are modified by this PR — the diff does not reference
`DNS_FLAG_BAD_NAME`, `dns_neg_cache_ttl`, `dns_cache_mk_bad_entry()` or
`dns_cache_do_request()` at all. So FQDNs the DNS does not answer for are still
queried once per `dns_neg_cache_ttl`, exactly as before.

`skip_record` was never driven by a DNS failure. `mark_skip_current_naptr()` was
called when `dns_srv_sip_resolve()` returned an error while *continuing* a
failover, i.e. when the handle had consumed all the SRV/A records of the current
NAPTR record after the destinations did not answer — its own log message says
"all IPs are unreachable". It recorded a *reachability* verdict in the DNS
cache, which is why it leaked across transactions and across the domains
CNAME-ing to the same target.

On the performance side, removing it does not add DNS traffic: when a second
transaction walks NAPTR A again, the SRV and A/AAAA records are served from the
DNS cache under their own TTLs (and from the negative cache when they do not
resolve), so this costs cache lookups, not queries.

What does legitimately change is that a second transaction will again *attempt*
destinations that a previous transaction found dead. Suppressing that is the job
of the destination blocklist, which is per destination (ip/port/proto) with its
own TTL and is consulted before sending a branch (`dst_is_blocklisted()` in
`t_fwd.c`, fed by `dst_blocklist_add()` on send error and on timeout in
`timer.c`) — unaffected by this PR. That is the right granularity for it: a
shared NAPTR flag cannot tell "this destination is dead" from "this destination
is perfectly fine for another customer domain that happens to CNAME to the same
NAPTR target", which is exactly the failure reported in #4911.

##### PR #2290 behaviour is preserved

`dns_naptr_sip_resolve()` still moves on to the next supported NAPTR record once
the current one has no destination left; it just asks `dns_naptr_sip_iterate()`
with the handle's own bitmap to know which records it has already used. NAPTR
order/preference selection (`naptr_choose()`), `MAX_NAPTR_RRS`, SRV load
balancing (`DNS_SRV_LB` — `h->srv_tried_rrs` is re-armed by
`dns_srv_resolve_nxt()` whenever `h->srv` is dropped), the IPv4/IPv6 flags and
the fallback to SRV then A/AAAA are unchanged.

##### Refcounting fix (first commit, independent)

Split out as its own commit so it can be reviewed and backported alone. It comes
from `34844f3b235235c9d44681f89098893967e3b408`, which started storing the NAPTR
entry in the handle:

* the handle keeps the reference returned by `dns_get_entry()`, but the success
  path also called `dns_hash_put(e)`, and `dns_srv_handle_put()` released it a
  second time later; on the calls continuing the same resolution `e` is
  `h->naptr` and no new reference is taken at all, so every one of them
  decremented it again;
* on the fallback path `dns_srv_handle_init()` only zeroed `h->srv` / `h->a`,
  leaking the entries left there by the failed `dns_srv_resolve_ip()` attempts;
  `dns_srv_handle_reset()` releases them and keeps `h->naptr`.

The second commit additionally makes a NAPTR record exhausted while continuing a
resolution report the last error instead of falling through and restarting
`dns_srv_sip_resolve()` from scratch.

The now-redundant second `str2ip()/str2ip6()` guard is dropped: the first-call
branch still returns early for IP literals *before* any `dns_get_entry()`, so
the fix from #2541 stays in place.

##### `rr_skip_record` compatibility (#4174)

`1f4e6795502626865d10c83b3880e2985eb94787` exposed the cached `skip_record`
through `dns_cache_print_entry()` as `rr_skip_record`. Rather than break the
`dns.view` output, the field is kept and now always reports `"no"`, with a
comment marking it deprecated: a cached NAPTR record is never skipped globally
anymore, the records already tried are tracked per `dns_srv_handle`. No RPC
consumer breaks, and the reported value is accurate. Happy to drop the field
entirely instead if that is preferred.

##### Tests

`test/misc/code/dns_naptr_iterate.c` drives `dns_naptr_sip_iterate()` with two
independent `dns_srv_handle` instances over the same hand-built cached NAPTR
record list, and checks that:

* one handle walks through every NAPTR record and then reports none left;
* `dns_srv_handle_reset()` does not lose the records already tried, while
  `dns_srv_handle_init()` does;
* a fresh handle still selects the record the first handle exhausted;
* two interleaved handles keep independent iteration state;
* the cached records are left byte for byte unchanged.

The core has no unit test harness, so the test links against the objects of a
normal build; `sh test/misc/code/dns_naptr_iterate.sh` builds and runs it
(19 checks, all passing). If a test outside `test/misc/code/` would be
preferred, or if it is not wanted at all, I am happy to move or drop that
commit.

Manual end-to-end check for the scenarios a unit test cannot cover: with
`use_dns_failover=yes` / `dns_try_naptr=yes` and a name with two NAPTR records
where the first one's SRV targets blackhole traffic, place a call (408 after
trying all of NAPTR A then NAPTR B), then immediately place a second call to the
same name or to another domain CNAME-ing to the same NAPTR target — before this
patch it fails instantly with a 478, after it the second call tries NAPTR A
normally. `kamcmd dns.view` shows the cached entry unchanged throughout.

##### Backporting

5.8, 6.0 and 6.1 do not have `34844f3b23`, so they have no `h->naptr` and still
call `dns_srv_handle_init(h)` inside the NAPTR loop, which would clear the new
bitmap. A backport needs either `34844f3b23` first, or that `init` replaced by a
reset that preserves `naptr_tried_rrs`. Glad to prepare the backport PRs once
the approach is agreed.

### DIFF
--- a/src/core/dns_cache.c
+++ b/src/core/dns_cache.c
@@ -2760,10 +2760,6 @@ struct naptr_rdata *dns_naptr_sip_iterate(struct dns_rr *naptr_head,
 		}
 		LM_DBG("found a valid sip NAPTR rr %.*s, proto %d\n", naptr->repl_len,
 				naptr->repl, (int)naptr_proto);
-		if(naptr->skip_record) {
-			i++;
-			continue;
-		}
 		if((naptr_proto_supported(naptr_proto))) {
 			if(naptr_choose(&naptr_saved, &saved_proto, naptr, naptr_proto))
 				idx = i;
@@ -3330,65 +3326,6 @@ inline static int dns_srv_sip_resolve(struct dns_srv_handle *h, str *name,
 #ifdef USE_NAPTR
 
 
-static void mark_skip_current_naptr(
-		struct dns_rr *naptr_head, struct dns_hash_entry *srv)
-{
-	int i;
-	struct dns_rr *l;
-	struct naptr_rdata *naptr;
-
-	if(!naptr_head || !srv) {
-		return;
-	}
-
-	for(l = naptr_head, i = 0; l && (i < MAX_NAPTR_RRS); l = l->next, i++) {
-		naptr = (struct naptr_rdata *)l->rdata;
-		if(naptr == 0) {
-			break;
-		}
-		if(naptr->skip_record) {
-			continue;
-		}
-		if(srv->name_len == naptr->repl_len
-				&& !memcmp(srv->name, naptr->repl, srv->name_len)) {
-			naptr->skip_record = 1;
-			LM_NOTICE("Mark to skip %.*s NAPTR record due to all IPs are "
-					  "unreachable\n",
-					naptr->repl_len, naptr->repl);
-			break;
-		}
-	}
-}
-
-
-inline static int have_more_active_naptr(struct dns_rr *naptr_head)
-{
-	int i, res = 0;
-	struct dns_rr *l;
-	struct naptr_rdata *naptr;
-	char naptr_proto;
-
-	if(!naptr_head) {
-		return res;
-	}
-
-	for(l = naptr_head, i = 0; l && (i < MAX_NAPTR_RRS); l = l->next, i++) {
-		naptr = (struct naptr_rdata *)l->rdata;
-		if(naptr == 0) {
-			break;
-		}
-		if(naptr->skip_record) {
-			continue;
-		} else if((naptr_proto = naptr_get_sip_proto(naptr)) <= 0) {
-			continue;
-		} else if((naptr_proto_supported(naptr_proto))) {
-			res = 1;
-			break;
-		}
-	}
-	return res;
-}
-
 /* resolves a host name trying:
  * - NAPTR lookup if the address is not an ip and proto!=0, port!=0
  *    *port==0 and *proto=0 and if flags allow NAPTR lookups
@@ -3407,13 +3344,11 @@ inline static int dns_naptr_sip_resolve(struct dns_srv_handle *h, str *name,
 {
 	struct hostent *he;
 	struct ip_addr *tmp_ip;
-	naptr_bmp_t tried_bmp;
 	struct dns_hash_entry *e;
 	char n_proto, origproto;
 	str srv_name;
 	int ret;
-	int res;
-	int try_lookup_naptr = 0;
+	int continued;
 
 	ret = -E_DNS_NO_NAPTR;
 	if(proto)
@@ -3430,6 +3365,7 @@ inline static int dns_naptr_sip_resolve(struct dns_srv_handle *h, str *name,
 		}
 		return -E_DNS_NO_NAPTR;
 	}
+	continued = 0;
 	if(((h->srv == 0) && (h->a == 0) && (h->naptr == 0)) && /* first call */
 			proto && port && (*proto == 0) && (*port == 0)) {
 		*proto = PROTO_UDP; /* just in case we don't find another */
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by miconda** [comment]
Does this affect the process of marking the records invalid because they are not resolvable. This is a feature to avoid queries for FQDNs for which the dns is not responding and can deteriorate the performances.

Also, I would suggest to split into a separate commit the part for the fix referred by:

```
While there, fix the reference counting of h->naptr: since it is stored
in the handle it must be released by dns_srv_handle_put() only, the
dns_hash_put() calls on the success path decremented the refcount of an
entry the handle still pointed to.
```

So it is clearly identified and can be reviewed/backported alone.

Gskill75:
Thanks for the review.

**On marking records invalid / not resolvable — that path is untouched.**

Negative caching is a separate mechanism and this PR does not modify it. An
unresolvable name is still marked in `dns_get_entry()` / `dns_cache_do_request()`:
when the resolver returns nothing, `dns_cache_mk_bad_entry()` creates a
`DNS_FLAG_BAD_NAME` entry with `dns_neg_cache_ttl`, and every later lookup hits it
(`dns_hash_get()` -> "negative cache => not resolvable") and returns without
querying the DNS. The diff does not reference `DNS_FLAG_BAD_NAME`,
`dns_neg_cache_ttl`, `dns_cache_mk_bad_entry()` or `dns_cache_do_request()` at all,
so FQDNs the DNS does not answer for are still queried at most once per
`dns_neg_cache_ttl`, exactly as today.

`skip_record` was never driven by a DNS failure. `mark_skip_current_naptr()` was
called when `dns_srv_sip_resolve()` returned an error while *continuing* a
failover, i.e. once the handle had consumed all the SRV/A records of the current
NAPTR record because the destinations did not answer — its own log message says
"all IPs are unreachable". It recorded a reachability verdict in the DNS cache,
which is precisely why it leaked into unrelated transactions and into other
domains CNAME-ing to the same NAPTR target.

No extra DNS traffic results either: when a second transaction walks that NAPTR
record again, the SRV and A/AAAA lookups are served from the DNS cache under their
own TTLs (and from the negative cache when they do not resolve), so it costs cache
lookups, not queries.

What does change is that a second transaction will again *attempt* destinations a
previous transaction found dead. Suppressing that is the destination blocklist's
job — per ip/port/proto, with its own TTL, consulted before sending a branch
(`dst_is_blocklisted()` in `t_fwd.c`, fed by `dst_blocklist_add()` on send error
and on timeout in `timer.c`) — and it is unaffected here. That is also the right
granularity: a shared NAPTR flag cannot distinguish "this destination is dead"
from "this destination is fine for another customer domain that happens to CNAME
to the same NAPTR target", which is the failure reported in #4911.

I have expanded the PR description with the same explanation.

**On splitting the refcounting fix** — done, it is now the first commit
(`core: dns - fix the refcounting of the entries kept in dns_srv_handle`), touching
only `dns_naptr_sip_resolve()` and independent of the rest, so it can be reviewed
and backported on its own. It also covers a second leak on the same path:
`dns_srv_handle_init()` on the fallback only zeroed `h->srv` / `h->a` instead of
releasing them.

That makes `core` two commits rather than one, contrary to the checklist — tell me
if you would rather have them squashed.

miconda:
I pushed manually the commits for the core.

The test would need first some decision about that folder and an eventual updated framework -- it is rather old an not properly structured.

Thanks!

PR #4917, opened 2026-09-08
SIP parsing enhancements (core, pua, sipcapture, sanity)

david-trihy:
#### Pre-Submission Checklist

  from source code folder

#### Type Of Change

#### Checklist:

- [ ] Tested changes locally

#### Description
This PR is to make the sip parsing more deterministic when in both strict and non strict mode.

**Strict mode changes**

When in strict parsing mode ie. `sip_parser_mode=1` when receiving a new message the Kamailio core will test for what it calls the transaction quadruple, which is the `To`, `From`, `Call-ID`, `CSeq` and fail to enter the route if the SIP messages does not contain those headers. The parsing of the message is deliberately lazy in nature as opposed to eager parsing but it can leave the core in an odd state when strict parsing is set on.

Example invalid message which with strict parsing on will fail to enter the route due to a duplicate quadruple(To in this case)

```
INVITE sip:user@domain.com SIP 2.0
To: <valid nameaddr>
To: <valid nameaddr>
From: <valid nameaddr>
Call-ID: valid-callid
CSeq: INVITE 1
X-Custom-Header: value
....
Content-Length: 0
```

Example of a request which will enter the route but result in undefined behaviour

```
INVITE sip:user@domain.com SIP 2.0
To: <valid nameaddr>
From: <valid nameaddr>
Call-ID: valid-callid
CSeq: INVITE 1
X-Custom-Header: value
To: <valid nameaddr>
....
Content-Length: 0
```

The reason why this happens is because the second To header is not parsed as the check_required_headers keeps going until it finds an instance of the quadruple and it will enter the route and depending on how you are using the route you could fail at another point where the rest of the headers would be parsed or when construct the SIP reply or request once the route finishes, given the expectation would be that strict parsing would fail straight away but in this case it would not. 

The change turns the strict parsing mode into an eager parser so it fails early, this shouldn't meaningfully change the behaviour of most configuration but it does in the case outlined above so it technically creates a backwards incompatible change.

**Non strict mode changes**

When the `sip_parser_mode=0` and there is a duplicate check it will only copy the first valid instance of the quadruple in the reply, the current behaviour is such that when constructing a reply eg. `400 Bad Request` both To headers are copied, this change ensures only one of them is copied which is the first instance of that header during parsing.

One other change is when in this mode if there is a message such as this

```
INVITE sip:user@domain.com SIP 2.0
To: <invalid nameaddr>
To: <valid nameaddr>
From: <valid nameaddr>
Call-ID: valid-callid
CSeq: INVITE 1
X-Custom-Header: value
....
Content-Length: 0
```

It will still parse the quadruple but will take into account the valid `To` header only which in this case is the second instance of the header but the first instance is tracked as an error so that when calling `sanity_check` in the `sanity` module it can behave correctly if the check for required headers is set in the modparam.

The reason for fixing this case is that if the request contains at least one valid instance of the quadruple then it can be used to generate a reply. In the above message case the core would drop the request.

However the below example would not drop and then the sanity check would fail at this point and due to how the check required headers works it means we would not be able to reply but we would do the sanity check.

```
INVITE sip:user@domain.com SIP 2.0
To: <valid nameaddr>
From: <valid nameaddr>
Call-ID: valid-callid
CSeq: INVITE 1
X-Custom-Header: value
To: <invalid nameaddr>
....
Content-Length: 0
```

This just once again tightens up the behaviour so that is deterministic and the script writer can reasonably infer how the parsing will behave based on their requirements.

#### Testing

I have a suite of SIPp tests that can verify this behaviour, what is the recommended project to do those tests in?

### DIFF
--- a/src/core/parser/msg_parser.c
+++ b/src/core/parser/msg_parser.c
@@ -69,12 +69,26 @@ msg_flags_t global_req_flags = 0;
 
 int ksr_sip_parser_mode = KSR_SIP_PARSER_MODE_STRICT;
 
-/* returns pointer to next header line, and fill hdr_f ;
- * if at end of header returns pointer to the last crlf  (always buf)*/
-char *get_hdr_field(
-		char *const buf, char *const end, struct hdr_field *const hdr)
-{
+/* advance _p to the next header, including folded continuation lines.
+ * _match is NULL if no '\n' is found. */
+#define GET_HDR_NEXT(_p, _end, _match)                                   \
+	do {                                                                 \
+		do {                                                             \
+			(_match) = q_memchr((_p), '\n', (_end) - (_p));              \
+			if((_match) == NULL)                                         \
+				break;                                                   \
+			(_match)++;                                                  \
+			(_p) = (_match);                                             \
+		} while(((_p) < (_end)) && ((*(_p) == ' ') || (*(_p) == '\t'))); \
+	} while(0)
 
+/* fill hdr and set *next to the following header line (or the EOH crlf).
+ * next may be NULL if the caller does not need that pointer.
+ * returns 0 on success, -1 on error. hdr->type is HDR_ERROR_T only when
+ * parse_hname failed; a body parse error keeps the name type. */
+int get_hdr_field(char *const buf, char *const end, struct hdr_field *const hdr,
+		char **next)
+{
 	char *tmp = 0;
 	char *match;
 	struct via_body *vb;
@@ -82,6 +96,9 @@ char *get_hdr_field(
 	struct to_body *to_b;
 	int integer, err;
 	unsigned uval;
+
+	if(next)
+		*next = NULL;
 
 	if(!buf) {
 		DBG("null buffer pointer\n");
@@ -92,7 +109,9 @@ char *get_hdr_field(
 		/* double crlf or lflf or crcr */
 		DBG("found end of header\n");
 		hdr->type = HDR_EOH_T;
-		return buf;
+		if(next)
+			*next = buf;
+		return 0;
 	}
 
 	tmp = parse_hname(buf, end, hdr);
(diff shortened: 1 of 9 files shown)


miconda:
Thanks for contributing, the PR will be reviewed.

For tests, if you want to contribute and what you have suits to that framework, look at the repository:

src/core/parser/msg_parser.c
```diff
@@ -336,12 +372,39 @@ int parse_headers(
...
 		hf->type = HDR_ERROR_T;
-		rest = get_hdr_field(tmp, end, hf);
+		rest = NULL;
+		if(get_hdr_field(tmp, end, hf, &rest) < 0) {
+			LOG(cfg_get(core, core_cfg, sip_parser_log),
+					"bad header field [%.*s]\n",
+					(end - tmp > 100) ? 100 : (int)(end - tmp), tmp);
+
+			if((ksr_sip_parser_mode & KSR_SIP_PARSER_MODE_STRICT)
+					|| (ser_error == E_OUT_OF_MEM)) {
```
miconda, line 382: Upon a quick review, I am not sure it really has any impact here, but for proper clarification, is get_hdr_field() re-/setting ser_error? ser_error is global and can stay set from another part of code, which might make the above condition true.

david-trihy: Yes that's correct, my idea at the time was that we needed the ability to stop parsing on pkg_malloc failures when in permissive mode instead of trying to continually parse. Looking at the code I'm unsure why I just didn't add another error code to get_hdr_field and return -2 when there's a memory allocation error and used this pattern instead, need to reread it to see if there was a particular reason.

I still need to port the tests to the framework you posted, I have them running against our internal tooling but they're pretty basic, just a simple sanity_check() call in the script when permissive parsing is on.

PR #4918, opened 2026-09-08
dispatcher: trigger event_routes for disabled/inactive destinations on init/reload in mode 2

### LINKED ISSUE #4906: dispatcher: event_callback_mode=2 (DS_EVRTMODE_INIT) does not trigger event_route on startup/reload for disabled destinations
According to the official documentation for module `dispatcher` ([dispatcher_admin.xml](https://kamailio.org/docs/modules/stable/modules/dispatcher.html#dispatcher.p.event_callback_mode)):

> *"If set to 2, the event routes will always be executed when a state change occurs, even if no previous state exists (**for example, on startup or reload**)."*

However, in `dispatch.c`, `ds_run_route()` is only ever executed inside `ds_update_state()`, which is triggered upon receiving SIP probe (OPTIONS) replies or when an explicit runtime state update command is processed.

Because disabled or inactive destinations (`flags = 4`, `DS_DISABLED_DST` or `flags = 1`, `DS_INACTIVE_DST`) are intentionally skipped during keepalive probing in `ds_ping()` (`if(ds_skip_dst(...)) continue;`), `ds_update_state()` is never invoked for them.

Consequently, initial event routes (`dispatcher:dst-down`) are never fired on startup or reload for these destinations. In environments where monitoring systems (e.g. Prometheus via `xhttp_prom` or Zabbix) rely on `event_route[dispatcher:dst-down]` to set gateway metric gauges, initially disabled gateways are completely absent from exported metrics, breaking Low-Level Discovery (LLD) and observability.

#### Reproduction

1. Configure `dispatcher` with:
```ini
modparam("dispatcher", "event_callback_mode", 2)
```

2. Define an event route to export/log status:
```ini
event_route[dispatcher:dst-down] {
    xinfo("Destination DOWN: $dsv(uri) flags: $dsv(flags)\n");
}
event_route[dispatcher:dst-up] {
    xinfo("Destination UP: $dsv(uri)\n");
}
```

3. Add a destination with disabled flag `4` in `dispatcher.list` or database:
```text
1 sip:192.0.2.10:5060 4 0
```

4. Start Kamailio or issue `kamcmd dispatcher.reload`.
5. Observe that `dispatcher:dst-down` is never executed for `sip:192.0.2.10:5060`, and no state is reported until a manual state change occurs.

### Possible Solutions

Trigger initial event routes during startup and post-reload for all configured destinations when `event_callback_mode` is set to `2` (`DS_EVRTMODE_INIT`):

1. Iterate over `_ds_list` using `ds_iter_set()` to execute `dispatcher:dst-down` (for inactive/disabled nodes) or `dispatcher:dst-up` (for active nodes) with appropriate `$dsv(flags)`.
2. Execution context:
   - For startup: execute in post-fork worker context (such as `PROC_SIPINIT` / rank 1 or `PROC_POSTCHILDINIT`) rather than `PROC_INIT` (pre-fork) to ensure safe execution of `event_route`.
   - For reload: trigger post-reload in `dispatcher.reload`.

I have a working proof-of-concept patch tested in our environment. I would appreciate feedback on:
- Whether this behavior aligns with the original intent of `DS_EVRTMODE_INIT`.
- The preferred process rank for the startup callback execution.

I will gladly submit a clean PR once the architectural approach is confirmed.

### Additional Information

- [x] LLM/AI Assistants were involved in discovery, analysis or submission of the issue
- [ ] It happened in a production deployment
- [x] It happened in a testing or development deployment
- [ ] Testing or fuzzing tools were used to generate SIP traffic when it happened

* **Kamailio Version** - output of `kamailio -v`:

```text
kamailio 6.1.4 (x86_64/Linux)
```

* **Operating System**:

```text
Linux (Debian 13 x86_64)
```

lean1ee:
### Description

When `event_callback_mode = 2` (`DS_EVRTMODE_INIT`) is enabled, event routes are executed for active destinations on the first probe reply (via `old_state == 0`). However, disabled (`flags = 4`) and inactive (`flags = 1`) destinations are intentionally skipped from OPTIONS keepalive probing in `ds_ping_set()`, and neither `child_init` nor `dispatcher_rpc_reload` invoked `ds_update_state()`.

Consequently, initial event routes (`dispatcher:dst-down`) were never fired for these destinations on startup or reload, leaving monitoring systems (e.g. Prometheus via `xhttp_prom` or Zabbix) without metrics for disabled gateways.

This patch adds `ds_run_init_event_routes()`, which iterates over `_ds_list` using `ds_iter_set()` to trigger `dispatcher:dst-down` for disabled and inactive destinations:
- On startup: in `child_init()` for `rank == PROC_SIPINIT` (first worker).
- On reload: in `dispatcher_rpc_reload()` right after lists are successfully loaded into shared memory.

Active destinations continue to be triggered by keepalive replies as originally introduced in #4460, avoiding duplicate events.

Closes #4906

### DIFF
--- a/src/modules/dispatcher/dispatch.c
+++ b/src/modules/dispatcher/dispatch.c
@@ -4202,11 +4202,11 @@ int ds_update_state(sip_msg_t *msg, int group, str *address, str *iuid,
 
 
 			if((ds_event_callback_mode == DS_EVRTMODE_RUNTIME)
-					|| (ds_event_callback_mode == DS_EVRTMODE_INIT)
+					|| ((ds_event_callback_mode & DS_EVRTMODE_OPTIONS) == 0)
 					|| ((mode & DS_STATE_MODE_FUNC) == 0)) {
 				was_down = ds_skip_dst(old_state);
 				is_down = ds_skip_dst(idx->dlist[i].flags);
-				if(ds_event_callback_mode == DS_EVRTMODE_INIT) {
+				if(ds_event_callback_mode & DS_EVRTMODE_INIT) {
 					if((!was_down && is_down) || (old_state == 0 && is_down)) {
 						ds_run_route(msg, address, "dispatcher:dst-down", rctx);
 					} else if((was_down && !is_down)
@@ -4495,6 +4495,47 @@ int ds_reinit_state_all(int group, int state)
 					old_state, idx->dlist[i].flags, idx);
 		}
 	}
+	ds_put_list(list);
+	return 0;
+}
+
+/**
+ * Callback for iterating over destinations to trigger init event routes
+ * for disabled and inactive nodes
+ */
+static void ds_run_init_dst_cb(ds_set_t *node, int i, void *arg)
+{
+	ds_rctx_t rctx;
+
+	if(!ds_skip_dst(node->dlist[i].flags)) {
+		return;
+	}
+
+	memset(&rctx, 0, sizeof(ds_rctx_t));
+	rctx.flags = node->dlist[i].flags;
+	rctx.setid = node->id;
+	rctx.uri = node->dlist[i].uri;
+	rctx.code = 0;
+
+	ds_run_route(NULL, &node->dlist[i].uri, "dispatcher:dst-down", &rctx);
+}
+
+/**
+ * Execute event routes for disabled/inactive destinations when DS_EVRTMODE_LOAD is enabled
+ */
+int ds_run_init_event_routes(void)
+{
+	ds_list_t *list;
+
+	if((ds_event_callback_mode & DS_EVRTMODE_LOAD) == 0)
+		return 0;
+
+	list = ds_get_list();
+	if(list == NULL || list->head == NULL)
+		return 0;
+
+	LM_DBG("executing init event routes for disabled/inactive destinations\n");
+	ds_iter_set(list->head, &ds_run_init_dst_cb, NULL);
 	ds_put_list(list);
 	return 0;
 }
--- a/src/modules/dispatcher/dispatch.h
+++ b/src/modules/dispatcher/dispatch.h
@@ -91,8 +91,9 @@
 #define DS_STATE_MODE_FUNC (1<<1)
 
 #define DS_EVRTMODE_RUNTIME 0
-#define DS_EVRTMODE_OPTIONS 1
-#define DS_EVRTMODE_INIT 2
+#define DS_EVRTMODE_OPTIONS (1 << 0)
+#define DS_EVRTMODE_INIT    (1 << 1)
+#define DS_EVRTMODE_LOAD    (1 << 2)
 
 #define DS_SELRES_FAILED (ds_selres_t){0}
 
@@ -181,6 +182,7 @@ int ds_update_state(sip_msg_t *msg, int group, str *address, str *iuid,
 		int state, int mode, ds_rctx_t *rctx);
 int ds_reinit_state(int group, str *address, str *iuid, int state);
 int ds_reinit_state_all(int group, int state);
+int ds_run_init_event_routes(void);
 int ds_reinit_duid_state(int group, str *vduid, int state);
 int ds_mark_dst(struct sip_msg *msg, int state);
 int ds_mark_dst_mode(struct sip_msg *msg, int state, int mode);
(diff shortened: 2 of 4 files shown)


miconda:
I think this should be introduced as mode 4 and the event_callback_mode be tested as a flag/bit, then various modes can be combined.

By just pushing like it is, the behaviour for mode 2 changes, because this is not done on a state change, this is done always with the initial state.

lean1ee:
Thanks @miconda for the guidance!
I have updated the PR accordingly:
- Evaluated `event_callback_mode` as bitmask flags (`1 << 0`, `1 << 1`, `1 << 2`).
- Added mode 4 (`DS_EVRTMODE_LOAD`, `1 << 2`) to execute `dispatcher:dst-down` for initial state of disabled and inactive destinations on startup and RPC reload.
- Mode 2 remains backwards compatible and modes can now be combined (e.g., `2 + 4 = 6`).
- Updated the documentation in `dispatcher_admin.xml`.

src/modules/dispatcher/dispatch.c
```diff
@@ -4201,12 +4201,11 @@ int ds_update_state(sip_msg_t *msg, int group, str *address, str *iuid,
 			}
 
 
-			if((ds_event_callback_mode == DS_EVRTMODE_RUNTIME)
-					|| (ds_event_callback_mode == DS_EVRTMODE_INIT)
+			if(((ds_event_callback_mode & DS_EVRTMODE_OPTIONS) == 0)
```
miconda: The condition for `ds_event_callback_mode == DS_EVRTMODE_RUNTIME` (for value 0) was removed. Was it intentional?

lean1ee:
It wasn't meant to drop mode 0 support conceptually — the idea was that since mode 1 (`DS_EVRTMODE_OPTIONS`, `1 << 0`) restricts execution to OPTIONS keepalives only, testing `((ds_event_callback_mode & DS_EVRTMODE_OPTIONS) == 0)` evaluates to true whenever bit 0 is unset. This covers mode 0 (`DS_EVRTMODE_RUNTIME` where all bits are 0), as well as modes 2, 4, and 6.

However, I realize this makes mode 0 implicit and leaves `DS_EVRTMODE_RUNTIME` unused in the source.

Would you prefer keeping `(ds_event_callback_mode == DS_EVRTMODE_RUNTIME)` explicitly, like this:

```c
if((ds_event_callback_mode == DS_EVRTMODE_RUNTIME)
		|| ((ds_event_callback_mode & DS_EVRTMODE_OPTIONS) == 0)
		|| ((mode & DS_STATE_MODE_FUNC) == 0)) {
```

Or would you prefer structuring the condition differently?

src/modules/dispatcher/dispatch.c
lean1ee: It wasn't meant to drop mode 0 support conceptually — the idea was that since mode 1 (`DS_EVRTMODE_OPTIONS`, `1 << 0`) restricts execution to OPTIONS keepalives only, testing `((ds_event_callback_mode & DS_EVRTMODE_OPTIONS) == 0)` evaluates to true whenever bit 0 is unset. This covers mode 0 (`DS_EVRTMODE_RUNTIME` where all bits are 0), as well as modes 2, 4, and 6.

However, I realize this makes mode 0 implicit and leaves `DS_EVRTMODE_RUNTIME` unused in the source.

Would you prefer keeping `(ds_event_callback_mode == DS_EVRTMODE_RUNTIME)` explicitly, like this:

```c
if((ds_event_callback_mode == DS_EVRTMODE_RUNTIME)
		|| ((ds_event_callback_mode & DS_EVRTMODE_OPTIONS) == 0)
		|| ((mode & DS_STATE_MODE_FUNC) == 0)) {
```

Or would you prefer structuring the condition differently?

miconda: I think it is better and safer in long term to keep that condition explicitly. If other changes are done, then the mode condition can be found quickly.

I am thinking that, after a proper review, maybe all possible modes should be changed to be flags and initialized to what would be the new value for DS_EVRTMODE_RUNTIME

lean1ee: Updated the PR to keep `(ds_event_callback_mode == DS_EVRTMODE_RUNTIME)` explicitly as discussed. Thanks!

PR #4925, opened 2026-09-09
sdpops: fix $sdp(m0:*) to use the first m-line

Sig-thd:
#### Pre-Submission Checklist
  from source code folder

#### Type Of Change

#### Checklist:
- [x] Tested changes locally

#### Description

The `m0` variants of `$sdp(...)` return values taken from the **last** media stream of an SDP
body instead of the first, whenever the body carries more than one `m=` line.

The SDP parser prepends each parsed stream to the session list (`src/core/parser/sdp/sdp.c`,
"Insert the new stream": `stream->next = _session->streams; _session->streams = stream;`), so
`sdp->sessions->streams` points at the **last** m-line parsed. `pv_get_sdp()` dereferenced that
pointer directly for every `m0`-named case, so the returned value came from the wrong stream.
A body with a single stream is unaffected, which is why this is easy to miss.

The fix selects the stream with `get_sdp_stream_sdp(sdp, 0, 0)`, which matches on the
parse-ordered `stream_num`, so index 0 is the first `m=` line as the variable name implies.

Affected variables: `$sdp(m0:rtp:port)`, `$sdp(m0:rtcp:port)`, `$sdp(m0:raw)`, `$sdp(m0:b:AS)`,
`$sdp(m0:b:RR)`, `$sdp(m0:b:RS)`.

**Reproduction.** With a two-stream offer such as an IMS audio+video INVITE:

```
v=0
o=- 0 0 IN IP4 198.51.100.1
s=-
c=IN IP4 198.51.100.1
t=0 0
m=audio 49170 RTP/AVP 96
a=rtpmap:96 AMR-WB/16000
m=video 51372 RTP/AVP 97
a=rtpmap:97 H264/90000
```

`$sdp(m0:rtp:port)` yields `51372` (the video stream) where `49170` is expected. The same offset
applies to the other `m0` variants.

**Impact seen in practice.** Found in a 5G SA / IMS deployment where the P-CSCF reads
`$sdp(m0:...)` while building the media description for QoS authorisation. On audio-only calls
the values were correct; as soon as video was offered, the first-stream lookup silently returned
the video stream's values.

**Testing.** The change has been running in a production IMS deployment carrying VoNR and ViNR
traffic. Voice, video and SMS-over-IMS calls were exercised after the change with no regression,
and single-stream bodies behave exactly as before. Verified against current `master`;
`clang-format` clean using the in-tree `.clang-format` with clang-format 18, the version the
`pull_request` workflow pins.

### DIFF
--- a/src/modules/sdpops/sdpops_mod.c
+++ b/src/modules/sdpops/sdpops_mod.c
@@ -2135,6 +2135,7 @@ int sdpops_attr_val(str *payload, str *attr, str *val)
 static int pv_get_sdp(sip_msg_t *msg, pv_param_t *param, pv_value_t *res)
 {
 	sdp_info_t *sdp = NULL;
+	sdp_stream_cell_t *stream0 = NULL;
 	str sess_version = STR_NULL;
 	long sess_version_num = 0;
 	unsigned int uport = 0;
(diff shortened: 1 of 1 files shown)


miconda:
Thanks!

PR #4937, opened 2026-09-18
registrar, ims_registrar_scscf: reject REGISTER with malformed Contact URI

### LINKED ISSUE #4926: registrar: malformed SIP Contact URI can be accepted and stored
### Description

A malformed SIP Contact URI can be accepted and stored by the registrar. Kamailio returns 200 OK and includes the malformed URI in subsequent REGISTER responses for the same address of record. 
If a response contains five Contacts and one is invalid, a receiving SIP implementation may reject the entire response, affecting devices with otherwise valid registrations.

Reproduced on upstream Kamailio with a minimal UDP registrar configuration using in-memory storage.

 Example malformed Contact:

  ```Contact: <sip:alice@192.0.2.10:5060transport=udp>;expires=3600```

  Expected behavior: reject a REGISTER containing a malformed SIP/SIPS Contact URI with 400 Bad Request before changing any bindings.            
  The invalid Contact must not be stored or included in responses to subsequent valid registrations.

  RFC 3261 §25.1 defines the SIP URI port as digits and requires a semicolon
  before URI parameters. Section 21.4.1 defines 400 for malformed syntax.

  References:
  - https://www.rfc-editor.org/rfc/rfc3261.html#section-25.1
  - https://www.rfc-editor.org/rfc/rfc3261.html#section-21.4.1

#### Reproduction

  Configure a minimal registrar using registrar and usrloc.
   Send a REGISTER with a malformed Contact URI:

  Contact: <sip:alice@192.0.2.10:5060transport=udp>;expires=3600

  Expected: 400 Bad Request.

  Reported behavior: the malformed Contact is stored and included in subsequent REGISTER 200 OK responses.

 Was reproduced by me locally

#### Debugging Data

  - src/modules/registrar/sip_msg.c:
    check_contacts() checks wildcard rules and Contact lengths, but does
    not parse each Contact URI.

  - src/modules/registrar/save.c:
    Contact URIs are passed to usrloc for storage. The conditional URI
    parsing used for TCP handling only logs a parsing failure.

  - src/modules/usrloc/ucontact.c:
    new_ucontact() copies the Contact URI string into the binding.

  - src/modules/registrar/reply.c:
    The response builder copies the stored Contact URI into the response.

#### SIP Traffic

Anonymized SIP message

```
  SIP/2.0 200 OK
  Via: SIP/2.0/UDP 192.0.2.15:5092;branch=z9hG4bKexample
  To: <sip:alice@example.com>;tag=server-tag
  From: <sip:alice@example.com>;tag=client-tag
  Call-ID: example-registration@example.com
  CSeq: 42 REGISTER
  Contact: <sip:alice-device1@192.0.2.11:5060;transport=udp>;q=0;expires=1484, <sip:alice-device2@192.0.2.12:5060;transport=udp>;q=0;expires=1627, <sip:alice-device3@192.0.2.13:5092;transport=udp>;q=0;expires=558, <sip:alice-device4@192.0.2.14:5060transport=udp>;q=0;expires=3055, <sip:alice-device5@192.0.2.15:5092;transport=udp>;expires=502
  Content-Length: 0

```

### Possible Solutions
 In src/modules/registrar/sip_msg.c, add URI validation inside the Contact loop in check_contacts(), before any bindings are saved:
```
  if(((c->uri.len >= 4 && strncasecmp(c->uri.s, "sip:", 4) == 0)
             || (c->uri.len >= 5 && strncasecmp(c->uri.s, "sips:", 5) == 0))
          && parse_uri(c->uri.s, c->uri.len, &uri) < 0) {
      rerrno = R_PARSE_CONT;
      return 1;
  }
```

[registrar-contact-uri-validation.patch](https://github.com/user-attachments/files/32015519/registrar-contact-uri-validation.patch)

### Additional Information

- [x] LLM/AI Assistants were involved in discovery, analysis or submission of the issue
- [x] It happened in a production deployment
- [] It happened in a testing or development deployment
- [ ] Testing or fuzzing tools were used to generate SIP traffic when it happened

  * **Kamailio Version** - output of `kamailio -v`

```
latest Kamailio version is affected
```

* **Operating System**:

```
Not related, but linux
```

rizwan3659:
#### Pre-Submission Checklist
- [x] Code is formatted with `clang-format` using the config file `.clang-format` from source code folder
- [x] No commits to README files for modules

#### Type Of Change

#### Checklist:
- [x] Tested changes locally
- [x] Related to issue #4926

#### Description
Fixes #4926: a REGISTER with a malformed SIP Contact URI such as
`<sip:alice@192.0.2.10:5060transport=udp>` was accepted with 200 OK and
the binding was stored, then echoed back in later 200 OK replies.

**registrar**
- `check_contacts()` now runs `parse_uri()` on every `sip:`/`sips:` Contact
  URI and replies 400 (`R_PARSE_CONT`, "Contact parse error") before any
  binding is changed. Other schemes are left untouched.

**ims_registrar_scscf** (same validation, plus two related bugs found while porting it)
- the non-star branch of `check_contacts()` started from
  `_m->contact->next`, so the contacts of the *first* Contact header were never
  checked (neither length nor the new URI check). It now starts from `_m->contact`.
- `save()` overwrote `rerrno` with `R_SAR_FAILED` after `check_contacts()`
  failed, so the UE got 500 instead of the proper 400. The rerrno set by
  `check_contacts()` is now kept.

**Testing**
- both modules compile without warnings (`-Wall`)
- `parse_uri()` checked against the example URIs: `...:5060transport=udp`
  and `...:50a60` are rejected; `...:5060;transport=udp`, IPv6 `sips:` and
  IMS `;user=phone` URIs are accepted


henningw:
Thanks for the PR. Looks good to me. The whole message is already parsed, and also all contact bodies. So it will not introduce any significant parsing overhead. There is also no good reason to store invalid contacts, as they will fail during lookup and later routing anyway. Lets wait a bit before merging it.

miconda:
Fine to merge it.

PR #4940, opened 2026-09-21
add dispatcher.reload_id to the dispatcher

UnderControlled:
#### Pre-Submission Checklist

- [ ] Code is formatted with `clang-format` using the config file `.clang-format`
  from source code folder

#### Type Of Change

#### Checklist:

- [x] Tested changes locally

#### Description

Added a dispatcher rpc `dispatcher.reload_id <id>` to reload a single element given its database id. Used an unused flag `(1 << 6)` in the flags attribute. If that flag is enabled, the entry will be considered as deleted (tombstone mechanism). This is done because internal data structures don't keep the database id.


miconda:
I am not sure this brings any real benefit rcompared with the reload, because as I can see, the operation first removes the  old record from the memory structure then adds a new one loaded from db. But a single record remove practically clones all the records without the one to be removed, being similar to a reload. The same is with adding a single record, all in memory records are cloned in a new structure where the new record is added as well. Over all, this operation results in two similar-to-reload operations. The db query is not returning the same amount of records, it can be a plus, but the additional "reload" operation can add to penalties. Besides that, this is diverging from the file backend. The id in the database table was not added for internal dispatcher needs, but only for direct database management.

Personally, I don't consider this useful.

Eventually, to get something similar, it could be useful dispatcher.update that combines the dispatcher.add and dispatcher.remove in a single (clone once) operation.

henningw:
Yes, I also don't see the a substantial benefit of doing individual reloads. The DB id should stay also isolated from the actual module logic, similar as for other modules. Set it to draft for now.

UnderControlled:
The intended use-case was to prevent split brain between the DB (at-least in DB-centered solutions) and the module. There is a scenario where I update the DB, and I have to either do "add + remove", which is not atomic, or I have to reload the whole DB, which will cost me move every destination to the state of Trying. This reload_id does handle the atomic update making sure no "remove without add" happens and also it helps to use the database as the single source of truth, so the Kamailio will eventually be consistent with the DB (It will never happen that I have added/removed/updated a record via a Kamailio RPC but the DB has failed to capture that, or vice versa), and finally, my single-record update won't change the state of any other destination by any accident.

If that is not the intended purpose of the module, or the atomicity + single-source-of-truth is possible via another means, please suggest me the solution.

Issue #4938, opened 2026-09-19
Dialog DMQ replication of timeout flags


scharrua:
### Description
Network description: 
    - Both Kamailio nodes share same loopback interface IP 10.242.17.147 and a VIP managed by Keepalived at 10.242.17.107. 
    - The VIP is used as a route gateway address for network router to send traffic to any of the Kamailio nodes. 
    - This way, the route table is set to send any traffic to 10.242.17.147 where VIP 10.242.17.107 is available and both Kamailio nodes are listening and advertising on 10.242.17.147. 
    - Having loopback interface and VIP used for SIP traffic allows sub-second failover reducing retransmissions.

### Expected behavior
Goal: test call timeout limited to 10 mins, so, the call should not be more than 10mins. In production, limits can go up to 24h

#### Actual observed behavior

1. VIP is set on SIP Node 1 who is **ACTIVE** (and SIP Node 2 is **PASSIVE**)
2. Endpoints send SIP traffic to 10.242.17.147 only.
3. SIP Node 1 receives initial INVITE on 10.242.17.107 and gets routes from routing engine, who instructs to limit call duration to 600secs (10mins)
4. SIP Node 1 will relay the call to egress; call is answered, and  SIP Node 1 flags it with timeout to current timestamp + 600s (as far as I understand it)
5. At 4 minutes Keepalived is killed on SIP Node 1 (systemctl kill keepalived)
6. SIP Node 2 is now ACTIVE (SIP Node 1 is not reachable yet because VIP is set on SIP 2). Remember that SIP 2 also has 10.242.17.107 address and that VIP is now pointing to SIP node 2.
7. After 10 mins, SIP Node 1 (who initially handled the call is not the ACTIVE node) sends correctly the BYE message to both endpoint to signal the timeout. 
8. The BYE is sent by SIP Node 1 via address 10.242.17.147 because Kamailio listens and advertises on this address. VIP is still on SIP Node 2 at this time.
9. SIP Node 2 who is receiving ALL SIP traffic (because it is the **ACITVE** node, VIP is present on Node 2) will then receive the 200 OK from endpoints who send replies to 10.242.17.107
10. SIP Node 1 who doesn't receive 200 OK (because VIP is active on SIP 2) from endpoints will keep retransmitting the BYE message. 
11. The CDR is closed, on billing services, but the SIP signalling is "weird" to say the least....
12. 

[...]

PR #4943, opened 2026-09-22
dialog: added dmq_timeout_iflags_sync modparam

rizwan3659:
#### Pre-Submission Checklist
  from source code folder

#### Type Of Change

#### Checklist:
- [ ] Tested changes locally
- [x] Related to issue #4938

#### Description
The timeout BYE flag (`DLG_IFLAG_TIMEOUTBYE`) is not replicated over DMQ, only the keep-alive flags are (with `dmq_ka_iflags_sync`). So when another node takes over the dialog, or the node that created it restarts and gets the dialogs back from its peer, the call is never ended with BYE at the timeout.

This adds a new modparam `dmq_timeout_iflags_sync`, same style as `dmq_ka_iflags_sync`. If set to 1, the owner sends its iflags in the DMQ update and the receiving node applies `DLG_IFLAG_TIMEOUTBYE`. The default is 0, so nothing changes for existing setups.

With it enabled, every node that has the dialog will send BYE on timeout, so if only the active node should do it, that has to be handled in the config (e.g. `dlg_filter_mode` or the timeout event route), as noted in the docs.

The dialog module builds without warnings. I have not tested it yet on a two node DMQ setup.

Code and docs are in separate commits, I can squash them into one if preferred.


henningw:
Thanks for the PR, would be great if you could also do some basic tests, e.g. with a two node DMQ setup.

